PR #1, opened 2019-07-05
add Dockefile to create image to use prusti with for projects

karroffel:
Hey! This is a really exciting project, I was interested so I wanted to try it out and didn't find a docker image to just boot up and use it with.

I added a Dockerfile which is based on `fpoli/prusti-base`, fetches the repository via git and then builds it and links the binaries properly. With this docker image it is possible to simply create a (or mount an existing) project and use the familiar `cargo` commands - prusti works out of the box.

I am not very experienced with docker, so maybe it would be a better idea to copy the current project into the image instead of fetching from git, if so I'd be happy to make changes!

I also added this image to the `publish-docker-images` make target, I don't know if that's desired, here as well I'd be happy to change it.

------------

When preparing the branch I noticed that some files where marked as "changed" even though I just freshly cloned the repository. This is because of CRLF -> LF conversion.

-----------

I love this project, thanks for working on this! :blue_heart:


vakaras:
Thank you for your pull request! We are currently quite buried under work, so most likely we will not be able to review it until the end of the next week.

karroffel:
I saw that efec88b8356bded53637948318dcdbf199b0e81a does something very similar but without setting the `RUSTC_WRAPPER` and some other envs.

If that image would be published then I'd say this PR is not really needed anymore.

karroffel:
It seems like c9baeef67061d99b8b06bfcf79ada9a00e058b04 should support this exact functionality that this PR adds, awesome :)

fpoli:
Hi, thanks for the PR! Indeed, we needed to publish a very similar Docker image. In the future we would like to publish pre-compiled binaries, so trying Prusti should become much easier.

I guess you already found that the `cargo-prusti` binary internally sets `RUSTC_WRAPPER` and calls `cargo check`. If you have any problem feel free to open an issue!

karroffel:
I will play around with the artifact image, thanks!

I tested out prusti a bit over the past few days and noticed some things. I am aware that all of this is still very much in development, should I report issues for things that are not yet "acknowledged" by the compiler anyway?

(For example I wouldn't open an issue on arrays not being supported since the compiler notes that. I found that when trying to use a `#[pure]` function from another *crate* it just "silently" says that the function is not marked as pure (even if it is), this seems to not be "acknowledged" so I would think it's worth a report )

fpoli:
In general, if `cargo-prusti` or `prusti-rustc` crashes without showing "partially supported/unsupported" warning/error messages, or if the tool does not crash but the result is somehow unexpected, then opening an issue makes sense because it's something that we either want to warn about or fix.

Regarding your example, I would not open an issue for arrays because the tool should already report that they are unsupported before crashing. However, an issue for the usage of "#[pure]" from another crate makes sense: I would also expect that the specification is picked up. (We haven't tested usage of specifications across crates so far.)

That said, opening an issue even just for a question is perfectly fine :)

PR #5, opened 2020-01-30
Fix $SHELL Makefile variable

alevy:
This is just a small nit to help the build system work on niche Linux setups (like mine).

On some Unix's (specifically NixOS), bash is not in `/bin/bash` (which is actually _not_ a standard POSIX requirement), so use the standard `/usr/bin/env bash` instead.


fpoli:
Thank you!

PR #14, opened 2020-03-23
Add basic docker build

xldenis:
Hi,

I tried building this on my Mac and encountered a bunch of errors related to Scala and other things so I gave up and made a basic docker build instead. I've just confirmed that `make test` works on my computer.

The `Dockerfile` is very bare-bones but successfully installs viper and builds prusti.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM openjdk:buster
+
+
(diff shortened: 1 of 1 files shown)


xldenis:
never mind, I see there is an included dockerfile already, it just wasn't at the top-level.

PR #21, opened 2020-04-10
Adding a first, partial support for arrays and slices

mario-bucev:
Hello,
This PR adds some support for arrays and slices.

***Type encoding***
The array predicate is encoding as follow:
```silver
field val_array: Seq[Ref]

predicate array$ElemType$size(self: Ref) {
  acc(self.val_array, write) 
  && |self.val_array| == size 
  && forall i: Int :: { self.val_array[i] } 0 <= i && i < |self.val_array| ==> acc(self.val_array[i].val_ref, write) 
  && forall i: Int :: { self.val_array[i] } 0 <= i && i < |self.val_array| ==> acc(ElemType(self.val_array[i].val_ref), write)
}
```
Slices are encoded the same way, but without specifying the size of the sequence.

***Sequence indexing and length operations encoding***
`SeqIndex` and `SeqLen` are introduced in `vir::ast::expr::Expr` for that purpose.
The translation of array/slice indexing in Rust to Silver is naturally handled.

For slice length operation, we "intercept" the call and translate this function call to a `SeqLen` expression.

Note: the `.len()` method on arrays is unsupported because it produces a cast of `Unsized` nature.

***Handling quantified resource accesses***
The type predicates for slices and arrays use quantified field and predicate accesses. `QuantifiedResourceAccess` structure in `vir::ast::expr` is introduced to work with them.

***Limitations***
Many limitations are to be expected, but these ones are the most severe:
* Encoded programs generally fail to be verified in Silicon (but many can be verified in Carbon).
* Arrays or slices of `struct` are unsupported
* Repeat operations (e.g. `[0usize; 64]`) are unsupported
* Arrays whose size is not a literal (for instance, an unevaluated constant expression) are unsupported.

I am open to any feedback and questions :) I hope that this PR meets the standards of the project.

Thank you


fpoli:
Thank you for the PR! It seems you learned Prusti's internals quite well! :)

The failing tests are very interesting. Regarding those that take too much time to verify, do you know if they all work with Carbon? Also, do you have an intuition of what might be the cause of the slowdown (e.g. an unusually large amount of fold/unfold in the encoding)? In the long run we would like to synthesize some worst-case Viper examples and report them upstream in Silicon, so any hint from your experience is useful.

We are also aware of some inefficiencies of Silicon with regard to Viper sequences, so if that turns out to be a bottleneck we could try to encode arrays using a [Viper domain](http://viper.ethz.ch/tutorial/?page=1&section=#array-domain) instead of a Viper `Seq`. Once we merge this PR we should open an issue to remember to try the encoding with domains.

We are a bit busy these weeks, so we might not be very quick in reviewing and merging the PR, but from a quick look it looks very good :+1:

mario-bucev:
Hi!
Regarding the tests that take too much time to verify, I meant that Carbon took too much time (I stop the verification process when reaching ~3mins). Silicon rejects these tests. For the tests it can verify, Silicon is pretty fast though.
To me, it seems that Silicon handles very well `&mut` array or slices, but struggles (by rejecting) programs that have shared array/slices. I think it has the same root issue as https://github.com/viperproject/silicon/issues/479.

vakaras:
Hello @mario-bucev, thank you for the pull request and sorry that it took me so long to come back to you!

I am wondering if it would not be better to encode arrays and slices by using abstract predicates (I am ignoring read permissions here):

```viper
predicate Array(self: Ref);
predicate Slice(self: Ref);

function Array$len(self: Ref): Int
  requires acc(Array(self))
  ensures result >= 0
function Array$lookup(self: Ref, index: Int): Int
  requires acc(Array(self)) && 0 <= index && index < Array$len(self)

// Same two functions for `Slice`.
```

With using `len` and `lookup` functions, it should be possible to provide specs in the same way as for the `Vec` in the [Selection sort example](https://github.com/viperproject/prusti-dev/blob/d8fd79355c4ff2a84f9fa21a0d5278effe720297/prusti/tests/verify/pass-overflow/rosetta/Selection_sort.rs#L22-L71). In this way, it should be possible to reuse the existing Prusti infrastructure and avoid changing VIR and the fold-unfold algorithm.

vakaras:
By the way, I am going to start upgrading Prusti to work with the new compiler version, which most likely will move around and break a lot of things.

mario-bucev:
Indeed it would be simpler; I'll start looking at this approach once the upgrading is done :)

fpoli:
@mario-bucev you can resume working on this PR if you wish. The compiler upgrade is done. We are just fixing some regressions.

mario-bucev:
Hello @fpoli, thanks for letting me know! I hope to resume working on it once I have enough time.
Cheers

fpoli:
@mario-bucev do you plan to work on this? If you don't mind, we would like to include the topic of encoding arrays with the `predicate Array(self: Ref)` in a thesis that is about to begin.

mario-bucev:
@fpoli I fear I do not have much time in my hands for the upcoming months, so I'll close the PR. I hope that a simpler and better alternative will be found, as mine is quite cumbersome (and outdated) :p

fpoli:
Thanks for the quick reply! I'm sure we can still include some of your work, like the tests that you wrote.

Issue #27, opened 2020-05-02
Temporary borrows in condition to break out of loop causes panic


f4z3r:
Temporary borrows such as used by pattern matching that wrap a `break` statement inside a loop cause Prusti to panic.

## Rust Code

```rust
#![allow(path_statements)]
fn main () {
    loop {
        if let Some(_) = get_a_number() {
            1;
        } else {
            break;
        }
    }
}

fn get_a_number() -> Option<u32> {
        Some(5)
}
```

Note that changing the condition to something like `if get_a_number().is_some()`, does not exhibit the problem. Changing the `break` statement to anything else (even `continue`) also does not exhibit the problem.

## Prusti log

```
  __          __        __  ___             
 |__)  _\/_  |__) |  | /__`  |   ____\/_  |  Hash:  ccf572597ff91ca6c7c
 |      /\   |  \ \__/ .__/  |       /\   |  Build: 2020-04-24 10:57:06

 INFO 2020-05-02T13:43:41Z: prusti::compiler_calls: Parsing of annotations successful (0.0 seconds)
 INFO 2020-05-02T13:43:41Z: prusti::compiler_calls: Type-checking of annotations successful (0.0 seconds)
Verification of 2 items...
 INFO 2020-05-02T13:43:41Z: viper::viper: Using JVM OpenJDK 64-Bit Server VM, Java 11.0.7
 INFO 2020-05-02T13:43:41Z: prusti::verifier: JVM startup (0.4 seconds)
 INFO 2020-05-02T13:43:41Z: viper::verification_context: Using Z3 path: '/usr/bin/viper-z3'
 INFO 2020-05-02T13:43:41Z: viper::verifier: Using backend Silicon version <unknown-build-version>
 INFO 2020-05-02T13:43:43Z: prusti::verifier: Verifier startup (1.67 seconds)
 INFO 2020-05-02T13:43:43Z: prusti_viper::verifier: Received 2 items to be verified:
 INFO 2020-05-02T13:43:43Z: prusti_viper::verifier:  - code::main from code.rs:2:1: 10:2 (code::main[0])
 INFO 2020-05-02T13:43:43Z: prusti_viper::verifier:  - code::get_a_number from code.rs:12:1: 14:2 (code::get_a_number[0])
warning: [Prusti] this is partially supported, because it causes abrupt loop terminations
 --> code.rs:4:16
  |
4 |         if let Some(_) = get_a_number() {
  |                ^^^^^^^

[...]

PR #28, opened 2020-05-06
Report nice errors when impure functions are used

### LINKED ISSUE #11: Warn about non-pure function calls in specifications
Prusti should detect calls of non-pure functions in specifications, and generate a proper error message for them.

Currently, Prusti panics on programs like the following.

```rust

// #[pure] // good: mark this function as pure
fn get_true() -> bool {
    true
}

#[requires="get_true()"] // bad: this contract calls a non-pure function
fn main(){
    assert!(true);
}
```

sapphire-arches:
Fixes #11.

This approach allows lowering to continue, despite it generating invalid VIR. However, we never pass that code to Viper since we exit the verifier early if any errors are reported during the lowering process.

### DIFF
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -13,6 +13,7 @@ use encoder::foldunfold;
 use encoder::places;
 use encoder::procedure_encoder::ProcedureEncoder;
 use encoder::pure_function_encoder::PureFunctionEncoder;
+use encoder::stub_function_encoder::StubFunctionEncoder;
 use encoder::spec_encoder::SpecEncoder;
 use encoder::type_encoder::{
     compute_discriminant_values, compute_discriminant_bounds, TypeEncoder};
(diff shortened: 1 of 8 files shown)


fpoli:
Nice, thanks! I guess the same approach can be used to fix #27.

When the "procedure is not marked as pure" error is raised, I think it would be nice to generate anyway a valid and trivial VIR (e.g. a body-less pure function with `false` as pre- and post-condition) to let Viper verify the rest of the project.

@vakaras @cmatheja any thoughts?

cmatheja:
Thanks, the solution looks good to me.

I agree with @fpoli that we should still pass something to Viper such that it can catch issues with other functions.

sapphire-arches:
Thanks for the feedback =). Ideally the approach we settle on here can be generalized to handle all errors which arise during encoding, since I assume there's more than just this one and #27. I'll look in to generating some stub functions to allow verification to continue.

sapphire-arches:
I've added some code which generates stub functions like:

```
function m_non_pure_function$$foo$opensqu$0$closesqu$$$get_false$opensqu$0$closesqu$_stub(): Bool
  requires false
  ensures false
```
which allows verification to continue, while keeping the slightly more detailed error reporting.

vakaras (approved):
Thank you for your pull request! Looks good to me.

PR #40, opened 2020-05-30
fix trait refinement post-condition inhale and pre-condition exhale

f4z3r:


f4z3r:
The current status of this PR only fixes this when using methods with a `self` parameter. Static functions such as in the following snippet are not supported. Therefore the snippet fails to verify.

```rust


trait Foo {
    #[ensures="result > 0"]
    fn foo(_a: i32) -> i32;
}

struct Dummy { }

impl Foo for Dummy {
    #[ensures="result == 5"]
    fn foo(_a: i32) -> i32 {
        5
    }
}

fn main() {
    let val = Dummy::foo(3);
    assert!(val == 5);
}
```

fpoli:
Is this ready to be merged, as it's not marked as "draft"?

f4z3r:
> Is this ready to be merged, as it's not marked as "draft"?

It is ready with the limitation commented above. Maybe someone wants to review it though?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -344,18 +344,59 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
+
+        // get receiver object base type
+        let tcx = self.env().tcx();
+        let mut impl_spec = SpecificationSet::Procedure(vec![], vec![]);
+        let sig = tcx.fn_sig(proc_def_id);
+
+        if sig.inputs().skip_binder().len() > 0 {
+            let self_ty_binder = sig.input(0);
+            let mut self_ty = self_ty_binder.skip_binder().clone();
+
```
fpoli, line 373: Nit: could you avoid the double empty line?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -344,18 +344,59 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
+            }
+
+            // get specification on trait impl if exists, and has receiver object
+            if self_ty.is_self() {
+                if let Some(ty) = tymap[0].get(self_ty) {
+                    if let Some(id) = tcx.trait_of_item(proc_def_id) {
+                        let proc_name = tcx.item_name(proc_def_id).to_string();
+                        tcx.for_each_relevant_impl(id, ty, |impl_id| {
+                            for assoc_item in tcx.associated_items(impl_id) {
+                                if assoc_item.name == proc_name {
```
fpoli: What you are doing here sounds generic enough to be worth a method in `Environment`.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -344,18 +344,59 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
+                                    if let Some(spec) = self.get_spec_by_def_id(assoc_item.def_id) {
```
fpoli: In one execution of `for_each_relevant_impl`, do you expect this line (`get_spec_by_def_id`) to be executed just once or multiple times (for different `assoc_item` or `impl_id`)?

In the former case, it might be better to split the implementation in two steps:
1. Find the appropriate `assoc_item` with a call to `for_each_relevant_impl` (or with the method suggested in the previous comment).
2. Call `get_spec_by_def_id` on the `assoc_item` that you found.

fpoli (changes requested):
Very good, thanks! I think we can simplify a bit the `tcx.for_each_relevant_impl` call.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -374,16 +373,13 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
                         tcx.for_each_relevant_impl(id, ty, |impl_id| {
-                            for assoc_item in tcx.associated_items(impl_id) {
-                                if assoc_item.name == proc_name {
-                                    if let Some(spec) = self.get_spec_by_def_id(assoc_item.def_id) {
-                                        impl_spec = spec.clone();
-                                    } else {
-                                        debug!("Procedure {:?} has no specification", assoc_item.def_id);
-                                    }
-                                    break;
+                            if let Some(item) = self.env().get_assoc_item(impl_id, proc_name) {
```
fpoli: I meant to move `for_each_relevant_impl` as well to a `Environment` method.

If I understand correctly, lines 377-378 iterate over all implementations of a trait and selects one associated item. So, this could become a `get_associated_item_impl(trait_id, ty, def_id) -> Option<...>` method in Environment. Using it, it would be explicit that `impl_spec = ...` cannot be executed more than once.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -344,18 +344,55 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
+                self_ty = true_ty;
+            }
+
+            // get specification on trait impl if exists, and has receiver object
+            if self_ty.is_self() {
+                if let Some(ty) = tymap[0].get(self_ty) {
+                    if let Some(id) = tcx.trait_of_item(proc_def_id) {
+                        let proc_name = tcx.item_name(proc_def_id).as_symbol();
+                        tcx.for_each_relevant_impl(id, ty, |impl_id| {
+                            if let Some(item) = self.env().get_assoc_item(impl_id, proc_name) {
```
fpoli: I meant to move `for_each_relevant_impl` as well to a `Environment` method.

If I understand correctly, lines 377-378 iterate over all implementations of a trait and selects one associated item. So, this could become a `get_associated_item_impl(trait_id, ty, def_id) -> Option<...>` method in Environment. Using it, it would be explicit that `impl_spec = ...` cannot be executed more than once.

prusti-interface/src/environment/mod.rs
```diff
@@ -200,4 +201,21 @@ impl<'r, 'a, 'tcx> Environment<'r, 'a, 'tcx> {
...
+        self.tcx().associated_items(id).find(|assoc_item| assoc_item.name == name)
+    }
+
+    /// Get a trait method declaration by name for type.
+    pub fn get_trait_method_decl_for_type(&self, typ: ty::Ty<'tcx>, trait_id: DefId, name: Symbol) -> Option<ty::AssociatedItem> {
+        let mut result = None;
+        self.tcx().for_each_relevant_impl(trait_id, typ, |impl_id| {
+            let item = self.get_assoc_item(impl_id, name);
+            if item.is_some() {
+                result = item;
```
fpoli: Last thing: could you `assert!(result.is_none())` in this branch?

fpoli (changes requested):

prusti-interface/src/environment/mod.rs
f4z3r: Sure will, do. I will also try to change this PR to not have the above mentioned restriction. I will have an in depth look into this next week.

fpoli: Which restriction? The `assert!(result.is_none())` would not be a negative thing in my opinion.

f4z3r: The restriction with the `self` parameter. The current solution does not work for functions without an implicit `self` type.

f4z3r:
The only not supported feature of trait refinement left to do is:

```rust
impl From<i32> for Dummy { ... }
impl From<bool> for Dummy { ... }

fn foo(...) {
    let d = Dummy::from(val);
}
```

In the case above, an info message is provided (that trait refinement on generic traits is not supported), but any trait refinements on the `from` method in the trait definitions is ignored, as Prusti has trouble determining which contracts are relevant for the call and which are not.

Note that if only a single generic trait implementation is relevant for some _type_, it will still be supported, as the method to pick for the trait refinement is not ambiguous.

prusti-interface/src/environment/mod.rs
f4z3r: I did not add `assert!(result.is_none())` as the code currently has issues differentiating between generic traits implemented with different associated types. See my new comment below for an example. Putting this assert here would create a panic in such situations, even though they can be handled just as they were before; i.e. we simply do not support trait refinement checking on such traits if several implementations exist for a single type.

fpoli: I see that you already changed the return type to `Result<..>`; perfect! I was about to ask that.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -382,11 +382,16 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
             if let Some(id) = self.env().tcx().trait_of_item(proc_def_id) {
                 let proc_name = self.env().tcx().item_name(proc_def_id).as_symbol();
-                if let Some(item) = self.env().get_trait_method_decl_for_type(ty, id, proc_name) {
-                    if let Some(spec) = self.get_spec_by_def_id(item.def_id) {
-                        impl_spec = spec.clone();
-                    } else {
-                        debug!("Procedure {:?} has no specification", item.def_id);
+                let proc_res = self.env().get_trait_method_decl_for_type(ty, id, proc_name);
+                if proc_res.is_err() {
+                    info!("trait refinement not supported on generic traits");
```
fpoli: It would be great if you add a variant to `EncodingError` (even just a `Generic(String)` one) and call `self.register_encoding_error(..)` here, instead of `info!(..)`.

f4z3r: Sure, but this will make some existing tests fail. Some tests implement traits with different associated items, and can be verified, as they don't rely on contract refinement on those trait methods. If I register an error there, it will fail compilation and therefore fail the test.

See test `008-num-traits-pow.rs`.

f4z3r: I will add the error, as I guess it would be useful also for other parts of the code, and push it here. Then we can see which tests will fail.

prusti-viper/src/encoder/error_manager/mod.rs
```diff
@@ -98,6 +99,13 @@ pub enum ErrorCtxt {
 pub enum EncodingError {
     Procedure(ProcedureEncodingError),
     PureFunction(PureFunctionEncodingError),
+    Generic(GenericEncodingError),
```
fpoli: I think `Generic(String, Span)` is good enough. I find the additional `GenericEncodingError` a bit redundant.

fpoli (approved):

PR #42, opened 2020-06-01
fix compiler warnings

karlosss:


karlosss:
also add info about how to run tests to `CONTRIBUTING.md`

prusti-filter/src/validators/support_status.rs
```diff
@@ -111,14 +111,6 @@ impl SupportStatus {
...
     }
 
-    pub fn is_partially_supported(&self) -> bool {
-        !self.precise_restrictions.is_empty()
-            && self
-                .precise_restrictions
-                .iter()
-                .all(|s| s.is_partially_supported())
-    }
-
```
fpoli: I would prefer marking this as `#[allow(dead_code)]` instead of removing it, to keep the API uniform as we also have `is_supported`.

prusti-viper/src/encoder/mir_encoder.rs
```diff
@@ -104,12 +105,15 @@ impl<'p, 'v: 'p, 'r: 'v, 'a: 'r, 'tcx: 'a> MirEncoder<'p, 'v, 'r, 'a, 'tcx> {
     /// - `vir::Expr`: the place of the projection;
     /// - `ty::Ty<'tcx>`: the type of the place;
     /// - `Option<usize>`: optionally, the variant of the enum.
-    fn encode_projection(
+    pub fn encode_projection(
         &self,
         place_projection: &mir::PlaceProjection<'tcx>,
+        encoded_place: Option<(vir::Expr, ty::Ty<'tcx>, Option<usize>)>
```
fpoli: Could you explain this change in the rustdoc (the `/// ...`)?

fpoli (changes requested):
Nice, thanks! Just two small changes, then we can merge this.

prusti-viper/src/encoder/mir_encoder.rs
karlosss: There is no change, is there? Currently, it just describes the return values, which remain unchanged.

fpoli: I'm referring to the new parameter on line 111. Could you add a note in the rustdoc explaining how the `encoded_place` parameter should be used? For example, it seems that `encoded_place` can be used to replace the encoding of `place_projection.base`, and this seems an important thing to mention. Also, what about renaming the parameter to `encoded_base_place`? It might be easier to understand that it's just the encoding of the base of the projection.

karlosss: Alright, it should be changed now.

PR #43, opened 2020-06-02
Prusti Server Mode

juliand665:
As part of my Bachelor's Thesis (under Federico Poli), I'm implementing a "server mode" for Prusti in which it keeps around a launched JVM as well as caching verifier instances to get rid of startup delays in verification and take load off clients.

This is a WIP draft (not ready for review yet), mostly for tracking purposes.


prusti-common/README.md
```diff
@@ -0,0 +1,3 @@
+# prusti-common
+
+Common utilities used across the Prusti project.
```
fpoli: I would add "compiler-agnostic" (utilities), as the idea is to never use rustc as a library in this crate.

prusti-common/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+            "{} ({}.{} seconds)",
+            $desc,
+            duration.as_secs(),
+            duration.subsec_millis() / 10
+        );
+    };
+}
+
+/// Runs a given function, timing and logging (info-level) how long it took, returning the function's result.
+pub fn run_timed<F: FnOnce() -> T, T>(desc: &'static str, task: F) -> T {
```
fpoli: Does it also work if `desc: &'static str` is declared as `desc: &str`?

prusti-interface/src/utils.rs
```diff
@@ -4,7 +4,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-//! Various helper functions for working with `mir::Place`.
+//! Various helpers functions for working with `mir::Place`.
```
fpoli: Could you check the correct spelling?

prusti-server/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+        program: Program,
+        backend: VerificationBackend,
+    ) -> FutVerificationResult {
+        self.get_or_create_thread(backend).verify(program)
+    }
+
+    fn get_or_create_thread(&mut self, backend: VerificationBackend) -> &VerifierThread {
+        if !self.threads.contains_key(&backend) {
+            let thread = VerifierThread::new(self.verifier_builder.clone(), backend);
+            self.threads.insert(backend, thread);
```
fpoli: You still need to implement a limit on the thread pool size, right?

prusti-server/src/service.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn new_from_string(server_address: String) -> Self {
+        Self::new(server_address.parse().unwrap())
+    }
+
+    pub fn new(server_address: SocketAddr) -> Self {
+        let client = SyncClient::connect(server_address, client::Options::default()).unwrap();
+        Self { client }
+    }
+
+    pub fn new_fake() -> Self {
```
fpoli: Could you add brief a rustdoc explaining the `_fake`?

prusti-server/src/verifier_thread.rs
```diff
@@ -0,0 +1,61 @@
...
+        let builder = thread::Builder::new().name(format!("Verifier thread running {}", backend));
+
+        builder
+            .spawn(move || {
+                let context = verifier_builder.new_verification_context();
+                let verifier = context.new_viper_verifier(backend);
+                let ast_factory = context.new_ast_factory();
+                while let Ok(request) = request_receiver.recv() {
+                    let viper_program = request.program.to_viper(&ast_factory);
+                    let result = verifier.verify(viper_program);
```
fpoli: It would be nice to measure and log timings for the calls to the JVM.

prusti-viper/src/encoder/vir/cfg/method.rs
```diff
@@ -12,30 +12,35 @@ use uuid::Uuid;
 
 pub(super) const RETURN_LABEL: &str = "end_of_method";
 
-#[derive(Debug, Clone)]
-pub struct CfgMethod {
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CfgMethod { // TODO: Should labels and reserved_labels be ignored?
```
fpoli: Regarding the TODO: the best thing here would be to
1. rename `CfgMethod` to `CfgMethodBuilder`, and remove the serialize/deserialize from it
2. add a new, much simpler, `CfgMethod` (serializable, without `labels`, `reserved_labels`, ...)
3. add a `fn build(self) -> CfgMethod` to `CfgMethodBuilder`
4. move any `to_viper` method from `CfgMethodBuilder` to `CfgMethod`

If we don't do this, then skipping label fields is ok as a temporary solution (if they are not used in the conversion to Viper).

prusti-viper/src/encoder/vir/program.rs
```diff
@@ -0,0 +1,94 @@
...
+pub struct Program {
+    pub fields: Vec<Field>,
+    pub builtin_methods: Vec<BodylessMethod>,
+    pub methods: Vec<CfgMethod>,
+    pub functions: Vec<Function>,
+    pub viper_predicates: Vec<Predicate>,
+}
+
+impl Program {
+    pub fn optimized(mut self) -> Self {
```
fpoli: Could you move
```rust
impl Program {
    pub fn optimized(...) -> Self { ...}
}
```
to the `mod.rs` of the `optimizations` module? Such that `vir` doesn't use stuff from `optimizations`.

prusti-viper/src/encoder/vir/program.rs
```diff
@@ -0,0 +1,94 @@
...
+            })
+            .collect();
+        self.functions = new_functions
+            .into_iter()
+            .map(|f| optimizations::folding::FoldingOptimizer::optimize(f))
+            .collect();
+        self
+    }
+
+    pub fn to_viper<'v>(self, ast: &AstFactory<'v>) -> viper::Program<'v> {
```
fpoli: Could you move this method to the `to_viper` module, implementing the `ToViper` trait for `Program`?

prusti-viper/src/verification_service.rs
```diff
@@ -0,0 +1,12 @@
...
+
+pub trait VerificationService {
+    fn verify(&self, program: Program, config: ViperBackendConfig) -> viper::VerificationResult;
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ViperBackendConfig {
+    Silicon(Vec<String>),
+    Carbon(Vec<String>),
+}
```
fpoli: Could you add a brief rustdoc that explains what is a `ViperBackendConfig` and what are the `Vec<String>`?

fpoli (commented):
I added some comments.

prusti-server/src/lib.rs
juliand665: That's correct; I'll add a `TODO` note.

prusti-interface/src/utils.rs
juliand665: oops, no idea how that slipped through!

prusti-server/src/service.rs
juliand665: This whole method is a little questionable; I've renamed it and expanded in the comments/docs

prusti-server/src/verifier_thread.rs
juliand665: Indeed. In the newest commits, this will be taken care of.

prusti-viper/src/encoder/vir/program.rs
juliand665: Oh yeah, for some reason I got confused a while back when this idea first came up. Because the code uses `into_iter()`, which requires ownership, I thought I couldn't easily implement `ToViper`, but it turns out it works just fine with regular old `iter()`, so I've implemented the trait now.

fpoli:
I had a thought on `run_timed!`, as it's a bit strange to have a sequence of statements in between `(...)` parenthesis, and `let` statements in an indented block are especially confusing.

What if we define a `Stopwatch` type with an API like the following?
* `fn start(&mut self, msg: &str) -> &mut Self` starts a measurement,
* `fn done(&mut self) -> &mut Self` stops the timer and prints the measurement,
* on `drop()`, if the stopwatch is still running, it panics or calls `error!(..)` with "please always call .done()".

Example usage:
```rust
let stopwatch = Stopwatch::new().start("phase 1");
// ...
stopwatch.done().start("phase 2");
// ...
stopwatch.done().start("phase 3");
// ...
stopwatch.done();
```

Alternative API:
```rust
let stopwatch = Stopwatch::start("phase 1");
// ...
stopwatch.restart("phase 2");
// ...
stopwatch.restart("phase 3");
// ...
stopwatch.done();
```

juliand665:
That's not bad! I like how the macro makes invalid usage impossible, but in the end it was probably fighting the compiler a bit too much (I wish I could just take a block and paste it in without enclosing it in a scope). Your solution is much more idiomatic and doesn't increase indentation. In your first API example, we could probably have a generic argument `State` that makes it impossible to call `done` on not-running or `start` on running stopwatches, although the alternative API simply makes that inexpressible at the cost of forcing you to time contiguous sections, hmm. Going more into the latter API, we could even make `done` consume self. I'll experiment with some different approaches and see what feels right in our codebase.

prusti-common/src/config.rs
```diff
@@ -184,6 +178,19 @@ pub fn report_support_status() -> bool {
...
 
+/// The maximum amount of instantiated viper verifiers the server will keep around for reuse.
+///
+/// **Note:** This does _not_ limit how many verification requests the server handles concurrently, only the size of what is essentially its verifier cache.
+pub fn server_max_stored_verifiers() -> usize {
+    read_setting("SERVER_MAX_STORED_VERIFIERS")
+}
+
+/// When set, Prusti will connect to this server and use it for its verification backend (i.e. the things using the JVM/Viper).
+/// Set to "MOCK" to run the server off-thread, effectively mocking connecting to a server without having to start it up separately.
```
fpoli, line 192: Is `<ip>:<port>` a valid `SERVER_ADDRESS`?

prusti-server/src/verifier_runner.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn verify(&self, program: Program, program_name: &str) -> VerificationResult {
+        let mut stopwatch = Stopwatch::start("construction of JVM objects");
+        let viper_program = program.to_viper(&self.ast_factory);
+        if config::dump_viper_program() {
+            self.dump(viper_program, program_name);
+        }
+
+        stopwatch.start_next_section("verification");
+        self.verifier.verify(viper_program)
+    }
```
fpoli: Could you measure and log the duration of `to_viper`, `dump` and `verify` with the `Stopwatch`?

prusti-viper/src/verifier.rs
```diff
@@ -228,148 +231,28 @@ impl<'v, 'r, 'a, 'tcx> Verifier<'v, 'r, 'a, 'tcx> {
...
 
-        let duration = start.elapsed();
-        info!(
-            "Verification complete ({}.{} seconds)",
-            duration.as_secs(),
-            duration.subsec_millis() / 10
-        );
+        let mut program = self.encoder.get_viper_program();
+        if config::simplify_encoding() {
+            program = program.optimized();
```
fpoli, line 125: Could you measure `optimized` with a `Stopwatch`?

prusti-viper/src/verifier.rs
```diff
@@ -228,148 +231,28 @@ impl<'v, 'r, 'a, 'tcx> Verifier<'v, 'r, 'a, 'tcx> {
...
+        }
+        let verification_result = verify(program);
 
         let verification_errors = match verification_result {
             viper::VerificationResult::Failure(errors) => errors,
-            _ => vec![],
+            viper::VerificationResult::ConsistencyErrors(errors) => {
+                errors.iter().for_each(|e| error!("{}", e));
+                panic!("Consistency errors. The encoded Viper program is incorrect.");
```
fpoli: Instead of `panic`, could you try using `PrustiError::internal("...", DUMMY_SP).emit();`, returning `Failure`?

fpoli:
I would like to merge this early, leaving further improvements and features to new PRs. Could you rebase on `master` (or pull changes from there)?

juliand665:
> I would like to merge this early, leaving further improvements and features to new PRs. Could you rebase on `master` (or pull changes from there)?

Sounds good, actually. Apart from some usability concerns & optimization opportunities with the purely additive features (the standalone server executable), it seems to be functioning rather well.
I've updated from master. Assuming the test suite passes (and that github is lying to me about the conflicts i just fixed), I think we can merge now—I'll mark it as such.

fpoli:
Thanks! I forgot that we recently started encoding Viper `Domains`, so they should be added to the `Program` structure to pass the test suite. One more reason to merge early.

https://github.com/viperproject/prusti-dev/blob/7b5b2f677d4809b6a110d578572ccf638a535ca8/prusti-common/src/vir/ast/domain.rs#L10-L16

https://github.com/viperproject/prusti-dev/blob/5295a5858d27ef5e180c1fd26fa4062dc348db8d/prusti-viper/src/encoder/encoder.rs#L200

https://github.com/viperproject/prusti-dev/blob/7b5b2f677d4809b6a110d578572ccf638a535ca8/prusti-viper/src/verifier.rs#L242-L245

fpoli:
I tried the branch, and on simple programs the runtime goes down from ~10s to 1.6s! :tada::tada::tada:

I also tried `make run-release-profile` and `make run-release-flamegraph`: they run perfectly fine when using the server. On medium-size tests a lot of time seems to be spent in cloning `PathCtxt` and in calling `compute_definitely_initialized`, but it varies a lot. If you choose to work on profiling, I suggest to profile on big programs that look as much real-world as possible. For example, by instrumenting `cargo-prusti` and profiling Prusti while it verifies the core proof of real crates like the one mentioned in #77. The outcome should be a report (or Github issue) with e.g. the top-10 performance issues of Prusti, with a quantitative motivation (e.g. how you measured it). Suggesting how to fix the performance issues would be a nice plus, but there is no need to actually fix them during the thesis.

juliand665:
Sounds good! Thanks for the ideas; I'll look into it as I go.

I think we can merge now.

PR #44, opened 2020-06-03
Double pass parser for improved marker trait invariants

f4z3r:


f4z3r:
Moved to new version, with complete rewrite, see PR #48

PR #46, opened 2020-06-04
Support loop conditions with side effects, and improve error reporting

fpoli:


fpoli:
Support for *any* loop condition is not yet complete, but I'm merging this PR as it already has many improvements for the error reporting.

PR #54, opened 2020-06-11
Pledges for multiple input and output references.

frececroka:
This implements support for re-borrowing functions with multiple input and output references. To see what is supported, check out the test cases for the [callees](https://github.com/frececroka/prusti-dev/blob/lorenz/thesis/prusti-tests/tests/verify/pass/pledges/callees.rs) and [callers](https://github.com/frececroka/prusti-dev/blob/lorenz/thesis/prusti-tests/tests/verify/pass/pledges/callers.rs). To see the error reporting in action, check out the test cases for invalid pledges [here](https://github.com/frececroka/prusti-dev/blob/lorenz/thesis/prusti-tests/tests/parse/ui/pledges-callees-fail.rs) and [here](https://github.com/frececroka/prusti-dev/blob/lorenz/thesis/prusti-tests/tests/parse/ui/pledges-callers-fail.rs).

There are still some things that are not implemented. Most notably:

+ Shared re-borrows of mutable references.
+ Re-borrows in loops.
+ The `after_expiry_if` pledges (necessary for type invariants, as far as I know).
+ Quantifiers don't work within pledges. Supporting them in general requires a new feature in Viper. Supporting them for pledges that only talk about a single input and output reference would in theory be possible with some special handling.
+ The way copy assignments are modeled requires a workaround in order to be compatible with pledges. But this workaround falls short in certain circumstances, for example when incrementing a reference returned by a re-borrowing function, like `*px += 1`.
+ When a function calls two re-borrowing functions, the expiration tools for these two calls will collide in the names they use.


prusti-interface/src/environment/polonius_info.rs
```diff
@@ -429,24 +447,41 @@ fn add_fake_facts<'a, 'tcx: 'a>(
...
+                    };
+                let lhs_regions = lhs_places.iter()
+                    .filter_map(|p| place_regions.for_place(p));
+                for lhs_region in lhs_regions {
+                    let loan = new_loan();
+                    reference_moves.push(loan);
+                    borrow_region.push((lhs_region, loan, point));
+                    debug!("Adding generic: _ {:?} {:?} {:?}",
+                        lhs_region, location, loan);
+                }
```
frececroka: Fake loans for assignments are created here. Right now it requires that the LHS is a local variable. For the following assignment:

    _2 = (move _3, move _4),

it lists all places below the `_2` place that have region information available -- because `_2` is a tuple, the places below `_2` are `_2.0` and `_2.1`. It then creates a fake loan for every one of these places that has a region attached.

For an assignment that does not involve tuples, like this:

    _2 = move _3,

this should work just as before. It creates a single fake loan for the region of the local variable on the LHS.

What I don't understand is why this new fake loan is not associated with the region of the variable on the RHS that is moved into the LHS.

prusti-interface/src/environment/borrowck/regions.rs
```diff
@@ -4,33 +4,88 @@
...
+                    _ => unreachable!()
+                };
+                (local, fields)
+            },
+            _ => unreachable!("unsupported place: {:?}", place)
+        }
     }
+}
+
+pub fn load_place_regions(path: &Path) -> io::Result<PlaceRegions> {
```
frececroka: I changed this to load the regions not only for local variables, but also for places inside tuples. So if there is a local variable like this:

    let _5: (&'6rv mut i32, &'7rv mut i32),

then it remembers that the place `_5.0` has region 6 and the place `_5.1` has region 7. Internally, places are represented as a local variable plus a vector of field projections (because I can't construct the `mir::Place` without type information), but the `PlaceRegions` struct (that stores the region information) provides a way to query the region for a given `mir::Local` or `mir::Place`.

Note that this only works for tuples that are not nested, i.e., a type like this is not supported:

    let _5: (&'6rv mut i32, (&'7rv mut i32, &'8rv mut i32)).

prusti-filter/src/validators/procedure_validator.rs
```diff
@@ -58,14 +58,7 @@ impl<'a, 'tcx: 'a> CommonValidator<'a, 'tcx> for ProcedureValidator<'a, 'tcx> {
 
     fn check_inner_ty(&mut self, ty: ty::Ty<'tcx>, span: Span) {
         skip_visited_inner_type_variant!(self, &ty.sty);
-
         self.check_ty(ty, span);
-
-        match ty.sty {
-            ty::TypeVariants::TyRef(..) => partially!(self, span, "uses reference-typed fields"),
```
frececroka: I just allow everything that uses reference-typed fields, but eventually this should be changed to only allow references inside tuples.

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -1029,25 +1065,56 @@ impl<'a, 'tcx: 'a> PoloniusInfo<'a, 'tcx> {
...
-                ref x => {
-                    debug!("Borrow starts at statement {:?}", x);
-                    None
-                }
+        let location = self.loan_position[loan];
+        let assignment = self.get_assignment_for_loan(*loan)?;
+        let (dest, source) = assignment.as_assign().unwrap();
+        let dest = dest.clone();
+        let source = source.clone();
+        Some(LoanPlaces { dest, source, location })
```
frececroka, line 1073: Implementing `get_loan_places` is a bit more complicated when there are tuples. This is because an assignment like

    _3 = (move _4, move _5)

creates two loans on the same line (if both fields of `_3` are references), let's call them `L0` and `L1`. This means we can't just inspect the assign statement to figure out the loan places for `L0` -- we also have to find out which field of `_3` `L0` is associated with. This happens in `get_assignment_for_loan`, which returns an atomic assignment for a given loan. For the example above, `get_assignment_for_loan(L0)` would be `_3.0 = move _4` and `get_assignment_for_loan(L1)` would be `_3.1 = move _5`. Using these atomic assignments, we can simply read off the loan places as before.

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -1029,25 +1065,56 @@ impl<'a, 'tcx: 'a> PoloniusInfo<'a, 'tcx> {
...
+        });
+
+        if assignments.len() > 1 {
+            unreachable!(
+                "Cannot determine assignment for loan {:?}, because there are multiple possible options: {:?}",
+                loan, assignments);
         }
+
+        assignments.pop()
+    }
```
frececroka: The idea here is to split the assignment that created a loan, for example

    _3 = (move _4, move _5)

into its atomic parts, which are

    _3.0 = move _4
    _3.1 = move _5

in this case. The LHS of the atomic assignment that created the loan must have a region that requires the terms of the loan to be enforced. If there is exactly one such atomic assignment, we can return it. If there are more (I'm not sure whether this can happen), we panic.

I'm not sure whether this works in general.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3112,21 +3123,36 @@ impl<'p, 'v: 'p, 'r: 'v, 'a: 'r, 'tcx: 'a> ProcedureEncoder<'p, 'v, 'r, 'a, 'tcx
...
+                let region = self.polonius_info().place_regions.for_place(&blocker).unwrap();
+                // TODO: Check if it really is always start and not the mid point.
+                let start_point = self.polonius_info()
+                    .get_point(location, facts::PointType::Start);
+                let (all_loans, zombie_loans) = self.polonius_info()
+                    .get_all_loans_kept_alive_by(start_point, region);
+                let mut new_package_stmts = self.encode_expiration_of_loans(
+                    all_loans, &zombie_loans, location, None);
+                package_stmts.append(&mut new_package_stmts);
+            }
```
frececroka: This is not done yet.

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -1029,25 +1065,56 @@ impl<'a, 'tcx: 'a> PoloniusInfo<'a, 'tcx> {
...
-                    debug!("Borrow starts at statement {:?}", x);
-                    None
-                }
+        let location = self.loan_position[loan];
+        let assignment = self.get_assignment_for_loan(*loan)?;
+        let (dest, source) = assignment.as_assign().unwrap();
+        let dest = dest.clone();
+        let source = source.clone();
+        Some(LoanPlaces { dest, source, location })
+    }
```
frececroka: Implementing `get_loan_places` is a bit more complicated when there are tuples. This is because an assignment like

    _3 = (move _4, move _5)

creates two loans on the same line (if both fields of `_3` are references), let's call them `L0` and `L1`. This means we can't just inspect the assign statement to figure out the loan places for `L0` -- we also have to find out which field of `_3` `L0` is associated with. This happens in `get_assignment_for_loan`, which returns an atomic assignment for a given loan. For the example above, `get_assignment_for_loan(L0)` would be `_3.0 = move _4` and `get_assignment_for_loan(L1)` would be `_3.1 = move _5`. Using these atomic assignments, we can simply read off the loan places as before.

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -1029,25 +1065,56 @@ impl<'a, 'tcx: 'a> PoloniusInfo<'a, 'tcx> {
...
-                    debug!("Borrow starts at statement {:?}", x);
-                    None
-                }
+        let assignment = self.get_assignment_for_loan(*loan)?;
+        let (dest, source) = assignment.as_assign().unwrap();
+        let dest = dest.clone();
+        let source = source.clone();
+        let location = self.loan_position[loan];
+        Some(LoanPlaces { dest, source, location })
+    }
```
frececroka: Implementing `get_loan_places` is a bit more complicated when there are tuples. This is because an assignment like

    _3 = (move _4, move _5)

creates two loans on the same line (if both fields of `_3` are references), let's call them `L0` and `L1`. This means we can't just inspect the assign statement to figure out the loan places for `L0` -- we also have to find out which field of `_3` `L0` is associated with. This happens in `get_assignment_for_loan`, which returns an atomic assignment for a given loan. For the example above, `get_assignment_for_loan(L0)` would be `_3.0 = move _4` and `get_assignment_for_loan(L1)` would be `_3.1 = move _5`. Using these atomic assignments, we can simply read off the loan places as before.

prusti-interface/src/environment/borrowck/regions.rs
vakaras: Sounds good.

By the way, I think it would be good to add such comments to the code so that it is easier to understand it.

prusti-interface/src/environment/mir_utils.rs
```diff
@@ -5,3 +5,142 @@
...
+        if location.statement_index < block.statements.len() {
+            Some(&block.statements[location.statement_index])
+        } else {
+            None
+        }
+    }
+}
+
+pub trait AllPlaces<'tcx> {
+    /// Returns all places that are below the given local variable. Right now, this only handles tuples. For a local variable `_2: u32`, `Place::Local(_2).all_places(&mir) == [_2]`. For a local variable `_2: (u32, u32)`, `Place::Local(_2).all_places(&mir) == [_2, _2.0, _2.1]`.
```
vakaras: It would be nice to keep the lines under 100 characters long. Most IDEs have support for automatic word wrap. For example, VS Code: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap.

prusti-interface/src/environment/polonius_info.rs
vakaras: > What I don't understand is why this new fake loan is not associated with the region of the variable on the RHS that is moved into the LHS.

Could you elaborate what do you expect to get and what you are actually getting?

prusti-interface/src/environment/polonius_info.rs
vakaras: Could you please add this comment to the code?

vakaras (commented):
The approach looks good. Could you please add the comments you posted on Github as code comments? Also, it would be good to format the code comments to be no more than 100 characters because otherwise it is really painful to read them on Github.

Thanks!

Aurel300:
Closing as very out of date and probably not the approach we'll take.

PR #58, opened 2020-06-15
Jakob/marker trait contract refinement

f4z3r:
## Trait Contract Refinement

In some cases, marker traits simply modify the behavior of methods in their super-traits. For instance consider the `PartialEq<T>` and `Eq` traits. In order to consider this additional behavior for verification, traits support contract refinement on trait level:

```rust
pub trait PartialEq<Rhs: ?Sized = Self> {
    #[ensures="<partial equivalence formulas>"]
    fn eq(&self, other: &Rhs) -> bool;
}

#[refine_ensures(PartialEq::eq="self.eq(&self)")]
pub trait Eq: PartialEq<Self> {}
```

Thus, any client implementing `Eq` on a custom type can take advantage of the additional semantics of the total equivalence. Similarly `#[refine_requires]` can be used to refine the precondition of a super-trait.

> Such trait refinement is not scoped. Therefore, considering the previous example, implementing `Eq` on a type implies that the total equivalence contract is always considered on the type, irrespective of whether `Eq` is in scope or not.

### DIFF
--- a/prusti-interface/src/parser.rs
+++ b/prusti-interface/src/parser.rs
@@ -1366,11 +1366,11 @@ impl<'tcx> SpecParser<'tcx> {
         let preconditions: Vec<_> = specs
             .clone()
             .into_iter()
-            .filter(|spec| spec.typ == SpecType::Precondition)
+            .filter(|spec| spec.typ.is_precondition())
             .collect();
         let postconditions: Vec<_> = specs
             .into_iter()
-            .filter(|spec| spec.typ == SpecType::Postcondition)
+            .filter(|spec| spec.typ.is_postcondition())
             .collect();
         let spec_set = SpecificationSet::Procedure(preconditions.clone(), postconditions.clone());
 
@@ -1473,7 +1473,7 @@ impl<'tcx> SpecParser<'tcx> {
                 let spec_trees: Vec<TokenTree> = token_stream.trees().collect();
                 if spec_trees.len() == 1 {
                     spec_trees[0].clone()
-                } else if spec_trees.len() == 5 && attribute.path.to_string().starts_with("refines_") {
+                } else if spec_trees.len() == 5 && attribute.path.to_string().starts_with("refine_") {
                     let trait_symbol = if let TokenTree::Token(_, token::Token::Ident(ref ident, _)) = spec_trees[0] {
                         ident.name.clone()
                     } else {
--- a/prusti-interface/src/parser.rs
+++ b/prusti-interface/src/parser.rs
@@ -1475,6 +1475,8 @@ impl<'tcx> SpecParser<'tcx> {
                 if spec_trees.len() == 1 {
                     spec_trees[0].clone()
                 } else if spec_trees.len() == 5 && attribute.path.to_string().starts_with("refine_") {
+                    // len 5 because 5 inner tokens in #[refine_...(Trait::method="contract")]
+                    //                                              111112233333345555555555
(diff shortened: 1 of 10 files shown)


f4z3r:
Finalized contract refinement for marker traits. This has the limitation that scoping is not supported. In other words, the refined contract is considered irrespective of whether the sub trait is in scope or not. This is caused by a limitation in the design of Prusti, as each AST item can only be annotated by a single specification. For scoping to work, the specification would need to be annotated to the trait. However, this would only allow a single specification refinement per trait, and only when no invariant is defined on the trait.

This limitation can be removed with a slightly larger change in the way annotations are performed, and how SpecIDs are defined (one would need a SpecID that allows to refer to AST nodes). As this is just a prototype implementation for now, I did not go through the effort of such a major change.

f4z3r:
@fpoli

fpoli:
Thanks! I'll look at this tomorrow

fpoli:
(closed by mistake)

prusti-interface/src/parser.rs
```diff
@@ -1428,14 +1472,50 @@ impl<'tcx> SpecParser<'tcx> {
             TokenTree::Delimited(_, Delimited { delim: token::DelimToken::Paren, ref tts } ) => {
                 let token_stream: TokenStream = tts.clone().into();
                 let spec_trees: Vec<TokenTree> = token_stream.trees().collect();
-                if spec_trees.len() != 1 {
+                if spec_trees.len() == 1 {
+                    spec_trees[0].clone()
+                } else if spec_trees.len() == 5 && attribute.path.to_string().starts_with("refine_") {
```
fpoli, line 1477: Could you add a short comment to document why `5` is a special length?

prusti-interface/src/parser.rs
```diff
@@ -2162,15 +2257,34 @@ impl<'tcx> Folder for SpecParser<'tcx> {
...
+                    let mut new_trait_items = vec![];
                     for trait_item in trait_items.into_iter() {
                         match trait_item.node {
                             ast::TraitItemKind::Method(..) => {
-                                new_trait_items.extend(self.rewrite_trait_item_method(trait_item));
+                                let method_symbol = trait_item.ident.name.clone();
+                                let func_ref = (trait_symbol.clone(), method_symbol);
+                                let reg = self.register.lock().unwrap();
+                                let func_attrs = reg.get_attrs_refine(&func_ref);
+                                drop(reg);
```
fpoli: I would prefer to add a scope instead of having to call `drop`. Would this work?
```
let func_attrs = {
    let reg = self.register.lock().unwrap();
    reg.get_attrs_refine(&func_ref)
};
```

fpoli:
Nice! Could you also add to the description of the PR a markdown paragraph that explains the new annotations? Such that I can then add it to the tutorial in the Prusti wiki as soon as we merge the PR.

f4z3r:
> Nice! Could you also add to the description of the PR a markdown paragraph that explains the new annotations? Such that I can then add it to the tutorial in the Prusti wiki as soon as we merge the PR.

Ok I added a short paragraph in the description, and fixed the other suggestions. Let me know if the description is enough.

Note that we do not perform "Selective Replacement" as for normal trait refinement from Matthias Erdin (actually, I am pretty sure this is not the case in the code anyways, in the code at least supplementation is implemented, if not even conditional strengthening), but the implementation here performs the "supplementing" strategy (same as for invariants).

PR #61, opened 2020-06-16
Use reborrows_direct directly.

frececroka:
This is something I came across by chance, and I think using `reborrows_direct` is much easier to understand. The performance hit should be negligible.


fpoli:
Really nice, thanks! @vakaras would you like to give it a look? If not, I'll merge as it looks good to me.

PR #62, opened 2020-06-16
WIP: parser for the new compiler version

karlosss:
Do not merge, meant only for discussion.


new/prusti-specs/src/specifications/untyped.rs
```diff
@@ -95,6 +92,15 @@ impl AssignExpressionId<AssertionKind> for common::AssertionKind<(), syn::Expr,
         use common::AssertionKind::*;
         match self {
             Expr(expr) => Expr(expr.assign_id(spec_id, id_generator)),
+            And(vec_assertions) => {
+                let mut v = vec![];
```
vakaras: This, I think would be cleaner:
```rust
And(assertions) => And(assertions.into_iter().map(|assertion| Assertion { kind: ... }).collect())
```

new/prusti-specs/src/specifications/untyped.rs
```diff
@@ -147,6 +153,11 @@ impl EncodeTypeCheck for Assertion {
             AssertionKind::Expr(expression) => {
                 expression.encode_type_check(tokens);
             }
+            AssertionKind::And(vec_assertions) => {
```
vakaras: It would be more consistent with the rest of Prusti to use `assertions` instead of `vec_assertions`.

new/prusti/src/specs/typed.rs
```diff
@@ -47,6 +47,10 @@ impl StructuralToTyped<AssertionKind> for json::AssertionKind {
         use json::AssertionKind::*;
         match self {
             Expr(expr) => AssertionKind::Expr(expr.to_typed(typed_expressions)),
+            And(vec_assertions) => {
+                // ???
+                AssertionKind::And()
```
vakaras: The following should do the job:
```rust
AssertionKind::And(assertions.into_iter().map(|assertion| assertion.to_typed(typed_expressions))
```

karlosss: This fails with
```
expected struct `std::vec::Vec`, found struct `std::iter::Map`
```
Does it mean that I should change the enum field in `json::AssertionKind` accordingly?

vakaras: I forgot `.collect()` after `.map(...)`.

PR #66, opened 2020-06-18
Move VIR to prusti-common crate

juliand665:
Work on the server mode has (all but) necessitated this change, along with some others. Since the server should not depend on compiler parts, it can't use prusti-interface or prusti-viper, so VIR and some shared logic is being moved to a dedicated crate. There are several notable changes:
- The Makefile has changed a bit to support macOS, have consistent formatting, and let you customize the arguments it passes to cargo (which is useful e.g. when configuring an IDE extension).
- There is a new crate, prusti-common, which is a home for things used all over prusti (including some things that were previously placed in prusti-interface instead, like `config` and `report`).
- You'll notice lots of formatting changes. My IDE is set to format on save, and I'm using the latest (-ish) nightly release of rustfmt, so some parts have received noticeably better formatting. I figured, since we're making sweeping changes already, why not reformat the code I touch now while we break all automatic merges anyway.

I'm pretty sure everything should be fine (unit tests will confirm or deny that of course).

Some details I felt were worth bringing up:
- `borrows::Borrow` is no longer a type alias to `facts::Loan`, since that crate dependency is inverted (prusti-interface now depends on prusti-common). Instead, I essentially copied the `Loan` type to VIR, however I of course could not implement `polonius_engine::Atom` for it this way, so now both types exist, along with `Loan` implementing `Into<Borrow>`.
- Because this is pre-rust-1.41.0 (and even before the corresponding unstable feature), the orphan rule is so strict that we can't e.g. implement `From<Loan>` for `Borrow` or `CheckNoOpAction` for `Vec<Action>`, so some things are a bit awkward right now, like the `convert_loans_to_borrows` free function or the `ActionVec` newtype. Once we update our rust version, we should be able to get rid of those workarounds.
- By virtue of VIR being in a different crate, some impls adding functions to VIR types had to be re-expressed as a trait and its implementation, most notably in `prusti_viper::encoder::foldunfold::permissions`.


fpoli (approved):

fpoli:
Thanks for summarizing the changes in the description of the PR.

To the team: `prusti_common` is meant to contain **compiler-agnostic** utilities used across the Prusti project.

PR #70, opened 2020-06-19
Fix the indentation of the first item in the path context state.

frececroka:
It's a tiny change ;)

Before, it looked like this:

    acc: {
        _0.tuple_0.val_ref: write,
      _0.tuple_0: write,
      _0.tuple_1.val_ref: write,
      _0.tuple_1: write,
    }
    pred: {
    
    }

Now, it looks like this:

    acc: {
      _0.tuple_0.val_ref: write,
      _0.tuple_0: write,
      _0.tuple_1.val_ref: write,
      _0.tuple_1: write,
    }
    pred: {
    
    }


fpoli:
Nice, thanks! If you need to debug the state, you might also find useful the new `FOLDUNFOLD_STATE_FILTER` config flag, which filters the dump of the state in the `graphviz_method_during_foldunfold/*.dot` files.

PR #76, opened 2020-06-23
Test branches with slashes.

frececroka:
The single `*` only matches branches and tags without slashes.

See https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions.


fpoli:
Thanks!

PR #79, opened 2020-06-23
Test programs from RustHorn's benchmark

fpoli:
In this PR:
* Add a `tests/core_proof` folder, to encode and verify only the core proof (that is, Prusti will skip unsupported functions and will not check absence of panics).
* Add to `tests/core_proof` the examples taken from [RustHorn's benchmark](https://github.com/hopv/rust-horn/tree/a72fc0dd4caa556fed65a0bee244033bb552d74e/toplas2021/benchmarks/rust-horn). Some of the examples take minutes to verify (e.g. programs with many branches or loops), so we should remove them from the test suite, to keep CI's time low.


fpoli:
Some programs use `&mut &mut T` types, which are not yet supported in Prusti. I'll remove them.

fpoli:
@frececroka The examples in which I added `ignore-test` (https://github.com/viperproject/prusti-dev/pull/79/commits/9f4b6a11295fea384691bd6c4d09c4e6a41b18a1) might be interesting to you, as they use tuples of references.

frececroka:
Thanks, I'll have a look. Examples for tuples with references are great!

Another general thought I had, because you mentioned that some examples take a long time to verify. Maybe it is useful to have two CI actions, one that runs some quick tests and finishes within 10-15 minutes, and another one that runs the full test suite and takes much longer. This way we'd get fast feedback if something obvious is broken, even if some tests take longer to verify.

fpoli:
> Another general thought I had, because you mentioned that some examples take a long time to verify. Maybe it is useful to have two CI actions, one that runs some quick tests and finishes within 10-15 minutes, and another one that runs the full test suite and takes much longer. This way we'd get fast feedback if something obvious is broken, even if some tests take longer to verify.

Yes, that would be very useful! I think we can internally split the "Test" workflow in a "build" and two dependant "test" and "quick-test" Github jobs, which Github should run in parallel. The quick-test could be just `TESTNAME=quick/ cargo test -p prusti`, such that it only runs a selection of the tests -- those that we moved to a "quick" subfolder, like `tests/verify/pass/quick`. Feel free to open a PR if you wish; I'm currently busy on another task.

PR #80, opened 2020-06-23
Datatype equality

cmatheja:
This PR adds basic support for datatype equality. For example, the following program now verifies:

```rust

#[derive(Clone,PartialEq,Eq)]
struct A {
    i: i32,
}

#[pure]
fn get_value(x: &A) -> i32 {
    x.i 
}


#[requires="x == y"]
#[ensures="result == 2*get_value(x)"]
fn test_eq_propagation(x: &A, y: &A) -> i32 {
    get_value(x) + get_value(y)
}


```


prusti-common/src/vir/optimisations/methods/purifier.rs
```diff
@@ -419,8 +419,8 @@ impl ast::StmtFolder for VarPurifier {
                 ast::Type::Int => "builtin$havoc_int",
                 ast::Type::Bool => "builtin$havoc_bool",
                 ast::Type::TypedRef(_) => "builtin$havoc_ref",
-            }
-            .to_string();
+                ast::Type::Domain(_) => unimplemented!(),
```
vakaras: Is it really `unimplemented`, or should it be `unreachable`?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -543,6 +575,24 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
         vir::ExprIterator::conjoin(&mut conjuncts.into_iter())
     }
 
+    /*
```
vakaras: Why this commented out function is here?

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -115,7 +115,16 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> TypeEncoder<'p, 'v, 'r, 'a, 'tcx> {
                 vir::Type::TypedRef(type_name)
             }
 
-            ty::TypeVariants::TyAdt(_, _) | ty::TypeVariants::TyTuple(_) => unreachable!(),
+            ty::TypeVariants::TyAdt(_, _) => {
+                let type_name = self.encoder.encode_type_predicate_use(self.ty);
+                match &self.encoder.encode_get_domain_type(type_name) {
+                    None => unimplemented!(),
```
vakaras: Shouldn't this be `unreachable` instead of `unimplemented`?

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -115,7 +115,16 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> TypeEncoder<'p, 'v, 'r, 'a, 'tcx> {
...
+                    Some(domain_type) => domain_type.clone()
+                }
+
+            },
+
+            ty::TypeVariants::TyTuple(_) => unimplemented!(),
```
vakaras: Shouldn't this be `unreachable` instead of `unimplemented`?

vakaras (approved):
LGTM.

cmatheja:
Thanks for the feedback; it's incorporated in the last commit. Once the CI is done, we can merge the PR.

prusti-common/src/vir/ast/expr.rs
```diff
@@ -160,6 +164,18 @@ impl fmt::Display for Expr {
                     .collect::<Vec<String>>()
                     .join(", "),
             ),
+            Expr::DomainFuncApp(ref function, ref args, ref _pos) => write!(
+                f,
+                "{}({})",
+                function,
```
fpoli, line 170: `function.name`? I think this prints the full function signature; I'm not sure if this is what you wanted.

prusti-common/src/vir/to_viper.rs
```diff
@@ -382,7 +383,24 @@ impl<'v> ToViper<'v, viper::Expr<'v>> for Expr {
...
                 )
-            }
+            },
+            &Expr::DomainFuncApp(
+                ref function,
+                ref args,
+                ref _pos,
+            ) => {
+                ast.domain_func_app(
+                    function.to_viper(ast),
```
fpoli, line 393: By calling `function.to_viper(ast)` we build an AST node for a domain function *definition*  each time we apply the function. It currently works, but only because Viper doesn't see that there are multiple definitions for the same domain function.

When I modified `func_app` to take a String, in commit 2215efa72f13f75bd14d6ef4eb4dc890dfb04e2e, I didn't care about also modifying `domain_func_app` as we were not using it, but I think it would be better to do so.

This change would remove the `function` argument from `domain_func_app`, and would add arguments for the name of the domain function, the name of the domain, and the type of the result.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1256,11 +1342,125 @@ impl<'v, 'r, 'a, 'tcx> Encoder<'v, 'r, 'a, 'tcx> {
...
+                        },
+                        vir::Type::Domain(_) => unreachable!(),
+                        t=> t,
+                    }
+                })
+                .collect();
+
+            let mirror_function = vir::DomainFunc {
+                name: format!("mirror${}", // nasty hack as Viper does not like :: in functions
+                              self.env.get_item_name(def_id)).replace("::", "_"),
```
fpoli, line 1444: What about using `self.encode_item_name(def_id)` instead of `self.env.get_item_name(def_id)`?

prusti-common/src/vir/to_viper.rs
cmatheja: That makes sense. I was wondering why func_app domain_func_app take different arguments (except a Function vs. a DomainFunc) anyway.

prusti-viper/src/encoder/encoder.rs
cmatheja: Good point, thanks!

prusti-common/src/vir/ast/expr.rs
cmatheja: Right. This needs to change anyway as DomainFuncApp should mirror the signature of FuncApp.

prusti-common/src/vir/to_viper.rs
cmatheja: @fpoli 
Your suggested change of ```ast.domain_func_app```would also require a change of the automatically generated file ```DomainFuncApp_object_wrapper.rs``` as ```call_apply``` takes a domain function (```Lviper/silver/ast/DomainFunc```) as its first argument. Can you give me a pointer where this needs to be changed in jni-gen? Thanks!

fpoli: I think there is already a constructor in `DomainFuncApp_wrapper.rs`

PR #85, opened 2020-06-29
Fix forall in parser, typecheck forall, add some testcases

karlosss:


new/prusti-interface/src/specs/typed.rs
```diff
@@ -42,6 +44,24 @@ impl StructuralToTyped<Expression> for json::Expression {
     }
 }
 
+impl StructuralToTyped<TriggerSet> for json::TriggerSet {
+    fn to_typed(self, typed_expressions: &HashMap<String, rustc_hir::BodyId>) -> TriggerSet {
+        TriggerSet {
+            0: vec![]
```
vakaras: `TriggerSet(Vec::new())`

new/prusti-specs/src/specifications/untyped.rs
```diff
@@ -169,6 +170,9 @@ impl AssignExpressionId<Assertion> for common::Assertion<(), syn::Expr, Arg> {
 /// A trait for encoding the statements for type-checking the spec.
 pub trait EncodeTypeCheck {
     fn encode_type_check(&self, tokens: &mut TokenStream);
+    fn encode_type_check_any_type(&self, tokens: &mut TokenStream) {
```
vakaras: Could you add doc comments to both methods and also explain why it is fine to have an explicit panic here?

new/prusti-specs/src/specifications/untyped.rs
```diff
@@ -169,6 +170,9 @@ impl AssignExpressionId<Assertion> for common::Assertion<(), syn::Expr, Arg> {
...
+        unimplemented!();
```
vakaras: `unimplemented!` means that something is not yet implemented. Is this really the case, or should it be `unreachable!`?

new/prusti/tests/pass/parse/composite.rs
```diff
@@ -8,49 +7,49 @@
...
-#[requires(true && forall(|a: i32| a == 5))]
+#[requires(true && forall(|a: i32| a == 5) && true && true && true && true && true && true && true && true && true && true)]
 fn test11() {}
 
-#[requires(forall(|a: i32| a == 5) && true)]
+#[requires(forall(|a: i32| a == 5) && true && true && true && true && true && true && true && true && true && true && true && true)]
 fn test12() {}
 
-#[requires(true ==> forall(|a: i32| a == 5) ==> true)]
+#[requires(true ==> forall(|a: i32| a == 5) ==> true && true && true && true && true && true && true && true && true && true && true && true && true)]
```
vakaras: This looks pretty terrible. Could you please add code that removes UUIDs before sorting? It would be best to put it under a command-line option.

vakaras (changes requested):

new/prusti-specs/src/specifications/untyped.rs
karlosss: I introduced this because of two types of expressions: one of them has to be bools and the other one not (triggers) so i need two ways to encode; the second one is relevant only for expressions (i am not sure if it will always be the case), therefore i picked a default implementation

new/prusti/tests/pass/parse/composite.rs
karlosss: i will, what do you mean by "It would be best to put it under a command-line option."? like i need to run `cargo test --some-option`?

new/prusti-specs/src/specifications/untyped.rs
vakaras: Actually, I thought a bit and I think it would be cleaner to have a separate type for trigger instead of `Expression` so that you can implement a custom `encode_type_check` and avoid `encode_type_check_any_type`.

new/prusti/tests/pass/parse/composite.rs
vakaras: At the beginning of the test file, you have a special comment that contains the additional command-line options:
```rust
 // compile-flags: -Zprint-desugared-specs -Zprint-typeckd-specs -Zskip-verify
```

These are parsed [here](https://github.com/viperproject/prusti-dev/blob/296891d28e007edf7ce1f4f5701aa382b3378cf8/new/prusti/src/bin/prusti-driver.rs#L35-L49)

new/prusti-interface/src/specs/typed.rs
karlosss: does not work: `expected function, tuple struct or tuple variant, found type alias `TriggerSet``

vakaras: How about `Self(Vec::new())`?

karlosss: expected struct `prusti_specs::specifications::common::TriggerSet`, found struct `prusti_specs::specifications::json::TriggerSet`

vakaras: Ok.

PR #86, opened 2020-06-29
Split test CI action in quick/full test.

frececroka:
This would split the test action into two jobs. One is only testing the examples from `prusti/tests/verify/pass/quick/`, the other one is running the full test suite (Including the examples from `prusti/tests/verify/pass/quick/`, they should probably be excluded). How useful this is depends on how representative the quick tests can be, i.e., how confident we are that the changes are correct after the quick tests pass.

Another option would be to do the split the other way round, i.e., keep the test suite mostly as it is today and move the tests that take too long to a separate job for long-running tests.

Still another way would be to split the tests into multiple categories, for example

- hand rolled test that test specific aspects of the encoding, meaning the ones from `prusti/tests/verify/pass/expiring-loans`, `prusti/tests/verify/pass/equality`, and so on.
- RustHorn tests
- Rosetta tests
- Crate tests
- ...

This way we actually get faster results because the different categories run in parallel (and we see results from categories that finish quickly early).

Building Prusti once and reusing the result for all test jobs is possible, but not trivial. The issue is that two jobs don't share the filesystem. The only way to share files is via the `upload-artifact` and `download-artifact` actions. Just uploading/downloading the whole `target/` folder is not viable, because it is way too big. We'd have to only upload/download the individual test binaries.

### DIFF
--- a/.github/workflows/prepare-environment.sh
+++ b/.github/workflows/prepare-environment.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+sudo apt-get install -y build-essential pkg-config gcc libssl-dev openjdk-8-jdk
+
+echo "::set-env name=LD_LIBRARY_PATH::"$(dirname "$(find "$JAVA_HOME" -name "libjvm.so")")""
+
+wget -q "http://viper.ethz.ch/downloads/ViperToolsNightlyLinux.zip"
+unzip ViperToolsNightlyLinux.zip -d viper_tools
+rm ViperToolsNightlyLinux.zip
+echo "::set-env name=VIPER_HOME::$(pwd)/viper_tools/backends/"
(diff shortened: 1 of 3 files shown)


fpoli:
Thanks! I didn't know that Github jobs don't share the file system. Uploading the Prusti binaries, Z3 and Viper jars as artifact sounds like something we'll want to do in the future. At that point, we could move the tests to a separate `prusti-tests` crate to compile and test it without recompiling the whole Prusti.

frececroka:
Factoring out `prusti-test` sounds like a good idea. This would make it very easy to have a common build phase and then multiple subsequent test phases. In the meantime, we can symlink some tests into the `quick/` folder to make the quick test useful. I already added the `basic.rs` tests in PR https://github.com/viperproject/prusti-dev/pull/89.

PR #90, opened 2020-07-01
 Fix forall in parser, typecheck forall, add some testcases

karlosss:


new/prusti-interface/src/specs/mod.rs
```diff
@@ -17,11 +17,16 @@ struct SpecItem {
     specification: JsonAssertion,
 }
 
+pub struct Node {
+    pub hir_id: rustc_hir::hir_id::HirId,
+    pub body_id: rustc_hir::BodyId
```
vakaras: Is `BodyId` useful? If not, it would be better to remove it completely.

new/prusti-interface/src/specs/typed.rs
```diff
@@ -1,39 +1,42 @@
...
+pub trait StructuralToTyped<'tcx, Target> {
+    fn to_typed(self, typed_expressions: &HashMap<String, Node>, tcx: TyCtxt<'tcx>) -> Target;
 }
 
-impl StructuralToTyped<Expression> for json::Expression {
-    fn to_typed(self, typed_expressions: &HashMap<String, rustc_hir::BodyId>) -> Expression {
-        let body_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)];
+impl<'tcx> StructuralToTyped<'tcx, Expression> for json::Expression {
+    fn to_typed(self, typed_expressions: &HashMap<String, Node>, tcx: TyCtxt<'tcx>) -> Expression {
+        let body_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)].body_id;
```
vakaras: Is `body_id` here sufficient to obtain MIR from `mir_validated`? If not, we need to fix that.

new/prusti-interface/src/specs/typed.rs
```diff
@@ -42,33 +45,77 @@ impl StructuralToTyped<Expression> for json::Expression {
...
+
+impl<'tcx> StructuralToTyped<'tcx, ForAllVars<'tcx>> for json::ForAllVars {
+    fn to_typed(self, typed_expressions: &HashMap<String, Node>, tcx: TyCtxt<'tcx>) -> ForAllVars<'tcx> {
+        let hir_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)].hir_id;
+        let local_id = tcx.hir().local_def_id(hir_id);
+        let (body, _) = tcx.mir_validated(local_id);
+        let body = body.borrow();
+
+        let mut vars = vec![];
+        for local_decl in body.local_decls.iter() {
```
vakaras: Shouldn't this be traversal over closure arguments only and not all variables defined in the closure?

vakaras (changes requested):
Looks mostly good.

new/prusti-interface/src/specs/mod.rs
karlosss: yes, the `ToStructure` code for expressions needs that.

new/prusti-interface/src/specs/typed.rs
karlosss: This piece of code is originally written by (probably) you, I thought it is just correct. If not, good that I know it, I will test it.

new/prusti-interface/src/specs/typed.rs
karlosss: yeah, but how to get those? looking at the debug-printed body, this is probably the least obscure way to do that...

karlosss: I tried with `body.args_iter()`, but that does not work, since even the count of variables obtained from the parser differs from the `arg_count` of the body. So I'd keep this as is, at least for now.

new/prusti-interface/src/specs/typed.rs
```diff
@@ -1,74 +1,124 @@
...
+pub trait StructuralToTyped<'tcx, Target> {
+    fn to_typed(self, typed_expressions: &HashMap<String, HirId>, tcx: TyCtxt<'tcx>) -> Target;
 }
 
-impl StructuralToTyped<Expression> for json::Expression {
-    fn to_typed(self, typed_expressions: &HashMap<String, rustc_hir::BodyId>) -> Expression {
-        let body_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)];
+impl<'tcx> StructuralToTyped<'tcx, Expression> for json::Expression {
+    fn to_typed(self, typed_expressions: &HashMap<String, HirId>, tcx: TyCtxt<'tcx>) -> Expression {
+        let hir_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)];
```
vakaras: Wouldn't it be better to store `LocalDefId` in `typed_expressions`, which would allow avoiding the `tcx` argument here? (Visitor is created from `tcx`, so it seems more natural to me to have the argument there.)

new/prusti-interface/src/specs/typed.rs
vakaras: Why it differs? It should be the same.

vakaras (changes requested):

new/prusti-interface/src/specs/typed.rs
karlosss: but then you need the `tcx` to get the `Body`: `body = tcx.mir_validated(local_def_id)`

vakaras: Yes, but that is the third piece of code which needs `tcx` anyway.

new/prusti-interface/src/specs/typed.rs
karlosss: If it is a bit more readable, this works:
```rust
let mut vars = vec![];
for local_decl in body.local_decls.iter() {
    if local_decl.is_user_variable() {
        vars.push(local_decl.ty.kind.clone())
    }
}
```
However, [this iterator](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/mir/mod.rs.html#295-300) does not: it sometimes misses a variable.

karlosss: or maybe this is even a bit more compact:
```rust
        let vars: Vec<TyKind> = body
            .local_decls
            .iter()
            .filter(|x| x.is_user_variable())
            .map(|x| x.ty.kind.clone())
            .collect();
```

new/prusti-interface/src/specs/typed.rs
karlosss: how would this work then?
```rust
fn to_typed(self, typed_expressions: &HashMap<String, HirId>, tcx: TyCtxt<'tcx>) -> ForAllVars<'tcx> {
    let hir_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)];
    let local_id = tcx.hir().local_def_id(hir_id);
    let (body, _) = tcx.mir_validated(local_id);
```
copying the thick `Body` does not seem like a good idea to me...

new/prusti-interface/src/specs/typed.rs
vakaras: But this sounds wrong. You need arguments, not just variables.

new/prusti-interface/src/specs/typed.rs
vakaras: True. However, I think that it would still be better to store `LocalDefId` instead of `HirId`.

new/prusti-interface/src/specs/typed.rs
vakaras: Are there more or fewer arguments than expected?

new/prusti-interface/src/specs/typed.rs
karlosss: That's something we can probably do. I will change it.

new/prusti-interface/src/specs/typed.rs
karlosss: The iterator misses some, i.e. reports 0 instead of 1. And I think that `local_decls` holds them all, doesn't it?

new/prusti-interface/src/specs/typed.rs
vakaras: Thanks!

new/prusti-interface/src/specs/typed.rs
vakaras: `local_decls` contains arguments, variables, and the return value while we need only arguments.

vakaras: Looking at the `args_iter` [implementation](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/mir/mod.rs.html#334-337), it should return the exact number of closure arguments. If it does not, then something is wrong with the encoding.

karlosss: Apparently it is not the case, when the `assert!(self.count == body.arg_count)` (i.e. check the number of variables from the parser == the number of arguments to the body) fails.

vakaras: Basically, my point is that if `args_iter` does not return what we expect, we have to understand why.

vakaras: > Apparently it is not the case, when the `assert!(self.count == body.arg_count)` (i.e. check the number of variables from the parser == the number of arguments to the body) fails.

What is the exact difference?

karlosss: > Basically, my point is that if `args_iter` does not return what we expect, we have to understand why.

I am just trying to investigate this. Will let you in a few moments once I find something out.

karlosss: So APPARENTLY (i.e. not confirmed by looking into the source code) the first argument is the closure itself. This holds:
```rust
assert!(body.arg_count-1 == self.count);
```
and this works:
```rust
let mut vars = vec![];
for arg in body.args_iter() {
    let decl = body.local_decls.get(arg).unwrap();
    if !decl.is_user_variable() {
        continue;
    }
    vars.push(decl.ty.kind.clone());
}
```
before I had a wrong iterator in the code... \*facepalm\*

vakaras: Looks good. Could you also add an assert that `vars.len() == self.count`?

karlosss: This one is already there, that is the most basic one.

new/prusti-interface/src/specs/mod.rs
```diff
@@ -4,6 +4,8 @@ use rustc_hir::intravisit;
 use rustc_middle::hir::map::Map;
 use rustc_middle::ty::TyCtxt;
 use rustc_span::Span;
+use rustc_hir::hir_id::HirId;
```
vakaras: Is this import still necessary?

new/prusti-interface/src/specs/typed.rs
```diff
@@ -1,74 +1,127 @@
 use prusti_specs::specifications::common;
 use prusti_specs::specifications::json;
 use rustc_hir::BodyId;
+use rustc_hir::def_id::LocalDefId;
+use rustc_middle::ty::{TyCtxt, TyKind};
 use std::collections::HashMap;
+use rustc_hir::hir_id::HirId;
```
vakaras: Is this import still necessary?

new/prusti-interface/src/specs/typed.rs
```diff
@@ -1,74 +1,127 @@
...
+
+impl<'tcx> StructuralToTyped<'tcx, ForAllVars<'tcx>> for json::ForAllVars {
+    fn to_typed(self, typed_expressions: &HashMap<String, LocalDefId>, tcx: TyCtxt<'tcx>) -> ForAllVars<'tcx> {
+        let local_id = typed_expressions[&format!("{}_{}", self.spec_id, self.expr_id)];
+        let (body, _) = tcx.mir_validated(local_id);
+        let body = body.borrow();
+
+        let vars: Vec<TyKind> = body
+            .args_iter()
+            .skip(1)
```
vakaras, line 81: Could you add a comment that explains why we skip the first argument?

vakaras (changes requested):

vakaras (approved):

PR #112, opened 2020-07-20
Add else branch to Viper if statements.

frececroka:
This adds the option to have else branches in the `vir::If` AST nodes, which will be very useful for my work. The problem is that I don't know to what extent this affects other parts of the code, especially the fold/unfold algorithm. The tests are green, but this is not too surprising because it is not used anywhere right now.

Let me know if you think this change is useful.

P.S.: A related change that would also be useful is to have else-if branches, i.e., `if (e1) { ... } elseif (e2) { ... } else { ... }`.


vakaras (approved):
LGTM

PR #114, opened 2020-07-21
Support "if" statement in foldunfold/semantics.rs.

frececroka:
This PR extends the support for "if" statements further. Now, I define the semantics of if statements w.r.t. the fold/unfold algorithm. The basic idea is that applying an if statement `if (e) { b1 } else { b2 }` in a fold/unfold state `s` can result in one of two possible states:
- First, we could end up in the state `s1` that is the result of applying `b1` in the state `s` (this happens when `e` evaluates to true).
- Second, we could end up in the state `s2` that is the result of applying `b2` in the state `s` (this happens when `e` evaluates to false).

Because we don't know statically what `e` will evaluate to, the result `s'` of applying the whole if statement in state `s` is the greatest lower bound of `s1` and `s2`. Concretely, this means the access permissions we hold for a place `p` in `s'` is the minimum of what we hold in `s1` and of what we hold in `s2`. The predicate permissions for `s'` are computed analogously. The places that are considered moved permanently in `s'` are the places that are moved permanently in `s1` or `s2` (i.e., the set union).

The properties `framing_stack` and `dropped` are reset to their default empty values. I'm not sure if this is the right thing to do.

In general I do not know if my changes are actually doing the right thing, though my current understanding of the fold/unfold algorithm makes me think they do. If the changes seem sensible I would be happy if they are merged now, otherwise we can postpone merging until I can confirm that they actually work correctly. But it will take some time until I can confirm this, because there are other issues I have to work on first in order to be able to test it.

### DIFF
--- a/prusti-viper/src/encoder/foldunfold/log.rs
+++ b/prusti-viper/src/encoder/foldunfold/log.rs
@@ -15,6 +15,7 @@ use prusti_common::vir;
 use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::rc::Rc;
+use std::sync::RwLock;
 
 // Note: Now every PathCtxt has its own EventLog, because a Borrow no longer unique
 // (e.g. we duplicate the evaluation of the loop condition in the encoding of loops).
@@ -25,7 +26,7 @@ pub(super) struct EventLog {
     /// Actions performed by the fold-unfold algorithm before the join. We can use a single
     /// CfgBlockIndex because fold-unfold algorithms generates a new basic block for dropped
     /// permissions.
-    prejoin_actions: HashMap<vir::CfgBlockIndex, Rc<Vec<Action>>>,
+    prejoin_actions: HashMap<vir::CfgBlockIndex, Rc<RwLock<Vec<Action>>>>,
 
     /// A list of accessibility predicates for which we inhaled `Read`
     /// permission when creating a borrow and original places from which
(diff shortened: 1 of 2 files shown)


fpoli:
The greatest_lower_bound that you describe should be what PathCtxt::join already does (plus something more). Could you check whether there is code to be reused?

frececroka:
Ah thanks for pointing me to this. So it looks like this is somewhat more complicated than I initially anticipated. In this case I will postpone this for now.

frececroka:
Naturally, the last attempt was more or less nonsense ;)

I now handle `if` statements differently. When `FoldUnfold::replace_stmt` is invoked with an `if` statement, two clones of the current path context are produced, one called `then_pctxt` for the `then` branch, one called `else_pctxt` for the `else` branch. Then `replace_stmt` is called for every statement of the `then` branch using the cloned path context `then_pctxt`. The same is done for the statements of the `else` branch. Finally, `then_pctxt` and `else_pctxt` are joined using `FoldUnfold::prepend_join` and the current path context is set to the joined path context.

At least for my use case, this seems to work fine.

prusti-viper/src/encoder/foldunfold/log.rs
```diff
@@ -25,7 +26,7 @@ pub(super) struct EventLog {
     /// Actions performed by the fold-unfold algorithm before the join. We can use a single
     /// CfgBlockIndex because fold-unfold algorithms generates a new basic block for dropped
     /// permissions.
-    prejoin_actions: HashMap<vir::CfgBlockIndex, Rc<Vec<Action>>>,
+    prejoin_actions: HashMap<vir::CfgBlockIndex, Rc<RwLock<Vec<Action>>>>,
```
frececroka, line 29: This is necessary because after a `PathCtxt` is cloned, the `Rc` forbids write access (because now there are two `Rc`s pointing to the same memory). Is the `Rc` still the right thing to have (instead of a plain `Vec<Action>`)? The way it is now the `then_pctxt` and the `else_pctxt` have the same `prejoin_actions`, which does seem odd.

PR #120, opened 2020-08-11
Support Mac in x.py

Aurel300:
`JAVA_HOME` is detected using `/usr/libexec/java_home`, and the library is a `.dylib` rather than an `.so`. I also added `DYLD_LIBRARY_PATH` in addition to `LD_LIBRARY_PATH`, since the former is what Mac uses for configuring the dynamic linker. It seems to work with only the latter, but that might be a different problem.

I also wonder if we should update:

 - [ ] `Makefile` - remove it?
 - [ ] `docker/artefact.Dockerfile` - does a `make release` on L12
 - [ ] `evaluation/artifact/preparation.rst` - `make release` (this file is quite old as well)
 - [ ] `README.md` - `make run RUN_FILE=path/to/the/program_to_be_verified.rs`

### DIFF
--- a/x.py
+++ b/x.py
@@ -57,6 +57,35 @@ def get_linux_env():
     return variables
 
 
+def get_mac_env():
+    """Get environment variables for Mac."""
+    java_home = get_var_or('JAVA_HOME', None)
+    if java_home is None:
+        java_home = subprocess.run(["/usr/libexec/java_home"], stdout=subprocess.PIPE).stdout.strip()
+    variables = [
+        ('JAVA_HOME', java_home),
+        ('RUST_TEST_THREADS', '1'),
+    ]
+    if os.path.exists(java_home):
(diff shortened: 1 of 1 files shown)


vakaras (approved):
Thanks!

vakaras:
> I also wonder if we should update:

I agree that we should fix these too. Also, we should enable CI for Mac OS now.

Aurel300:
> I agree that we should fix these too. Also, we should enable CI for Mac OS now.

Sounds good. Was it disabled in a particular commit? Or should I just copy the `rustc-update` job in the GH workflow?

vakaras:
We never had Mac OS CI on GitHub.

fpoli:
> * `Makefile` - remove it?

I think yes. It's much easier to maintain the `x.py` cross-platform than the `Makefile`.

PR #121, opened 2020-08-11
Delete unused files in prusti{,-interface}/src/

FabianWolff:
It is quite confusing to have all these files lying around that are actually dead code. Also, it is unnecessary to keep them; that's what Git is there for.

This pull request therefore deletes a bunch of files that aren't even used when building Prusti.


Aurel300:
I think `prusti-macros` can also be completely removed then. `prusti-filter` may be a candidate too.

FabianWolff:
> I think `prusti-macros` can also be completely removed then. `prusti-filter` may be a candidate too.

I think you're right. I have removed these two directories as well and adjusted several `Cargo.toml` files as necessary.

cmatheja:
`prusti-filter` should indeed be removed eventually.

Most of the mentioned files have become dead code due to the recent compiler upgrade. Since we still have quite a few regressions, it may be preferable to keep the files for now (as we may have to include some of the code again - at least for debugging) and remove them once all tests pass again.

@vakaras Would it be convenient for you to keep the files for now? Otherwise I'm fine with merging this.

prusti-interface/src/trait_register.rs
```diff
@@ -1,246 +0,0 @@
-// © 2020, ETH Zurich
-//
-// This Source Code Form is subject to the terms of the Mozilla Public
-// License, v. 2.0. If a copy of the MPL was not distributed with this
```
vakaras: @cmatheja If I am not mistaken, you are supervising @jakobbeckmann, right? I think it would be better to fix the broken functionality instead of deleting it.

vakaras (commented):
I think we should fix `prusti-interface/src/trait_register.rs` instead of deleting it. Otherwise, I am fine with merging this.

FabianWolff:
> I think we should fix `prusti-interface/src/trait_register.rs` instead of deleting it. Otherwise, I am fine with merging this.

I noticed that it was commented out in `prusti-interface/src/lib.rs`, and thus was not compiled, and therefore I assumed that it is no longer needed. I've restored that file now; it still isn't compiled, but you can fix it and then add it to `lib.rs` again.

prusti-interface/src/trait_register.rs
cmatheja: Right, I'll give it a look. Thanks for spotting this.

vakaras (approved):
LGTM

PR #122, opened 2020-08-11
Enable CI on Mac

Aurel300:
(draft for now to see if CI passes)

### DIFF
--- a/prusti/src/bin/prusti-rustc.rs
+++ b/prusti/src/bin/prusti-rustc.rs
@@ -64,8 +64,13 @@ fn process(mut args: Vec<String>) -> Result<(), i32> {
         let prusti_contracts_path =
             get_latest_crate_artefact(&prusti_home, "prusti_contracts", "rlib");
         cmd.arg(format!("prusti_contracts={}", prusti_contracts_path));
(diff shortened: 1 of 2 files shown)


Aurel300:
The failure seems to stem from the hard-coded `so` extension here:

https://github.com/viperproject/prusti-dev/blob/9a57e3d78d2539072d1b341efd34cfdb997fbadd/prusti/src/bin/prusti-rustc.rs#L68

prusti/src/bin/prusti-rustc.rs
```diff
@@ -64,8 +64,13 @@ fn process(mut args: Vec<String>) -> Result<(), i32> {
         let prusti_contracts_path =
             get_latest_crate_artefact(&prusti_home, "prusti_contracts", "rlib");
         cmd.arg(format!("prusti_contracts={}", prusti_contracts_path));
+        let dylib_extension = if cfg!(target_os = "macos") {
+            "dylib"
```
vakaras, line 68: Isn't there a function in the standard library that returns the correct extension?

vakaras: Unfortunately, I could not find any.

vakaras (approved):
LGTM

PR #123, opened 2020-08-11
Add generic linux option to x.py

karlosss:


x.py
```diff
@@ -11,6 +11,14 @@
 dry_run = False
 
 
+def default_java_loc():
```
vakaras: Could you rename this to `default_linux_java_loc` since it is Linux specific?

vakaras (changes requested):

x.py
karlosss: Sure.

vakaras (approved):
LGTM

PR #124, opened 2020-08-12
Prusti dev guide

Aurel300:
[**Link to rendered book**](http://media.thenet.sk/aca/prusti-dev-guide/)

Here is a rough initial version of a "developer guide". It mostly focuses on walking through the steps of verifying a file.

- [x]  Does this supersede `design`? That book is mostly empty but it seems the contents would be roughly the same.
- [x] Do we want a `doc` directory for non-code things?

Things that might be good to add:

 - [x] Configuration options
 - [ ] How to contribute? Walkthrough of adding a simple Rust feature into Prusti-supported code.
 - [ ] Duplicate appendix A of the publication, re-structure and update based on features that were not described in the publication yet.
 - [ ] Rough description of current (student) projects and their progress.
 - (feel free to suggest/add more here)


dev-guide/src/intro.md
```diff
@@ -0,0 +1,5 @@
+# Introduction
+
+This is the Prusti dev guide, intended to make Prusti more approachable for new contributors.
+
+> Direct links to code are provided in boxes like this. These links are based on the [`9ca9cd1b`](https://github.com/viperproject/prusti-dev/commit/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af) commit, to allow linking to specific lines and functions. When reading through the code using this guide, make sure to check if the code has since changed.
```
vakaras: Maybe, instead of naming the exact commit, you could just say that to make the links more robust they are pointing to some specific commit, but the reader should check that the code in question has not changed since the last version. This way, everyone who contributes to the guide does not need to go over all of it and update the links to an updated version.

vakaras: An even better way maybe would be to run rustdoc on Prusti, publish the generated docs as GitHub pages and then link to the generated docs.

dev-guide/src/pipeline/summary.md
```diff
@@ -0,0 +1,26 @@
+# Verification pipeline
+
+At a high level, Prusti is a tool that converts Rust code enriched with Prusti specifications into Viper code, verifies the code with an external verifier, and then reports the resutsl back to the user.
```
vakaras:
```suggestion
At a high level, Prusti is a tool that converts Rust code enriched with Prusti specifications into [Viper code](https://viper.ethz.ch), verifies the code with an external verifier, and then reports the results back to the user.
```

dev-guide/src/pipeline/summary.md
```diff
@@ -0,0 +1,26 @@
...
+
+This section summarises the steps that take place when the user runs Prusti on a given Rust file. The steps are described in greater detail in the subsequent chapters. Although we present the steps in separate "stages", this distinction only exists for the purposes of this guide, and is not clearly mirrored in the codebase.
```
vakaras:
```suggestion
This section summarises the steps that take place when the user runs Prusti on a given Rust file. The steps are described in greater detail in the subsequent chapters. Although we present the steps in separate “stages”, this distinction only exists for the purposes of this guide, and is not clearly mirrored in the codebase.
```

dev-guide/src/pipeline/viper.md
```diff
@@ -0,0 +1,20 @@
+# Viper verification stage
+
+- (With Prusti server only) Send VIR to the server.
+- VIR is encoded into Viper.
+- Viper verifier is called and the results are obtained.
+- (With Prusti server only) Send verification results back to the client.
+
+## Prusti server
+
+[Prusti server](https://github.com/viperproject/prusti-dev/pull/43) is an optional component of Prusti that can significantly reduce compilation times by running a background process. The background process keeps an instance of JVM open, which is what Viper backends use to perform verification of Viper code. With the server enabled, a client only needs to send VIR to the server and receive the results once they are ready.
```
vakaras:
```suggestion
[Prusti server](https://github.com/viperproject/prusti-dev/pull/43) is an optional component of Prusti that can significantly reduce verification times by running a background process. The background process keeps an instance of JVM open, which is what Viper backends use to perform verification of Viper code. With the server enabled, a client only needs to send VIR to the server and receive the results once they are ready.
```

dev-guide/src/layout.md
```diff
@@ -0,0 +1,52 @@
...
+ - [`prusti-contracts-internal/`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-contracts-internal) - a procedural macro crate that forwards calls to `prusti-specs`.
+ - [`prusti-specs/`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-specs) - does the specification rewriting.
+ - [`prusti-contracts-test`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-contracts-test) – a minimal test that checks that `prusti-contracts` can be used with a regular compiler.
+
+## Common library code
+
+These crates contain the majority of Prusti's code.
+
+ - [`prusti-common/`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-common) - common modules used across Prusti (code independent from Rust internals).
+   - [`src/vir/`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-common/src/vir) - Viper IR.
```
vakaras:
```suggestion
   - [`src/vir/`](https://github.com/viperproject/prusti-dev/tree/9ca9cd1b9bcfd9870691fa5a7a957a90987ba4af/prusti-common/src/vir) - VIR.
```

I would prefer to use *VIR* everywhere instead of *Viper IR* because even though VIR got its name as an acronym of Viper Intermediate Representation, VIR is going further and further away from the Viper AST.

vakaras (changes requested):
> Does this supersede `design`? That book is mostly empty but it seems the contents would be roughly the same.

Yes, I think `design` can be deleted. The only thing there that could be useful is the `dependency_graph.py` script that generates the Prusti component diagram as a graphviz, which we could include in the documentation to give an overview of Prusti structure.

> Do we want a doc directory for non-code things?

Probably it would make sense to use this name.

> Things that might be good to add:

I agree. For student projects, we maybe could just link to the page in our group website that lists the student projects. Otherwise, I am afraid that the list will be permanently outdated.

> (feel free to suggest/add more here)

I think we should also document:

1.  Testing and what [special flags](https://github.com/viperproject/prusti-dev/blob/c7908f9bc27c0038ad8b3e1b11db6ccd0c2819df/prusti-common/src/config.rs#L15-L26) are supported by Prusti for that and how to use them.
2.  The various [configuration options](https://github.com/viperproject/prusti-dev/blob/c7908f9bc27c0038ad8b3e1b11db6ccd0c2819df/prusti-common/src/config.rs#L34-L68), especially the ones that help with debugging Prusti.

cmatheja:
Thanks, that looks quite useful.

> Things that might be good to add:

We may want to add a few examples of Rust code as well as sanitized versions of the corresponding MIR, VIR, and Viper code. This does not replace a full documentation of the translation but is probably easier to keep up to date.

doc/dev-guide/src/encoding/types.md
```diff
@@ -0,0 +1,128 @@
...
+```viper
+// with ENCODE_UNSIGNED_NUM_CONSTRAINT
+predicate u32(self: Ref) {
+  acc(self.val_u32) && 0 <= self.val_u32
+}
+```
+
+## Composite types
+
+There are no classes in Viper (every `Ref` instance has all fields), so non-primitive Rust types are encoded using [predicates](http://viper.ethz.ch/tutorial/?page=1&section=#predicates) and `Ref`s.
```
cmatheja: This is correct, but we somewhat have (or will have) two representations for composite types:

1. a `Ref` variable and a predicate specifying all the fields, and 
2. a "snapshot", which essentially encodes the composite type in a deep value using Viper domains.

The second representation is independent of the Viper heap; it can thus, for example, be returned by a Viper function, which is not possible using the first approach.

fpoli:
Hi Aurel! :wave:

We also have some documentation in the [wiki](https://github.com/viperproject/prusti-dev/wiki). We could move part (or even all) of the wiki to the dev guide.

Aurel300:
@fpoli Good idea. I added content from the "Home" and "Development Setup" pages. The build setup is now described with `x.py` in mind.

The main `README` should also be updated once the dev guide has a proper (github.io) URL.

fpoli:
We also have some instructions in the `CONTRIBUTING.md` file. We could move those into the guide, leaving a link to a relevant page of the guide.

PR #129, opened 2020-08-16
Loop body invariant

vakaras:
Rename `invariant!` to `body_invariant!`.

Motivation:
* The loop invariant in Prusti has unusual semantics that it is only a body invariant. I think naming the macro `body_invariant` makes this point clearer.
* Prusti has not only loop, but also object and trait invariants. While the former are implemented as function-like macros `macro!(...)`, the latter are attribute macros `#[macro(...)]` and, as a result, need to be named differently. Therefore, my proposal is to call the loop invariants `body_invariant!(...)` and trait invariants `#[trait_invariant(...)]`. The additional advantage of having unique names is that it is going to be easier for the users to find the relevant parts of the documentation (once we have it).

cc @karlosss @fpoli @cmatheja @Aurel300

### DIFF
--- a/prusti-contracts-impl/src/lib.rs
+++ b/prusti-contracts-impl/src/lib.rs
@@ -34,6 +34,6 @@ pub fn trusted(_attr: TokenStream, tokens: TokenStream) -> TokenStream {
 }
 
 #[proc_macro]
-pub fn invariant(_tokens: TokenStream) -> TokenStream {
+pub fn body_invariant(_tokens: TokenStream) -> TokenStream {
     (quote! { () }).into()
 }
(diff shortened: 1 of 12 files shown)


Aurel300:
Seems good to clarify this. Any outstanding issues/TODOs you can think of? The [loop invariant section in the tutorial](https://github.com/viperproject/prusti-dev/wiki/tutorial#loop-invariants) should be updated after this is merged.

This looks like a breaking change but I don't know how seriously we treat these.

vakaras:
> Any outstanding issues/TODOs you can think of?

No.

> The loop invariant section in the tutorial should be updated after this is merged.

The entire tutorial needs to be updated to match the new specification syntax.

> This looks like a breaking change but I don't know how seriously we treat these.

If we do this now, it is not a breaking change because we just switched to a new way of parsing specifications and this has not reached our users yet.

cmatheja:
I agree with the renaming - this way we could also reintroduce standard loop invariants at some later point without another breaking change (the new loop encoding).

PR #131, opened 2020-08-18
[WIP] Start working on closure support

FabianWolff:
Vytautas suggested to open a pull request with my changes so far, so that we don't run into conflicts. Everything is still work in progress, closure support doesn't actually work yet, but all of the tests should pass.

I would have run `cargo fmt` to get the coding style corrected, but this incurred _many_ additional, unrelated changes, so I refrained, although I suggest that it be run at some point so that contributors can use it without creating a huge diff.


prusti-specs/src/lib.rs
```diff
@@ -128,3 +128,150 @@ pub fn body_invariant(tokens: TokenStream) -> TokenStream {
         }
     }
 }
+
+/** Unlike the functions above, which are only called from
+ *  prusti-contracts-internal, this function also needs to be called
```
vakaras: In Rust, the convention is to use `///` for doc comments.

prusti-interface/src/environment/mod.rs
```diff
@@ -157,7 +162,13 @@ impl<'tcx> Environment<'tcx> {
         let tcx = self.tcx;
         let mut visitor = CollectPrustiSpecVisitor::new(self);
         tcx.hir().krate().visit_all_item_likes(&mut visitor);
-        visitor.get_annotated_procedures()
+
+        let mut cl_visitor = CollectClosureDefsVisitor::new(self);
```
vakaras, line 166: Am I correct, that you are verifying closure bodies without taking into account the state of the captured variables? That is, verifying closures as if they were functions with the state struct passed as an additional argument about which we know nothing?

prusti-specs/src/lib.rs
```diff
@@ -128,3 +128,150 @@ pub fn body_invariant(tokens: TokenStream) -> TokenStream {
...
+ *  drop it (for -impl).
+ */
+pub fn closure(tokens: TokenStream, drop_spec: bool) -> TokenStream {
+    let mut requires: Vec<TokenTree> = vec! [];
+    let mut ensures: Vec<TokenTree> = vec! [];
+    let mut cl: Option<syn::ExprClosure> = None;
+
+    let mut iter = tokens.clone ().into_iter();
+
+    while let Some(tok) = iter.next() {
```
vakaras: The parsing code should be either in `prusti-dev/prusti-specs/src/specifications/preparser.rs` or in `prusti-dev/prusti-specs/src/specifications/untyped.rs`. Also, is there any reason why you are using the tokens directly and not the higher level parsing API provided by the `syn` crate?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4805,7 +4808,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
             }
 
             &mir::AggregateKind::Closure(def_id, _substs) => {
-                assert!(self.encoder.is_spec_closure(def_id), "closure: {:?}", def_id);
+                //assert!(self.encoder.is_spec_closure(def_id), "closure: {:?}", def_id);
                 // Specification only. Just ignore in the encoding.
                 // FIXME: Filtering of specification blocks is broken, so we need to handle this here.
                 Vec::new()
```
vakaras, line 4817: Shouldn't this be `unimplemented!()`?

prusti/src/bin/prusti-driver.rs
```diff
@@ -57,6 +57,7 @@ fn main() {
     args.push("-Zdump-mir-dir=log/mir/".to_owned());
     args.push("-Zdump-mir=renumber".to_owned());
     args.push("-Zalways-encode-mir".to_owned());
+    args.push("-Zcrate-attr=feature(stmt_expr_attributes)".to_owned());
```
vakaras: Why do you need this? Relying on unstable features can cause serious maintenance problems.

prusti-specs/src/lib.rs
FabianWolff: Fixed.

prusti-interface/src/environment/mod.rs
FabianWolff: Well, yes, how else are you going to do it if a) closures are encoded as functions in the MIR and b) the MIR is the basis for verification in Prusti?

This doesn't make the approach less powerful, however, only a little more verbose. Consider this (using informal syntax):
```Rust
let x = 42;
let cl = /* ensures: result == 42 */ || x;
```
This won't verify, because we do indeed lose information about the captured state. However, we can define a view:
```Rust
let x = 42;
let cl =
    // view: capt_x: i32 = x
    // ensures: result == capt_x
    || x;
assert_eq!(cl(), 42);
```
This will verify (once everything is implemented, I mean), because we can prove the specification separately _and_ use the knowledge about the captured state available at the instantiation-site to deduce the actual return value. I suppose you could even give an invariant:
```Rust
let x = 42;
let cl =
    // view: capt_x: i32 = x
    // invariant: capt_x == 42
    // ensures: result == 42
    || x;
```
Then we can prove the original postcondition, although of course the only interesting aspect about this "invariant" is that it is established; we get preservation for free.

prusti-specs/src/lib.rs
FabianWolff: But I'm not parsing a specification (well, sort of, but not in the sense of `preparser.rs`), I'm parsing the macro _containing_ the specification, so I think putting it into `prusti-specs/src/specifications/` would not be appropriate.

And I _am_ using the higher-level API when parsing the closure, the rest is just reading in the pre- and postconditions.

prusti-viper/src/encoder/procedure_encoder.rs
FabianWolff: No. Arguably, maybe there needs to be something else there, but it's not "unimplemented", it's in the process of being implemented. If I put `unimplemented!()` there, I'll get a panic every time I run Prusti on some code that has a non-spec-only closure in it, which is precisely the feature I'm trying to implement.

Or did you mean I should _temporarily_ put `unimplemented!()` there for this pull request and remove it on my fork again immediately afterwards?

prusti-interface/src/environment/mod.rs
vakaras: The implementation of closures in Nagini used to essentially inline the closure bodies into the parent function so that they could take the captured state into account. However, your approach with views also looks reasonable. Thanks for explaining it!

prusti/src/bin/prusti-driver.rs
FabianWolff: > Why do you need this? Relying on unstable features can cause serious maintenance problems.

Well, so can linking against the nightly compiler. But you see, I have to annotate closures with specifications, and closures are expressions, hence the feature that allows me to add annotations to expressions. Specifically, I'm currently using the feature like this when rewriting the `closure!()` macro:
```Rust
#[prusti::post_spec_id_ref = "3fa334f2e58746f98899c3e81532413f"]
|a: i32, b: i32| { ... }
```
But if this is a problem, feel free to suggest an alternative solution!

prusti-specs/src/lib.rs
vakaras: > I think putting it into `prusti-specs/src/specifications/` would not be appropriate.

You could also put it in a separate file. If I am not mistaken, this function is the only one that does parsing in this file and, therefore, does not really fit.

> And I am using the higher-level API when parsing the closure, the rest is just reading in the pre- and postconditions.

I would still suggest defining something like:
```rust
struct ClosureWithSpecs {
  pres: ...,
  posts: ...,
  closure: ...,
}
```
and implementing `Parse` for it because it results in significantly more readable and maintainable code.

prusti-viper/src/encoder/procedure_encoder.rs
vakaras: Ideally, for this PR it should be something like this:
```suggestion
              if self.encoder.is_spec_closure(def_id) {
                // Specification only. Just ignore in the encoding.
                // FIXME: Filtering of specification blocks is broken, so we need to handle this here.
                Vec::new()
                } else {
                  unimplemented!();
                }
```

prusti/src/bin/prusti-driver.rs
vakaras: The specifications are also desugared into closures and adding attributes on them does not seem to cause problems. Maybe, something like this would work:
```rust
{
  #[prusti::post_spec_id_ref = "3fa334f2e58746f98899c3e81532413f"]
  let cl = |a: i32, b: i32| { ... };
  cl
}
```

prusti-specs/src/lib.rs
FabianWolff: Fixed.

prusti/src/bin/prusti-driver.rs
FabianWolff: > The specifications are also desugared into closures and adding attributes on them does not seem to cause problems. Maybe, something like this would work:
> 
> ```rust
> {
>   #[prusti::post_spec_id_ref = "3fa334f2e58746f98899c3e81532413f"]
>   let cl = |a: i32, b: i32| { ... };
>   cl
> }
> ```

Unfortunately not:
```
thread 'rustc' panicked at 'not implemented: Closure(DefId(0:7 ~ closure_true[317d]::test[0]::{
{closure}}[0]), [i8, extern "rust-call" fn((i32, i32)) -> i32, ()])', prusti-viper/src/encoder/
procedure_encoder.rs:4719:22
```
Seems like it has something to do with the assignment (which we're going to have to implement at some point, anyways, but we're not there yet).

vakaras: > which we're going to have to implement at some point, anyways, but we're not there yet

Correct me if I am wrong, but you will have to implement assignment support before you can do anything with closures because the Rust compiler always emits temporary variables and many assignments.

FabianWolff: > Correct me if I am wrong, but you will have to implement assignment support before you can do anything with closures because the Rust compiler always emits temporary variables and many assignments.

Yes, but so far, I have not yet succeeded in verifying the closure _bodies_, which I decided would be my first goal. And for that, you don't need assignments (unless you have higher-order closures or something).

FabianWolff: Then again, it doesn't work currently anyway, so I dropped the experimental feature for this PR in my latest commit.

PR #135, opened 2020-08-24
Tuples with references.

frececroka:
This implements references inside tuples. At least with the old compiler version, the tests were green.

Not everything is supported:

- They cannot be nested inside of each other, like `(&mut u32, (&mut u32, &mut u32))`.
- They cannot be passed as arguments to a function.
- They can be returned from a function, but receiving them from a function probably does not work.

To see what is supported, have a look at the two test cases.


prusti/tests/verify/pass/mut-borrows/refs-in-tuples-1.rs
```diff
@@ -0,0 +1,19 @@
+fn main() {
```
vakaras: Could you move these tests to `prusti/tests/pass/parse/` so that they are executed by CI?

prusti-interface/src/environment/mir_utils/all_places.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+impl<'tcx> AllPlaces<'tcx> for mir::Local {
+    fn all_places(self, tcx: ty::TyCtxt<'tcx>, mir: &mir::Body<'tcx>) -> Vec<mir::Place<'tcx>> {
+        let mut places = vec![self.as_place()];
+        let ty = mir.local_decls[self].ty;
+        if let ty::TyKind::Tuple(types) = ty.kind {
+            for (i, ty) in types.iter().enumerate() {
+                let field = mir::Field::new(i);
+                // TODO: Can expect_ty() fail?
```
vakaras: It should be safe to call `expect_ty` here.

prusti-interface/src/environment/mir_utils/as_place.rs
```diff
@@ -0,0 +1,12 @@
+use rustc_middle::mir;
+use rustc_middle::ty;
+
+pub trait AsPlace<'tcx> {
+    fn as_place(&self) -> mir::Place<'tcx>;
```
vakaras: Couldn't you use `into` instead? For example:
```rust
let x: mir::Local = ...
let y: mir::Place = x.into();
```

prusti-interface/src/environment/mir_utils/place_add_field.rs
```diff
@@ -0,0 +1,19 @@
+use std::iter::once;
+
+use rustc_middle::mir;
+use rustc_middle::ty;
+
+pub trait PlaceAddField<'tcx> {
+    fn field(self, tcx: ty::TyCtxt<'tcx>, field: mir::Field, ty: ty::Ty<'tcx>) -> mir::Place<'tcx>;
```
vakaras: I think [`tcx.mk_place_field`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.mk_place_field) does the same.

prusti-interface/src/environment/mir_utils/place_as_local.rs
```diff
@@ -0,0 +1,16 @@
+use rustc_middle::mir;
+
+pub trait PlaceAsLocal {
+    /// Tries to interpret the given `mir::Place` as a local variable. If this fails, it returns `None`.
+    fn as_local(&self) -> Option<mir::Local>;
```
vakaras: How is this different from [`as_local`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Place.html#method.as_local)?

prusti-interface/src/environment/mir_utils/split_aggregate_assignment.rs
```diff
@@ -0,0 +1,73 @@
...
+            mir::Rvalue::Aggregate(box kind, operands) => {
+                assert_eq!(kind, mir::AggregateKind::Tuple,
+                    "The only supported aggregates are tuples.");
+                let local = lhs.as_local().unwrap();
+                let items_ty = mir.local_decls[local].ty.tuple_items().unwrap();
+                operands.into_iter().zip(items_ty.into_iter())
+                    .enumerate()
+                    .map(|(i, (rhs, ty))| {
+                        let field = mir::Field::new(i);
+                        let lhs = local.as_place().field(tcx, field, ty);
```
vakaras:
```suggestion
                        let lhs = tcx.mk_place_field(local.into(), field, ty);
```

prusti-interface/src/environment/mir_utils/statement_at.rs
```diff
@@ -0,0 +1,17 @@
+use rustc_middle::mir;
+
+pub trait StatementAt<'tcx> {
+    fn statement_at(&self, location: mir::Location) -> Option<&mir::Statement<'tcx>>;
+}
+
+impl<'tcx> StatementAt<'tcx> for mir::Body<'tcx> {
+    fn statement_at(&self, location: mir::Location) -> Option<&mir::Statement<'tcx>> {
+        let block = &self[location.block];
+        if location.statement_index < block.statements.len() {
```
vakaras, line 10:
```suggestion
        if location.statement_index != block.statements.len() {
```
It would be better to panic if the location for some reason is completely off.

prusti-interface/src/environment/mir_utils/tuple_items_for_ty.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// Tries to interpret the given `mir::Ty` as a tuple type. If this succeeds, it returns the
+    /// nested types making up the tuple. If this failes, it returns `None`.
+    fn tuple_items(&self) -> Option<Vec<ty::Ty<'tcx>>>;
+}
+
+impl<'tcx> TupleItemsForTy<'tcx> for ty::Ty<'tcx> {
+    fn tuple_items(&self) -> Option<Vec<ty::Ty<'tcx>>> {
+        if let ty::TyKind::Tuple(items) = self.kind {
+            let types = items.iter()
+                // TODO: Can expect_ty() fail?
```
vakaras: No.

prusti-interface/src/environment/mir_utils/ty_as_ref.rs
```diff
@@ -0,0 +1,18 @@
+use rustc_middle::mir::terminator::Mutability;
+use rustc_middle::ty::Region;
+use rustc_middle::ty::Ty;
+use rustc_middle::ty::TyKind;
+
+pub trait TyAsRef<'tcx> {
+    fn as_ty_ref(&self) -> Option<(Region<'tcx>, Ty<'tcx>, Mutability)>;
```
vakaras: Why the function name does not match the file name?

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -4,24 +4,45 @@
...
+use super::borrowck::facts;
+use super::borrowck::regions;
+use super::loops;
+use super::mir_analyses::initialization::compute_definitely_initialized;
+use super::mir_analyses::initialization::DefinitelyInitializedAnalysisResult;
+use super::mir_analyses::liveness::compute_liveness;
+use super::mir_analyses::liveness::LivenessAnalysisResult;
+use super::procedure::Procedure;
+
+/// This represents the assignment in which a loan was created. The `source` will contain the creation of the loan, while the `dest` will store the created reference.
```
vakaras: Could you make the doc comments no longer than 80 symbols?

vakaras (changes requested):

prusti-interface/src/environment/mir_utils/place_as_local.rs
frececroka: Ah thanks, the previous compiler version didn't have `as_local`.

prusti-interface/src/environment/mir_utils/ty_as_ref.rs
frececroka: I renamed the file to `ty_as_ty_ref.rs`. The file name should convey that a `ty::Ty` is turned into a `ty::TyKind::Ref`. The function name should convey that the receiver (which is a `ty::Ty`) is turned into a `ty::TyKind::Ref`.

vakaras (approved):
LGTM.

PR #138, opened 2020-08-28
Verification of closure bodies

FabianWolff:
With the changes in this pull request, I am able to verify examples like this:
```Rust
#![feature(register_tool)]
#![register_tool(prusti)]

use prusti_contracts::*;

fn test() {
    let _cl = closure! (
        requires (a > b),
        ensures (result > b),
        |a: i32, b: i32| -> i32 { a }
    );
}

fn main() {}
```
If I change the postcondition to, say, `result > a`, I get the appropriate error message:
```
  --> /tmp/closure-fail.rs:10:18
   |
10 |         ensures (result > a),
   |                  ^^^^^^^^^^
   |
```
The closure _must_ have type annotations for now (as in the example), which of course won't work for higher-order closures, but I'd say it's an acceptable restriction for the time being.

I'm admittedly still relying on the `stmt_expr_attributes` feature; I have made some attempts to circumvent this, but it hasn't worked. Patches are welcome if this is an issue.


prusti-interface/src/environment/polonius_info.rs
```diff
@@ -370,6 +370,16 @@ fn load_polonius_facts<'tcx>(
     debug!("Reading facts from: {:?}", dir_path);
     let mut facts_loader = facts::FactLoader::new();
     facts_loader.load_all_facts(&dir_path);
+
+    let all_facts = &facts_loader.facts;
+    all_facts.borrow_region.iter()
+        .for_each(|(origin, loan, point): &(facts::Region, facts::Loan, facts::PointIndex)| -> () {
```
vakaras: I think in this case a simple for loop would be much more readable:
```rust
for (origin, loan, point) in &all_facts.borrow_region {
              trace!("load_polonius_facts: borrow_region: ({:?}, {:?}, {:?})",
                   origin, loan, facts_loader.interner.get_point(*point));
}
```

prusti-specs/src/lib.rs
```diff
@@ -175,10 +175,20 @@ pub fn closure(tokens: TokenStream, drop_spec: bool) -> TokenStream {
             });
         }
 
-        let (spec_toks_pre, spec_toks_post) = rewriter.generate_cl_spec(preconds, postconds);
         let syn::ExprClosure { attrs, asyncness, movability, capture, or1_token,
                                inputs, or2_token, output, body } = cl_spec.cl;
 
+        let output_type: syn::Type = match output {
+            syn::ReturnType::Default => {
+                return syn::Error::new(output.span(), "closure must specify return type, for now")
```
vakaras:
```suggestion
                return syn::Error::new(output.span(), "closure must specify return type")
```

prusti-specs/src/lib.rs
```diff
@@ -187,20 +197,18 @@ pub fn closure(tokens: TokenStream, drop_spec: bool) -> TokenStream {
         quote! {
             {
                 #cl_annotations #attrs_ts
-                let _prusti_closure =
```
vakaras, line 302: Why did you need to change this?

prusti/tests/pass/typecheck/nested_forall.stdout
```diff
@@ -4,13 +4,17 @@
 
 #![feature(register_tool)]
 #![register_tool(prusti)]
+
+
+
+
+
+#![feature(stmt_expr_attributes)]
```
vakaras: So, why is this needed?

prusti/src/bin/prusti-driver.rs
```diff
@@ -57,6 +57,7 @@ fn main() {
     args.push("-Zdump-mir-dir=log/mir/".to_owned());
     args.push("-Zdump-mir=renumber".to_owned());
     args.push("-Zalways-encode-mir".to_owned());
+    args.push("-Zcrate-attr=feature(stmt_expr_attributes)".to_owned());
```
vakaras: Why is this feature strictly necessary?

vakaras (changes requested):

prusti-specs/src/lib.rs
FabianWolff: Because `get_opt_spec_id()` in `prusti-viper/src/encoder/encoder.rs` receives the `DefId` of the function/closure and then checks the attributes associated with that `DefId` in the `TyCtxt`. In particular, this means that if we add the pre-/postcondition annotation to the assignment, as you suggest, then the encoder won't find the closure specification.

prusti/tests/pass/typecheck/nested_forall.stdout
FabianWolff: See my explanation above for why this feature is necessary; I needed to change the `*.stdout` test files because the `#![feature(stmt_expr_attributes)]` line is added to the after-parse output when using `-Zcrate-attr=feature(stmt_expr_attributes)`.

prusti/src/bin/prusti-driver.rs
FabianWolff: I suppose the reason why it works for spec-only closures without this feature is that spec-only closures never get used:
```Rust
#[prusti::spec_only]
#[prusti::expr_id = "34265d3e9c6942f1b164a73d3a866e2c_101"]
|| -> bool { a > b };
```
i.e. the attribute will be added to the closure here, because the statement consists only of the closure, whereas I need to do something with the closure, e.g.
```Rust
#[some_annotation]
let cl = || { ... };
```
meaning the annotation will be added to the _assignment_ rather than the closure (from my understanding; at least it's definitely not added to the closure, I've checked), and so the encoder won't find it.

prusti-specs/src/lib.rs
vakaras: `get_opt_spec_id` is intended for the top-level items. For example, it does not handle loop invariants. Unfortunately, I did not get around to fix them.

vakaras: Do you have a strong reason for this to get merged soon? Otherwise, I would prefer to wait until we fix support for the loop invariants, which should also solve your problem of attaching specs to closures.

FabianWolff: > Do you have a strong reason for this to get merged soon?

No, I just thought it would make sense to open a PR, given that I reached my first milestone (verification of closure bodies), to keep you updated and my branch from diverging too far.

On the other hand, given that "looking for the specification" is a somewhat unrelated problem from closure verification in particular, perhaps whoever fixes the loop invariants could also fix this issue while they're at it? I mean, it's not even "fixing", it's just replacing the `stmt_expr_attributes` solution with a different one.

vakaras: > No, I just thought it would make sense to open a PR, given that I reached my first milestone (verification of closure bodies), to keep you updated and my branch from diverging too far.

Having an open PR definitely helps for everyone to keep track of what is happening.

> On the other hand, given that "looking for the specification" is a somewhat unrelated problem from closure verification in particular, perhaps whoever fixes the loop invariants could also fix this issue while they're at it?

I think it will be much easier to fix the PR where we can clearly see the changed files than to clean it once the code is merged. Hopefully, this will not take too long.

FabianWolff:
Closure calls now also work. For instance, I am able to verify this example:
```Rust
use prusti_contracts::*;

fn _test() {
    let cl = closure! (
        requires (a > b),
        ensures (result > b),
        |a: i32, b: i32| -> i32 { a }
    );

    let a = 5;
    let b = 3;
    let c = cl (a, b);
    assert! (c > b);
}

fn main() {}
```
Changing the closure call to, say, `cl (b, b)` results (correctly) in an error:
```
error: [Prusti: verification error] precondition might not hold.
  --> /tmp/closure.rs:13:13
   |
13 |     let c = cl (b, b);
   |             ^^^^^^^^^
   |
note: the failing assertion is here
  --> /tmp/closure.rs:6:19
   |
6  |         requires (a > b),
   |                   ^^^^^
```
Similarly, changing the `assert!()` to, say, `c > a` also yields an error:
```
error: [Prusti: verification error] statement might panic
  --> /tmp/closure.rs:14:5
   |
14 |     assert! (c > a);
   |     ^^^^^^^^^^^^^^^^
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

fpoli:
To solve the conflicts it should be enough to move the tests from `prusti/tests/pass/parse/` to `prusti-tests/tests/parse/pass/` and from `prusti/tests/pass/typecheck/` to `prusti-tests/tests/typecheck/pass/`.

fpoli:
`prusti/src/bin/prusti-driver.rs` has been moved to `prusti/src/driver.rs`

Aurel300:
Hey Fabian, I have some questions about this.

https://github.com/FabianWolff/prusti-dev/blob/5e91a795ec26e375accf5fea798b7ed08c9034a5/prusti-specs/src/lib.rs#L134-L141

Is it a reasonable assumption that any function containing a closure with a spec will also itself have a specification? In which case, can we not look for closure specifications whenever we are looking at a function through the existing `proc_macro` attributes? That is, can the closure spec parsing code not be invoked from `prusti_specs::requires` or `prusit_specs::ensures`?

---

Somewhat related to the previous question, if we parse closure-specs during `prusti_specs::requires` and `prusti_specs::ensures`, we could do without the `closure!` macro, and specify closures using the usual syntax.

```rust
#[requires(...)] #[ensures(...)] let x = |args| ...;
```

We get the `TokenStream`, so we can rewrite it to whatever we need. I also suspect that we could place the closure-specs directly on the closure (`let x = #[requires(...)] |args| ...`), without enabling `stmt_expr_attributes`, because we can rewrite it to a compatible syntax.

This would unfortunately mean that `-impl` would not be a no-op but it might be convenient anyway.

---

> The closure must have type annotations for now (as in the example), which of course won't work for higher-order closures, but I'd say it's an acceptable restriction for the time being.

Is this because the rewriter produces a spec-only function?

https://github.com/FabianWolff/prusti-dev/blob/5e91a795ec26e375accf5fea798b7ed08c9034a5/prusti-specs/src/rewriter.rs#L174-L182

Can the attribute not be attached to the closure as well? Or is there any other reason for producing a `fn` instead of a closure?

---

Edit: after a bit more thought, actually why not re-use the existing attributes for closure specs (without re-writing them as I suggest or using `closure!`)? Is the issue that it is hard to detect whether the spec is meant for a closure or for a top-level function?

FabianWolff:
> Hey Fabian, I have some questions about this.

Sure!

> https://github.com/FabianWolff/prusti-dev/blob/5e91a795ec26e375accf5fea798b7ed08c9034a5/prusti-specs/src/lib.rs#L134-L141
> 
> Is it a reasonable assumption that any function containing a closure with a spec will also itself have a specification? In which case, can we not look for closure specifications whenever we are looking at a function through the existing `proc_macro` attributes? That is, can the closure spec parsing code not be invoked from `prusti_specs::requires` or `prusit_specs::ensures`?

I don't get it. What does the _containing_ function's pre-/postcondition have to do with the _closure's_ specification? Do you mean that while processing the outer function, we might visit its entire body and look for closures with specifications, and then rewrite those?

> Somewhat related to the previous question, if we parse closure-specs during `prusti_specs::requires` and `prusti_specs::ensures`, we could do without the `closure!` macro, and specify closures using the usual syntax.
> 
> ```rust
> #[requires(...)] #[ensures(...)] let x = |args| ...;
> ```
> 
> We get the `TokenStream`, so we can rewrite it to whatever we need. I also suspect that we could place the closure-specs directly on the closure (`let x = #[requires(...)] |args| ...`), without enabling `stmt_expr_attributes`, because we can rewrite it to a compatible syntax.

I think I see your point, but that sound like a lot of work just for replacing the `closure!()` macro. Just think of examples like this:
```rust
let a: i32 = 42;
let b: i32 = 137;
let c = 1 |b| a;
let d = move |b| a;
d (42);
```
i.e. keeping expressions like the right-hand sides of the assignments to `c` and `d` apart would probably be quite difficult and brittle if we parse everything manually (though by no means impossible, of course, it's just that somebody would have to do it, whereas using `closure!()`, we can use `syn::parse` and be done with it).

> 
> This would unfortunately mean that `-impl` would not be a no-op but it might be convenient anyway.
> 
> > The closure must have type annotations for now (as in the example), which of course won't work for higher-order closures, but I'd say it's an acceptable restriction for the time being.
> 
> Is this because the rewriter produces a spec-only function?
> 
> https://github.com/FabianWolff/prusti-dev/blob/5e91a795ec26e375accf5fea798b7ed08c9034a5/prusti-specs/src/rewriter.rs#L174-L182
> 
> Can the attribute not be attached to the closure as well? Or is there any other reason for producing a `fn` instead of a closure?

Producing a `fn` allowed me to reuse many of the existing specification encoding functions in Prusti, so that's what I went for initially, but you're right, at some point one might want to change this. One problem that occurs when you rewrite into a closure instead of a function is that the closure can "accidentally" capture additional variables it's not supposed to capture, whereas a function will create an error in that case (e.g. if the precondition refers to a variable from the enclosing scope that neither occurs in the body nor is shadowed by the closure arguments).

Aurel300:
Ok, I wasn't entirely familiar with the long story behing Rust's attributes. I think that specs attached directly to the closure seem most natural:

```rust
let x = #[requires(...)] |args| ...;
```

But this seems to require both `#![feature(stmt_expr_attributes)]` and `#![feature(proc_macro_hygiene)]` for now. @vakaras You were opposed to the former at least, in the PR comments. Is it just because you don't want to rely on any unstable features at all? Both seem to be pretty old features by now, I would expect them to stabilise…

By the way, the reason I am asking these questions is because I think it would be great if "abstractions for computation" had a unified specification language. That is:

 - methods
 - closures
 - thread spawning
 - futures/async blocks

All of these have some type of post-condition, some may have pre-conditions (unless inline-defined and immediately invoked).

Aurel300:
Also @FabianWolff just to make sure I'm not interfering with your work – have you tried parsing higher-order specifications yet? E.g.:

```rust
#[requires(arg |= [requires(arg_0 > 0), ensures(result < 0)])]
fn foo(arg: Fn (i32) -> i32) { ... }
```

If not I will try to work on this.

FabianWolff:
> Also @FabianWolff just to make sure I'm not interfering with your work – have you tried parsing higher-order specifications yet? E.g.:
> 
> ```rust
> #[requires(arg |= [requires(arg_0 > 0), ensures(result < 0)])]
> fn foo(arg: Fn (i32) -> i32) { ... }
> ```
> 
> If not I will try to work on this.

I have just pushed my current progress with this. Yes, I am working on it, although I suppose that the encoding part is relatively independent from the parsing of the user-level syntax, so if you want to work on the latter, I suppose you can give it a try. On the other hand, though, there could be some interference with other features that are not yet implemented at all, such as views for the captured state, invariants, `outer()`, etc., so we should at least get in touch e.g. via email and discuss the syntax in some detail.

prusti-specs/src/specifications/common.rs
```diff
@@ -187,6 +187,13 @@ pub struct ForAllVars<EID, AT> {
     pub vars: Vec<AT>,
 }
 
+#[derive(Debug, Clone)]
+pub struct SpecEntVars<EID, AT> {
+    pub spec_id: SpecificationId,
+    pub id: EID,
+    pub vars: Vec<AT>,
+}
```
Aurel300: I wonder if we really need a separate `struct` for this? `TypeCond` also uses `ForAllVars`. Maybe we could just rename the struct to something like `SpecVars` and use it for all three?

FabianWolff: I agree that this version of `SpecEntVars` looked superfluous. But keep in mind that type checking for closure specifications is slightly more intricate than for quantifiers, because we can't type-check the pre- and postconditions in one go: The postcondition may refer to `result` but the precondition may not. For this reason, my current implementation defines `SpecEntVars` like this:
```rust
#[derive(Debug, Clone)]
pub struct SpecEntVars<EID, AT> {
    pub spec_id: SpecificationId,
    pub pre_id: EID,
    pub post_id: EID,
    pub args: Vec<AT>,
    pub result: AT,
}
```
i.e. we additionally need the result type, and _two_ IDs, one for type-checking the pre- and one for type-checking the postconditions.

To be honest, `result` is currently hard-coded as `result: i32`, but we could change the syntax to `cl |= |i: i32| -> T [ ... ]` to fix this problem. What doesn't work in this setting is higher-order closures, because if `cl` returns a closure type, we can't write that down (as `T`), because it's anonymous. This is unfortunate but seems to be an intrinsic limitation to the way type checking is implemented right now (note that closures can't be generic, so we can't write `|| -> impl Fn () -> i32`, as we could for functions).

FabianWolff:
@Aurel300 @vakaras I need some help resolving generic type parameters. Consider this example:
```rust
#[requires(f |= |i: i32| [ requires(i >= 0), ensures(true) ])]
fn test<F: Fn (i32) -> i32> (f: F) {}

let f = ...;
test(f);
```
I need to encode `test`'s precondition in two different ways (once for proving `test`, and once for proving the call-site); in particular, at the call-site, I need to know the _concrete_ type of `f` in order to be able to use the correct specification functions.

But in `spec_encoder.rs`, I don't see a way to get this information; in particular, I can get `F` as the type of `f`, but not the concrete closure type at the call-site. So, how can I get the concrete types of generic type parameters in the `SpecEncoder`?

prusti-specs/src/specifications/common.rs
Aurel300: Ok, if `SpecEntVars` ends up different from `ForAllVars`, it makes sense.

As for the return type, we could indeed make the syntax **optionally** accept `-> X` after the arguments. When parsing, it would still be unambiguous, I believe. The higher-order closure-returning-a-closure situation is interesting. We could pretend that the type annotation is valid even if Rust would not accept it, but I predict this would get rather unwieldy very quickly.

I expect that closures like this are not extremely common. Maybe eventually (as a result of your work or later still) closures will not require explicit type annotation in Prusti, both in the specifications and the definitions. Then we could generate type checks that rely on type inference just like what the user wrote in the first place.

Aurel300:
> in particular, I can get `F` as the type of `f`, but not the concrete closure type at the call-site.

Do you mean you can get the type `F` at the call-site as well? Is this by checking the type of `test`? I would have assumed in the MIR you should find a call to the monomorphised version of `test`, not one that still references `F`.

FabianWolff:
> > in particular, I can get `F` as the type of `f`, but not the concrete closure type at the call-site.
> 
> Do you mean you can get the type `F` at the call-site as well? Is this by checking the type of `test`? I would have assumed in the MIR you should find a call to the monomorphised version of `test`, not one that still references `F`.

Yes, the call is to a monomorphized version, but when type-checking the specification entailment, the closure expression will have some generic type, and I didn't know how to reconcile the two. But then I found `typaram_repl` in `Encoder`, which seems to provide just the information I need, so that's what I ended up using. Still, thanks for your reply!

vakaras:
> But this seems to require both `#![feature(stmt_expr_attributes)]` and `#![feature(proc_macro_hygiene)]` for now. @vakaras You were opposed to the former at least, in the PR comments. Is it just because you don't want to rely on any unstable features at all? Both seem to be pretty old features by now, I would expect them to stabilise…

I think that the code must be compilable with the stable Rust compiler to make Prusti adoptable in real codebases.

Aurel300:
I'm working on understanding this better and resolving the conflicts so we can merge it into the codebase. I don't know how I feel about the extra `trace!` in many places. I assume they were there to help you @FabianWolff during debugging, but maybe they don't need to stay.

Aurel300:
Oh also @FabianWolff is there any work you still need to push?

FabianWolff:
> I'm working on understanding this better and resolving the conflicts so we can merge it into the codebase. I don't know how I feel about the extra `trace!` in many places. I assume they were there to help you @FabianWolff during debugging, but maybe they don't need to stay.

Sure, feel free to drop the ones you don't like (you [should have](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) write access to my forked repository, or at least to the relevant branch for this PR).

> Oh also @FabianWolff is there any work you still need to push?

No, I only have some stashed changes left that are broken and didn't get anywhere.

Aurel300:
Well after the rebase this compiles and has no conflicts, the tests run, but unfortunately I see there are no closure-related tests. When testings a simple example like:

```rust
#![feature(stmt_expr_attributes)]

use prusti_contracts::*;

fn main() {
    let f = closure!(
        requires(i >= 0),
        ensures(result == i + 1),
        |i: i32| -> i32 { i + 1 }
    );
    f(0);
}
```

I get:

```
error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 40 }), val_int: Int, Position { line: 9, column: 18, id: 40 }), read)))
error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 181 }), val_int: Int, Position { line: 9, column: 18, id: 181 }), read)))
error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@10.17)
error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@9.18)
```

So it seems something broke due to the Prusti changes, or simply me making a mistake in the rebasing process. @FabianWolff If you have not reset your repo yet after the force push, could you please confirm that the example works with your version? Have you faced the same errors when working on this?

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -171,6 +172,111 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
                     .collect(),
                 self.encode_assertion(body),
             ),
+            box typed::AssertionKind::SpecEnt(ref cl, ref vars, ref pres, ref posts) => {
```
fpoli: It would be better to move the implementation of this case to a separate function, to avoid reaching 12 levels of indentation.

fpoli:
> ```
> error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 40 }), val_int: Int, Position { line: 9, column: 18, id: 40 }), read)))
> error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 181 }), val_int: Int, Position { line: 9, column: 18, id: 181 }), read)))
> error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@10.17)
> error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@9.18)
> ```

This looks similar to some bugs that I faced while upgrading the implementation of `SpecEncoder::encode_assertion`, to make loop invariants work. I suggest to start debugging by looking at a dump of the Viper encoding.

FabianWolff:
> Well after the rebase this compiles and has no conflicts, the tests run, but unfortunately I see there are no closure-related tests. When testings a simple example like:
> 
> ```rust
> #![feature(stmt_expr_attributes)]
> 
> extern crate prusti_contracts;
> 
> use prusti_contracts::*;
> 
> fn main() {
>     let f = closure!(
>         requires(i >= 0),
>         ensures(result == i + 1),
>         |i: i32| -> i32 { i + 1 }
>     );
>     f(0);
> }
> ```
> 
> I get:
> 
> ```
> error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 40 }), val_int: Int, Position { line: 9, column: 18, id: 40 }), read)))
> error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(i32), Position { line: 9, column: 18, id: 181 }), val_int: Int, Position { line: 9, column: 18, id: 181 }), read)))
> error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@10.17)
> error: [Prusti internal error] consistency error: Consistency error: Local variable _1 not found. (@9.18)
> ```
> 
> So it seems something broke due to the Prusti changes, or simply me making a mistake in the rebasing process. @FabianWolff If you have not reset your repo yet after the force push, could you please confirm that the example works with your version? Have you faced the same errors when working on this?

Yes, I can confirm that the example verifies successfully with my version. I could also reproduce your problem with the rebased version, although I could not immediately find the reason for it by looking at the Viper code. It might have something to do with how the arguments are named, i.e. my code assumes that the arguments are called `_1`, `_2`, etc. in the MIR, and if that has changed somehow in the rest of the encoding, that's where the problem could come from.

fpoli:
> I see there are no closure-related tests.

@Aurel300 (or @FabianWolff) Could you add them, such that we can see the CI tests failing?

> Yes, I can confirm that the example verifies successfully with my version. I could also reproduce your problem with the rebased version, although I could not immediately find the reason for it by looking at the Viper code. It might have something to do with how the arguments are named, i.e. my code assumes that the arguments are called _1, _2, etc. in the MIR, and if that has changed somehow in the rest of the encoding, that's where the problem could come from.

Have your tried dumping the viper encoding with both versions, comparing them with e.g. [`meld`](https://meldmerge.org/) (it'a a bit better than just `diff` when only part of a line changed)? I found it extremely useful a few weeks ago.

@FabianWolff If you could tell a bit more about the assumptions on the name of the Viper variables it would be great (the alternative is to read it from the code). A short Rust+Viper example would be even better, such that we can check whether the assumptions are still valid.

Aurel300:
I pushed some very very basic tests. The parsers tests are fine, since that is not affected by the Prusti changes. The verification tests should hopefully work as intended.

FabianWolff:
> @FabianWolff If you could tell a bit more about the assumptions on the name of the Viper variables it would be great (the alternative is to read it from the code). A short Rust+Viper example would be even better, such that we can check whether the assumptions are still valid.

The MIR for the closure in Aurel's example looks like this:
```Rust
fn main::{closure#0}(_1: &'_#4r [closure@../examples/some-example.rs:8:13: 12:6], _2: i32) -> i32 { ... }
```
`_0` is the return value (implicitly), `_1` is the closure's `self` parameter, and `_2` is the actual argument. For type-checking (neglecting the views, which have not been implemented properly yet), there will be a helper function `prusti_post_closure_[...]` (run Prusti with `-Zprint-desugared-specs` to see this) with the following MIR signature:
```Rust
fn main::{closure#0}::prusti_post_closure_66d32e9672e84fcc87b6ed2689d0705c(_1: i32, _2: i32) -> () { ... }
```
Here, `_1` represents the closure argument (`i: i32` in the example) and `_2` represents the result (also an `i32`). In other words, when encoding the postcondition assertion, `_1` should become `_2` and `_2` should become `_0` to match the closure's signature. This is achieved in `SpecFunctionEncoder` by calling `Encoder::encode_assertion()` with the respective values for `target_args` and `target_return`. This has worked fine for me but seems to be broken after the rebase, because the following is generated for the postcondition specification function in Viper:
```
function sf_post_some_example$$main$$$opencur$closure$sharp$0$closecur$__$TY$__$int$$$int$$$bool$(_2: Int, _0: Int): Bool
{
    _2 == _1.val_int + 1
}
```
i.e. the postcondition specification function correctly receives the closure argument (`_2`) and its result (`_0`) as arguments, but the assertion is encoded incorrectly (the body should be `_0 == _2 + 1`). This means that something must have changed in the way snapshots are encoded (why does it say `_1.val_int` if the argument is supposed to be a snapshot, i.e. an `Int` and not a `Ref`?) and resolved (`_1` should be `_2` and `_2` should be `_0`, as noted above).

I have looked into this, but couldn't seem to find the reason for this. Do you have any ideas? Hopefully this explanation clears things up somewhat for you.

prusti-viper/src/encoder/spec_function_encoder.rs
```diff
@@ -0,0 +1,156 @@
...
+            .args
+            .iter()
+            .map(|local| self.encode_local(local.clone().into()).into())
+            .collect();
+
+        for item in contract.functional_precondition() {
+            func_spec.push(self.encoder.encode_assertion(
+                &item,
+                &self.mir,
+                Some(&""),
```
fpoli: Should be `None`?

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -171,6 +172,111 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
...
+                let tcx = self.encoder.env().tcx();
+                let (mir, _) = tcx.mir_promoted(ty::WithOptConstParam::unknown(cl.expr));
```
fpoli: We have `self.encoder.env().mir(...)`

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -171,6 +172,111 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
...
+                            );
+
+                            vec![pre_conjunct, post_conjunct]
+                                .into_iter()
+                                .conjoin()
+                        }
+                        _ => unreachable!()
+                    }
+                } else {
+                    // TODO
```
fpoli, line 374: Could someone describe what is left to do?

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -375,6 +481,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
             curr_def_id = outer_def_id;
         }
 
+        let skip_first = self.encoder.encode_item_name(curr_def_id).contains("_closure_");
```
fpoli, line 594: This needs a comment. Why is this needed?

Aurel300: I think this is to skip the closure's "self" argument. I am currently trying to figure out how the arguments get remapped, which should solve at least the "Local variable _1 not found." errors.

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -387,6 +495,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
         replacements.extend(
             mir.args_iter()
                 .zip(self.target_args)
+                .skip(if skip_first { 1 } else { 0 })
```
fpoli: Is it correct to ignore the first element of `self.target_args` when `skip_first` is `true`? If not, should the `.skip(..)` be before the `.zip(..)`?

Aurel300: This panics when the order is swapped, so I think it is correct like this. The `self` argument is also skipped in `spec_function_encoder` when creating `formal_args`.

fpoli:
> I have looked into this, but couldn't seem to find the reason for this. Do you have any ideas? Hopefully this explanation clears things up somewhat for you.

Thanks for the explanation! From what you describe I see nothing wrong. However, see the comment on `skip_first`, as it sounds a bit strange to me.

It may be worth adding a `debug!(..., replacements)` after this, to check that the replacements are correct:

https://github.com/viperproject/prusti-dev/blob/4425d92a7d135ed6906a7b7203e6a8c8bcc5d9c9/prusti-viper/src/encoder/spec_encoder.rs#L386-L404

Aurel300:
> It may be worth adding a `debug!(..., replacements) after` this, to check that the replacements are correct:

Yeah, this one seems wrong. It does:

 - `_2.val_int` -> `_2`
 - `_3.val_int` -> `_3`
 - `_3.val_int` -> `_0`

(to be clear here I was testing a closure with 2 arguments)

Aurel300:
Some minor notes as I am working on this.

I tried to hardcode the replacements vector to see what would happen if it were correct. I.e. `_1.val_int` -> `_2` and `_2.val_int` -> `_0`. The function body after the replacements is correct (`_0 == _2 + 1`). However, Viper then complains "No matching local variable _0 found with type Int, instead found Ref." and "No matching local variable _2 found with type Int, instead found Ref.".

I was a bit confused also because the `log/vir_program_beofre_viper/foo.rs.vir` file did not contain the specification function(s), even though the file *was* regenerated with each compilation. With `PRUSTI_LOG=debug`, I can see the Viper program dumped though. It does contain the specification function as expected:

```
function sf_post_foo$$main$$$opencur$closure$sharp$0$closecur$__$TY$__$int$$$int$$$bool$(_2: Int, _0: Int): Bool
{
  _0 == _2 + 1
}
```

But it also contains this, probably responsible for the error, inside `m_foo$$main$$$opencur$closure$sharp$0$closecur$()`:

```
  // ...
  fold acc(i32(_0), write)
  // obtain acc(i32(_0), write)
  
  assert _0 == _2 + 1 // !!!
  assert true
  exhale acc(closure$0_6$3$6795211212542883138(_old$pre$0), read$())
  exhale acc(i32(_0), write)
  // ...
```

So it seems that the replaced body is making it into the assertion there. Maybe something should be `clone()`d but isn't? Also @FabianWolff, is the specification function actually meant to be called somewhere in Viper? From the dumped Viper code it looks like an unused function.

FabianWolff:
> But it also contains this, probably responsible for the error, inside `m_foo$$main` (where I actually call the closure):
> 
> ```
>   // ...
>   fold acc(i32(_0), write)
>   // obtain acc(i32(_0), write)
>   
>   assert _0 == _2 + 1 // !!!
>   assert true
>   exhale acc(closure$0_6$3$6795211212542883138(_old$pre$0), read$())
>   exhale acc(i32(_0), write)
>   // ...
> ```
> 
> So it seems that the replaced body is making it into the assertion there. Maybe something should be `clone()`d but isn't? Also @FabianWolff, is the specification function actually meant to be called somewhere in Viper? From the dumped Viper code it looks like an unused function.

Yes, the idea is that where you wrote `!!!`, it should say something like `assert pre$cl(_2, _0)`. I implemented closure calls before the specification functions, and did not get around to fixing this in the end, but it should be a minor change.

Specification functions are currently called when encoding specification entailments; for instance, try this example and see how the higher-order function call is encoded:
```Rust
#![feature(stmt_expr_attributes)]
use prusti_contracts::*;

fn test() {
    let f = closure! (
        requires (a > b),
        ensures (result > b),
        |a: i32, b: i32| -> i32 { a }
    );

    hof (f);
}


#[requires(f |= |a: i32, b: i32| [
    requires(a == 5), requires(b == 4),
    ensures(result > 4)
]) ]
fn hof<F: Fn (i32, i32) -> i32> (f: F) {}

fn main() {}
```

Aurel300:
> Yes, the idea is that where you wrote `!!!`, it should say something like `assert pre$cl(_2, _0)`. I implemented closure calls before the specification functions, and did not get around to fixing this in the end, but it should be a minor change.

Ok, but then I wonder how this worked before the rebase?

FabianWolff:
> > Yes, the idea is that where you wrote `!!!`, it should say something like `assert pre$cl(_2, _0)`. I implemented closure calls before the specification functions, and did not get around to fixing this in the end, but it should be a minor change.
> 
> Ok, but then I wonder how this worked before the rebase?

This is how the call was encoded prior to the rebase:
```
      _4 := builtin$havoc_ref()
      inhale acc(tuple1$i32(_4), write)
      _aux_havoc_i32 := builtin$havoc_ref()
      unfold acc(tuple1$i32(_4), write)
      _4.tuple_0 := _aux_havoc_i32
      inhale acc(_4.tuple_0.val_int, write)
      _4.tuple_0.val_int := 0
      // [mir] _2 = const <[closure@../examples/some-example.rs:8:13: 12:6] as std::ops::Fn<(i32,)>>::call(move _3, move _4) -> [return: bb2, unwind: bb1]
      
      label l1
      assert _4.tuple_0.val_int >= 0
      fold acc(closure$0_6$3$8955822482898200535(_3.val_ref), write)
      fold acc(i32(_4.tuple_0), write)
      assert true
      exhale acc(_3.val_ref, write) && acc(i32(_4.tuple_0), write)
      _2 := builtin$havoc_ref()
      inhale true
      inhale acc(i32(_2), write)
      // transfer perm _3.val_ref --> old[l1](_3.val_ref) // unchecked: true
      
      inhale true
      inhale (unfolding acc(i32(_2), write) in _2.val_int == old[l1](_4.tuple_0.val_int) + 1)
```
In the future, I suggest you refrain from rebasing: If you had just merged in the master branch, you would have been able now to check out the old, working commits, but now, they're gone, and my rebased commits appear nonsensical because they have been developed for some old version of the master branch, and the same changes don't necessarily make sense/work on top of the newer master branch, as you have noticed.

Aurel300:
> This is how the call was encoded prior to the rebase [...]

Interesting. Actually the call looks very similar. However, instead of:

```
inhale (unfolding acc(i32(_2), write) in _2.val_int == old[l1](_4.tuple_0.val_int) + 1)
```

We get:

```
inhale _0 == _2 + 1
```

The assertion is in the body of the closure function itself, but even there it is wrong. The local names look correct, but they also require `unfolding`.

> In the future, I suggest you refrain from rebasing: If you had just merged in the master branch, you would have been able now to check out the old, working commits, but now, they're gone, and my rebased commits appear nonsensical because they have been developed for some old version of the master branch, and the same changes don't necessarily make sense/work on top of the newer master branch, as you have noticed.

They are not gone, `reflog` helped. Here we are: https://github.com/viperproject/prusti-dev/commits/closures-before-rebase

The problem is that even with having the old version available I am not sure how useful it is; the changes from the rebase still need to be applied, and since e.g. the replacement in `spec_encoder` works quite differently, it wouldn't be any better trying to merge rather than rebase.

fpoli:
If you want to disable the optimizations that inline pure function calls you can set `PRUSTI_SIMPLIFY_ENCODING=false`. It should be easier to debug the encoding this way. For example, I suspect that the `assert _0 == _2 + 1 // !!!` was a pure function call.

Aurel300:
Ok, now I have both the old version and the rebased version built.

> Have your tried dumping the viper encoding with both versions, comparing them with e.g. [`meld`](https://meldmerge.org/)

Yes, after hard-coding the replacements for the Viper program to even contain these parts, the non-cosmetic differences between the two versions are:

(At the end of the closure body)

```diff
360c360
< assert (!false ? (unfolding acc(i32(_0), write) in _0.val_int) == old[pre]((unfolding acc(i32(_2), write) in _2.val_int)) + 1 : false) && true
---
> assert (!false ? _0 == _2 + 1 : false) && true
```

(At the call site)

```diff
440c440
< inhale (!false ? (unfolding acc(i32(_2), write) in _2.val_int) == old[l1](_4.tuple_0.val_int) + 1 : false) && true
---
> inhale (!false ? _0 == _2 + 1 : false) && true
```

FabianWolff:
I just pushed some changes that seem to fix the variable replacement issue for me. One observation was that the replacement vector in `SpecEncoder::encode_expression()` sometimes contains multiple replacements for the same variable, and the correct one is the *last* one. I can now verify Aurel's example again:
```Rust
#![feature(stmt_expr_attributes)]

use prusti_contracts::*;

fn main() {
    let f = closure!(
        requires(i >= 0),
        ensures(result == i + 1),
        |i: i32| -> i32 { i + 1 }
    );
    f(0);
}
```
However, the following still fails:
```Rust
#![feature(stmt_expr_attributes)]

use prusti_contracts::*;

fn test() {
    let f = closure! (
        requires (a > b),
        ensures (result > b),
        |a: i32, b: i32| -> i32 { a }
    );

    let a = 5;
    let b = 3;
    let c = f (a, b);
    assert! (c > b);

    test2 (f);
}


#[requires(f |= |a: i32, b: i32| [ requires(a == 5), requires(b == 4), ensures(result > 4)])]
fn test2<F: Fn (i32, i32) -> i32> (f: F) {}

fn main() {}
```
with
```
error: [Prusti internal error] consistency error: Consistency error: Local variable _2_forall not found. (@22.12)
error: [Prusti internal error] consistency error: Consistency error: Local variable _3_forall not found. (@22.12)
error: [Prusti internal error] consistency error: Consistency error: Local variable _4_forall_8664c4c4daa04f9180a1981aab41eaf6_103 not found. (@22.80)
```
The offending assertion looks like this:
```
assert (forall _2_forall_8664c4c4daa04f9180a1981aab41eaf6_102: Int, _3_forall_8664c4c4daa04f9180a1981aab41eaf6_102: Int :: _2_forall_8664c4c4daa04f9180a1981aab41eaf6_102 == 5 && _3_forall_8664c4c4daa04f9180a1981aab41eaf6_102 == 4 ==> sf_pre_another_example$$test$$$opencur$closure$sharp$0$closecur$__$TY$__$int$$$int$$$bool$(_2_forall, _3_forall)) && (forall _2_forall_8664c4c4daa04f9180a1981aab41eaf6_102: Int, _3_forall_8664c4c4daa04f9180a1981aab41eaf6_102: Int, _4_forall: Int :: _2_forall_8664c4c4daa04f9180a1981aab41eaf6_102 == 5 && _3_forall_8664c4c4daa04f9180a1981aab41eaf6_102 == 4 ==> sf_post_another_example$$test$$$opencur$closure$sharp$0$closecur$__$TY$__$int$$$int$$$int$$$bool$(_2_forall, _3_forall, _4_forall) ==> _4_forall_8664c4c4daa04f9180a1981aab41eaf6_103 > 4)
```
This looks to me as though it doesn't have anything to do with the closure/specification entailment encoding, because the numbering is correct; the issue is that sometimes a suffix is added to the quantified variables (e.g. `_3_forall` becomes `_3_forall_8664c4c4daa04f9180a1981aab41eaf6_102`), but not consistently. This must happen at a later stage of Prusti, because my encoding generates and uses only the simple names (`_3_forall` etc., as generated by `encode_forall_arg()` in `SpecEncoder` when encoding specification entailments).

fpoli:
> One observation was that the replacement vector in `SpecEncoder::encode_expression()` sometimes contains multiple replacements for the same variable, and the correct one is the last one.

Regarding that, I consider it a bug of `encode_expression`.

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -171,6 +172,111 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
...
+                            let pre_conjunct = vir::Expr::forall(
+                                vars.args.iter().map(|(arg, arg_ty)| self.encode_forall_arg(*arg, arg_ty, &format!("{}_{}", vars.spec_id, vars.pre_id))).collect(),
+                                vec![], // TODO: encode triggers
+                                vir::Expr::implies(
+                                    encoded_pres.clone(),
+                                    vir::Expr::FuncApp(
+                                        sf_pre_name,
+                                        (0 .. vars.args.len())
+                                            .map(|i| vir::Expr::Local(
+                                                vir::LocalVar::new(format!("_{}_forall", i + 2), vir::Type::Int),
```
fpoli: Should this call `encode_forall_arg` like 8 lines above?

Aurel300: You are right, the forall encoding has changed since this was written. This fixes part of the issue.

prusti-viper/src/encoder/spec_encoder.rs
```diff
@@ -225,31 +226,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecEncoder<'p, 'v, 'tcx> {
...
-                                    .iter()
-                                    .map(|(arg, arg_ty)| self.encode_forall_arg(*arg, arg_ty, &format!("{}_{}", vars.spec_id, vars.pre_id)))
-                                    .chain(std::iter::once(vir::LocalVar::new(result_name.clone(), vir::Type::Int)))
-                                    .collect(),
+                                qvars_post.clone(),
                                 vec![], // TODO: encode triggers
                                 vir::Expr::implies(
-                                    encoded_pres,
+                                    // The quantified variables in the precondition have been encoded using
+                                    // different IDs (vars.pre_id vs. vars.post_id), so we need to fix them
```
Aurel300, line 340: I was in the middle of working on this (should have self-assigned!), but your solution is nicer anyway. I was trying to reencode `encoded_pres` under a different `def_id`. Thanks.

FabianWolff: Ah, sorry about that! I thought I'd try to help you out a bit on this.

Anyway, the second example in my comment yesterday now also verifies successfully!

prusti-viper/src/encoder/spec_encoder.rs
FabianWolff: Indeed, that was part of the problem. Thanks for the hint @fpoli!

prusti-viper/src/encoder/spec_function_encoder.rs
```diff
@@ -0,0 +1,156 @@
...
+            .iter()
+            .map(|local| self.encode_local(local.clone().into()).into())
+            .collect();
+        let encoded_return = self.encode_local(contract.returned_value.clone().into());
+
+        for item in contract.functional_postcondition() {
+            func_spec.push(self.encoder.encode_assertion(
+                &item,
+                &self.mir,
+                Some(&""),
```
fpoli: `None`?

Aurel300: I think you somehow reviewed an old version? The `None`s were added in cb3591ee0bc4616336223143d9a88c166c40ca63.

prusti-viper/src/encoder/spec_encoder.rs
fpoli: I'm still not convinced about this. Maybe it was needed to avoid a bug that has been fixed in the meantime.

If before the replacement the (encoding of the) `self` argument is used in `curr_expr` then after the replacement it will still be there because of the `skip(1)`, but that would almost certainly be wrong because the remaining replacements could introduce a `_1` with a different type/meaning. So I guess that before the replacement `self` is not supposed to be in `curr_expr`. In that case, the `skip(1)` should have no effect, because the first replacement wouldn't match any place.

From what I understand we could remove `skip_first` because it's dead code. ~~However, I cannot find an usage of `encode_assertion` which actually passes the encoding of the `self` argument that should be skipped by `skip(1)`. In my view the implementation of `encode_{pre|post}_spec_func` should prepend `encoded_args` with a dummy variable named like `_nonexistent_`, but that does not happen.~~ (I was wrong about `encode_{pre|post}_spec_func`; they encode `self`.)

(As a side note, I would prefer having closure-specific logic on the caller side of `encode_assertion`, where it should be clear whether we are encoding a closure, rather than inside `encode_assertion` where we can only use the heuristic on the encoded item name.)

Aurel300:
Ok, now I think the PR is back to supporting the features shown during Fabian's presentation. I think it might be good to merge this before working on additional features, so that we don't desync from master again. However, before that happens, I'd like to:

 - [x] see if we can get rid of the need for `stmt_expr_attributes`
 - [x] check the duplicate replacements situation in `encode_expression` (I thought this is because the last argument is checked twice if `target_return` is `Some(_)`, in which case the reverse iteration order when replacing should not be needed)
 - [x] create a test suite from Fabian's thesis examples, testing the basic features in various patterns, but also tests for the not-yet-supported features with `ignore-test`
 - [ ] clean up, reduce code duplication, add comments and explanations where appropriate (especially for the TODOs)

Replacing `closure!(...)` with a more method-like syntax is also on my agenda, but since I have a feeling this would definitely require `stmt_expr_attributes` (or a more involved preparser), let's leave that for later.

prusti-viper/src/encoder/spec_encoder.rs
Aurel300: > [...] I would prefer having closure-specific logic on the caller side of `encode_assertion`, where it should be clear whether we are encoding a closure [...]

I agree. `skip_first` will eventually not be needed, once we figure out how to properly encode the closure's self type with snapshots.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4780,15 +4889,14 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-                //assert!(self.encoder.is_spec_closure(def_id), "closure: {:?}", def_id);
-                // Specification only. Just ignore in the encoding.
-                // FIXME: Filtering of specification blocks is broken, so we need to handle this here.
-              if self.encoder.is_spec_closure(def_id) {
-                // Specification only. Just ignore in the encoding.
-                // FIXME: Filtering of specification blocks is broken, so we need to handle this here.
-                Vec::new()
+                if self.encoder.is_spec_closure(def_id) {
+                    // Specification only. Just ignore in the encoding.
+                    // FIXME: Filtering of specification blocks is broken, so we need to handle this here.
```
Aurel300: I guess this should be resolved since #176 ?

prusti-viper/src/encoder/spec_encoder.rs
Aurel300: I'm not exactly sure. The condition itself is checking for the result type (which is the closure type) being in the last entry in `typaram_repl`. I assume this *should* be using `self.encoder.current_tymap()` instead, <s>which would probably mean this condition can be dropped</s>. So this does happen in a basic example, where a type parameter (representing an input function) is not known. I need to check better when this happens and when it is in the type map.

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: This branch should be unreachable, because we don't encode to Viper the basic blocks used for specifications:

https://github.com/viperproject/prusti-dev/blob/d27d8fa49e08a058ac369be503825d4f283013b7/prusti-viper/src/encoder/procedure_encoder.rs#L900-L906

I would try replacing the `if` with a `debug_assert!(!self.encoder.is_spec_closure(def_id))`.

prusti-viper/src/encoder/spec_function_encoder.rs
```diff
@@ -150,8 +155,11 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecFunctionEncoder<'p, 'v, 'tcx> {
             posts: Vec::new(),
             body: Some(func_spec.into_iter()
                                 .map(|post| SnapshotSpecPatcher::new(self.encoder).patch_spec(post))
+                                .collect::<Result<Vec<vir::Expr>, _>>()
+                                .map_err(|err| err.with_span(self.procedure.get_span()))?
+                                .into_iter()
```
Aurel300: These extra calls are a bit messy. @fpoli am I missing something with the positionless errors? Is this the best way I can map them?

prusti-interface/src/specs/typed.rs
```diff
@@ -30,6 +30,8 @@ pub type Expression = common::Expression<ExpressionId, LocalDefId>;
 pub type TriggerSet = common::TriggerSet<ExpressionId, LocalDefId>;
 /// For all variables that have no types associated with it.
 pub type ForAllVars<'tcx> = common::ForAllVars<ExpressionId, (mir::Local, ty::Ty<'tcx>)>;
+/// Specification entailment variables that have no types associated.
+pub type SpecEntVars<'tcx> = common::SpecEntVars<ExpressionId, (mir::Local, ty::Ty<'tcx>)>;
```
vakaras: Somehow “Ent” for me does come across as a shorthand for “Entailment”. I think it would be better to spell out the full word.

prusti-specs/src/specifications/common.rs
```diff
@@ -240,6 +253,13 @@ pub enum AssertionKind<EID, ET, AT> {
         TriggerSet<EID, ET>,
         Assertion<EID, ET, AT>,
     ),
+    /// Specification entailment: spec_ent(closure, argument binders, pres, posts)
+    SpecEnt(
```
vakaras: Since from the types it is not obvious the meaning of each field, I would suggest to explicitly name them:
```rust
    SpecEnt{
        closure: Expression<EID, ET>,
        arg_binders: SpecEntVars<EID, AT>,
        pres: Vec<Assertion<EID, ET, AT>>,
        posts: Vec<Assertion<EID, ET, AT>>,
    }
```

prusti-specs/src/specifications/json.rs
```diff
@@ -13,6 +13,7 @@ pub enum AssertionKind {
     And(Vec<Assertion>),
     Implies(Assertion, Assertion),
     ForAll(ForAllVars, Assertion, TriggerSet),
+    SpecEnt(Expression, SpecEntVars, Vec<Assertion>, Vec<Assertion>),
```
vakaras: Same comment here about naming the fields.

vakaras (approved):
LGTM.

DenisGorbachev:
The page at https://viperproject.github.io/prusti-dev/user-guide/verify/spec_ent.html says "This feature is not yet supported in Prusti". However, this PR is merged.

If this feature is actually supported, maybe it's possible to remove the "not yet supported" message from the documentation?

PR #141, opened 2020-08-31
Use less fake arguments and targets for function calls.

frececroka:
For call-site contracts, a fake local variable is created for every argument and the return destination. This means we don't need special handling for arguments that are constants or projected locals, which is very convenient.

However, in some cases it is necessary to access the real arguments passed into the function. This change uses the real local variable for arguments and return destinations that are just local variables. Arguments and return places that are constants or projected locals are handled as before, by replacing them with a fake local variable.


vakaras (approved):
LGTM. Thanks!

PR #145, opened 2020-09-01
External specs

karenhong:
Adds the ability to specify external methods and top-level functions.


prusti-contracts/src/lib.rs
```diff
@@ -25,6 +25,9 @@ mod private {
 
     /// A macro for defining a closure with a specification.
     pub use prusti_contracts_impl::closure;
+
+    /// A macro for specifying external functions
```
Aurel300:
```suggestion
    /// A macro for specifying external functions.
```

(Nit-picky, sorry!)

prusti-contracts/src/lib.rs
```diff
@@ -52,6 +55,9 @@ mod private {
 
     /// A macro for defining a closure with a specification.
     pub use prusti_contracts_internal::closure;
+
+    /// A macro for specifying external functions
```
Aurel300:
```suggestion
    /// A macro for specifying external functions.
```

prusti/tests/extern_spec/fail/mismatched-sigs.rs
```diff
@@ -0,0 +1,15 @@
+extern crate prusti_contracts;
+use prusti_contracts::*;
+
+#[extern_spec]
+impl<T> std::vec::Vec<T> {
```
Aurel300: As a test, this should fail if either of the specifications work (when they shouldn't). Will this test as a whole not fail even if e.g. `push` happens to verify?

prusti/tests/extern_spec/pass/linked-list-i32.rs
```diff
@@ -0,0 +1,116 @@
...
+    get(self, i) == old(get(self, i))))]
+    pub fn pop_back(&mut self) -> Option<i32>;
+
+
+    #[ensures(self.len() == old(self.len() + other.len()))]
+    #[ensures(forall (|i: usize| (i < old(self.len())) ==>
+    get(self, i) == old(get(self, i))))]
+    #[ensures(forall (|j: usize| (old(self.len()) <= j && j < self.len()) ==>
+        get(self, j) == old(get(other, j - self.len()))))]
+    pub fn append(&mut self, other: &mut LinkedList<i32>);
```
Aurel300: `#[ensures(other.len() == 0)]` ?

prusti/tests/extern_spec/pass/set.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[extern_spec]
+impl<T> Option<T> {
+    #[pure]
+    #[ensures(matches!(*self, Some(_)) == result)]
+    pub fn is_some(&self) -> bool;
+
+    #[pure]
+    #[ensures(self.is_some() ==> !result)]
+    #[ensures(!self.is_some() ==> result)]
```
Aurel300: Maybe this doesn't make a difference in practice, but why not `result ==> !self.is_some()`? I think you wrote it in that order in the `LinkedList` spec.

prusti/tests/extern_spec/pass/swap.rs
```diff
@@ -0,0 +1,28 @@
+#![feature(type_ascription)]
```
Aurel300: Is this used (in the generated code)?

Aurel300:
Regarding one of the future goals of this feature, the one of providing specifications for the standard library (and maybe crates in general).

One thing to consider is proper versioning. APIs change across library versions, and it is probably not sufficient to keep an up-to-date specification of the latest version of a particular library. If a client project is currently locked into an older version, it would be extremely difficult piecing together the correct specification from different commits of the repo holding the external specs.

Some mechanism of tagging a specification with a version (major + minor, or a range of semvers) would be useful. Ideally we could have e.g.:

```rust
#[extern_spec("foobar", "1.0")]
impl Foobar {
  // specs ...
}

#[extern_spec("foobar", "2.0")]
impl Foobar {
  // different specs for the newer version ...
}
```

Such specifications could then only be used when a matching version of the library is in use, and would result in verification failure if no matching version is found.

---

Edit: a similar project to this is [`typeshed`](https://github.com/python/typeshed), which provides static type information for Python (both its standard library and some third-party libraries). [Here](https://github.com/python/typeshed/blob/8642d2aa97e6fbb87c06c169c72a0ebf8d1f42f1/stdlib/3/os/__init__.pyi#L275) is an example where the defined type is different depending on the runtime version.

prusti/tests/extern_spec/fail/mismatched-sigs.rs
karenhong: You are correct in that the test should fail if either of the specifications work. However, it will currently pass even if one happens to verify. I'll split this test case so that both cases are checked for correctly.

prusti/tests/extern_spec/pass/linked-list-i32.rs
karenhong: Thanks, I'll add this

prusti/tests/extern_spec/pass/set.rs
karenhong: Hmm I'm not sure why I wrote it this way, but no, I don't think that in practice it makes a difference. I think I will just change this to `result != self.is_some()`

prusti/tests/extern_spec/pass/swap.rs
karenhong: It is needed for this particular test case

fpoli:
Two comments regarding the new tests:
* We recently moved all tests from `prusti/tests/` to `prusti-tests/tests/`.
* Who executes the `extern_spec` tests? I think you lost some changes while merging with master. Unless you need to set special Prusti configuration flags, I suggest to put your tests under `prusti-tests/tests/verify/{pass|fail}/extern-spec/` (as you initially did) such that they will be executed as part of the `verify` test group.

cmatheja:
> Some mechanism of tagging a specification with a version (major + minor, or a range of semvers) would be useful. 
> ...
> Such specifications could then only be used when a matching version of the library is in use, and would result in verification failure if no matching version is found.

This is a very good idea. 
I would suggest one small modification: If no version is found, verification probably shouldn't fail; we can revert to the current sound but incomplete status, i.e. both the pre- and postcondition of such functions are just true.

Aurel300:
> I would suggest one small modification: If no version is found, verification probably shouldn't fail; we can revert to the current sound but incomplete status, i.e. both the pre- and postcondition of such functions are just true.

I agree, verification failure is probably too disruptive. We could introduce a configuration flag to warn for any un-specified functions, including ones which are specified but have the wrong version.

Aurel300:
Ok, I rebased and moved the tests to the new `prusti-tests` hierarchy. Some things to fix yet:

 - `ignore-test` on `tests/verify/fail/extern-spec/duplicate-2.rs` and `tests/verify/fail/extern-spec/variants.rs` - these fail as intended, but the error reporting method makes the compiler crash rather than exit gracefully. Relevant call is [here](https://github.com/karenhong/prusti-dev/blob/737cd7935ba3862fdd15fbba2c16c6880da4deb1/prusti-interface/src/specs/external.rs#L86-L88), using `TyCtxt::ty_error_with_message`.
 - `ignore-test` on `tests/verify/extern-spec/linked-list-i32.rs` – this is an actual verification failure, where the precondition of a method call cannot be verified (even though it directly follows from the precondition of the current method). Unsure why this happens.
 - `ignore-test` on `tests/verify/extern-spec/traits-3.rs` - a crash related to encoding the postcondition of a trait method.
 - `tests/verify/fail/extern-spec/duplicate-1.rs` - this one also fails with the correct error message, but something is wrong with the spans, so the error is reported on line 1, even though it actually happens on line 23.

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -124,10 +124,14 @@ impl<'tcx> ErrorManager<'tcx>
...
         let pos = if let Some(primary_span) = span.primary_span() {
-            let lines_info = self
+            let lines_info_res = self
                 .codemap
-                .span_to_lines(primary_span.source_callsite())
-                .unwrap();
+                .span_to_lines(primary_span.source_callsite());
+            if lines_info_res.is_err() {
+                debug!("Error converting span to lines {:?}", lines_info_res.err());
+                return Position::new(0, 0, pos_id.clone());
```
Aurel300, line 134: Might be related to the bad error position in some of the tests.

fpoli:
> ignore-test on tests/verify/fail/extern-spec/duplicate-2.rs and tests/verify/fail/extern-spec/variants.rs - these fail as intended, but the error reporting method makes the compiler crash rather than exit gracefully. Relevant call is here, using TyCtxt::ty_error_with_message.

We could accumulate error messages in a new field of `ExternSpecResolver`, instead of raising them, such that whoever calls `get_extern_fn_map` can decide how to report error messages. It would be nice to end up using `Prusti::incorrect(..)` for that.

prusti-interface/src/specs/external.rs
```diff
@@ -0,0 +1,98 @@
...
+
+                let res = self.tcx.typeck(callee_expr.hir_id.owner).qpath_res(qself, callee_expr.hir_id);
+                if let rustc_hir::def::Res::Def(_, def_id) = res {
+                    if self.extern_fn_map.contains_key(&def_id) &&
+                            self.extern_fn_map.get(&def_id).unwrap().0 == impl_ty {
+                        // FIXME: this kind of error causes a compiler panic
+                        // after it is reported
+                        self.tcx.ty_error_with_message(ex.span.source_callsite(),
+                                                       &format!("duplicate specification for {:?}",
+                                                                def_id));
```
fpoli: `ExternSpecResolver` could accumulate duplicate specifications instead of raising errors, such that whoever calls `get_extern_fn_map` can decide whether to report error messages or to conjoin the duplicate specifications. In case an error needs to be reported, it would be best to end up calling `Prusti::incorrect(..).emit()`.

prusti-specs/src/specifications/common.rs
```diff
@@ -111,6 +111,34 @@ impl Into<u64> for ExpressionId {
...
+                return Err("expected a path".to_string());
+            }
+        };
+        let uuid = Uuid::new_v4().to_simple();
+
+        Ok(format!("PrustiStruct{}{}", path_str, uuid))
+    }
+
+    pub(crate) fn generate_mod_name(&self, ident: &syn::Ident) -> String {
+        let uuid = Uuid::new_v4().to_simple();
```
fpoli, line 137: `NameGenerator` could use an internal counter (e.g. with `RefCell`) instead of calling `Uuid::new_v4()` to generate fresh names. This is necessary (but not sufficient) to make Prusti's encoding deterministic.

prusti-interface/src/specs/external.rs
```diff
@@ -0,0 +1,98 @@
...
+use rustc_span::Span;
+
+use std::collections::HashMap;
+use crate::specs::typed::ExternSpecificationMap;
+
+pub struct ExternSpecResolver<'tcx> {
+    tcx: TyCtxt<'tcx>,
+    /// Map of def id of the real function to a tuple containing the def id of the implementing type (or none)
+    /// mapped to the fake function with the specifications. Mapped to a tuple to account for trait implementations.
+    extern_fn_map: ExternSpecificationMap<'tcx>,
```
fpoli: The comment looks very confusing to me (is it describing three nested maps?). I think it should be rewritten.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -0,0 +1,211 @@
+use crate::specifications::common::NameGenerator;
+use proc_macro2::{TokenStream, TokenTree, Group};
+use quote::{quote, ToTokens};
+use syn::ImplItemMethod;
+use syn::spanned::Spanned;
+
+pub fn rewrite_mod(item_mod: &mut syn::ItemMod, path: &mut syn::Path) -> syn::Result<()> {
```
fpoli, line 13: It would be best to have a comment (here and in the following public functions) that explains what kind of rewriting is performed.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -202,6 +205,24 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         self.spec
     }
 
+    /// Returns the def_id of the element containing the specifications.
+    /// This can be different from the def_id that was passed in if the
+    /// specifications were externally declared.
+    pub fn def_id(&self, def_id: &'v ProcedureDefId) -> &'v ProcedureDefId {
```
fpoli: This method name is too generic. Anything else, e.g. `get_specification_def_id`, would be better.

fpoli:
I left comments for some changes that I think should be made. I might even do them myself, but not sure when.

prusti-interface/src/specs/mod.rs
```diff
@@ -109,34 +109,13 @@ fn reconstruct_typed_assertion<'tcx>(
...
-                    .with(|attr_name| attr_name == "spec_only")
-        }
-        _ => false,
-    })
-}
-
-/// Check if `prusti::extern_spec` is among the attributes.
-fn has_extern_spec_attr(attrs: &[ast::Attribute]) -> bool {
+/// Check if `prusti::<name>` is among the attributes.
+fn has_prusti_attr(attrs: &[ast::Attribute], name: &str) -> bool {
```
Aurel300: Maybe we should extract this change into `master`. We also do something very similar in https://github.com/viperproject/prusti-dev/pull/176/files#diff-1fb946419baeafc95d895e440b15e66eR247

Aurel300:
> I left comments for some changes that I think should be made. I might even do them myself, but not sure when.

I'll continue working on this now.

prusti-interface/src/specs/mod.rs
fpoli: I agree!

prusti-specs/src/specifications/common.rs
Aurel300: I *think* there are two reasons UUID is used:

 - `NameGenerator` is not shared, it is recreated per `mod` and `struct`. Without a global static state there may be "internal" collisions.
 - The external spec modules and structs are renamed and rewritten, but they still exist in the namespace. Renaming the module to `original_name<UUID>` is less likely to result in name collisions with actual code than `original_name1`, `original_name2`, etc.

Nevertheless, I don't see a reason why modules can't be renamed just like `structs`, with a `PrustiMod` prefix…

> This is necessary (but not sufficient) to make Prusti's encoding deterministic.

Could you elaborate on that? Did you have other things in mind to make it "sufficient"?

fpoli: > Could you elaborate on that? Did you have other things in mind to make it "sufficient"?

I mean that Prusti is already not deterministic for reasons that are orthogonal this PR. For example, the Viper code that we currently generate often contains `closure$...` predicates that have a seemingly random integer in their name.

fpoli: I found that the UUID technique is also used by `prusti-specs` work, and I don't have a nice solution to the problem of sharing a counter between procedural macros. So, I'm fine to leave this as it is.

prusti-interface/src/specs/mod.rs
Aurel300: #180

prusti-interface/src/specs/external.rs
Aurel300: How about this:

```
    /// Maps real functions (keyed by their `def_id`) to Prusti-generated fake
    /// functions with specifications. The mapping may also optionally contain
    /// the `def_id` of the implementing type to account for trait
    /// implementations.
```

fpoli: Much better :+1:

prusti-interface/src/specs/external.rs
Aurel300: Trying to make this work now. I am not sure we can do this without reorganising the crates a bit, or maybe moving `PrustiError` elsewhere. It lives in `prusti_viper`, but we want to emit this error from `prusti_interface` still. The description for `PrustiError:incorrect` is definitely what is happening here (incorrect usage of Prusti) though. Any suggestions?

fpoli: Moving `PrustiError` (not the error manager) to `prusti_interface` sounds good to me.

Aurel300:
@fpoli Please review again. I think we could merge this, so it doesn't become stale again. The two remaining `ignore-test`s (`linked-list-i32` and `traits-3`)  only concern external specs, so we can try to fix them later.

fpoli:
> @fpoli Please review again. I think we could merge this, so it doesn't become stale again. The two remaining `ignore-test`s (`linked-list-i32` and `traits-3`) only concern external specs, so we can try to fix them later.

It looks good to me, I'll merge. The failing tests may be cause by #179.

PR #146, opened 2020-09-02
Restore structure of the test suite

fpoli:
During the upgrade we lost some functionalities of the test suite. Among small cleanups, this PR restores the old structure of tests, `tests/<group>/<mode>`, where
* `<group>` can be `parse`, `typecheck`, `filter`, `verify`, `verify_overflow` or `core_proof`. Each group runs Prusti with a different set of configuration flags.
* `<mode>` can be `ui`, `pass` or `fail`. There are the same modes defined in [`compiletest-rs`](https://github.com/laumann/compiletest-rs/blob/4d129a6d854621e0798a1831b51fc73d014056d8/src/common.rs#L60-L74).

Moreover, to avoid some unnecessary dependencies between crates:
* the tests are moved from the `prusti` crate to the new `prusti-tests`,
* the `prusti-rustc` and `prusti-server` binaries are moved to the `prusti-launch` crate.


fpoli:
@vakaras @cmatheja @Aurel300 So far I have been merging my PRs waiting without much review. However, feel free to do so and to leave comments, even after the PR is merged. (I prefer to only wait for CI. I feel that enforcing a mandatory review pass for us is counterproductive, if we are confident in our changes.)

prusti-launch/src/bin/cargo-prusti.rs
```diff
@@ -0,0 +1,39 @@
+//  © 2020, ETH Zurich
+//
+//  This Source Code Form is subject to the terms of the Mozilla Public
+//  License, v. 2.0. If a copy of the MPL was not distributed with this
+//  file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
Aurel300: Two spaces after `//`? Same in the other copyright notices. In one file you actually replaced the same notice but added the space.

prusti-tests/tests/compiletest.rs
```diff
@@ -0,0 +1,184 @@
...
+    let path: PathBuf = ["tests", group_name, "fail"].iter().collect();
+    if path.exists() {
+        config.mode = common::Mode::CompileFail;
+        config.src_base = path;
+        run_tests(&config);
+    }
+}
+
+fn run_no_verification(group_name: &str, filter: &Option<String>) {
+    // Automatically restored to their original value at the end of the lexical scope
```
Aurel300: Seems like a comment for the `TemporaryEnvVar` struct itself, not places where it is used, especially since it is repeated many times.

prusti-tests/tests/compiletest.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    // Test the parsing of specifications. This doesn't run the verifier.
+    println!("[parse]");
+    run_no_verification("parse", &filter);
+
+    // Test the type-checking of specifications. This doesn't run the verifier.
+    println!("[typecheck]");
+    run_no_verification("typecheck", &filter);
+
+    // Test the error messages of prusti-filter
```
Aurel300:
```suggestion
    // Test the error messages of prusti-filter.
```

prusti-tests/tests/compiletest.rs
fpoli: Nice catch, thanks

prusti-launch/src/bin/cargo-prusti.rs
fpoli: Nice catch, thanks

prusti-tests/tests/compiletest.rs
fpoli: :+1:

prusti-tests/tests/compiletest.rs
fpoli: :+1:

PR #149, opened 2020-09-03
Refactor process_expire_borrows.

frececroka:
I moved the `process_expire_borrows` functions to a separate module and broke it up into multiple functions. Unless I messed up, this should not introduce any functional changes.

I will have to change the way `process_expire_borrows` works soon and I find the refactored version easier to work with. But of course this is subjective, and if the changes are not considered an improvement I don't mind if the PR is closed.


fpoli:
> I will have to change the way process_expire_borrows works soon and I find the refactored version easier to work with. But of course this is subjective, and if the changes are not considered an improvement I don't mind if the PR is closed.

I suggest to enable as many non-failing tests as possible (by moving them from the `prusti-test/test_old` to the `prusti-test/test` folder) before modifying the behaviour. We are restoring the test suite, but I'm not sure we'll be fast enough for you.

frececroka:
Yes this is a very good idea. I hope I'll be ready to test my changes by tomorrow or Monday.

The changes in this PR should be relatively low-risk though, because I really just moved code around.

PR #158, opened 2020-09-14
Minor fixes

Aurel300:
- use `box` syntax instead of `Box::new`
 - fix the `FIXME`s for `get_spans` in `AssertionKind`
 - use the new `strip_prefix` function in a server test

### DIFF
--- a/prusti-interface/src/specs/typed.rs
+++ b/prusti-interface/src/specs/typed.rs
@@ -48,7 +48,13 @@ pub type Trigger = common::Trigger<ExpressionId, LocalDefId>;
 /// A pledge in the postcondition.
 pub type Pledge<'tcx> = common::Pledge<ExpressionId, LocalDefId, (mir::Local, ty::Ty<'tcx>)>;
 
+/// This trait is implemented for specification-related types that have one or
+/// more associated spans (positions within the source code). The spans are not
+/// necessarily contiguous, and may be used for diagnostic reporting.
 pub trait Spanned<'tcx> {
+    /// Returns the spans for the given value. `mir` is the function body used
+    /// to resolve positions of `rustc_middle::mir::Local` indices, `tcx` is
+    /// used to resolve positions of global items.
     fn get_spans(&self, mir_body: &mir::Body<'tcx>, tcx: TyCtxt<'tcx>) -> Vec<Span>;
 }
 
--- a/prusti-interface/src/specs/typed.rs
+++ b/prusti-interface/src/specs/typed.rs
@@ -49,37 +49,55 @@ pub type Trigger = common::Trigger<ExpressionId, LocalDefId>;
 pub type Pledge<'tcx> = common::Pledge<ExpressionId, LocalDefId, (mir::Local, ty::Ty<'tcx>)>;
 
 pub trait Spanned<'tcx> {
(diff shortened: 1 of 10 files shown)


prusti-interface/src/specs/typed.rs
```diff
@@ -49,37 +49,55 @@ pub type Trigger = common::Trigger<ExpressionId, LocalDefId>;
 pub type Pledge<'tcx> = common::Pledge<ExpressionId, LocalDefId, (mir::Local, ty::Ty<'tcx>)>;
 
 pub trait Spanned<'tcx> {
-    fn get_spans(&self, tcx: TyCtxt<'tcx>) -> Vec<Span>;
+    fn get_spans(&self, mir_body: &mir::Body<'tcx>, tcx: TyCtxt<'tcx>) -> Vec<Span>;
```
fpoli, line 58: Could you add a comment to explain the parameters?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3831,15 +3831,15 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                                &self.mir,
                                 PRECONDITION_LABEL,
                                 &encoded_args,
                                 None,
                                 false,
                                 Some(loop_inv_block),
                                 ErrorCtxt::GenericExpression,
                             );
-                            let spec_spans = typed::Spanned::get_spans(assertion, self.encoder.env().tcx());
+                            let spec_spans = typed::Spanned::get_spans(assertion, &self.mir, self.encoder.env().tcx());
```
fpoli, line 3842: If this is a bugfix, is it easy to add a test to avoid future regressions?

prusti-interface/src/specs/typed.rs
Aurel300: For both of your comments: this argument was added to be able to get spans from `mir::Local`s. A local variable is an index that only makes sense within its MIR function and the only way to get position information is to check with the MIR body. Fortunately, all places that called `get_spans` already had a reference to the MIR body. I'll add a comment.

prusti-viper/src/encoder/procedure_encoder.rs
Aurel300: I don't think span positions are tested. The change to `get_spans` may just slightly improve error reporting?

prusti-interface/src/specs/typed.rs
fpoli: I see, thanks!

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: In that case, we should see a difference in a "ui" test.

Aurel300: We should, although the current test suite does not seem to have such a case. As you can see, the CI passes. To be honest, I am not completely sure how to cause an error that would be reported on the whole `ForAll` assertion, rather than its body.

fpoli: I see, I thought you did this to fix something in the output. As far as I know we always report failing `ForAll` pointing to their body, just because it's easy to reuse and merge together the spans that come from the compiler.

PR #159, opened 2020-09-15
Enable tests

fpoli:
* Update the syntax of annotations of the tests in the `prusti-tests/tests_old` folder.
* Enable non-failing tests by moving them to `prusti-tests/tests`.

To do:
- [x] Squash & reword commits to clean up the history


prusti-tests/tests/verify/fail/assert-false/error-position.rs
```diff
@@ -0,0 +1,49 @@
+use prusti_contracts::*;
+
+#[ensures(false)] //~ ERROR postcondition
```
Aurel300, line 3: Should all of these `//~ ERROR postcondition` not include the whole `postcondition might not hold` string?

prusti-tests/tests/typecheck/fail/specification_result_precondition.rs
```diff
@@ -1,8 +1,8 @@
 // This test checks that preconditions don't contain result (Issue #212)
 
-extern crate prusti_contracts;
+use prusti_contracts::*;
 
-#[requires="result > 0"]  //~ ERROR
+#[requires(result > 0)]  //~ ERROR
```
Aurel300, line 5: Might want to match the actual error raised?

prusti-tests/tests/verify/pass/loop-invs/nested-loops-3.rs
```diff
@@ -0,0 +1,41 @@
+use prusti_contracts::*;
+
+#[requires(n >= 0)]
+//#[ensures(result == old(n) * (1 + old(n) * (1 + old(n))))]
```
Aurel300, line 4: Why is this (+ the invariants below) commented out?

prusti-tests/tests_old/parse/ui/attribute_macros.stderr
```diff
@@ -13,7 +13,7 @@ error: malformed specification (expected single argument)
 error: malformed specification (expected token)
   --> $DIR/attribute_macros.rs:15:5
    |
-15 |     #[invariant["true"]]
+15 |     #[invariant["true)]]
```
Aurel300, line 16: Seems like a rogue regex replacement. I see this is in `tests_old`, but still. This file also has assertions in strings still.

prusti-tests/tests_old/parse/ui/simple_and.rs
```diff
@@ -1,11 +1,11 @@
...
 
-extern crate prusti_contracts;
+use prusti_contracts::*;
 
-#[requires="12345"]
+#[requires(12345)]
 pub fn test1a(x: i32) {}
 
-#[requires=     "12345"]
+#[requires=     "12345)]
```
Aurel300, line 8: Left quote not replaced.

prusti-tests/tests_old/test.rs
```diff
@@ -10,16 +10,16 @@ use std::{
 
 fn get_prusti_rustc_path() -> PathBuf {
     let local_prusti_rustc_path: PathBuf = if cfg!(windows) {
-        ["target", "debug", "prusti-rustc.exe"].iter().collect()
+        ["target", "debug", "prusti-rustc.exe)].iter().collect()
```
Aurel300, line 13: All of the replacements in this file are wrong. Maybe this file is actually good for deletion though.

prusti-tests/tests/verify/fail/assert-false/error-position.rs
fpoli: I'll add it, why not. I think we initially just wanted to test that the program does not verify, but now we can strength the check.

prusti-tests/tests/verify/pass/loop-invs/nested-loops-3.rs
fpoli: Good question. Maybe because Prusti was very slow with it, I'll check.

prusti-tests/tests_old/verify/pass/loop-invs/loop-fields.rs
```diff
@@ -8,14 +8,14 @@ struct H {
...
+#[requires(a.g.f < 5)]
+#[ensures(result.f == 5)]
 fn test1(a: H) -> T {
     let mut a = a;
     let mut cont = true;
-    #[invariant="a.g.f < 5"]
-    #[invariant="cont == (a.g.f < 5)"]
     while cont {
+        body_invariant(a.g.f < 5);
+        body_invariant(cont == (a.g.f < 5));
```
Aurel300, line 18: `body_invariant!` (missing exclamation mark)

prusti-tests/tests_old/verify/pass/loop-invs/loop_in_guard.rs
```diff
@@ -1,13 +1,13 @@
...
 fn test_loop_in_guard() {
     let mut i = 0;
 
-    #[invariant="i % 10 == 0 && i < 60"]
     while {
+        body_invariant(i % 10 == 0 && i < 60);
         let old_i = i;
-        #[invariant="i < old_i + 10"]
         while i < old_i + 10 {
+            body_invariant(i < old_i + 10);
```
Aurel300, line 10: Same.

prusti-tests/tests_old/parse/ui/attribute_macros.stderr
fpoli: True. I'll revert the change, but I'm not updating the annotations to the new syntax because I think it's easier to regenerate them by running Prusti and manually checking the output.

prusti-tests/tests_old/verify/pass/loop-invs/simple-loop-3.rs
```diff
@@ -1,31 +1,31 @@
-extern crate prusti_contracts;
+use prusti_contracts::*;
 
 fn test1() {
     let mut i = 0;
-    #[invariant="i < 10"] // ---+
+    #[invariant(i < 10)] // ---+
```
Aurel300, line 5: I guess this should become a `body_invariant!` too? (Also line 28)

prusti-tests/tests_old/verify/pass/loop-invs/simple-nested.rs
```diff
@@ -1,17 +1,17 @@
...
 
-    #[invariant="res == ia"]
-    #[invariant="ia < n"]
+    #[invariant(res == ia)]
+    #[invariant(ia < n)]
     while ia < n {
         res += 1;
 
-        #[invariant="true"]
+        #[invariant(true)]
```
Aurel300, line 14: `body_invariant!` ?

prusti-tests/tests_old/test.rs
fpoli: Uh, what a mess. Thanks! I'll just delete this file, as it's never used.

prusti-tests/tests_old/verify/ui/failing-postcondition.stderr
```diff
@@ -1,7 +1,7 @@
 error: [Prusti: verification error] postcondition might not hold.
  --> $DIR/failing-postcondition.rs:8:12
   |
-8 | #[ensures="something_true() && false"]
+8 | #[ensures(something_true() && false)]
   |            ^^^^^^^^^^^^^^^^    ^^^^^
```
Aurel300, line 5: Did not realise in the previous ones, but the position of the arrows should actually change in this case. (True for the `.stderr` files in general.)

Aurel300:
There are still a lot of cases of a `while` loop with `#[invariant(...)]` rather than `body_invariant!(...);`, I did not comment on most of them.

Also, I had the feeling when going through the `tests_old` files that some of the tests were duplicates of ones already in `tests`. Maybe worth looking into to reduce the clutter.

prusti-tests/tests/verify/pass/loop-invs/nested-loops-3.rs
fpoli: It fails for some reason. I'll make a new test out of it.

prusti-tests/tests_old/parse/ui/simple_and.rs
fpoli: As above

fpoli:
> There are still a lot of cases of a while loop with #[invariant(...)] rather than body_invariant!(...);, I did not comment on most of them.

Thanks. I must have grepped for `#[invariant(` in the wrong folder.

> Also, I had the feeling when going through the tests_old files that some of the tests were duplicates of ones already in tests. 

It's possible, we never tried to deduplicate.

> Maybe worth looking into to reduce the clutter.

I agree. Right now, however, I prefer to focus on enabling as much as them as possible.

PR #168, opened 2020-09-17
Fix and enable tests

fpoli:


prusti-tests/tests/parse/ui/alternatives.rs
```diff
@@ -0,0 +1,14 @@
+use prusti_contracts::*;
+
+#[ensures()]
+fn no_args() {}
+
+#[requires[true]]
+fn square_brackets() {
+    let mut i=0u32;
```
Aurel300:
```suggestion
    let mut i = 0u32;
```

Might as well?

prusti-tests/tests/parse/ui/simple_forall.stderr
```diff
@@ -0,0 +1,14 @@
+error: expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime
```
Aurel300, line 1: (For a different PR, but) maybe this error should be more specific? After the colon there, we should not be able to see most of what it lists.

fpoli: Could you open an "enhancement" issue for that? I'm not sure how hard it would be to implement.

PR #171, opened 2020-09-17
Fix detection of panic cause

fpoli:


prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -743,67 +743,30 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
-                                                    .macro_def_id
-                                                    .unwrap();
-                                            // HACK to match the filename of the span
-                                            let second_def_site_span = format!(
-                                                "{:?}",
-                                                macro_backtrace[1]
-                                                    .def_site
-                                            );
+                            // Look at the top 4 macro calls
+                            let lookup_size = 4;
```
Aurel300: Why 4?

fpoli: I rephrased as "To classify the cause of the panic it's enough to look at the top 3 macro calls"

Issue #162, opened 2020-09-16
Loop invariants don't work


fpoli:
A regression of the compiler upgrade is that loop invariants are ignored. For example, the following verifies, but it should generate a verification error.

```rust
use prusti_contracts::*;

pub fn simple_loop() {
    let mut x = 0;
    while x < 100 {
        body_invariant!(false); // Should fail
        x += 1;
    }
}

fn main() {}
```

Aurel300:
I'll try to investigate this. So far it seems that `ProcedureEncoder::get_loop_spec_blocks` returns an empty vector even if the desugared Rust shows the invariant generated.

Aurel300:
https://github.com/viperproject/prusti-dev/commit/813c127ae3da75a1c8ef0fb087367da48cb9931a#diff-1fb946419baeafc95d895e440b15e66eR280 might be the problem.

vakaras:
The problem is in [the visitor](https://github.com/viperproject/prusti-dev/blob/58c11b21d29aa9433a118008f768d4bc4b0bd2a9/prusti-interface/src/specs/mod.rs#L36) that collects specifications. That visitor currently collects specs only for items.

fpoli:
After PR #176 Prusti crashes instead of ignoring the invariants.

PR #176, opened 2020-09-21
Filter specification blocks

fpoli:


fpoli:
Cherry-picking 234f708 made all tests that contain a `body_invariant!(..)` fail, but loop invariants were not actually verified. So, I'm disabling those tests. This PR will probably affect #162.

PR #177, opened 2020-09-21
Make prusti-spec rewrite all attributes at once

### LINKED ISSUE #167: Incorrect span of failing pure function postcondition
The following program has a verification error that is reported in the wrong place.

```rust
use prusti_contracts::*;

#[pure]
#[requires(n > 0)]
#[ensures(result == 5)] // <-- The error should be reported here
fn magic(n: i32) -> i32 {
    n
}

fn main() {}
```

Output:
```
error: [Prusti: verification error] postcondition of pure function definition might not hold
 --> prusti-tests/tests_old/verify/fail/issues/issue-81.rs:1:1
  |
1 | / use prusti_contracts::*;
2 | |
3 | | #[pure]
4 | | #[requires(n > 0)]
  | |__________________^
  |
note: the error originates here
 --> prusti-tests/tests_old/verify/fail/issues/issue-81.rs:1:1
  |
1 | / use prusti_contracts::*;
2 | |
3 | | #[pure]
4 | | #[requires(n > 0)]
  | |__________________^
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

fpoli:
Make prusti-spec rewrite all attributes at once (fixes #167)


prusti-specs/src/lib.rs
```diff
@@ -21,34 +26,97 @@ macro_rules! handle_result {
...
+    // Start with the outer attribute
+    let mut prusti_attributes = vec![
+        (outer_attr_kind, outer_attr_tokens)
+    ];
+
+    // Collect the remaining Prusti attributes, removing them from `item`.
+    let remaining_prusti_attributes = item.attrs.drain_filter(
+        |attr|
+            attr.path.segments.len() == 1
+                && SpecAttributeKind::try_from(attr.path.segments[0].ident.to_string()).is_ok()
```
Aurel300, line 49: Does this not result in a (maybe unimportant) difference in behaviour? Originally Prusti code would only be triggered from its own attributes. What if I only imported `requires` from Prusti, but had `pure` from a different crate? Here the assumption is that attributes that have the same name as Prusti attributes are definitely the ones from Prusti.

prusti-specs/src/lib.rs
```diff
@@ -21,34 +26,110 @@ macro_rules! handle_result {
...
+    )?;
+    Ok((
+        vec![spec_item],
+        vec![parse_quote!(#[prusti::pre_spec_id_ref = #spec_id_str])],
+    ))
 }
 
-pub fn ensures(attr: TokenStream, tokens: TokenStream) -> TokenStream {
-    let item: syn::ItemFn = handle_result!(syn::parse2(tokens));
+/// As required by the "ensures" annotation, generate spec items and attributes.
```
Aurel300: I'm not sure how useful these comments are.

prusti-specs/src/lib.rs
fpoli: Indeed, what you describe might happen. We haven't found a better solution. The correct way would be to let the compiler resolve and execute each procedural macro individually, but that makes the compiler lose all span information (bug https://github.com/rust-lang/rust/issues/43081, open since 2017), which is not acceptable for us.

prusti-specs/src/lib.rs
fpoli: I'm too not happy about them

prusti-specs/src/lib.rs
fpoli: Another case (by Vytautas) which we would not handle correctly is the following.

```rust
use prusti_contracts::requires as pre;

#[pre(true)]
#[pre(true)]
#[pre(false)] // <-- This will not be handled correctly by Prusti.
fn test() { ... }
````
Here, the compiler would still execute the second and third `pre` as procedural macros, losing the span information as it happens in #167.

Comment (Aurel300) [reply prusti-specs/src/lib.rs:49]: Could we require all attributes after the first Prusti one to be valid and trigger errors if we don't recognise them?

Comment (fpoli) [reply prusti-specs/src/lib.rs:49]: That's a good idea!

Comment (fpoli) [reply prusti-specs/src/lib.rs:49]: Unfortunately, we don't  have information on which attributes come after or before a procedural macro attribute. Disallowing any other kind of attributes (including e.g. `#[allow(..)]` and doc comments) would be too restrictive.

Comment (Aurel300) [reply prusti-specs/src/lib.rs:49]: That's a pity. I assumed they would be processed "left to right", i.e. the first proc macro attribute that is processed would only see the attributes that follow it.

Comment (fpoli) [reply prusti-specs/src/lib.rs:49]: I thought the same, but apparently all past attributes are still in the `item.attrs` field.

File: prusti-specs/src/lib.rs [hunk 3]
```diff
@@ -60,18 +60,22 @@ pub fn rewrite_prusti_attributes(
...
     );
 
-    quote!{
+    let res = quote!{
         #(#generated_spec_items)*
         #(#generated_attributes)*
         #item
-    }
+    };
+    println!("Generated: {}", res);
```
Aurel300: Was this for debugging?

fpoli: Yep.

prusti-specs/src/lib.rs
fpoli: It should be slightly better now.

fpoli:
The disabled test is because `prusti-specs` only supports annotating items that are free-standing functions. Before this PR, `trusted` and `pure` (but not `requires` or `ensures`) used to work fine also on other items because there was no need to iterate over attributes, which the new technique now requires. Lifting this limitation is for another issue/PR.

PR #180, opened 2020-09-22
has_prusti_attr and has_spec_only_attr

Aurel300:
- added `prusti_interface::utils::{has_prusti_attr, has_spec_only_attr}`
 - cleaned up places where a duplicate function was found

There are still places where attributes are checked similarly, but not quite the same. `prusti_interface::specs::read_prusti_attr` checks for attributes with arguments, `prusti_viper::encoder::Encoder::get_opt_spec_id` likewise.

Once merged I will rebase #145.

### DIFF
--- a/prusti-interface/src/utils.rs
+++ b/prusti-interface/src/utils.rs
@@ -263,6 +263,34 @@ impl<'tcx> VecPlace<'tcx> {
     }
 }
 
+/// Check if `prusti::<name>` is among the attributes.
+/// The attribute must not have any arguments.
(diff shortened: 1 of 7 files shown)


prusti-interface/src/utils.rs
```diff
@@ -263,6 +263,34 @@ impl<'tcx> VecPlace<'tcx> {
     }
 }
 
+/// Check if `prusti::<name>` is among the attributes.
+/// The attribute must not have any arguments.
```
fpoli, line 267: What happens if the attribute has arguments? The function returns `false`?

Aurel300: Yes.

Issue #187, opened 2020-09-28
First class support for type parameters in VIR


fpoli:
Currently, the `vir` module in which statements and expressions are defined has no notion of type parameters. To support Rust code that uses type parameters, Prusti uses a fragile regexp-based approach implemented in `prusti-common/src/vir/ast/typaram.rs`, plus a variety of hacks that can be found by searching for one of the following comments.

* `// FIXME; hideous monstrosity...`
* `// FIXME oh dear...`
* `// FIXME: the following fields serve a grotesque hack.`
* `// FIXME: A hack. Replaces all generic types with their instantiations`

To avoid the regular expressions and the hacks, we could extend `vir::Expr` and `vir::Type` to make them aware of type parameters and type substitutions.

vakaras:
I think we should check whether it would be cleaner to have two layers of VIR: a generic VIR and monomorphic VIR that can be directly translated into Viper.

Aurel300:
Can you see other issues that can be addressed by having the poly-/monomorphic variants?

Aurel300:
I'll close this. We no longer have the fragile regexp-base hack in the codebase (since #899 and a bunch of work by @vl0w), so generics are much more robust, although associated types might still take some more work (possible changes in #980). I also think having "first-class type parameters" in the VIR is the wrong decision: this is the job of the compiler. VIR poly thus should (and does not) have type parameters. @vakaras might be using them in the unsafe core proof VIR layers for dealing with lifetimes.

PR #191, opened 2020-09-29
Loop invariants

### LINKED ISSUE #162: Loop invariants don't work

fpoli:
Fixes #162.

This PR:
* changes the desugaring of loop invariants from `let _prusti_loop_invariant = { .. }` to a closure `|| { .. }`,
* adds a `replace_multiple_places` method to `Expr`,
* simplifies the implementation of `SpecCollector`,
* refactors `spec_encoder.rs` and add a proper documentation to `encode_spec_assertion(..)`,
* removes the `namespace` field from `MirEncoder`.

To do:
* [x] Fix the remaining tests that fail due to replacements of expressions with type parameters (related: #187).
* [x] Remove the `PRUSTI_FALLBACK` feature flag and the `old_encode_expression` function, both used for debugging.
* [x] Remove the `namespace` field from `MirEncoder`.
* [x] Enable tests related to loops.
* [x] Add support for nested quantifiers


prusti-common/src/vir/ast/expr.rs
```diff
@@ -1025,7 +1025,136 @@ impl Expr {
...
+                    dst,
+                    src.get_type(),
+                    dst.get_type()
+                );
+            }
+        }
+
+        struct PlaceReplacer<'a> {
+            replacements: &'a [(Expr, Expr)],
+            // FIXME: the following fields serve a grotesque hack.
```
Aurel300, line 1050: Maybe we should replace all of these "grotesque hack" and similar vague descriptions with links to #187 now that it exists?

prusti-interface/src/utils.rs
```diff
@@ -297,28 +297,43 @@ pub fn has_extern_spec_attr(attrs: &[ast::Attribute]) -> bool {
...
+            use rustc_ast::token::Lit;
+            use rustc_ast::token::Token;
+            use rustc_ast::token::TokenKind;
+            use rustc_ast::tokenstream::TokenTree;
+            match &tokens.0[0].0 {
+                TokenTree::Token(Token {
+                                     kind: TokenKind::Literal(Lit { symbol, .. }),
+                                     ..
+                                 }) => {
+                    strings.push(symbol.as_str().replace("\\\"", "\""))
```
Aurel300, line 326: This seems a bit fragile?

prusti-interface/src/utils.rs
```diff
@@ -297,28 +297,43 @@ pub fn has_extern_spec_attr(attrs: &[ast::Attribute]) -> bool {
...
-//                 }
-//                 _ => unreachable!(),
-//             },
-//             _ => unreachable!(),
-//         },
-//         _ => unreachable!(),
-//     }
-// }
+/// Read the value stored in a single Prusti attribute (e.g. `prusti::<attr_name>="...")`.
+/// Panics if there is more than one such attribute.
```
Aurel300, line 336: Does it?

prusti-interface/src/utils.rs
fpoli: I agree, but I don't have a quick solution. In this PR I just changed the return type from `Option<_>` to `Vec<_>`, moving the function out of `prusti-interface/src/specs/mod.rs`.

prusti-interface/src/utils.rs
fpoli: No, I forgot to update the documentation.

PR #192, opened 2020-09-29
Try to fix #188.

### LINKED ISSUE #188: Prusti crashes while creating a stub due to non-local proc_def_id
Currently, Prusti crashes on the program below while attempting to create a *stub* for `partial_cmp`. The issue originates [here](https://github.com/viperproject/prusti-dev/blob/master/prusti-interface/src/environment/procedure.rs#L40) when creating a new `Procedure`: the `def_id` corresponding to `partial_cmp` is not local.

Do you know a workaround for this? It seems the only method in `TyCtxt` providing MIR that does not require a `LocalDefId` is [`optimized_mir`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.optimized_mir).

```rust
use prusti_contracts::*;
use std::cmp::Ordering;

#[derive(PartialEq)]
struct A {
    i: i32,
}

impl PartialOrd for A {
    fn partial_cmp(&self, other: &A) -> Option<Ordering> {
        self.i.partial_cmp(&other.i)
    }
}

#[requires(x < y)]
fn test(x: A, y: A) {
}
```

vakaras:
@cmatheja With this change, Prusti does not crash anymore but gives the following error for your example:

```plain
[Prusti: invalid specification] use of impure function "std::cmp::PartialOrd::lt" in assertion is not allowed
```

### DIFF
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -1632,8 +1632,10 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         mut proc_def_id: ProcedureDefId,
     ) -> (String, vir::Type) {
         proc_def_id = *self.get_specification_def_id(&proc_def_id);
-        let procedure = self.env.get_procedure(proc_def_id);
-        let encoder = StubFunctionEncoder::new(self, proc_def_id, procedure.get_mir());
+        // The stub function may come from another module for which we can have
+        // only optimized_mir.
+        let body = self.env.tcx().optimized_mir(proc_def_id);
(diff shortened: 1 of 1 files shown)


cmatheja:
Thanks, this error is to be expected as `PartialOrd::lt` is indeed neither pure nor is there special support for it (yet).
As long as it is indeed safe to switch to optimized mir, I can take care of the rest.

vakaras:
> As long as it is indeed safe to switch to optimized mir, I can take care of the rest.

`StubFunctionEncoder` uses MIR only to get the types of the arguments, so optimized MIR should work here.

PR #193, opened 2020-09-29
Parse trait specs.

### LINKED ISSUE #179: Parsing of Prusti attributes on methods
Since the 2018->2020 compiler upgrade, `prusti-spec` doesn't support `requires` and `ensures` attributes on methods in traits and in type implementations. After #177, the same happens for `pure` and `trusted`.

vakaras:
A work in progress to fix #179

### DIFF
--- a/prusti-viper/src/encoder/procedure_encoder.rs
+++ b/prusti-viper/src/encoder/procedure_encoder.rs
@@ -1774,6 +1774,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 ..
             } => {
                 if let ty::TyKind::FnDef(def_id, substs) = ty.kind() {
+                    let self_ty = {
+                        // If we are calling a trait method on a struct, self_ty
+                        // is the struct.
+                        let generics = self.encoder.env().tcx().generics_of(*def_id);
+                        if generics.has_self {
+                            Some(substs.type_at(0))
+                        } else {
+                            None
+                        }
+                    };
+
                     let def_id = *self.encoder.get_specification_def_id(def_id);
                     let full_func_proc_name: &str =
                         &self.encoder.env().tcx().def_path_str(def_id);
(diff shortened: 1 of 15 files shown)


fpoli:
Why the `verify-test` command? If it's for debugging we could put scripts like that in a `scripts` folder, to keep `x.py` lean. I also have some scripts which I could add, for example to move passing tests from `tests_old` to `tests`.

vakaras:
> Why the `verify-test` command? If it's for debugging we could put scripts like that in a `scripts` folder

`./x.py test <test>` prints error messages in json, which is very hard to read. So, running a specific test in a “normal” way seems to be a command that one would use often. Especially, if it also configures the necessary flags correctly like checking for overflows (not done yet).

> to keep `x.py` lean.

I would like to have `x.py` as a main entry point because it makes much easier to discover commands. If the file grows too big, we can just split into multiple Python modules.

> I also have some scripts which I could add, for example to move passing tests from `tests_old` to `tests`.

I think it would be useful if you shared them. We can delete them after we have finished the migration completely.

fpoli:
> `./x.py test <test>` prints error messages in json, which is very hard to read. So, running a specific test in a “normal” way seems to be a command that one would use often. Especially, if it also configures the necessary flags correctly like checking for overflows (not done yet).

I see the plan, the configuration flags would be useful.

> I think it would be useful if you shared them

Here: https://gist.github.com/fpoli/291f2a1dbabe33f0b0caed380123f69d. I'll add it in the next PR.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2064,6 +2076,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 args,
                 destination,
                 called_def_id,
+                None, // FIXME: This is almost definitely wrong.
```
Aurel300, line 2079: I think it would be good to mention why?

PR #204, opened 2020-10-01
Continuous deployment

fpoli:
In this PR:
* Fix the CI script, to actually test Prusti on Windows and MacOS and not 3 times on Ubuntu. :disappointed:
* Add a GitHub workflow to build and upload Prusti artifacts to a new GitHub release each time a commit is pushed to the master branch.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -83,7 +83,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./prusti-release-ubuntu/prusti.zip
+          asset_path: ./prusti-release-ubuntu-latest/prusti.zip
           asset_name: prusti-release-ubuntu.zip
           asset_content_type: application/zip
       - name: Upload release asset for Windows
@@ -92,7 +92,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./prusti-release-windows/prusti.zip
+          asset_path: ./prusti-release-windows-latest/prusti.zip
           asset_name: prusti-release-windows.zip
           asset_content_type: application/zip
       - name: Upload release asset for MacOS
@@ -101,6 +101,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./prusti-release-macos/prusti.zip
+          asset_path: ./prusti-release-macos-latest/prusti.zip
           asset_name: prusti-release-macos.zip
           asset_content_type: application/zip
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -58,10 +58,10 @@ jobs:
       - name: Package Prusti artifacts
         shell: bash
         run: |
-          for os in ubuntu windows macos
+          for os in ubuntu-latest windows-latest macos-latest
           do
             echo "Package Prusti artifact for $os"
-            cd prusti-release-$os-latest
+            cd prusti-release-$os
             mv target/release/* .
             rm -rf target
             zip -r prusti.zip *
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,106 @@
+name: Continuous deployment
+
+on:
+  push:
+    branches:
+      - master
+
+env:
+  RUST_BACKTRACE: 1
+  PRUSTI_ASSERT_TIMEOUT: 60000
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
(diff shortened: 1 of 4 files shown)


fpoli:
Please give a look at the `deploy.yml`, as I can't test it locally.

fpoli:
Could someone with MacOS debug why Prusti doesn't compile on that OS?

Aurel300:
It's strange, it compiled for me on Mac fine so far. I guess the difference is due to dependencies or the new toolchain?

fpoli:
> It's strange, it compiled for me on Mac fine so far. I guess the difference is due to dependencies or the new toolchain?

Yes, it might be due to this change in a dependency: https://github.com/jni-rs/jni-rs/pull/230

x.py
```diff
@@ -77,7 +77,7 @@ def get_mac_env():
     if os.path.exists(java_home):
         ld_library_path = None
         for root, _, files in os.walk(java_home):
-            if 'libjvm.dylib' in files:
+            if 'libjli.dylib' in files:
```
Aurel300, line 80: Also the error message on line 84.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: false
+      - name: Upload release asset for Ubuntu
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./prusti-release-ubuntu./prustizip
```
Aurel300:
```suggestion
          asset_path: ./prusti-release-ubuntu/prusti.zip
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+          asset_name: prusti-release-ubuntu.zip
+          asset_content_type: application/zip
+      - name: Upload release asset for Windows
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./prusti-release-windows/prusti.zip
```
Aurel300:
```suggestion
          asset_path: ./prusti-release-windows/prusti.zip
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+          asset_name: prusti-release-windows.zip
+          asset_content_type: application/zip
+      - name: Upload release asset for MacOS
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./prusti-release-macos.z/prustiip
```
Aurel300:
```suggestion
          asset_path: ./prusti-release-macos/prusti.zip
```

(What happened here :D)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+    steps:
+      - name: Download all Prusti artifacts
+        uses: actions/download-artifact@v2
+      - name: Package Prusti artifacts
+        shell: bash
+        run: |
+          for os in ubuntu windows macos
+          do
+            echo "Package Prusti artifact for $os"
+            cd prusti-release-$os-latest
```
Aurel300:
```suggestion
            cd prusti-release-$os
```

IIUC this has to be the same as the ones from upload artifact?

.github/workflows/deploy.yml
fpoli: What's the difference?

.github/workflows/deploy.yml
fpoli: Yes, it has to be the same. Line 61 doesn't have the `-latest` suffix.

.github/workflows/deploy.yml
Aurel300: Sorry, I added `-latest` to all of the `asset_path`s before I realised that's probably wrong, then I edited my suggestions and didn't realise that this one was now correct.

.github/workflows/deploy.yml
Aurel300: Right, but doesn't this have to match line 38? (Which also doesn't have the suffix?)

fpoli: > Right, but doesn't this have to match line 38? 

Yes.

> (Which also doesn't have the suffix?)

`matrix.os` iterates over line 16, so the suffix should be there.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+            cd ..
+          done
+      - name: Create GitHub release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
```
Aurel300, line 77: So this named it `Release refs/heads/master`, which is not ideal. I guess we should change this to `github.sha` to get the commit hash?

https://github.com/viperproject/prusti-dev/releases/tag/refs%2Fheads%2Fmaster

PR #211, opened 2020-10-05
Test crates

fpoli:
Work in progress for a test that runs Prusti on a collection of crates.

To do:
- [ ] (optional) make `cargo-prusti` skip verification of dependencies.


fpoli:
It seems that only 2 of the top-50 crates compile fine with the flags that Prusti uses.

fpoli:
I'm going to merge this tomorrow. We can skip the verification of dependencies later.

PR #213, opened 2020-10-06
Clean up and fix releases

fpoli:
* Remove unused script and ERROR_ON_PARTIALLY_SUPPORTED flag.
* Mark as pre-release the automatic deployments
* Fix globs to match .exe files on Windows.


fpoli:
We could add one more check after the `build` job to be sure that the files in the artifact are enough to run Prusti.

PR #214, opened 2020-10-07
Add an analysis crate

fpoli:
The plan is to move the existing liveness and definitely-initialized analysis to the `analysis` crate, avoiding code duplication whenever possible. We'll then add the PCS analysis to the `analysis` crate, using the result of that analysis to generate `fold/unfold` statements in Prusti's encoding.

### DIFF
--- a/analysis/src/analyzer.rs
+++ b/analysis/src/analyzer.rs
@@ -30,7 +30,7 @@ impl<'tcx> Analyzer<'tcx> {
         unimplemented!();
     }
 
-    pub fn definitely_initilized_analysis(
+    pub fn definitely_initialized_analysis(
         &self,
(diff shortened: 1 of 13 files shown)


analysis/src/analyzer.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    pub fn liveness_analysis(
+        &self,
+        mir: &mir::Body<'tcx>,
+    ) -> Result<PointwiseState<LivenessState>> {
+        unimplemented!();
+    }
+
+    pub fn definitely_initilized_analysis(
```
Aurel300:
```suggestion
    pub fn definitely_initialized_analysis(
```

PR #216, opened 2020-10-07
Fix after_expiry parsing.

vakaras:


prusti-tests/tests_old/verify/pass/quick/with-spec-list.rs
```diff
@@ -36,8 +36,7 @@ impl List {
     /// Returns the last node of the linked list. Recursive implementation.
     #[ensures(result.len() == 1)]
     #[ensures(result.value == old(self.lookup(self.len() - 1)))]
-    #[ensures(
-        after_expiry<result>(
+    #[after_expiry(result => (
```
Aurel300, line 39: Why not `#[after_expiry[result](...)]` ?

vakaras: Is this syntax supported by procedural macros? I thought the only allowed syntax is an identifier followed by a single pair of parenthesis.

Aurel300: Ah, you might be right. I was still thinking of the `#[ensures(after_expiry...)]` syntax.

PR #219, opened 2020-10-12
Fix library path on Windows

### LINKED ISSUE #52: Set toolchain from cargo-prusti
It would be nice to automatically set the `RUST_TOOLCHAIN` environment variable in the `cargo-prusti` binary.

fpoli:
Fix library path on Windows. Closes #52.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,10 +81,11 @@ jobs:
     - name: Check Prusti binaries on Windows
       if: matrix.os == 'windows-latest'
       run: |
-        :: Should succeed
-        target\debug\prusti-rustc.exe \
+        # Should succeed
+        target\debug\prusti-rustc.exe `
           prusti-tests\tests\verify\pass\no-annotations\assert-true.rs
-        :: Should fail
-        ! target\debug\prusti-rustc.exe \
+        if ($LASTEXITCODE -ne 0) { throw "Exit code is $LASTEXITCODE" }
+        # Should fail
+        target\debug\prusti-rustc.exe `
           prusti-tests\tests\verify\fail\no-annotations\assert-false.rs
-        if %errorlevel% eq 0 exit /b 1
+        if ($LASTEXITCODE -eq 0) { throw "Exit code is $LASTEXITCODE" }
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -69,3 +69,22 @@ jobs:
...
+    - name: Check Prusti binaries on Windows
+      if: matrix.os == 'windows-latest'
+      run: |
+        :: Should succeed
+        target\debug\prusti-rustc.exe \
+          prusti-tests\tests\verify\pass\no-annotations\assert-true.rs
+        :: Should fail
+        ! target\debug\prusti-rustc.exe \
+          prusti-tests\tests\verify\fail\no-annotations\assert-false.rs
+        if %errorlevel% eq 0 exit /b 1
```
Aurel300: I am confused about how this works. The `!` on line 88 should negate the result of running `prusti-rustcs`. But we are in `cmd.exe` (I assume), which does not have `!` syntax. And which is why you also check if the errorlevel was 0, and fail if so? Also, `eq` should be `equ`?

fpoli: Thanks for finding this, the `!` should not be there. The `eq` worked in the past, but `==` should be a bit more readable.

.github/workflows/test.yml
```diff
@@ -69,3 +69,22 @@ jobs:
...
+    - name: Check Prusti binaries on Windows
+      if: matrix.os == 'windows-latest'
+      run: |
+        :: Should succeed
+        target\debug\prusti-rustc.exe \
+          prusti-tests\tests\verify\pass\no-annotations\assert-true.rs
+        :: Should fail
+        target\debug\prusti-rustc.exe \
+          prusti-tests\tests\verify\fail\no-annotations\assert-false.rs
+        if %errorlevel%==0 (exit /b 1)
```
Aurel300: Sorry for the additional confusion from my previous comment. We are actually running in PowerShell, as indicated by the CI failure. I don't have a Windows machine to test this quickly, but maybe something like:

```suggestion
        target\debug\prusti-rustc.exe \
          prusti-tests\tests\verify\fail\no-annotations\assert-false.rs
        ! $?
```

Aurel300: On second thought, it might be that PowerShell supports the exact same syntax as Bash here? 😅

fpoli: > We are actually running in PowerShell, as indicated by the CI failure

Oh, and I'm copying these commands from a `.bat` script... Let's see if it works.

PR #221, opened 2020-10-13
Trait checks

cmatheja:
Add a function `type_implements_trait` that checks whether a given Rust type implements the trait given by a `DefId`.
In particular, we use this function for reporting an error if a pure function returns a type that does not implement the `Copy` trait.


prusti-tests/tests/verify/fail/equality/pure-post-1.rs
```diff
@@ -0,0 +1,17 @@
...
+// Copy is missing here
+#[derive(Clone,PartialEq,Eq)]
+struct A {
+    i: i32,
+}
+
+#[pure] 
+#[requires(_x == _y)]
+#[ensures(_x == _y)]
+fn get_value(_x: A, _y: A) -> A { //~ ERROR
```
Aurel300: Match a substring of the error?

PR #233, opened 2020-10-24
Implement simple optimizations for predicates

tillarnold:
@vakaras this is the pull request for the optimizations we talked about in our meeting. I implemented the body removal idea we discussed as `remove_body_of_predicates_if_possible`. However it appears to remove predicate bodies that are actually needed so I commented out the lines where it is applied. I tried implementing a more naive approach in `remove_unnecessary_bodies` which does appear to work.


prusti-common/src/vir/optimizations/mod.rs
```diff
@@ -28,6 +29,17 @@ impl Program {
             .into_iter()
             .map(|f| folding::FoldingOptimizer::optimize(f))
             .collect();
+        self.viper_predicates = predicates::delete_unused_predicates(
+            &self.methods,
+            &self.functions,
+            &self.viper_predicates,
```
vakaras: I think it would be better to take the vector by value so that we can avoid a potentially expensive `to_vec()` inside the body of `delete_unused_predicates`.

prusti-common/src/vir/optimizations/mod.rs
```diff
@@ -28,6 +29,17 @@ impl Program {
...
+        );
+        self.viper_predicates = predicates::remove_unnecessary_bodies(
+            &self.methods,
+            &self.functions,
+            &self.viper_predicates,
```
vakaras: Same here.

prusti-common/src/vir/optimizations/predicates/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+mod delete_unused_predicates;
+mod remove_unnecessary_bodies;
+
+pub use self::{
+    delete_unused_predicates::delete_unused_predicates,
+    remove_unnecessary_bodies::remove_unnecessary_bodies,
+};
+
+/// Walks all Statements and Expressions in the provided methods
+fn walk_methods(methods: &[CfgMethod], walker: &mut (impl StmtWalker + ExprWalker)) {
```
vakaras: I think that it would be good to place these methods in VIR utils module or somewhere like that.

@fpoli @cmatheja @Aurel300  What do you think?

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use crate::vir::{ast::*, cfg, cfg::CfgMethod, CfgBlock};
+use std::collections::BTreeSet;
+
+fn get_used_predicates(methods: &[CfgMethod], functions: &[Function]) -> BTreeSet<String> {
+    let mut collector = UsedPredicateCollector::new();
+    super::walk_methods(methods, &mut collector);
+    super::walk_functions(functions, &mut collector);
+
+    // DeadBorrowToken$ is a used predicate but it does not appear in VIR becaue it is only created when viper code is created from VIR
```
vakaras: Would be good to keep the lines containing comments no more than 80 characters long.

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
```diff
@@ -0,0 +1,191 @@
...
+
+fn get_used_predicates_in_predicates(predicates: &[Predicate]) -> BTreeSet<String> {
+    let mut collector = UsedPredicateCollector::new();
+
+    for pred in predicates {
+        match pred {
+            Predicate::Struct(StructPredicate { body: Some(e), .. }) => {
+                ExprWalker::walk(&mut collector, e)
+            }
+            Predicate::Struct(_) => { /* ignore */ }
```
vakaras: I think it would be better to make the code more explicit.
```suggestion
            Predicate::Struct(StructPredicate { body: None, .. }) => { /* ignore */ }
```

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
```diff
@@ -0,0 +1,191 @@
...
+        has_changed = false;
+
+        let predicates_used_in_predicates = get_used_predicates_in_predicates(&new_predicates);
+        debug!(
+            "The used predicates in predicates are {:?}",
+            &predicates_used_in_predicates
+        );
+        new_predicates = new_predicates
+            .into_iter()
+            .filter(|p| {
```
vakaras: You probably could use [`Vec::retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) here.

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
```diff
@@ -0,0 +1,191 @@
...
+                    debug!("The predicate {} was never used and thus removed", name);
+                    has_changed = true;
+                }
+
+                is_used
+            })
+            .collect();
+    }
+
+    // FIXME: This acctually removes bodies that are needed
```
vakaras: I would prefer to avoid the fixpoint if possible because it could void the gains of the optimization.

Have you checked where the predicates with removed bodies are used? Are they used only in `fold`, `unfold`, and `unfolding`? If yes, we could collect the predicates whose folding state is changed in a special list and remove only the bodies of predicates that are not mentioned in that list.

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    debug!(
+        "The used predicates in functions and methods are {:?}",
+        &used_preds
+    );
+
+    while has_changed {
+        has_changed = false;
+
+        let predicates_used_in_predicates = get_used_predicates_in_predicates(&new_predicates);
```
vakaras: I think it would be more performant to create a map `predicate_name → set of predicate names used in that predicate` that is computed once instead of recomputing this information over and over again in the loop.

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
vakaras: In other words: could the optimization implemented in `prusti-common/src/vir/optimizations/predicates/remove_unnecessary_bodies.rs` be merged into this file?

vakaras (commented):
Thank you! The start looks very good. I have left some comments.

prusti-common/src/vir/optimizations/predicates/mod.rs
fpoli: This looks generic enough to go in VIR :+1: (of course, trying to avoid `utils` as name of modules).

prusti-common/src/vir/optimizations/predicates/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+            cfg::Successor::GotoSwitch(conditional_targets, _) => {
+                for (expr, _) in conditional_targets {
+                    ExprWalker::walk(walker, expr);
+                }
+            }
+        });
+    }
+}
+
+/// Walks all Expressions in the provided functions (including pre and post condidions)
```
Aurel300:
```suggestion
/// Walks all Expressions in the provided functions (including pre and post conditions)
```

prusti-common/src/vir/optimizations/predicates/delete_unused_predicates.rs
tillarnold: I rearranged the optimisations such that remove_unnecessary_bodies is applied first. Even in that case it appears that around 5 of the tests in the test suite require more than 1 iteration to remove all unused predicates. Should I remove the fixpoint iteration since in most cases most unused predicates are removed that way or do we aim to really eliminate all unused predicates?

fpoli: When you remove an unused predicates `P` you know precisely which other predicates might become unused: they are the predicates that were used by `P`. So, you can keep a much smaller set of predicates to be checked in the next iteration.

vakaras:
@fpoli Do you have any clue why `quick-tests` are failing while `all-tests` are passing? I thought that `quick-tests` are a subset of `all-tests`.

fpoli:
> @fpoli Do you have any clue why `quick-tests` are failing while `all-tests` are passing? I thought that `quick-tests` are a subset of `all-tests`.

They are. However, test `mut-borrows-binary-search.rs` seems flaky and fails with a low probability, so we should disable it. In the meantime you can restart all the checks from GitHub.

vakaras (approved):
LGTM. However, we need to check why the tests are failing.

fpoli:
Btw, what is an example of unused predicate? I thought that we only encode them on-demand.

tillarnold:
For example in `knapsack.rs` it removes quite a lot of `closure` and `ref$closure` predicates. Also the process of removing the bodies of predicates that are never folded/unfolded makes some other predicates unused.

fpoli:
> For example in `knapsack.rs` it removes quite a lot of `closure` and `ref$closure` predicates. Also the process of removing the bodies of predicates that are never folded/unfolded makes some other predicates unused.

Thanks! The former could be considered a bug of Prusti, but the latter is definitely very common in real-world projects (e.g. the `ThreadRng` in the `prusti-tests/tests/verify/pass/simple-specs/rand-ensures.rs` test). I'm very happy about this optimization :)

PR #234, opened 2020-10-27
Fix collection of closures

fpoli:


Aurel300:
Nice work on re-enabling yet more tests. It seems pretty clear, the only thing I wonder about is the naming – how about `SpecClosureCollector` or similar? Just so it is not confused with actual user-code closures.

PR #235, opened 2020-10-27
Skip dependencies when running cargo-prusti

fpoli:
* Disable the `test-on-crates` CI job, which is broken because Prusti tries to write the `nll-facts` folder on a read-only file system.
* Add a (disabled) test that runs `cargo-prusti` on Prusti itself.
* Replace MANY panics with a `return Err(EncodingError::...);`.


fpoli:
@cmatheja I moved `memory_eq_*` functions to a separate file and added `Result<..>` types to report unsupported features.

PR #237, opened 2020-10-30
Prusti user guide

Aurel300:
[**Link to rendered preview**](https://media.thenet.sk/aca/prusti-user-guide) (as of f9dd69f)

Here is a guide that should replace the wiki tutorial. I tried to document our more recent features too. Let me know if I should add something.

### DIFF
--- a/user-guide/src/basic.md
+++ b/user-guide/src/basic.md
@@ -3,14 +3,64 @@
 When the Prusti Assistant extension is active, Rust files can be verified in one of the following ways:
 
 - By clicking the "Verify with Prusti" button in the status bar.
+- By opening the [Command Palette](https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette) running the command `Prusti: save and verify this file`
 - By saving a Rust document, if "Verify on save" is enabled.
 - By opening a Rust document, if "Verify on open" is enabled.
 
 See the [following chapter](verify/summary.md) for a list of verification features available in Prusti.
 
-## Example
+## Introductory Example
 
-The specification for a `max` function (returning the greater of its two inputs) could be:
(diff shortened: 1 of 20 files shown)


fpoli:
"When using the extension, ensure that ... is uses nightly" — I would add something like "If you wish to use the latest development version" somewhere.

fpoli:
In my opinion Prusti is also meant to be used from the command line by the users; not just via the IDE. It should be simple (download the release, then launch `prusti-rustc --edition=2018 my/program.rs`), but we haven't written a README yet.

The Installation page could have two sections: Prusti Assistant and the precompiled binaries:
* For Prusti Assistant I would put as few instructions as possible, even less than what is currently there. I would rather suggest to follow the README of the extension (which we could extend).
* For the procompiled binaries, I would mention that they can be downloaded from https://github.com/viperproject/prusti-dev/releases. Those marked as "release" are meant to be a bit more stable.

user-guide/src/basic.md
```diff
@@ -3,14 +3,64 @@
...
+        b
+    }
+}
+```
+
+When pasting the above code into a Rust file and then verifying it with Prusti (as outlined at the top of the page), Prusti should report that verification succeeded.
+This tells us that
+
+1. the file consists of valid Rust code that can be compiled successfuly, and
+2. no execution reaches an explicit call of the [`panic!()`](https://doc.rust-lang.org/std/macro.panic.html) macro.
```
Aurel300: I guess "explicit" is misleading, since a panic can be caused e.g. by an assertion failure.

Aurel300:
https://viperproject.github.io/prusti-dev/user-guide/

PR #243, opened 2020-11-03
Convert panics to encoding errors

fpoli:


test-crates/src/main.rs
```diff
@@ -106,7 +106,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         .into_iter()
         .map(|c| c.into())
         // For the moment, test only a few of the crates.
-        .take(50)
+        .take(1)
```
Aurel300, line 109: Is this temporary (removed before the PR is merged)?

fpoli: Yes, it's temporary, but the `test-crates` tests are anyway disabled. Supporting just the first crate will require quite some work.

fpoli: > removed before the PR is merged

Ah, now I get your comment. No, I'm thinking of leaving it at 1, because the test is unused and broken.

PR #246, opened 2020-11-13
Refactor mapping of DefId to specifications

Aurel300:
- [x] remove `extern_spec` map
- [x] add `def_spec` map
- [x] look up specifications in `def_spec`
- [x] merge extern specs into `def_spec`
- [x] add `trusted` and `pure` into `def_spec`
- [x] loop invariants

### Summary

The goal is for attribute lookups to become unnecessary in `encoder`. `SpecCollector` now collects all the specifications into a map of `DefId`s -> specs. This is also a prerequisite for a much nicer #138 without `#[stmt_expr_attributes]`.

### Future work

As discussed in our meetings, eventually we should convert all the MIR bodies we would need during verification (specifications and un-`trusted` function bodies) into VIR during  the `SpecCollector` visit. This would allow external specs to work without creating a wrapper function, and would result in a cleaner separation of the Viper encoder from the Rust internals. However, a more generic (incl. type parameters) VIR is required before such an encoding can be done.


prusti-viper/src/encoder/encoder.rs
```diff
@@ -528,21 +415,23 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
                     // one to pick.
                     let item = procs[0];
                     if let Some(spec) = self.get_procedure_specs(item.def_id) {
-                        impl_spec = spec.clone();
+                        impl_spec = spec;
                     } else {
                         debug!("Procedure {:?} has no specification", item.def_id);
                     }
+                } else {
+                    panic!("multiple trait method declarations found");
```
Aurel300: I added this `panic` to see whether some of the failing tests are caused by this. It caused more tests to fail though. I have not investigated why we get multiple method yet. Ideas?

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -151,26 +151,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
...
-                    debug!("Procedure {:?} has no specification", self.proc_def_id);
-                    typed::SpecificationSet::Procedure(typed::ProcedureSpecification::empty())
-                }
-            };
-            let tymap = self.encoder.current_tymap();
-            let contract = compute_procedure_contract(
-                self.proc_def_id,
-                self.encoder.env().tcx(),
-                fun_spec,
-                Some(&tymap),
```
Aurel300, line 170: As far as I can tell, this line is the only difference between the code in `encoder` and here. In `encoder`, the last argument is `None`, here the `tymap` is used. I am not sure what this implies, but I don't think it affected the test results at all.

Aurel300:
The latest error seems to be related to the upstream Scala upgrade … Locally, I get these test failures:

```
    [run-pass] pass/crates/001-libc-cpu-equal.rs
    [run-pass] pass/crates/008-num-traits-pow.rs
    [run-pass] pass/enums/builtin.rs
    [run-pass] pass/equality/generic-eq.rs
    [run-pass] pass/equality/simple-struct.rs
    [run-pass] pass/equality/simple-tuple.rs
    [run-pass] pass/erdinm/traits-basic-norm-prov-call-post.rs
    [run-pass] pass/erdinm/traits-basic-norm-prov-call-pre.rs
    [run-pass] pass/erdinm/traits-basic-norm-reqd-call-post.rs
    [run-pass] pass/erdinm/traits-basic-norm-reqd-call-pre.rs
    [run-pass] pass/erdinm/traits-basic-norm-reqd-impl-pre.rs
    [run-pass] pass/quick/trait-contracts-refinement.rs
    [run-pass] pass/trait-contracts-refinement/precondition-inhale.rs
```

Most are due to the `panic!("multiple trait method declarations found")`, which was not a panic before, and maybe it should not be? The ones that fail due to something going wrong with trait contract refinement are:

```
    [run-pass] pass/erdinm/traits-basic-norm-reqd-call-post.rs
    [run-pass] pass/erdinm/traits-basic-norm-reqd-impl-pre.rs
    [run-pass] pass/quick/trait-contracts-refinement.rs
    [run-pass] pass/trait-contracts-refinement/precondition-inhale.rs
```

prusti-interface/src/specs/external.rs
```diff
@@ -89,7 +88,7 @@ impl<'tcx> ExternSpecResolver<'tcx> {
         }
     }
 
-    pub fn get_extern_fn_map(&self) -> ExternSpecificationMap<'tcx> {
+    pub fn get_extern_fn_map(&self) -> HashMap<DefId, (Option<DefId>, DefId)> {
         self.extern_fn_map.clone()
```
vakaras: Is there a reason to `clone` here?

prusti-interface/src/specs/mod.rs
```diff
@@ -3,22 +3,28 @@ use rustc_ast::ast;
...
 pub mod external;
 pub mod typed;
 
 use typed::StructuralToTyped;
+use typed::SpecIdRef;
 use std::fmt;
 use crate::specs::external::ExternSpecResolver;
 use prusti_specs::specifications::common::SpecificationId;
 
+#[derive(Clone)]
```
vakaras: Why `Clone` is needed?

prusti-interface/src/specs/mod.rs
```diff
@@ -33,47 +39,188 @@ impl fmt::Debug for SpecItem {
...
             typed_expressions: HashMap::new(),
             extern_resolver: ExternSpecResolver::new(tcx),
         }
     }
 
-    pub fn determine_typed_procedure_specs(self) -> typed::SpecificationMap<'tcx> {
-        let typed_expressions = self.typed_expressions;
-        let tcx = self.tcx;
-        self.spec_items
+    pub fn determine_typed_procedure_specs(&self) -> typed::SpecificationMap<'tcx> {
```
vakaras: As far as I understand, now the main entry point is going to be `determine_def_specs`, right? Then, I would suggest:
1. making `determine_typed_procedure_specs` private;
2. changing the code to avoid unnecessary `clone` operations;
3. in ` prusti/src/callbacks.rs`, print the `DefSpecificationMap`.

prusti-specs/src/specifications/json.rs
```diff
@@ -2,38 +2,38 @@ use super::untyped;
 use serde::{Deserialize, Serialize};
 use super::common;
 
-#[derive(Serialize, Deserialize)]
+#[derive(Clone, Serialize, Deserialize)]
 pub struct Assertion {
```
vakaras: Why do you need `Clone`?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -28,12 +28,13 @@ use prusti_interface::data::ProcedureDefId;
 use prusti_interface::environment::Environment;
 use prusti_interface::specs::typed;
 use prusti_interface::specs::typed::SpecificationId;
-use prusti_interface::utils::{has_spec_only_attr, read_prusti_attrs, has_prusti_attr};
+use prusti_interface::utils::{has_spec_only_attr, read_prusti_attrs};
 use prusti_interface::PrustiError;
 // use prusti_interface::specs::{
```
vakaras: Probably would be good to delete the commented out code.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -365,112 +338,27 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
     /// `prusti::loop_body_invariant_spec` attribute.
-    pub fn get_loop_specs(&self, def_id: DefId) -> Vec<SpecificationId> {
-        let attrs = self.env().tcx().get_attrs(def_id);
-        debug_assert!(has_prusti_attr(attrs, "loop_body_invariant_spec"));
-        read_prusti_attrs("spec_id", attrs).into_iter().map(
-            |raw_spec_id| raw_spec_id.try_into().expect(
-                &format!("cannot parse the spec_id attached to {:?}", def_id)
-            )
-        ).collect()
+    pub fn get_loop_specs(&self, def_id: DefId) -> Option<Vec<typed::Assertion<'tcx>>> {
```
vakaras: Would it be possible to return a reference here?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -365,112 +338,27 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
-                            reference: None,    // FIXME: Currently only `result` is supported.
-                            lhs: lhs.map(|spec_id| self.spec().get(&spec_id).unwrap().clone()),
-                            rhs: self.spec().get(&rhs).unwrap().clone(),
-                        })
-                    }
-                }
-            }
-            Some(typed::SpecificationSet::Procedure(typed::ProcedureSpecification::new(pres, posts, pledges)))
-        }
+    pub fn get_procedure_specs(&self, def_id: DefId) -> Option<typed::ProcedureSpecification<'tcx>> {
```
vakaras, line 344: And also here.

prusti-viper/src/encoder/encoder.rs
vakaras: Nothing comes to my mind at the moment.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1416,11 +1308,11 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
     /// when the user tries to call an impure function in a context that requires a pure function.
     pub fn encode_stub_pure_function_use(
         &self,
-        mut proc_def_id: ProcedureDefId,
+        proc_def_id: ProcedureDefId,
     ) -> (String, vir::Type) {
-        proc_def_id = *self.get_specification_def_id(&proc_def_id);
         // The stub function may come from another module for which we can have
         // only optimized_mir.
+        //let body = self.env.tcx().optimized_mir(proc_def_id.expect_local()).borrow();
```
vakaras: Delete this line?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1456,12 +1348,12 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             let (proc_def_id, substs) = self.encoding_queue.borrow_mut().pop().unwrap();
             let proc_name = self.env.get_absolute_item_name(proc_def_id);
             let proc_def_path = self.env.get_item_def_path(proc_def_id);
-            let proc_span = self.env.get_item_span(proc_def_id);
+            //let proc_span = self.env.get_item_span(proc_def_id);
```
vakaras: I think you can delete the commented-out lines.

prusti/src/callbacks.rs
```diff
@@ -51,9 +51,9 @@ impl rustc_driver::Callbacks for PrustiCompilerCalls {
             let mut visitor = specs::SpecCollector::new(tcx);
             intravisit::walk_crate(&mut visitor, &krate);
             let env = Environment::new(tcx);
-            let extern_specs = visitor.determine_extern_procedure_specs(&env);
-            let type_map = visitor.determine_typed_procedure_specs();
+            let def_spec = visitor.determine_def_specs(&env);
             if self.flags.print_typeckd_specs {
+                let type_map = visitor.determine_typed_procedure_specs();
                 let mut values: Vec<_> = type_map
```
vakaras: As I mentioned above, here we could print `def_spec` instead.

vakaras (approved):
LGTM

prusti-interface/src/specs/mod.rs
Aurel300: Ok, I can do that. It will change the stdout for `ui` tests, but I guess that's ok.

prusti-viper/src/encoder/encoder.rs
Aurel300: It would be, but for the actual usages of both `get_loop_specs` and `get_procedure_specs`, the result is immediately cloned anyway (the procedure specs need to be put into a `typed::SpecificationSet::Procedure`, for example).

vakaras: I see.

prusti-viper/src/encoder/encoder.rs
fpoli: This panic is executed even when `procs.len() == 0`. Is it intentional? It doesn't match the panic message.

Aurel300: @fpoli You are right, about half of the cases are triggered with `procs.len() == 0`. I guess these cases should be ok, since we just don't have any spec to find. But if multiple methods are found, we risk selecting the wrong one or ignoring an existing spec. Anyway, for the time being I am disabling this panic, since this is not really part of this PR…

prusti-interface/src/specs/mod.rs
```diff
@@ -33,47 +38,189 @@ impl fmt::Debug for SpecItem {
...
             extern_resolver: ExternSpecResolver::new(tcx),
         }
     }
 
-    pub fn determine_typed_procedure_specs(self) -> typed::SpecificationMap<'tcx> {
-        let typed_expressions = self.typed_expressions;
-        let tcx = self.tcx;
-        self.spec_items
+    fn prepare_typed_procedure_specs(&mut self) {
+        let mut spec_items = self.spec_items.drain(..).collect::<Vec<SpecItem>>();
```
fpoli: Could be `let mut spec_items = mem::replace(&mut self.spec_items, vec![]);`

prusti-interface/src/specs/mod.rs
```diff
@@ -33,47 +38,189 @@ impl fmt::Debug for SpecItem {
...
+                    &self.typed_expressions,
+                    self.tcx
                 );
                 (spec_item.spec_id, assertion)
             })
             .collect()
     }
 
-    pub fn determine_extern_procedure_specs(&self, env: &Environment<'tcx>) -> typed::ExternSpecificationMap<'tcx> {
+    pub fn determine_def_specs(&mut self, env: &Environment<'tcx>) -> typed::DefSpecificationMap<'tcx> {
```
fpoli: API nit: As far as I understand `determine_def_specs` is only meant to be called once (because of `prepare_typed_procedure_specs`), so I would replace the `&mut self` argument with `self` and I would rename the method to `build_def_specs` to make it clear that it's an instance of the builder pattern.

prusti-viper/src/encoder/pure_function_encoder.rs
fpoli: From the implementation of `compute_procedure_contract` the `tymap` seems to be used for pure functions with type parameters; but I don't have an example to demonstrate it.

fpoli (approved):
LGTM :+1:

LDY1998:
> * [x]  remove `extern_spec` map
> * [x]  add `def_spec` map
> * [x]  look up specifications in `def_spec`
> * [x]  merge extern specs into `def_spec`
> * [x]  add `trusted` and `pure` into `def_spec`
> * [x]  loop invariants
> 
> ### Summary
> The goal is for attribute lookups to become unnecessary in `encoder`. `SpecCollector` now collects all the specifications into a map of `DefId`s -> specs. This is also a prerequisite for a much nicer #138 without `#[stmt_expr_attributes]`.
> 
> ### Future work
> As discussed in our meetings, eventually we should convert all the MIR bodies we would need during verification (specifications and un-`trusted` function bodies) into VIR during the `SpecCollector` visit. This would allow external specs to work without creating a wrapper function, and would result in a cleaner separation of the Viper encoder from the Rust internals. However, a more generic (incl. type parameters) VIR is required before such an encoding can be done.

How is the MIR encoded process affect the decision of creating wrapper or not? Why would moving the MIR->VIR encoding into SpecCollector visit solve the wrapper problem?

PR #248, opened 2020-11-17
Run performance tests in CI

tillarnold:
@vakaras  So far this is just the x.py function to run the benchmarks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,7 @@ jobs:
   benchmark:
     strategy:
       matrix:
-        os: [ macos-latest, ubuntu-latest, windows-latest ]
+        os: [ubuntu-latest]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,12 +90,12 @@ jobs:
     - name: Run benchmark
       run: python x.py run-benchmarks
     - name: Publish to GitHub pages
-        uses: peaceiris/actions-gh-pages@v3
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./benchmark-output
-          keep_files: true
-          destination_dir: benchmark-output
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ./benchmark-output
+        keep_files: true
+        destination_dir: benchmark-output
 
 
   # Run Prusti on itself. Currently disabled because of many bugs in Prusti.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,7 +86,7 @@ jobs:
     - name: Set up the environment
       run: python x.py setup
     - name: Build with cargo
-      run: python x.py build --all --verbose
+      run: python x.py build --release --all --verbose
     - name: Run benchmark
       run: python x.py run-benchmarks
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,6 +89,13 @@ jobs:
       run: python x.py build --release --all --verbose
     - name: Run benchmark
       run: python x.py run-benchmarks
+    - name: Publish to GitHub pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./benchmark-output
+          keep_files: true
+          destination_dir: benchmark-output
 
 
   # Run Prusti on itself. Currently disabled because of many bugs in Prusti.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,36 +68,6 @@ jobs:
         cd prusti-contracts-test/
         cargo build
 
-  benchmark:
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-      fail-fast: false
-    runs-on: ${{ matrix.os }}
-    steps:
-    - name: Check out the repo
-      uses: actions/checkout@v2
-    - name: Set up Python 3
-      uses: actions/setup-python@v2
-      with:
-        python-version: '3.x'
-    - name: Set up Rust
-      uses: actions-rs/toolchain@v1.0.6
-    - name: Set up the environment
-      run: python x.py setup
-    - name: Build with cargo
-      run: python x.py build --release --all --verbose
-    - name: Run benchmark
-      run: python x.py run-benchmarks
-    - name: Publish to GitHub pages
-      uses: peaceiris/actions-gh-pages@v3
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: ./benchmark-output
-        keep_files: true
-        destination_dir: benchmark-output
-
-
   # Run Prusti on itself. Currently disabled because of many bugs in Prusti.
   test-on-prusti:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


x.py
```diff
@@ -257,6 +261,55 @@ def ide(args):
...
+
+    env["PRUSTI_SERVER_ADDRESS"]="localhost:12345"
+    try:
+        with open('benchmarks.csv') as csv_file:
+            csv_reader = csv.reader(csv_file, delimiter=',')
+            for row in csv_reader:
+                file_path = row[0]
+                results[file_path] = []
+                report("Starting to benchmark {}", row)
+                for i in range(warmup_iterations):
```
vakaras: I would suggest moving the warm-up outside of the `csv_reader` loop.

x.py
```diff
@@ -257,6 +261,55 @@ def ide(args):
...
 
+def run_benchmarks(args):
+    warmup_iterations = 3
+    bench_iterations = 10
+    results = {}
+
+    report("running benchmarks")
+    env = get_env()
+    report("Starting prusti-server")
+    server_process = subprocess.Popen(["./target/release/prusti-server-driver","--port","12345"], env=env)
```
vakaras: The `prusti-server-driver` path is probably platform-specific, so would be good to add a way to support multiple OS versions.

vakaras (changes requested):
Looks mostly good. Could you please also add a corresponding GitHub action?

vakaras:
@aterga @fpoli @ArquintL Why does http://viper.ethz.ch/downloads/ViperToolsNightlyMac.zip have both `/boogie/Binaries/boogie` and `/boogie/Binaries/Boogie`, which conflict on a case-insensitive file system?

x.py
fpoli: Also, the binary to be executed should be `prusti-server` and it should not need to set environment variables.

x.py
```diff
@@ -257,6 +261,55 @@ def ide(args):
...
+    json_result = json.dumps(results, indent = 2)
+    timestamp = time.time()
+    with open("benchmark-output/benchmark" + str(timestamp) + ".json", "w") as outfile: 
+        outfile.write(json_result) 
+
+                
+
+def measure_time(path, env):
+    start_time = time.perf_counter()
+    run_command(["./target/debug/prusti-rustc","--edition=2018", path],env=env)
```
fpoli: Have you tried running this without setting the `env`? I think it should work.

x.py
```diff
@@ -257,6 +261,55 @@ def ide(args):
...
+        os.makedirs('benchmark-output')
+
+    json_result = json.dumps(results, indent = 2)
+    timestamp = time.time()
+    with open("benchmark-output/benchmark" + str(timestamp) + ".json", "w") as outfile: 
+        outfile.write(json_result) 
+
+                
+
+def measure_time(path, env):
```
fpoli: This function name looks a a bit too generic, what about `measure_prusti_time(input_path, env)`?

x.py
fpoli: I mean, passing an env that just contains `PRUSTI_SERVER_ADDRESS`, and not the `get_env()`.

x.py
vakaras: @fpoli How do you kill it then properly? The kill below kills `prusti-server`, but `prusti-server-driver` keeps running.

tillarnold:
@vakaras regarding the GitHub action: which branch should I push the results to? There might be problems with pushing to `gh-pages` because the action of `gh-pages-src` would probably overwrite that.

x.py
fpoli: Oh, I didn't notice that. `prusti-server` should propagate SIGINT to `prusti-server-driver` or do something to shutdown properly. I opened issue #249. It's also possible that `z3` instances keeps running, even when using `prusti-server-driver`.

fpoli:
> @aterga @fpoli @ArquintL Why does http://viper.ethz.ch/downloads/ViperToolsNightlyMac.zip have both `/boogie/Binaries/boogie` and `/boogie/Binaries/Boogie`, which conflict on a case-insensitive file system?

I don't know. It may be a bug of the Jenkins build that prepares the zip.

vakaras:
> @vakaras regarding the GitHub action: which branch should I push the results to? There might be problems with pushing to `gh-pages` because the action of `gh-pages-src` would probably overwrite that.

I think it would be best to push to `gh-pages` and change `gh-pages-src` to overwrite only specific directories. @Aurel300 Do you have any insights?

Aurel300:
> I think it would be best to push to `gh-pages` and change `gh-pages-src` to overwrite only specific directories. @Aurel300 Do you have any insights?

Sure, I guess we can simply enable [`keep-files`](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-keeping-existing-files-keep_files) on the `gh-pages-src` action. Pushing to `-src` from benchmarks is messy IMO.

ArquintL:
> @aterga @fpoli @ArquintL Why does http://viper.ethz.ch/downloads/ViperToolsNightlyMac.zip have both `/boogie/Binaries/boogie` and `/boogie/Binaries/Boogie`, which conflict on a case-insensitive file system?

I only see `boogie`. Is there a `Boogie` file other than `Boogie.exe`?

fpoli:
> I only see `boogie`. Is there a `Boogie` file other than `Boogie.exe`?

Yes

.github/workflows/test.yml
```diff
@@ -68,6 +68,36 @@ jobs:
         cd prusti-contracts-test/
         cargo build
 
+  benchmark:
+    strategy:
+      matrix:
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
```
vakaras: I think it would be fine to run benchmarks on `ubuntu-latest` only.

vakaras:
> > I think it would be best to push to `gh-pages` and change `gh-pages-src` to overwrite only specific directories. @Aurel300 Do you have any insights?
> 
> Sure, I guess we can simply enable [`keep-files`](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-keeping-existing-files-keep_files) on the `gh-pages-src` action. Pushing to `-src` from benchmarks is messy IMO.

Do I understand it correctly that with `keep-files=false`, the action overrides only the `destination_dir`? So, one option would be to:
1. change `gh-pages-src` to write not to root, but to a subdirectory;
2. add https://viperproject.github.io/ with a proper index page that links to Viper stuff and to a Prusti index page in a subdirectory.

Any concerns regarding this design?

Aurel300:
> Do I understand it correctly that with `keep-files=false`, the action overrides only the `destination_dir`?

Yes.

With `keep-files: true`, I think it merges the directory structure, overriding any files with the same filename but otherwise leaving everything intact.

> Any concerns regarding this design?

A little bit, yes. `https://viperproject.github.io/` would need to be a separate repository, but I guess that's fine. We could set the action in `gh-pages-src` to write to e.g. `prusti-dev/dev-guide` and `prusti-dev/user-guide`, but then it is not writing the index, so that would need to be done manually in `gh-pages`. A bit of a pain and I don't see the point.

So let's just change the GH action in `gh-pages-src` to also have `keep-files: true`.

tillarnold:
@vakaras the CI push action fails with `remote: Permission to viperproject/prusti-dev.git denied to github-actions[bot].` is this only because this is running in the context of this PR and thus does not have push access to the repository or is there something I need to change?

vakaras:
> @vakaras the CI push action fails with `remote: Permission to viperproject/prusti-dev.git denied to github-actions[bot].` is this only because this is running in the context of this PR and thus does not have push access to the repository or is there something I need to change?

One problem is that this job is configured to run on each pull request and each should not:

```yaml
on:
  push:
    branches: 'master'
  pull_request:
    branches: [ '**' ]
```

Could you move it to `benchmarks.yaml` and configure to run only on push to `master`?

vakaras (approved):
LGTM. Thanks for implementing this!

PR #250, opened 2020-11-18
Add OPTIMIZATIONS config flag

tillarnold:
@vakaras this is the config flag we discussed. I'm not sure if this is how you imagined the implementation. Also it appears that config-rs does not allow passing a list in an environment variable https://github.com/mehcode/config-rs/issues/117 ?

### DIFF
--- a/prusti-common/src/vir/optimizations/mod.rs
+++ b/prusti-common/src/vir/optimizations/mod.rs
@@ -7,6 +7,7 @@
 //! A module that contains various VIR optimizations.
 
 use vir::Program;
+use crate::config::{self, Optimizations};
 
 pub mod folding;
 pub mod functions;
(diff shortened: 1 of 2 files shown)


prusti-common/src/vir/optimizations/mod.rs
```diff
@@ -15,25 +16,33 @@ pub mod predicates;
...
-        self.functions = new_functions
-            .into_iter()
-            .map(|f| folding::FoldingOptimizer::optimize(f))
-            .collect();
-        self.viper_predicates = predicates::delete_unused_predicates(
-            &self.methods,
-            &self.functions,
-            self.viper_predicates,
-        );
+        if optimizations.contains(&Optimizations::All) || optimizations.contains(&Optimizations::InlineConstantFunctions) {
```
vakaras: I think it would be better to have something like `optimizations.is_inline_constant_functions_enabled()`.

vakaras:
> Also it appears that config-rs does not allow passing a list in an environment variable [mehcode/config-rs#117](https://github.com/mehcode/config-rs/issues/117) ?

I think supporting environmental variables is critical and, therefore, we probably have to do parsing manually.

prusti-common/src/vir/optimizations/mod.rs
vakaras: Since it seems that you need to do parsing manually anyway, you could probably have `Optimizations` as a struct with a boolean field for each optimization.

vakaras (approved):
LGTM

vakaras:
Thanks for implementing this!

PR #251, opened 2020-11-18
Verify The Jackbot 10684

kammola:


kammola:
@vakaras

fpoli:
Could you remove the `.DS_Store` files from the PR?

kammola:
> Could you remove the `.DS_Store` files from the PR?

Done.

vakaras:
@kammola The tests are failing because on CI we test with the overflow-checks enabled while IDE by default has them disabled. [Here](https://viperproject.github.io/prusti-dev/user-guide/verify/overflow.html) are the instructions on how to enable the checks in the IDE.

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
```diff
@@ -0,0 +1,120 @@
+//https://onlinejudge.org/index.php?option=com_onlinejudge&Itemid=8&category=24&page=show_problem&problem=1625
```
vakaras, line 1: It would be useful to mention which functions are “real” code and which are there just for verification purposes.

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
```diff
@@ -0,0 +1,120 @@
...
+        max(0, seq.lookup(idx))
+    }else {
+        max(seq.lookup(idx), the_jackpot_rec(seq, idx - 1))
+    }
+}
+
+#[requires (seq.len() > 0)]
+#[ensures (result == the_jackpot_rec(seq, seq.len() - 1))]
+fn the_jackpot(seq: &VecWrapperI32) -> i32 {
+    let dp = solve(seq);
```
vakaras, line 214: I am a bit confused here: why `solve` is called but never used?

kammola: It was a bug. FIXED now.

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
kammola: I rearranged the code. Now,  the mail solution function is under "Solution" comment

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+        a
+    }
+}
+
+#[pure]
+#[requires (i >= 0)]
+#[requires (i < seq.len())]
+#[requires (seq.len() > 0)]
+fn solve_rec(seq: &VecWrapperI32, i: usize) -> i32 {
```
vakaras: I would say that it is not obvious that this function is computing the result we want it to compute. Therefore, we need to prove that this is indeed the case. One idea would be to do this as follows:

1. Write a function `solve_naive(seq: &VecWrapperI32) -> i32` that solves the problem by using two nested loops (one loop checks all possible start positions of the subsequence, and the second one computes all possible sums that start at that position).
2. Prove that `solve_rec(seq, seq.len()) >= solve_naive(seq)` (it should be much easier to prove `>=` than `==`).

Is it clear how to do that?

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
vakaras: Thanks!

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
kammola: I will try to do it and see if I manage.

kammola:
I have changed it but I am facing overflow problems now. Is there a way to avoid that or should I also verify the overflow won’t happen?

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
vakaras: Thanks!

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684.rs
vakaras: If you get any questions, please let me know.

vakaras:
> I have changed it but I am facing overflow problems now. Is there a way to avoid that or should I also verify the overflow won’t happen?

I think the final version should be verified to be overflow free. Please let me know if you get problems with this.

kammola:
> > I have changed it but I am facing overflow problems now. Is there a way to avoid that or should I also verify the overflow won’t happen?
> 
> I think the final version should be verified to be overflow free. Please let me know if you get problems with this.

The final version gives me overflow problems

vakaras:
> > > I have changed it but I am facing overflow problems now. Is there a way to avoid that or should I also verify the overflow won’t happen?
> > 
> > 
> > I think the final version should be verified to be overflow free. Please let me know if you get problems with this.
> 
> The final version gives me overflow problems

The task description says that each number is not larger than 1000. Adding this to the precondition and propagating down should allow convincing Prusti that there are no overflows.

kammola:
> > > > I have changed it but I am facing overflow problems now. Is there a way to avoid that or should I also verify the overflow won’t happen?
> > > 
> > > 
> > > I think the final version should be verified to be overflow free. Please let me know if you get problems with this.
> > 
> > 
> > The final version gives me overflow problems
> 
> The task description says that each number is not larger than 1000. Adding this to the precondition and propagating down should allow convincing Prusti that there are no overflows.

I am trying to add this condition #[ensures (result >= -1000 && result <= i * 10000)] , but i is usize and result is i32. it. gives me the error expected i32 found usize. is there a way to do the multiplication in the condition?

fpoli:
> I am trying to add this condition #[ensures (result >= -1000 && result <= i * 10000)] , but i is usize and result is i32. it. gives me the error expected i32 found usize. is there a way to do the multiplication in the condition?

Have you tried `#[ensures (result >= -1000 && (result as u128) <= (i * 10000 as u128))]`?

kammola:
> > I am trying to add this condition #[ensures (result >= -1000 && result <= i * 10000)] , but i is usize and result is i32. it. gives me the error expected i32 found usize. is there a way to do the multiplication in the condition?
> 
> Have you tried `#[ensures (result >= -1000 && (result as u128) <= (i * 10000 as u128))]`?

Prusti complains without reporting the error. It gives an error at the beginning of the file whenever I use casting.

fpoli:
> Prusti complains without reporting the error. It gives an error at the beginning of the file whenever I use casting.

Uh, thanks for reporting that. PR #253 should fix the error reporting.

We actually don't support casting between usize and other integer types, because Rust gives no guarantee about the range of usize. I didn't remember that. The issue #252 that you opened mentions the proper solution.

vakaras:
@kammola `x as usize` should work now if you compile Prusti from source code. (You can find the instructions in the [developer guide](https://viperproject.github.io/prusti-dev/dev-guide/development/setup.html).)

vakaras:
@kammola I have added instructions on how to use a locally built version in Prusti Assistant: https://viperproject.github.io/prusti-dev/dev-guide/development/setup.html#using-locally-built-version-in-prusti-assistant

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/TheJackpot_10684_failing.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    #[trusted]
+    #[ensures (self.len() == old(self.len()) + 1)]
+    #[ensures (self.lookup(old(self.len())) == value)]
+    #[ensures (forall(|i: usize| (0 <= i && i < old(self.len())) ==>
+                    self.lookup(i) == old(self.lookup(i))))]
+    pub fn push(&mut self, value: i32) {
+        self.v.push(value);
+    }
+}
```
Aurel300: Does this work with an `#[extern_spec]` implementation, like [here](https://github.com/viperproject/prusti-dev/blob/21d7d5616b581fdf9741bd1da44f42dd1f75321d/prusti-tests/tests/verify/pass/extern-spec/vec-3.rs#L21-L42)?

vakaras: Does `#[extern_spec]` allow adding a pure `lookup` function? (As far as I remember there were some serious problems with supporting the `Index` and `IndexMut` traits.)

kammola: I tried but I also couldn't find a way to support lookup.

Aurel300: Ah yes, you're right, the encoding panics when trying to call such a function. I suppose you could write a trusted wrapper only for the lookup function but then it's more messy.

vakaras (approved):
LGTM

Aurel300:
Only noticed this now unfortunately; there is a typo in the folder name, should be "competitive".

PR #253, opened 2020-11-20
Fix reporting of error messages

fpoli:


fpoli:
The failing test case is because of the `boogie/Boogie` case conflict in the Viper-server artifact :angry:

PR #266, opened 2020-11-25
Fix warnings

vakaras:
Fix unused function/closure warnings shown for specification functions.


fpoli:
I'll need a `x.py --help` at some point to remember all the possible ways to launch it :laughing:

PR #268, opened 2020-11-25
Fix 263: Integer bounds in quantifiers and pure functions

vakaras:


Aurel300:
How does this interact with `ENCODE_UNSIGNED_NUM_CONSTRAINT`?

vakaras:
> How does this interact with `ENCODE_UNSIGNED_NUM_CONSTRAINT`?

Thanks! Should be taken into account now.

PR #285, opened 2020-11-27
Update flags in the guide

fpoli:


Aurel300:
In `user-guide/src/verify/overflow.md`, the second `CHECK_OVERFLOWS` should be `PRUSTI_CHECK_OVERFLOWS` because it is the environment variable. I missed this in the previous PR.

PR #289, opened 2020-11-30
Simplify encoding of discriminants

fpoli:
@cmatheja FYI

### DIFF
--- a/prusti-viper/src/encoder/type_encoder.rs
+++ b/prusti-viper/src/encoder/type_encoder.rs
@@ -23,6 +23,7 @@ use std::hash::{Hash, Hasher};
 use rustc_ast::ast;
 use prusti_interface::specs::typed;
 use rustc_attr::IntType::SignedInt;
+use rustc_target::abi::Integer;
 use log::{debug, trace};
 use crate::encoder::errors::{PositionlessEncodingError, PositionlessEncodingResult};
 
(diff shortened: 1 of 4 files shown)


prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -755,34 +756,22 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
 }
 
 /// Compute the values that a discriminant can take.
-pub fn compute_discriminant_values(adt_def: &ty::AdtDef, tcx: ty::TyCtxt) -> Vec<i128> {
+pub fn compute_discriminant_values<'tcx>(adt_def: &'tcx ty::AdtDef, tcx: ty::TyCtxt<'tcx>) -> Vec<i128> {
     let mut discr_values: Vec<i128> = vec![];
     // Handle *signed* discriminats
```
Aurel300: Is this comment still true? Are there unsigned discriminants; do we handle them elsewhere?

fpoli: Signed discriminants are possible. I just added a test that fails if we don't call `sign_extend`. As far as I know they are handled correctly also elsewhere.

fpoli: Ah, you meant why there is no code block to handle unsigned discriminants? I'm clarifying that.

PR #290, opened 2020-11-30
Verify Sum of Different Primes

kammola:


fpoli:
> I don't know if this is something we can obtain from the counter-example. I suppose if not, we could insert a dummy sequence into the VIR code to which we push the basic block index any time we enter a basic block?

@Aurel300 In the worst case, at least knowing the values of these flags would be useful:

https://github.com/viperproject/prusti-dev/blob/4a25125c148b0bdd16470994c8c002072a2d34f5/prusti-viper/src/encoder/procedure_encoder.rs#L390

prusti-tests/tests/verify_overflow/pass/competitve_programming/dp/SumOfDifferentPrimes_1213.rs
```diff
@@ -0,0 +1,237 @@
...
+
+// Recursive solution
+
+#[pure]
+#[requires(n <= 1120)]
+#[requires(k <= 14)]
+#[requires(primes.len() > 0)]
+#[requires(forall(|k: isize| (k >= 0 && k < primes.len()) ==> (primes.lookup(k) >= 2)))]
+//#[ensures(k > 0 && n >= 0 ==> result  == sum_of_different_primes_rec_helper(primes, n, k, primes.len() - 1))]
+fn sum_of_different_primes_rec(primes: &VecWrapperI32, n: isize, k: isize) -> isize {
```
kammola: @vakaras Here is the function that reports the error wrongly with checking overflow.

vakaras:
Subsumed by #513

PR #301, opened 2020-12-03
Fix encoding of unsupported types

fpoli:
With this PR, Prusti will not generate an error message for unsupported types, but it will fail in same way when the unsupported type is used. The error would probably show up as a fold-unfold error, because the encoding might try to use fields that don't exist in the encoding. In other words, it would be a Viper consistency error or a Java exception if not for the fold-unfold pass.

It is very easy to encode all unsupported types using the *same* abstract predicate (a "predicate unsupportedType(self: Ref)"), which is what this PR is doing. I don't see a clear benefit in encoding different unsupported types with different abstract predicates, but I'm open to discuss this.

### DIFF
--- a/prusti-viper/src/encoder/type_encoder.rs
+++ b/prusti-viper/src/encoder/type_encoder.rs
@@ -505,38 +505,13 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
                 format!("__TYPARAM__${}$__", param_ty.name.as_str())
             }
 
-            ty::TyKind::Dynamic(..) => {
-                return Err(PositionlessEncodingError::unsupported(
-                    "trait objects are not supported"
-                ));
-            }
-
-            ty::TyKind::FnPtr(..) => {
-                return Err(PositionlessEncodingError::unsupported(
-                    "function pointer types are not supported"
-                ));
-            }
-
-            ty::TyKind::FnDef(..) => {
-                return Err(PositionlessEncodingError::unsupported(
-                    "function types are not supported"
-                ));
-            }
-
             ty::TyKind::Projection(ty::ProjectionTy { item_def_id, .. }) => {
                 self.encoder.encode_item_name(*item_def_id)
             }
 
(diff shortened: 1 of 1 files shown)


Aurel300:
> I don't see a clear benefit in encoding different unsupported types with different abstract predicates, but I'm open to discuss this.

I guess it's minor, but putting the type name into the predicate might be useful when reading/debugging the generated Viper program?

fpoli:
> I guess it's minor, but putting the type name into the predicate might be useful when reading/debugging the generated Viper program?

In some cases like `str` yes, but in other cases like `closure$...$...$...` or types with type parameters it's unreadable anyway. In the end I guess that yes, we'll still want separate predicates for debugging and to have Viper checks to validate the encoding when e.g. we move unsupported types between fields.

vakaras:
Didn't we have code that already does something similar to this? For example, I believe that we always encoded `Vec` as an abstract predicate, didn't we?

fpoli:
> Didn't we have code that already does something similar to this? For example, I believe that we always encoded `Vec` as an abstract predicate, didn't we?

For arrays yes, for `Vec` I don't think so.

vakaras:
> > Didn't we have code that already does something similar to this? For example, I believe that we always encoded `Vec` as an abstract predicate, didn't we?
> 
> For arrays yes, for `Vec` I don't think so.

Ok, so we had [this](https://github.com/viperproject/prusti-dev/blob/a341ba337762886056754da94805d1e48c5c47fb/prusti-viper/src/encoder/type_encoder.rs#L382). Anyway, I think that for debugging it would be useful to be able to distinguish different types.

fpoli:
> Ok, so we had [this](https://github.com/viperproject/prusti-dev/blob/a341ba337762886056754da94805d1e48c5c47fb/prusti-viper/src/encoder/type_encoder.rs#L382). Anyway, I think that for debugging it would be useful to be able to distinguish different types.

Yes, that's still there, but Prusti was panicking before reaching that line because of the unimplemented cases in `encode_predicate_use`. This PR fixes `encode_predicate_use`.

PR #303, opened 2020-12-03
Remove block invs

vakaras:
Remove unused `invs` field from `vir::cfg::BasicBlock`.


Aurel300:
Looks good; since you use the `cfg_cleaner`, might as well close #302?

prusti-common/src/vir/optimizations/methods/cfg_cleaner.rs
```diff
@@ -0,0 +1,50 @@
...
+    let mut new_indices = HashMap::new();
+
+    for cfg::CfgBlockIndex{ block_index, .. } in traversal_order {
+        if let Some(mut basic_block) = basic_blocks.remove(&block_index) {
+            let new_index = new_basic_blocks.len();
+            assert!(new_indices.insert(block_index, new_index).is_none());
+            while let vir::cfg::Successor::Goto(target) = &basic_block.successor {
+                // If the current basic block has only one successor.
+                if predecessors[&target.block_index].len() == 1 {
+                    // If the successor block has only one predecessor.
```
Aurel300: These comment seem a bit confusing (to me); it looks like the "If the successor block has only one predecessor." one is about the assertion. Move each one line up?

PR #305, opened 2020-12-04
Improve purifier

vakaras:


prusti-common/src/vir/optimizations/methods/mod.rs
```diff
@@ -11,9 +11,38 @@ mod empty_if_remover;
...
+use self::empty_if_remover::remove_empty_if;
+use self::assert_remover::remove_trivial_assertions;
+use self::var_remover::remove_unused_vars;
+use self::purifier::purify_vars;
+use self::quantifier_fixer::fix_quantifiers;
+
+pub fn optimize_method_encoding(cfg: CfgMethod, source_file_name: &str, optimizations: &Optimizations) -> CfgMethod {
+    macro_rules! apply {
+        ($optimization: ident, $cfg: ident) => {
+            if optimizations.$optimization {
```
Aurel300, line 30:
```suggestion
            let $cfg = if optimizations.$optimization {
```

And then remove the `let cfg = ` bits below?

vakaras: I wrote the current version because it makes the effects of a macro call clear without looking into its definition.

PR #309, opened 2020-12-04
Snapshots/enums

cmatheja:
Extension of the snapshot encoding for structs to enums.

### DIFF
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -434,13 +434,6 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
-    pub fn int(value: i64) -> Self {
-        Expr::Const(
-            Const::Int(value),
-            Position::default(),
-        )
-    }
-
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
@@ -471,7 +464,9 @@ impl Expr {
     pub fn domain_func_app(
         func: DomainFunc,
         args: Vec<Expr>,
-    ) -> Self { Expr::DomainFuncApp(func, args, Position::default()) }
+    ) -> Self {
+        Expr::DomainFuncApp(func, args, Position::default())
+    }
 
     pub fn magic_wand(lhs: Expr, rhs: Expr, borrow: Option<Borrow>) -> Self {
         Expr::MagicWand(box lhs, box rhs, borrow, Position::default())
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -320,6 +320,11 @@ impl Expr {
         Expr::PredicateAccessPredicate(name.to_string(), box place, perm, pos)
     }
 
+    pub fn field_access_predicate(place: Expr, perm: PermAmount) -> Self {
+        let pos = place.pos();
+        Expr::FieldAccessPredicate(box place, perm, pos)
+    }
+
     pub fn pred_permission(place: Expr, perm: PermAmount) -> Option<Self> {
         place
             .typed_ref_name()
@@ -429,6 +434,13 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
+    pub fn int(value: i64) -> Self {
+        Expr::Const(
+            Const::Int(value),
+            Position::default(),
+        )
+    }
+
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
@@ -455,6 +467,11 @@ impl Expr {
     ) -> Self {
         Expr::FuncApp(name, args, internal_args, return_type, pos)
     }
+
+    pub fn domain_func_app(
+        func: DomainFunc,
+        args: Vec<Expr>,
+    ) -> Self { Expr::DomainFuncApp(func, args, Position::default()) }
 
     pub fn magic_wand(lhs: Expr, rhs: Expr, borrow: Option<Borrow>) -> Self {
         Expr::MagicWand(box lhs, box rhs, borrow, Position::default())
--- a/prusti-common/src/vir/optimizations/folding/expressions.rs
+++ b/prusti-common/src/vir/optimizations/folding/expressions.rs
@@ -242,7 +242,7 @@ fn find_common_unfoldings2(
 /// Find unfoldings that are in all three sets.
 fn find_common_unfoldings3<'a>(
     mut first: UnfoldingMap,
-    mut first_reqs: &'a RequirementSet,
+    mut _first_reqs: &'a RequirementSet,
     mut second: UnfoldingMap,
     mut second_reqs: &'a RequirementSet,
     mut third: UnfoldingMap,
@@ -250,15 +250,10 @@ fn find_common_unfoldings3<'a>(
 ) -> (UnfoldingMap, UnfoldingMap, UnfoldingMap, UnfoldingMap) {
     let mut common = HashMap::new();
     let mut new_first = HashMap::new();
-    let swap = first.is_empty();
-    if swap {
-        mem::swap(&mut first, &mut second);
-        mem::swap(&mut first_reqs, &mut second_reqs);
-    }
     for (place, data) in first {
-        let second_agrees = second.contains_key(&place) ||
+        let second_agrees = second.contains_key(&place); ||
             second_reqs.iter().all(|p| !p.has_prefix(&place));
-        let third_agrees = third.contains_key(&place) ||
+        let third_agrees = third.contains_key(&place); ||
             third_reqs.iter().all(|p| !p.has_prefix(&place));
         if second_agrees && third_agrees {
             second.remove(&place);
@@ -268,11 +263,7 @@ fn find_common_unfoldings3<'a>(
             new_first.insert(place, data);
         }
     }
-    if swap {
-        (common, second, new_first, third)
-    } else {
-        (common, new_first, second, third)
-    }
+    (common, new_first, second, third)
 }
 
 fn update_requirements(requirements: &mut RequirementSet, mut new_requirements: Vec<ast::Expr>) {
@@ -349,7 +340,7 @@ impl ast::ExprFolder for ExprOptimizer {
                 new_expr
             }
             _ => {
-                if expr.is_place() {
+                if expr.is_place() { // CMTODO this might be the issue
                     self.requirements.insert(expr.clone());
                     expr
                 } else {
@@ -440,6 +431,7 @@ impl ast::ExprFolder for ExprOptimizer {
         let g = guard.clone();
         let f = then_expr.clone();
         let s = else_expr.clone();
+
         let guard_folded = self.fold_boxed(guard);
         let guard_unfoldings = self.get_unfoldings();
         let guard_requirements = self.get_requirements();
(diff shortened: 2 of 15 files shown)


prusti-common/src/vir/optimizations/folding/expressions.rs
```diff
@@ -242,23 +242,18 @@ fn find_common_unfoldings2(
...
-        mem::swap(&mut first, &mut second);
-        mem::swap(&mut first_reqs, &mut second_reqs);
-    }
     for (place, data) in first {
-        let second_agrees = second.contains_key(&place) ||
+        let second_agrees = second.contains_key(&place); ||
             second_reqs.iter().all(|p| !p.has_prefix(&place));
-        let third_agrees = third.contains_key(&place) ||
+        let third_agrees = third.contains_key(&place); ||
             third_reqs.iter().all(|p| !p.has_prefix(&place));
```
Aurel300: This looks like a syntax error, but perhaps it is interpreted as a closure after the semicolon?

prusti-common/src/vir/optimizations/folding/expressions.rs
```diff
@@ -349,7 +340,7 @@ impl ast::ExprFolder for ExprOptimizer {
                 new_expr
             }
             _ => {
-                if expr.is_place() {
+                if expr.is_place() { // CMTODO this might be the issue
```
Aurel300: Can you elaborate? (:

prusti-tests/tests/verify/pass/equality/enum-empty.rs
```diff
@@ -0,0 +1,14 @@
+use prusti_contracts::*;
+
+#[derive(Clone,PartialEq,Eq)]
+enum A {
+}
```
Aurel300, line 5: This is an interesting case, equivalent to [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html). You can never construct a value of this type, so a method taking `A`/`Infallible` as arguments can never be called. Maybe we should report this as a precondition failure.

Also relevant about `match` on this: https://stackoverflow.com/questions/47589690/what-does-a-match-on-an-empty-enum-return#47589799

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1421,7 +1421,9 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         proc_def_id = *self.get_specification_def_id(&proc_def_id);
         // The stub function may come from another module for which we can have
         // only optimized_mir.
+        //let body = self.env.tcx().optimized_mir(proc_def_id); // FIXME: https://github.com/viperproject/prusti-dev/issues/286
```
Aurel300: Have you checked what happens when you do use the `optimized_mir`? Do you also get errors about the MIR being stolen? (That is something I ran into when working on #246)

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -758,6 +758,7 @@ pub fn compute_discriminant_values<'tcx>(adt_def: &'tcx ty::AdtDef, tcx: ty::TyC
     discr_values
 }
 
+
```
Aurel300: There are many of these whitespace-only changes in this PR, maybe could be removed?

prusti-tests/tests/verify/pass/equality/simple-enum.rs
```diff
@@ -0,0 +1,16 @@
...
+#[derive(Clone,PartialEq,Eq)]
+enum A {
+    ANone(u32),
+    ASome(i32),
+}
+
+#[requires(_x == _y)]
+#[ensures(result == 17)]
+fn test(_x: &A, _y: &A) -> i32 {
+    17
```
Aurel300, line 12: I assume it works, but could we have tests (both passing and failing) that also call these `test` methods with some instances of the enums?

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -177,23 +180,27 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
-                    warn!("Generating equality tests for enums is not supported");
-                }
-                self.encode_snap_struct()?
+                self.encode_generic()?
             }
             ty::TyKind::Tuple(_) => {
-                self.encode_snap_struct()?
+                self.encode_tuple()?
+            }
+            ty::TyKind::Adt(adt_def, subst) if !adt_def.is_box() => {
```
Aurel300, line 175: Wouldn't a `box` just be a recursive call to the encoder, or is there anything special about it? I assume something like `Box<dyn ...>` should fail in the recursive call because of the `dyn`, not because of the box.

prusti-tests/tests/verify/pass/equality/enum-empty.rs
vakaras: Another example is `Void`, which we also already have in our test suite. Having an instance of `Void` is equivalent to being in a contradictory state. Therefore, for your `test` function, we should be able to verify `#[ensures(false)]`. However, I would be fine if this is not implemented yet as part of this PR because we are already discussing how to support this with @tillarnold.

prusti-common/src/vir/ast/expr.rs
```diff
@@ -425,6 +430,13 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
+    pub fn int(value: i64) -> Self {
+        Expr::Const(
+            Const::Int(value),
+            Position::default(),
+        )
```
fpoli: We have `impl From<i64> for Expr` which achieves the same thing, so instead of `Expr::int(123)` you could use `Expr::from(123)`.

prusti-viper/src/encoder/snapshot_encoder.rs
fpoli: [Box](https://doc.rust-lang.org/src/alloc/boxed.rs.html#160) is a wrapper around a [raw pointer](https://doc.rust-lang.org/1.21.0/src/core/ptr.rs.html#1082-1090). Our encoding is a bit special (see below) and I think the compiler special-cases it too.

https://github.com/viperproject/prusti-dev/blob/ee5def9423f99af974300d8d7d1a43377354e8cf/prusti-viper/src/encoder/type_encoder.rs#L360-L368

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -642,24 +643,352 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
     fn encode_ref_snapshot_call<S: Into<String>>(&self, arg_name: S) -> vir::Expr {
         let name = arg_name.into();
-        let arg = self.get_ref_field(self.encode_snap_arg_var(name.to_string()));
+        let arg = self.get_ref_field(self.encode_arg_var(name.to_string()));
         self.encode_snapshot_call(name, arg)
     }
 }
+
+
+impl<'s, 'p: 's, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotAdtEncoder<'s, 'p, 'v, 'tcx> {
```
fpoli: It should be possible to remove `'a` and `'r`.

vakaras:
@cmatheja Any news regarding this PR? It seems that we need it for https://github.com/kammola/prusti-dev/commit/213ebe890d588df64f51b89f0af6bd8d4e76f9d9.

cc @kammola

cmatheja:
> @cmatheja Any news regarding this PR? It seems that we need it for [kammola@213ebe8](https://github.com/kammola/prusti-dev/commit/213ebe890d588df64f51b89f0af6bd8d4e76f9d9).

I plan to finish this PR this week. However, I'm not sure this PR solves the issue you referenced; returning Copy types is independent and should already (at least partially) be supported.

vakaras:
> > @cmatheja Any news regarding this PR? It seems that we need it for [kammola@213ebe8](https://github.com/kammola/prusti-dev/commit/213ebe890d588df64f51b89f0af6bd8d4e76f9d9).
> 
> I plan to finish this PR this week.

That would be great!

>  However, I'm not sure this PR solves the issue you referenced; returning Copy types is independent and should already (at least partially) be supported.

We need to return `Copy` enums and I thought that this PR is exactly what adds this support. Am I missing something?

cmatheja:
> We need to return `Copy` enums and I thought that this PR is exactly what adds this support. Am I missing something?

No, you're right. The piece of code you linked above contained only structs, so I did not spot why you need enum support :)

prusti-tests/tests/verify/pass/equality/simple-enum.rs
Aurel300: I see, this one will not work because we can't encode reference copy types.

vakaras: > we can't encode reference copy types

Why? Shared references should not be a problem. Why3 does that.

Aurel300: Maybe it's just in specific circumstances. The error when I tried this is the following:

https://github.com/viperproject/prusti-dev/blob/95409028b9acbd77fa5ca516ad0d2e4f1d21b298/prusti-viper/src/encoder/procedure_encoder.rs#L4960-L4969

vakaras: Something is strange here. If I recall correctly, this error is only for cases when reference-typed fields are used, which should not be the case here.

Aurel300:
Writing this down so we don't forget, the failing tests are:

- `pass/loop-invs/result_in_guard.rs`
- `pass/rosetta/Knights_tour.rs`

The former crashes [here](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/type_encoder.rs#L119-L131). If instead of the stdlib `Result`, we define `MyResult` with the same variants, the test passes, which suggests that the encoding only fails (is not "defined") with external types. Knight's tour also uses `Result`, so it's possible that it fails for the same reason, although I haven't tested this yet.

fpoli:
Is this PR superseded by #326? If so, we can close this and #326 should consider all the comments that we added here.

vakaras:
> Is this PR superseded by #326? If so, we can close this and #326 should consider all the comments that we added here.

@fpoli No. That PR is dependent on this one.

prusti-common/src/vir/optimizations/folding/expressions.rs
cmatheja: Yes, that's certainly a typo.

prusti-common/src/vir/optimizations/folding/expressions.rs
cmatheja: Just a note to myself that I forgot to delete. Some time ago the set of expressions considered as places was very limited (in particular it did not include unfolding).

prusti-tests/tests/verify/pass/equality/enum-empty.rs
cmatheja: I agree that we leave that corner case to @vakaras and @tillarnold.
We will keep the test to ensure that Prusti does not falsely assume that enums have at least one variant.

cmatheja:
Moved remaining comments and feature requests to #343 such that we can merge this PR and unblock #326.

PR #315, opened 2020-12-09
Verify frog jumping

kammola:


prusti-tests/tests/verify_overflow/pass/competitive_programming/greedy/FrogJumping.rs
```diff
@@ -0,0 +1,111 @@
...
+        let take = solve_rec(positions, last_positions, next, next + 1) + 1;
+        let leave = solve_rec(positions, last_positions, idx, next + 1);
+        let x1 = last_positions.lookup(idx) + 1;
+        let tmp1  = solve_rec(positions, last_positions, last_positions.lookup(idx), x1);
+        assert!(leave ==  tmp1 + 1);
+        let x2 = next + 1;
+        assert!(x2 ==  next + 1);
+        assert!(last_positions.lookup(idx) >  next && last_positions.lookup(idx) < positions.len());
+        let tmp2 = solve_rec(positions, last_positions, next, x2);
+        assert!(tmp2 >= tmp1);
```
kammola: @vakaras This assertion is failing. But I think it shouldn't because the post condition at line 81 should be used to verify it. We have `x2 == next + 1`, that will lead to `forall(|j:  isize| (j > idx && j < positions.len()) ==> 
            solve_rec(positions, last_positions, next, x2) >= solve_rec(positions, last_positions, j, j + 1))` and by using `last_positions.lookup(idx)` as `j` then we should get `tmp2 >= tmp1`. Do you think I am missing something? Also, I tried to increase the assert timeout but it fails in less time.

vakaras: One thing that comes to my mind is that the quantifier on line 81 does not have a good trigger because `j+1` cannot be used inside a trigger. Did you try to use the trick with pulling out the addition into a separate pure function? You will probably also need to change `assert!(x2 ==  next + 1);` into `assert!(x2 ==  inc(next));`.

kammola: I changed that and the verification succeeded.

vakaras: Great!

vakaras:
Superseded by #509.

Issue #320, opened 2020-12-16
How do I enforce an invariant on a struct?


finegeometer:
Suppose I have this code:
```
#[derive(Copy, Clone)]
struct Vector(i32, i32);

impl Vector {
    fn neg(self) -> Self {
        Self(-self.0, -self.1)
    }
}
```

I would like to specify that `v : Vector` is valid iff neither `v.0` nor `v.1` are `i32::MIN`. That way, the overflow checks will pass.

How do I do this?

---

If this is possible, the syntax should be documented somewhere. I can't find it, after several hours of searching.
If it is not possible, then that fact should be documented somewhere. I automatically assumed that it would be possible, so I wasted several hours searching.

---

~There is a workaround~:
```rust
use prusti_contracts::*;

#[derive(Copy, Clone)]
struct Vector(i32, i32);

impl Vector {
    #[pure]
    fn valid(self) -> bool {
        self.0 != i32::MIN && self.1 != i32::MIN
    }

    #[requires(self.valid())]
    #[ensures(self.valid())]
    fn neg(self) -> Self {
        Self(-self.0, -self.1)
    }
}
```

~But this workaround would fail if `neg` was in an `impl std::ops::Neg for Vector`.~

Edit: I just noticed that Prusti is giving me an error when I attempt this workaround. But VSCode won't tell me the error message, and I haven't installed the command-line version of Prusti.

fpoli:
Thanks for the report! With the workaround it seems that you hit a Prusti bug which originates from the following line:

https://github.com/viperproject/prusti-dev/blob/e85739dadbcce2fbee46f8a2f0f904d7c819a618/prusti-viper/src/encoder/type_encoder.rs#L152

@cmatheja have you already seen this? I'll at least replace that panic with an internal error message.

Regarding the difference between the command-line version of Prusti and VSCode, can it be that in VSCode you are using the "latest release" build channel and not "latest dev"? See https://github.com/viperproject/prusti-assistant#configuration.

Aurel300:
Thanks for the report.

The feature you are looking for, type invariants, is something we used to support in Prusti. The syntax would be `#[invariant(...)]` attached to the `struct`. However, this was broken (by the recent compiler upgrade I think), and the user guide was written after, so it is not documented. We plan on adding this back.

Your program has one syntax error `i32.MIN` should be `i32::MIN`.

[...]

PR #322, opened 2020-12-16
Convert more panics to error messages

fpoli:
With this PR the following program (from #320) no longer panics, but reports the error message "unexpected failure while encoding the snapshot of the 'Vector' type".

```rust
use prusti_contracts::*;

#[derive(Copy, Clone)]
struct Vector(i32, i32);

impl Vector {
    #[pure]
    fn valid(self) -> bool {
        self.0 != i32::MIN && self.1 != i32::MIN
    }

    #[requires(self.valid())] //~ ERROR unexpected failure while encoding the snapshot of the 'Vector' type
    #[ensures(self.valid())]
    fn neg(self) -> Self {
        Self(-self.0, -self.1)
    }
}

fn main(){}
```

### DIFF
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -50,6 +50,7 @@ use std::borrow::Borrow;
 use crate::encoder::specs_closures_collector::SpecsClosuresCollector;
 use crate::encoder::memory_eq_encoder::MemoryEqEncoder;
 use rustc_span::MultiSpan;
+use crate::encoder::utils::transpose;
 
 const SNAPSHOT_MIRROR_DOMAIN: &str = "$SnapshotMirrors$";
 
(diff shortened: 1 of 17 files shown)


prusti-viper/src/encoder/encoder.rs
```diff
@@ -634,29 +635,31 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         builtin_encoder.encode_builtin_function_name(&function_kind)
     }
 
-    pub fn encode_cast_function_use(&self, src_ty: ty::Ty<'tcx>, dst_ty: ty::Ty<'tcx>) -> String {
+    pub fn encode_cast_function_use(&self, src_ty: ty::Ty<'tcx>, dst_ty: ty::Ty<'tcx>)
+        -> Result<String, PositionlessEncodingError>
```
Aurel300: Why not use the `PositionlessEncodingResult` and `EncodingResult` aliases you have defined?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1280,18 +1288,18 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         proc_def_id: ProcedureDefId,
         arg_ty: ty::Ty<'tcx>, // type arguments
         is_equality: bool // true = equality, false = disequality
-    ) -> (String, vir::Type) {
+    ) -> Result<(String, vir::Type), EncodingError> {
         let snapshot_res = self.encode_snapshot(&arg_ty);
```
Aurel300, line 1294:
```suggestion
        let snapshot_res = self.encode_snapshot(&arg_ty)?;
```

And remove the `is_ok` check etc?

fpoli: It's not very readable, but the point is that encoding errors of the snapshot should be catched and converted to stubs, while errors raising from the encoding of the stub should not be catched.

Aurel300: Ah ok. Without the additional `is_defined` check I would suggest this be a `.map_or_else` operation. I wonder what it means to successfully encode a snapshot that is not "defined". @cmatheja ?

fpoli: I wonder the same. I think that there is some overlapping between returning a "not defined" snapshot and a snapshot encoding error. I was the one who added `EncodingError`s to the snapshot encoder to fix some panics, and I did so trying not to change the behavior of the encoder. Now that the encoding errors are usable, we could get rid of the `is_defined()` technique, but we should check with Christoph to be sure.

PR #323, opened 2020-12-17
Enable encoding errors with optional span information

fpoli:
This PR allows `PositionlessEncodingError` to hold an *optional* span information. To reflect this in the name of the types, it also renames `EncodingError => SpannedEncodingError` and `PositionlessEncodingError => EncodingError`.

To sum up:
* `SpannedEncodingError` is to be used when it's mandatory to have a span information attached to the error. For example, in the outermost calls to the encoder, where errors are catched and reported via the compiler's API.
* `EncodingError` is to be used everywhere else, when it's possible for the error to have a span. This is the case of the encoding of rustc's types: only some of them have a `def_id` from which we can get the span.

Examples:
```rust
fn from_spanned<T>(res: SpannedEncodingResult<T>) -> EncodingResult<T> {
    let value = res?;
    Ok(value)
}

fn to_spanned<T>(res: EncodingResult<T>, span: Span) -> SpannedEncodingResult<T> {
    let value = res.with_span(span)?;
    Ok(value)
}

fn to_spanned_bis<T>(res: SpannedEncodingResult<T>, span: Span) -> SpannedEncodingResult<T> {
    let value = res.with_span(span)?;
    Ok(value)
}
```

One advantage of the new design is that it's now possible to return a `SpannedEncodingError` when an `EncodingError` is expected without losing span information. Furthermore, this design will make very easy in the future to *stack* in `SpannedEncodingError` the spans set via `with_span(..)`, such that Prusti can report an even better error message showing a stack of the items under encoding.


prusti-interface/src/prusti_error.rs
```diff
@@ -13,11 +13,11 @@ use ::log::warn;
 ///
 /// A Prusti message can originate from:
 /// * invalid usage detected during specification collection
-/// * an encoding error (see the `EncodingError` type)
+/// * an encoding error (see the `SpannedEncodingError` type)
 /// * a Viper verification error
 ///
 /// A `PrustiError` can be displayed as a *warning* to the user. (We should rename `PrustiError`,
-/// `EncodingError` and similar types to something less confusing.)
+/// `SpannedEncodingError` and similar types to something less confusing.)
```
Aurel300, line 20: I think the usual term is "diagnostics".

fpoli: So, `PrustiDiagnostic`, `EncodingDiagnostic` and `SpannedEncodingDiagnostic`? It's 5 characters longer then `...Error`, but I like it because the type alias would hide it. @vakaras @cmatheja any opinion before I do the renaming?

vakaras: Diagnostic in the Rust compiler is the object used to construct the final message to the user. Here, however, it is an error of the computation. Therefore, I would prefer to keep `Error`.

fpoli: I'll keep as it is for the moment.

PR #324, opened 2020-12-18
(Closures) Support raw functions passed to Fn args

Aurel300:
Add support for encoding regular functions passed to `Fn` arguments.

```rust
// specs...
fn call_add<F: Fn (i32, i32) -> i32>(add: F) -> i32 { ... }
// specs...
fn add(a: i32, b: i32) -> i32 { a + b }
fn main() { call_add(add); }
```

### DIFF
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -85,6 +85,8 @@ pub enum Const {
     Bool(bool),
     Int(i64),
     BigInt(String),
+    /// Used for encoding function pointers.
+    Dummy,
 }
 
 impl fmt::Display for Expr {
@@ -236,6 +238,7 @@ impl fmt::Display for Const {
             &Const::Bool(val) => write!(f, "{}", val),
             &Const::Int(val) => write!(f, "{}", val),
             &Const::BigInt(ref val) => write!(f, "{}", val),
+            &Const::Dummy => write!(f, "Dummy"),
         }
     }
 }
@@ -867,6 +870,7 @@ impl Expr {
                 match constant {
                     Const::Bool(..) => &Type::Bool,
                     Const::Int(..) | Const::BigInt(..) => &Type::Int,
+                    Const::Dummy => &Type::Int,
                 }
             }
             Expr::BinOp(ref kind, box ref base1, box ref base2, _pos) => {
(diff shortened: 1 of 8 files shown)


prusti-common/src/vir/ast/expr.rs
```diff
@@ -85,6 +85,8 @@ pub enum Const {
     Bool(bool),
     Int(i64),
     BigInt(String),
+    /// Used for encoding function pointers.
+    Dummy,
```
vakaras: This looks really hacky. Can we at least name it `FnPtr` so that is clear what is the purpose of it?

prusti-common/src/vir/ast/expr.rs
```diff
@@ -867,6 +870,7 @@ impl Expr {
                 match constant {
                     Const::Bool(..) => &Type::Bool,
                     Const::Int(..) | Const::BigInt(..) => &Type::Int,
+                    Const::Dummy => &Type::Int,
```
vakaras: Why is this an `Int`? If it just a temporary workaround, would be good to add a comment explaining that.

vakaras (approved):
Except for the `Dummy` thing, looks good.

prusti-common/src/vir/ast/expr.rs
Aurel300: I guess I think there might be more types where we just need a dummy value, but yes, for now it can just be `FnPtr`.

prusti-common/src/vir/ast/expr.rs
Aurel300: I chose `Int` as a "simple" type. Maybe it could be a `TypedRef` to some unique, abstract instance …

The alternative for passing a dummy value that is always `0` (for example) is to process function signatures such that we completely skip encoding the argument. That would be more work, more likely to result in errors, and maybe it would cause more conflicts with closures, where the value must be passed.

To be clear:

```rust
fn foo<F: Fn...>(x: F) { ... }
```

We can always just pass a dummy value for `x`, because the function is actually identified by the type, and so with monomorphisation and spec. entailments, we have everything we need both at call site and declaration site.

vakaras: I would either pick `TypedRef("FnPtr")` that is always set to Viper `null` or `Domain("FnPtr")` that is set to `fn_ptr()`. Returning an `Int` can trigger problems for optimization passes because they optimize simple types.

Issue #184, opened 2020-09-24
Unify handling of configuration flags


fpoli:
Currently, there are two ways to set configuration flags: the `-Z` arguments and the flags defined using the `config` crate. This leads to some code duplication and maintenance is error prone.

It would be nice to automatically read configuration flags from the `-Z` arguments by implementing the [Source](https://docs.rs/config/0.10.1/config/trait.Source.html) trait of the `config` crate.

fpoli:
This is a good first issue for new contributors! :muscle:

Please ask for more information in case you are interested.

gavinleroy:
I am interested in working on this issue. Is there any additional information I can get to work on it?

vakaras:
Hello @gavinleroy!
All configuration options that Prusti uses are defined in [config.rs](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/config.rs). As you can see from the file, there are two types of them:
1. [Command line flags](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/config.rs#L59-L70): each field in that struct is [parsed by the driver](https://github.com/viperproject/prusti-dev/blob/864ff5677901367501a01b7f2b65e689ff37953e/prusti/src/driver.rs#L150-L163) and fed down.
2. [SETTINGS lazy object](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/config.rs#L74-L113) that is initialized from `Prusti.toml` and environment variables.

Until recently, these two sets of settings were completely disjoint. However, recently I had to allow changing one of the `SETTINGS` via [a command-line flag](https://github.com/viperproject/prusti-dev/blob/864ff5677901367501a01b7f2b65e689ff37953e/prusti/src/driver.rs#L162-L163). Therefore, we think it would be good to merge command-line flags into the `SETTINGS` object by implementing the aforementioned `Source` trait.

@fpoli Did I forget anything?

@gavinleroy Is it clear how to proceed?

gavinleroy:
@vakaras Yes, that is clear. 

I do have two questions regarding changes to the `ConfigFlags` struct: 

[...]

PR #325, opened 2020-12-20
Unify handling of configuration flags

gavinleroy:
This PR is to close issue #184.

Configuration settings are automatically read from flags specified from the `-P` arguments. The shape of command-line arguments for configuration is `-P<identifier>=<value>`.

For example, you could pass the argument `-Puse_more_complete_exhale=false` to override this flag. 

It is notable to mention that command-line arguments use an underscore '_' instead of a dash '-' to separate words. 
Additionally, there is no longer a need for both the `skip_verify` and `no_verify` flag, I opted to remove `skip_verify` as it was part of the old `ConfigFlags` struct and remain with the `no_verify` configuration flag. All UI tests have been updated to reflect this change.


prusti-common/src/config.rs
```diff
@@ -105,6 +92,12 @@ lazy_static! {
         settings.set_default("JSON_COMMUNICATION", false).unwrap();
         settings.set_default("OPTIMIZATIONS","all").unwrap();
 
+        // Default cl arguments from the old ConfigFlags struct
```
vakaras:
```suggestion

```
Since `ConfigFlags` is being deleted, this comment is not too helpful.

vakaras (approved):
Looks very good! Thank you!

PR #326, opened 2020-12-22
Initial Purification Optimization

tillarnold:
@vakaras this is some initial code for the Purification Optimization. The way that the bodies of the automatized functions is generated here is wrong I'm pretty sure.

### DIFF
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -434,13 +434,6 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
-    pub fn int(value: i64) -> Self {
-        Expr::Const(
-            Const::Int(value),
-            Position::default(),
-        )
-    }
-
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
@@ -471,7 +464,9 @@ impl Expr {
     pub fn domain_func_app(
         func: DomainFunc,
         args: Vec<Expr>,
-    ) -> Self { Expr::DomainFuncApp(func, args, Position::default()) }
+    ) -> Self {
+        Expr::DomainFuncApp(func, args, Position::default())
+    }
 
     pub fn magic_wand(lhs: Expr, rhs: Expr, borrow: Option<Borrow>) -> Self {
         Expr::MagicWand(box lhs, box rhs, borrow, Position::default())
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -320,6 +320,11 @@ impl Expr {
         Expr::PredicateAccessPredicate(name.to_string(), box place, perm, pos)
     }
 
+    pub fn field_access_predicate(place: Expr, perm: PermAmount) -> Self {
+        let pos = place.pos();
+        Expr::FieldAccessPredicate(box place, perm, pos)
+    }
+
     pub fn pred_permission(place: Expr, perm: PermAmount) -> Option<Self> {
         place
             .typed_ref_name()
@@ -429,6 +434,13 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
+    pub fn int(value: i64) -> Self {
+        Expr::Const(
+            Const::Int(value),
+            Position::default(),
+        )
+    }
+
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
@@ -455,6 +467,11 @@ impl Expr {
     ) -> Self {
         Expr::FuncApp(name, args, internal_args, return_type, pos)
     }
+
+    pub fn domain_func_app(
+        func: DomainFunc,
+        args: Vec<Expr>,
+    ) -> Self { Expr::DomainFuncApp(func, args, Position::default()) }
 
     pub fn magic_wand(lhs: Expr, rhs: Expr, borrow: Option<Borrow>) -> Self {
         Expr::MagicWand(box lhs, box rhs, borrow, Position::default())
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -434,6 +434,13 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
+    pub fn int(value: i64) -> Self {
+        Expr::Const(
+            Const::Int(value),
+            Position::default(),
+        )
+    }
+
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -434,13 +434,6 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
-    pub fn int(value: i64) -> Self {
-        Expr::Const(
-            Const::Int(value),
-            Position::default(),
-        )
-    }
-
     pub fn unfolding(
         pred_name: String,
         args: Vec<Expr>,
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -307,7 +307,7 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
     
         for f in self.pure_functions.borrow().values().into_iter() {
             let old_formal_args = f.formal_args.clone();
-            let mut purifier = ExprPurifier {
+            let mut purifier = snapshot::ExprPurifier {
                 old_formal_args: old_formal_args.clone(),
                 snapshots: snapshots_info.clone(),
             };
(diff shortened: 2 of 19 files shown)


prusti-viper/src/encoder/encoder.rs
```diff
@@ -274,10 +276,103 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+        .borrow()
+        .clone();
+
+        for f in self.pure_functions.borrow().values().into_iter() {
+            let old_formal_args = f.formal_args.clone();
+            let mut purifier = ExprPurifier {old_formal_args: old_formal_args.clone(), snapshots: snapshots_info.clone()};
+
+            let mut formal_args: Vec<vir::LocalVar> = f.formal_args.iter().map(|e| {
+                    let old_type = e.typ.clone();
+                    let new_type =  purifier.transalte_type(old_type);
```
vakaras:
```suggestion
                    let new_type =  purifier.translate_type(old_type);
```

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1551,3 +1646,104 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         substs.join(";")
     }
 }
+
+
+
+struct ExprPurifier {
```
vakaras: Could you please move this to a separate file?

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -32,20 +33,21 @@ pub struct SnapshotEncoder<'p, 'v: 'p, 'tcx: 'v> {
...
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct SnapshotDomain {
     pub domain: vir::Domain,
     pub equals_func: vir::Function,
     pub equals_func_ref: vir::Function,
     pub not_equals_func: vir::Function,
     pub not_equals_func_ref: vir::Function,
+    pub fields: Vec<String>,
```
vakaras: Why do you need this?

prusti-common/src/vir/ast/expr.rs
```diff
@@ -425,6 +430,13 @@ impl Expr {
         Expr::Cond(box guard, box left, box right, Position::default())
     }
 
+    pub fn int(value: i64) -> Self {
+        Expr::Const(
+            Const::Int(value),
+            Position::default(),
+        )
+    }
+
```
fpoli, line 439: We already have `Expr::from(value)`, which is implemented for many integer types.

prusti-viper/src/encoder/snapshot/purifier.rs
```diff
@@ -56,16 +86,15 @@ impl ExprFolder for ExprPurifier {
     fn fold_local(&mut self, v: LocalVar, p: Position) -> Expr {
         if v.name == "__result" {
             self.self_function.clone()
+        } else {
+            Expr::Local(
+                LocalVar {
+                    name: v.name,
+                    typ: super::transalte_type(v.typ, &self.snapshots),
```
Aurel300:
```suggestion
                    typ: super::translate_type(v.typ, &self.snapshots),
```

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -223,25 +245,31 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
                         }
                     }
                 }
 
                 true
             }
+            ty::TyKind::Adt(adt_def, subst) if adt_def.is_box() => {
+                parent_boxes.push(ty);
+                let is = self.is_ty_supported(ty.boxed_ty(), parent_boxes);
+                parent_boxes.retain(|&x| x != ty);
```
Aurel300: Is it not enough to just `pop` here?

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -630,10 +658,56 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
     {
-        let snapshot = self.encoder.encode_snapshot(field_ty)?;
-        Ok(snapshot.snap_call(
-            self.encode_arg_field(location, field)
-        ))
+        warn!("encode_snap_arg begin");
+        let res = match field_ty.kind() {
+        ty::TyKind::Adt(adt_def, _) if adt_def.is_box() => { //TODO this breaks for lots of boxes. e.g. Box<Box<T>> or Box<i32>
+            let boxed_ty = field_ty.boxed_ty();
+            warn!("enocding boxed ty {:?}", boxed_ty);
```
Aurel300: Can you invoke the snapshot encoding recursively here, i.e. `self.encoder.encode_snapshot(boxed_ty)`?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -249,10 +259,248 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+            domains.push(self.axiomatized_function_domain.borrow().clone());
+            domains.push(self.get_nat_domain());
+            domains.push(self.get_primitive_valid_domain())
+        }
+
         domains.sort_by_key(|d| d.get_identifier());
         domains
     }
 
+    fn get_primitive_valid_domain(&self) -> vir::Domain {
```
fpoli: This and the `get_nat_domain` implementation might go to `prusti-viper/src/encoder/builtin_encoder.rs` (e.g. adding a `BuiltinDomainKind`) to avoid making `encoder.rs` longer.

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -177,33 +180,61 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
+                            ).encode_snapshot()?
+                        } else {
+                            unreachable!()
+                        }
+                    }
+                    x => unimplemented!("{:?}", x),
+                }
             }
-
             x => unimplemented!("{:?}", x),
```
fpoli: It would be better to return `EncodingError::unsupported(..)` instead of calling `unimplemented!(..)`.

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -555,58 +630,84 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
+                            vir::Position::default(),
+                        ))
+
+                        
+                    } else {
+                        unreachable!()
+                    }
                 }
-                args
+                x => unimplemented!("{:?}", x),
```
fpoli: `EncodingError::unsupported(..)`?

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -647,24 +748,700 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
+                    let variant_name = variant.ident.name.to_ident_string();
+                    axioms.push(self.encode_enum_variant_cons_axiom(variant)?);
+                    for field in &variant.fields {
+                        funcs.push(self.encode_field_func(field, Some(variant_name.clone()))?);
+                        axioms.push(self.encode_field_axiom(field, Some(variant))?);
+                    }
+                }
+            }
+            _ => {
+                unimplemented!();
```
fpoli: `EncodingError::unsupported(..)`?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -249,10 +259,248 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+        vir::Domain {
+            name: nat_domain_name.to_owned(),
+            functions,
+            axioms: vec![],
+            type_vars: vec![],
+        }
+    }
+
+    fn encode_axiomatized_pure_function(&self, f: &vir::Function) {
+        let snapshots_info: HashMap<String, Box<Snapshot>> = self.snapshots.borrow().clone();
```
fpoli: I wonder if it's possible to avoid cloning the map. The `ExprPurifier` could take a reference to the map, for example.

prusti-viper/src/encoder/snapshot_encoder.rs
tillarnold: I think that would cause infinite recursion. For example if I have a struct
```rust
struct List {
    value: u32,
    next: Option<Box<List>>,
}
```
`encode_snapshot(List)` would call `encode_snapshot(Option<Box<List>>)` and then that would call `encode_snapshot(Box<List>)` which calls `encode_snap_arg(Box<List>)` and then `encode_snapshot(List)` (because we'd call `self.encoder.encode_snapshot(boxed_ty)` where the `boxed_ty` is `List` )

This admittedly rather ugly code I wrote is an attempt to prevent this infinite recursion by only computing the parts of the snapshot we actually need.

Aurel300: I understand the issue, but I just doing another (less complete) copy of the encoding for boxes is not really a solution. You could split ADT encoding into two phases, such that the `snapshot` map in `Encoder` contains an entry for the snapshot being encoded before the recursive case is hit.

prusti-viper/src/encoder/encoder.rs
tillarnold: That was in fact possible. Thank you for pointing this out.

prusti-viper/src/encoder/encoder.rs
tillarnold: As you suggested I have moved this to `builtin_encoder.rs` and added `BuiltinDomainKind`

vakaras:
The changes of this PR were merged as part of #354.

prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -647,24 +737,694 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
...
+                self.subst
+            )?;
+            Ok(
+                vir::Expr::unfolding(
+                    predicate_name,
+                    vec![variant_location],
+                    vir::Expr::domain_func_app(cons_func, args),
+                    vir::PermAmount::Read,
+                    enum_variant_index,
+                )
```
Aurel300, line 1392: @tillarnold Why is this needed? Does the fold/unfold step fail to generate the variant unfolding automatically?

tillarnold: I am not sure why this is needed. I think (if I did git blame correctly) this was originally added in commit https://github.com/viperproject/prusti-dev/commit/8e0bf653d7e114e25e7e7f36dbc16f9c497983b0 so I don't think I'm the original author of this.

PR #327, opened 2020-12-28
Analysis Framework

ELowis:
- Built a simple MIR body analyzer using abstract interpretation. It iterates through the program statements in program order (or rather basic block index order) and calls abstract transformers and lattice operations defined in abstract domains.
- Translated the analyses from prusti-interface (initialization -> definitely_initialized; liveness -> reaching_definitions) into the new framework as a proof of concept and to separate them crate-wise from Prusti, since they are independent (, and to be more modular).
- Implemented a way to test the analyses using serde for serialization. Some simple test cases are included.

The final goal for the future is to integrate the PCS analysis from https://github.com/viperproject/rust-pcs into the framework and use it to generate fold/unfold statements in Prusti.

This work was requested and supervised by @fpoli.


fpoli:
Thanks for the PR! I'll review it in the next few days.

analysis/src/abstract_domains/definitely_initialized.rs
```diff
@@ -4,11 +4,323 @@
...
+                        place2
+                    );
+                }
+            }
+        }
+    }
+
+    /// Sets `place` as definitely initialized (see place_set/insert()
+    fn set_place_initialised(&mut self, place: &mir::Place<'tcx>) {
+        self.check_invariant();
```
fpoli: I guess that checks like this are expensive, so you could enclose them in `if cfg!(debug_assertions) { ... }` to run them only in debug mode (from [here](https://stackoverflow.com/questions/39204908/how-to-check-release-debug-builds-using-cfg-in-rust)).

Maybe replacing all `assert!` with `debug_assert!` inside `check_invariant` is actually enough to optimize away the call.

analysis/src/abstract_domains/definitely_initialized.rs
```diff
@@ -4,11 +4,323 @@
...
+            places.insert(local.clone().into());
+        }
+        Self {def_init_places: places, mir, tcx}
+    }
+
+    fn need_to_widen(_counter: &u32) -> bool {
+        false   //TODO: check
+    }
+
+    /// = intersection of place sets
```
fpoli: Nit: just for the `///` doc comments, could you start them with an uppercase letter and make them more sentence-like? (here and elsewhere)

analysis/src/abstract_domains/reaching_definitions.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// A set of definition locations and function parameter indices per Local,
+/// meaning that the Local might still have the value
+/// which was assigned at the location or passed as a parameter
+///
+/// derived from prusti-interface/.../mir_analyses/liveness
+#[derive(Clone)]
+pub struct ReachingDefsState<'a, 'tcx: 'a> {
+    // Local -> Location OR index of function parameter
+    reaching_assignments: HashMap<mir::Local, HashSet<Result<mir::Location, usize>>>,
```
fpoli: I would be better to define a new enumeration instead of reusing `Result`, because `Result` suggests that the second element describes an error.

analysis/src/abstract_domains/reaching_definitions.rs
```diff
@@ -0,0 +1,172 @@
...
+        f.debug_struct("ReachingDefsState")
+            .field("reaching_assignments", &self.reaching_assignments)
+            .finish()
+    }
+}
+
+impl<'a, 'tcx: 'a> PartialEq for ReachingDefsState<'a, 'tcx> {      // manual implementation needed, because not implemented for Body
+    fn eq(&self, other: &Self) -> bool {
+        self.reaching_assignments == other.reaching_assignments
+        // ignore Body
```
fpoli: Maybe you don't need it, but `Body` implements [`HashStable`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Body.html#impl-HashStable%3CStableHashingContext%3C%27__ctx%3E%3E) which provides an identifier that you can use in comparisons.

analysis/src/abstract_domains/reaching_definitions.rs
```diff
@@ -0,0 +1,172 @@
...
+            mir::StatementKind::Assign(box (ref target, _)) => {
+                if let Some(local) = target.as_local() {
+                    let location_set = self.reaching_assignments.entry(local).or_insert(HashSet::new());
+                    location_set.clear();
+                    location_set.insert(Ok(*location));
+                }
+                Ok(())
+            }
+            _ => {Ok(())}
+        }
```
fpoli: You could move the `Ok()` after the match.

analysis/src/abstract_state.rs
```diff
@@ -4,7 +4,46 @@
...
-    fn join(self, other: Self) -> Self;
-    // ...
+use rustc_middle::mir;
+use rustc_middle::ty::TyCtxt;
+use std::vec::Vec;
+use crate::AnalysisError;
+use serde::Serialize;
+
+
+pub trait AbstractState<'a, 'tcx: 'a>: Clone + Eq + Sized + Serialize {   // Sized needed for apply_terminator_effect's return type
```
fpoli, line 44: Could you add a doc comment with a summary of the properties that should be ensured by the implementation? Like "forall x :: join(bottom(), x) == x".

analysis/src/analyzer.rs
```diff
@@ -9,38 +9,137 @@ use rustc_middle::mir;
...
+                if prev_state.into_iter().any(|s| s == &current_state) {      //use .contains when it becomes stable
+                    // same state, don't need to reiterate
+                    continue 'block_loop;
+                }*/
+                p_state.set_before(&location, current_state.clone());
+                // normal statement
+                let res = current_state.apply_statement_effect(&location);
+                if res.is_err() {
+                    return Err(res.unwrap_err());
+                }
```
fpoli: Lines 91-94 can be replaced with `current_state.apply_statement_effect(&location)?;`. The [`?` operator](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html) makes the code look nicer.

analysis/src/analyzer.rs
```diff
@@ -9,38 +9,137 @@ use rustc_middle::mir;
...
+            let location = mir.terminator_loc(bb);
+            /* Too much performance overhead to check for every statement?
+            let prev_state = p_state.lookup_before(&location);
+            if prev_state.into_iter().any(|s| s == &current_state) {
+                // same state, don't need to reiterate
+                continue 'block_loop;
+            }*/
+            p_state.set_before(&location, current_state.clone());
+
+            let term_res = current_state.apply_terminator_effect(&location);
```
fpoli: As above: `let next_states = current_state.apply_terminator_effect(&location)?;`

analysis/src/serialization_utils.rs
```diff
@@ -0,0 +1,19 @@
...
+pub fn location_to_stmt_str(location: &mir::Location, mir: &mir::Body) -> String {
+    let bb_mir = &mir[location.block];
+    if location.statement_index < bb_mir.statements.len() {
+        let stmt = &bb_mir.statements[location.statement_index];
+        format!("{:?}", stmt)
+    } else { // location = terminator
+        let terminator = bb_mir.terminator();
+        format!("{:?}", terminator.kind)
+    }
+}
```
fpoli, line 19: Could you add a newline at the end (here and in other `.rs` files)?

fpoli (commented):
Looks good to me! I left some comments.

analysis/src/abstract_domains/reaching_definitions.rs
ELowis: I assumed that only states from the same analysis (with the same Body) would be compared. And the Body is not really part of the "actual state", it is only there to perform some lookup. But I can still include the Body in the comparison, if you wish.

fpoli: We can add the comparison in a `debug_assert!(..)` to just check it while testing Prusti.

ELowis:
Could you have a look at the new commits after your comments to see if they resolve the issues properly?

PR #328, opened 2020-12-31
Added hint to install rustup to readme

ELowis:
May be obvious, but for me it wasn't without looking at x.py.


fpoli:
I added a link to the user and developer guides to fix #338.

PR #329, opened 2020-12-31
x.py setup_win: Better handling of existing viper_tools folder

ELowis:
Prevent exiting with a cryptic FileNotFound error after downloading Viper if the folder already exists, instead exit before downloading with a FileExistsError.


fpoli:
Thanks

PR #336, opened 2021-01-03
update Readme to ensure the build instructions actually work

mabbamOG:
this silly oversight took me forever to debug, hopefully the update helps others successfully complete the build

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,4 +27,9 @@ The easiest way to try out Prusti is by using the ["Prusti Assistant"](https://m
 
 Alternatively, if you wish to use Prusti from the command line there are two options:
 * Download the precompiled binaries for Ubuntu, Windows, or MacOS from a [GitHub release](https://github.com/viperproject/prusti-dev/releases).
-* Compile from the source code, by first running `./x.py setup` and then afterwards `./x.py build --release`.
+* Compile from the source code, by running:
+    
+    ```bash
+    $ python3 x.py setup
+    $ env VIPER_HOME=$PWD/viper_tools/backends/ Z3_EXE=$PWD/viper_tools/z3/bin/ cargo build --release --all
+    ```


README.md
```diff
@@ -27,4 +27,9 @@ The easiest way to try out Prusti is by using the ["Prusti Assistant"](https://m
 
 Alternatively, if you wish to use Prusti from the command line there are two options:
 * Download the precompiled binaries for Ubuntu, Windows, or MacOS from a [GitHub release](https://github.com/viperproject/prusti-dev/releases).
-* Compile from the source code, by first running `./x.py setup` and then afterwards `./x.py build --release`.
+* Compile from the source code, by running:
+    
+    ```bash
+    $ python3 x.py setup
+    $ env VIPER_HOME=$PWD/viper_tools/backends/ Z3_EXE=$PWD/viper_tools/z3/bin/ cargo build --release --all
```
vakaras, line 34: The second command should be equivalent to `./x.py build --release --all`. Does it fail for you for some reason?

Also, `./x.py` has `#!/usr/bin/env python3` at the top, so `./x.py` should *just* work. Does it fail on your machine?

mabbamOG: it did fail for me, because VIPER_HOME and Z3_EXE were not set correctly, and this led to all sorts of errors.
it is also not clear to the user what they need to point to.

there might also be a bug in my commit, i think your suite wants Z3_EXE to point to a binary rather than a directory:

```
export Z3_EXE=$PWD/viper_tools/z3/bin/z3
```

for improved portability, i recommend using cargo build scripts in the future, rather than `x.py`.

vakaras: > it did fail for me, because VIPER_HOME and Z3_EXE were not set correctly

That is strange and should be fixed. Could you please post the output that gets printed by executing:
```
./x.py ++verbose build --all
```

>  it is also not clear to the user what they need to point to.

The whole point of having `x.py` is to make sure that the user does not need to think about this. If `x.py` fails, that is a bug and needs to be fixed.

> for improved portability, i recommend using cargo build scripts in the future

Cargo build scripts cannot (at least this was the case when we introduced `x.py`) set environment variables for dependencies. If you happen to know a workaround, please share it ☺.

mabbamOG: after running everything with `x.py` on a new directory, everything surprisingly builds just fine. 🤔 🤔 
given this result, i'm not sure why i had issues with the build in the first place... i suspect a problem during the setup, or when i configured the dependencies for the first time.

I suppose we can close the issue now, and hope no other users experience similar problems. thank you for the support @vakaras !

> Cargo build scripts cannot (at least this was the case when we introduced x.py) set environment variables for dependencies. If you happen to know a workaround, please share it ☺.
I see... perhaps the dependency on environment variables for the internal build can be removed altogether?

vakaras: >> Cargo build scripts cannot (at least this was the case when we introduced x.py) set environment variables for dependencies. If you happen to know a workaround, please share it relaxed.
>
> I see... perhaps the dependency on environment variables for the internal build can be removed altogether?

That would require to patch (if I am not mistaken) the `jni` crate. Maybe, that is the proper way of dealing with this.

PR #342, opened 2021-01-06
Verify Max Coins in Binary Tree

kammola:


kammola:
@vakaras However the file is verified successfully on my machine, it fails to pass the tests here. The error states some pre-conditions of some functions might not be true at two function calls. The first one is in line 151. However, the `add` function doesn't have any preconditions and I don't know why the other one fails at line 213.

vakaras:
This looks like a bug in Prusti and needs to be investigated. Just leave it as it is for now.

prusti-tests/tests/verify_overflow/pass/competitive_programming/dp-on-trees/MaxCoinsInBinaryTree.rs
```diff
@@ -0,0 +1,217 @@
...
+        None => {}
+        Some(box l) => {
+            result = add(result, solve2_rec(&l));
+        }
+    }
+
+    match &(*node).right {
+        None => {}
+        Some(box r) => {
+            result = add(result, solve2_rec(&r));
```
vakaras, line 151: I have looked into the generated Viper code. The error reported on the Viper level is that the return value of `solve2_rec` is not guaranteed to be in the range of `isize`, which is clearly a bug in either Prusti or Viper.

fpoli: Could it be related to https://github.com/viperproject/silicon/issues/367?

vakaras: > Could it be related to [viperproject/silicon#367](https://github.com/viperproject/silicon/issues/367)?

I do not see from where we could get non-linear arithmetic in this example, so it is probably unrelated.

vakaras:
Superseded by #508 .

Issue #249, opened 2020-11-18
Proper shutdown of prusti-server


fpoli:
Sending a SIGINT to `prusti-server` should ensure that all its (transitive) children are killed.

vakaras:
I think that `prusti-server` should ensure that **all** its (transitive) children are killed.

fpoli:
> I think that `prusti-server` should ensure that **all** its (transitive) children are killed.

Yes. Making this cross-platform might be tricky.

fpoli:
This is another good first issue for new contributors!

Please ask for more information in case you are interested.

gavinleroy:
Hi, I would be interested in working on this issue. 

Is there further information on the issue or a way to reproduce the bug?

fpoli:
On Ubuntu, here is how I can reproduce it:

1. Start the server in one terminal: `./target/debug/prusti-server --port 1234`
2. In another terminal, run `ps a | grep prusti | grep -v grep`. You should see a `prusti-server` process and its child `prusti-server-driver`. For example:
    ```
    18163 pts/19   S+     0:00 ./target/debug/prusti-server --port 1234
    18164 pts/19   Sl+    0:00 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port 1234
    ```
3. Send a `SIGINT` to the parent process with e.g. `kill -s SIGINT 18163`.
4. Rerun `ps a | grep prusti | grep -v grep`. You should see that `prusti-server-driver` is still alive.

Note that if instead of `kill -s SIGINT ...` you kill the server with `Ctrl-C` in the first terminal then all the processes die and you don't see the issue, because [`Ctrl-C` sends a `SIGINT` to all the processes in the foreground process group](https://stackoverflow.com/a/8406413/2491528).

fpoli:
When you verify a program while the server is running, `prusti-server-driver` runs Viper in a JVM thread, which in turn spawns additional `z3` or `boogie` processes. As far as I know, Viper doesn't guarantee to stop all those processes when the JVM gets stopped. So, both the Prusti and Viper IDE extensions use [tree-kill](https://github.com/pkrumins/node-tree-kill), whose implementation first obtains the tree of processes and then sends a signal to each of them. The ideal solution to this issue is to move that logic inside `prusti-server`, such that a `SIGINT` to the server guarantees to stop all subprocesses.

If you ever need it, a command to launch a long-running verification is the following
```
[...]

PR #344, opened 2021-01-08
Killing prusti-server process tree

gavinleroy:
This PR is to close issue #249.

SIGINT signals (CTRL_C_EVENT and CTRL_BREAK_EVENT on Windows) are handled by killing the process tree rooted at the prusti-server process to ensure that all children are killed. On *nix, this is done by simply killing the process group, and on Windows, it is handled by `taskkill /pid PID /t /f `.

### DIFF
--- a/prusti-launch/src/bin/prusti-server.rs
+++ b/prusti-launch/src/bin/prusti-server.rs
@@ -4,8 +4,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::{env, path::PathBuf, process::Command};
+use std::{env, path::PathBuf, process::{self, Command, Stdio}};
 use prusti_launch::{find_viper_home, find_z3_exe};
+#[cfg(target_family = "unix")]
+use nix::{unistd::getpgrp, sys::signal::{Signal, killpg}};
 
 fn main() {
     if let Err(code) = process(std::env::args().skip(1).collect()) {
@@ -62,6 +64,9 @@ fn process(args: Vec<String>) -> Result<(), i32> {
         }
     };
 
+    // Register the SIGINT handler; CTRL_C_EVENT or CTRL_BREAK_EVENT on Windows
+    ctrlc::set_handler(sigint_handler).expect("Error setting Ctrl-C handler");
+
     let exit_status = cmd.status().expect("could not run prusti-server-driver");
 
     if exit_status.success() {
@@ -69,4 +74,23 @@ fn process(args: Vec<String>) -> Result<(), i32> {
     } else {
         Err(exit_status.code().unwrap_or(-1))
     }
+}
+
+#[cfg(target_family = "unix")]
+fn sigint_handler() {
+    // Killing the process group terminates the process tree
+    killpg(getpgrp(), Signal::SIGKILL).expect("Error killing process tree.");
+}
+
+#[cfg(target_family = "windows")]
+fn sigint_handler() {
+    // Kill process tree rooted at prusti-server.exe
+    let kill_command: &str = &*format!("TASKKILL /PID {} /T /F", process::id());
+
+    Command::new("cmd")
+        .args(&["/C", kill_command])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()
+        .expect("Error killing process tree.");
 }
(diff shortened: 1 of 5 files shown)


prusti-launch/src/bin/prusti-server.rs
```diff
@@ -70,3 +75,22 @@ fn process(args: Vec<String>) -> Result<(), i32> {
...
+    // Kill process tree rooted at prusti-server.exe
+    let kill_command: &str = &*format!("TASKKILL /PID {} /T /F", process::id());
+
+    Command::new("cmd")
+        .args(&["/C", kill_command])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .spawn()
+        .expect("Error killing process tree.");
+}
```
Aurel300:
```suggestion
}

```

prusti-launch/src/bin/prusti-server.rs
```diff
@@ -70,3 +75,22 @@ fn process(args: Vec<String>) -> Result<(), i32> {
...
+    killpg(getpgrp(), Signal::SIGKILL).expect("Error killing process tree.");
+}
+
+#[cfg(target_family = "windows")]
+fn sigint_handler() {
+    // Kill process tree rooted at prusti-server.exe
+    let kill_command: &str = &*format!("TASKKILL /PID {} /T /F", process::id());
+
+    Command::new("cmd")
+        .args(&["/C", kill_command])
```
Aurel300: I wonder, why not just spawn `taskkill` with the arguments above, instead of spawning `cmd` first?

fpoli:
Thanks for the PR, the process group solution is really nice!

* Could you add the same handler to the `prusti-rustc` binary? I forgot to mention that we have the same kind of issues there (`prusti-rustc` spawns `prusti-driver`, which starts its own JVM with Viper if the server address is not configured).

* Shouldn't the `prusti-rustc`/`prusti-server` also create the process group with `setpgid` before setting the signal handler? My understanding is that when running from a terminal it's the terminal that creates the process group (the "notes" of [this](https://man7.org/linux/man-pages/man2/setpgid.2.html) mention that bash calls `setpgid` in order to implement shell job control). In Prusti we also run the server from a VS Code extension (nodejs) and from the `x.py` script. In those cases, the process group might not exist or might include more processes than just Prusti.

prusti-launch/src/bin/prusti-server.rs
gavinleroy: Both the above suggestions are good. First spawning `cmd` then passing `taskkill` as an argument was an artifact of a previous implementation that I forgot to change. It's now updated.

gavinleroy:
> Shouldn't the `prusti-rustc`/`prusti-server` also create the process group with `setpgid` before setting the signal handler? My understanding is that when running from a terminal it's the terminal that creates the process group (the "notes" of [this](https://man7.org/linux/man-pages/man2/setpgid.2.html) mention that bash calls `setpgid` in order to implement shell job control). In Prusti we also run the server from a VS Code extension (nodejs) and from the `x.py` script. In those cases, the process group might not exist or might include more processes than just Prusti.

Yes you are correct that `prusti-rustc`/`prusti-server` should be the group leader before setting the signal handler. I was (incorrectly) assuming that the server wouldn't be run as a child process. You can see this update reflected [here](https://github.com/gavinleroy/prusti-dev/blob/bb346bdb7e7efd9323753267bad6e030f114c611/prusti-launch/src/bin/prusti-server.rs#L68-L69). You'll notice that I am ignoring the `Result` from `setpgid`. From the [setpgid man page](https://man7.org/linux/man-pages/man2/setpgid.2.html) the possible errors are the following:

```
       EACCES An attempt was made to change the process group ID of one
              of the children of the calling process and the child had
              already performed an execve(2) (setpgid(), setpgrp()).

       EINVAL pgid is less than 0 (setpgid(), setpgrp()).

       EPERM  An attempt was made to move a process into a process group
              in a different session, or to change the process group ID
              of one of the children of the calling process and the
              child was in a different session, or to change the process
              group ID of a session leader (setpgid(), setpgrp()).

       ESRCH  For getpgid(): pid does not match any process.  For
              setpgid(): pid is not the calling process and not a child
```
Because I am setting the pgid to be the pid, the only applicable error from above is `EPERM` which can be thrown when the process is already the group leader. This is expected to happen sometimes depending on how prusti-server is run so I thought it best to ignore the result.

fpoli:
Thanks! I added a comment to remember why the result is ignored.

PR #351, opened 2021-01-11
Workaround for a snapshot/ref encoding bug

Aurel300:
This is a small, somewhat hacky fix to a problem that happens when ADTs with `Ref` fields are encoded into snapshots. For some reason, `unfolding` is not properly applied within the snapshot constructor arguments. With this PR, this example works:

```rust
use prusti_contracts::*;

#[derive(PartialEq, Eq)]
struct A<'a> {
    i: &'a mut i32,
}

#[requires(_x == _y)]
fn test_eq(_x: &A, _y: &A) {}

fn main() {}
```

### DIFF
--- a/prusti-viper/src/encoder/snapshot_encoder.rs
+++ b/prusti-viper/src/encoder/snapshot_encoder.rs
@@ -183,6 +183,7 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
                     unreachable!()
                 }
             }
+            // Refs should be stripped by now by [Encoder::encode_snapshot]
             x => unimplemented!("{:?}", x),
         })
     }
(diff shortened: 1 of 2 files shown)


prusti-viper/src/encoder/snapshot_encoder.rs
```diff
@@ -183,6 +183,7 @@ impl<'p, 'v, 'r: 'v, 'a: 'r, 'tcx: 'a> SnapshotEncoder<'p, 'v, 'tcx> {
                     unreachable!()
                 }
             }
+            // Refs should be stripped by now by [Encoder::encode_snapshot]
             x => unimplemented!("{:?}", x),
```
fpoli, line 187: (Not for this PR) at some point, we should replace this `unimplemented` with an  `EncodingError::internal` in case `x` is a reference and with `EncodingError::unsupported` in all the remaining cases.

fpoli:
> unfolding is not properly applied within the snapshot constructor arguments

When there will be time, I would like to see the VIR code on which the algorithm fails to add unfoldings.

prusti-viper/src/encoder/snapshot_encoder.rs
Aurel300: The case should actually never be reached because this is only invoked after the snapshot check for `is_ty_supported`. However, `is_ty_supported` returns `true` for `Ref`s (again because they are handled in `Encoder::encode_snapshot`, so I added an explanation for why they don't appear here.

Aurel300:
Thanks @cmatheja for the help :)

PR #355, opened 2021-01-14
Add interner to safely shorten Viper identifiers

fpoli:
The interner can be disabled with `PRUSTI_INTERN_NAMES=false`. It does a good job at shortening function names and types, except for the annoying `_beg_..._sep_..._end_` suffix that is kept for the encoding of generics.

### DIFF
--- a/prusti-viper/src/encoder/name_interner.rs
+++ b/prusti-viper/src/encoder/name_interner.rs
@@ -1,0 +1,129 @@
+// © 2021, ETH Zurich
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::{HashMap, HashSet};
+use log::debug;
+
+/// Name interner.
+/// This structure can be used to shorten long unique names without losing the uniqueness property.
+#[derive(Debug)]
+pub struct NameInterner {
+    name_to_symbol: HashMap<String, String>,
+    used_symbols: HashSet<String>,
+}
+
+impl NameInterner {
+    pub fn new() -> Self {
+        NameInterner {
+            name_to_symbol: HashMap::new(),
+            used_symbols: HashSet::new(),
+        }
+    }
+
+    /// Intern a full unique name, returning a possibly readable string that uniquely identifies it.
+    /// The `readable_names` must not collide with past or future `full_unique_name`s.
+    pub fn intern<S: ToString>(&mut self, full_unique_name: S, readable_names: &[S]) -> String {
+        let full_unique_name = full_unique_name.to_string();
+        let readable_names: Vec<_> = readable_names.iter().map(|s| s.to_string()).collect();
+        assert!(!readable_names.contains(&"".to_string()));
+
+        // Return the symbol, if we already interned the full name
+        if let Some(symbol) = self.name_to_symbol.get(&full_unique_name) {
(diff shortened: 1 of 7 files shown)


prusti-viper/src/encoder/name_interner.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[test]
+    fn test_name_interner() {
+        let mut interner = NameInterner::new();
+        assert_eq!(interner.intern("unreadable$name", &["name"]), "name");
+        assert_eq!(interner.intern("unreadable$name", &["name"]), "name");
+        assert_eq!(interner.intern("another$name", &["name"]), "another$name");
+        assert_eq!(interner.intern("another$name", &["name"]), "another$name");
+        assert_eq!(interner.intern("unreadable$name", &["name"]), "name");
+        assert_eq!(interner.intern("another$name", &["name"]), "another$name");
+        assert_eq!(interner.intern("third$name", "third"), "third");
```
Aurel300: Is this not a type error in the second `intern` argument?

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -0,0 +1,129 @@
...
+        if let Some(symbol) = self.name_to_symbol.get(&full_unique_name) {
+            debug!("Interning of {:?} is {:?}", &full_unique_name, symbol);
+            return symbol.clone();
+        }
+
+        // Check that the readable name is not equal to full names passed in the past.
+        assert!(!readable_names.iter().any(|r| self.name_to_symbol.contains_key(r)));
+
+        // Check that readable names passed in the past are not equal to the current full name.
+        assert!(!self.used_symbols.contains(&full_unique_name));
```
Aurel300: These look more like debug assertions? Surely we shouldn't need to check this for every `intern` call.

prusti-viper/src/encoder/name_interner.rs
fpoli: Yes, it is :/

prusti-viper/src/encoder/name_interner.rs
fpoli: Well, it's not ensured by the API of the `NameInterner` so it should always be checked, but I'm fine with making them `debug_asser`.

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -0,0 +1,129 @@
...
+            used_symbols: HashSet::new(),
+        }
+    }
+
+    /// Intern a full unique name, returning a possibly readable string that uniquely identifies it.
+    /// The `readable_names` must not collide with past or future `full_unique_name`s.
+    pub fn intern<S: ToString>(&mut self, full_unique_name: S, readable_names: &[S]) -> String {
+        let full_unique_name = full_unique_name.to_string();
+        let readable_names: Vec<_> = readable_names.iter().map(|s| s.to_string()).collect();
+        assert!(!readable_names.contains(&"".to_string()));
```
Aurel300: `readable_names` is not used until after the `if let` check; it can be calculated futher down. The assertion is a debug assertion?

fpoli: Regarding debug, as above. Regarding the position, I think it's best to check them as early as possible. So, I can move one of the following assertions; the other needs to stay after the `if let`.

fpoli: Actually, both the last two assertions need to stay after the `if let`.

fpoli: In the end I'm using `AsRef<str>` instead of `ToString`, to avoid the allocation at all.

PR #356, opened 2021-01-15
Update rustc to nightly-2021-01-15

viper-admin:
* Update rustc to `nightly-2021-01-15`.

@vakaras could you take care of this?


vakaras:
@Aurel300 Since the change seems to be minimal, I am going to merge it.

PR #390, opened 2021-02-05
Add predicate syntax

dario23:


dario23:
still missing:

 - [ ] check that the predicate functions are never actually called

prusti-specs/src/lib.rs
```diff
@@ -421,3 +421,37 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
         _ => { unimplemented!() }
     }
 }
+
+pub fn predicate(attr: TokenStream, tokens: TokenStream) -> TokenStream {
+    if !attr.is_empty() {
+        // TODO: what here? error? re-emit the previous attributes? not sure
+        //       what the semantics for that are supposed to be..
```
fpoli: Do you know an example that triggers this? In the other macros we just call the first parameter `_attr` and we never use it.

prusti-specs/src/lib.rs
```diff
@@ -421,3 +421,37 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+            } else {
+                unreachable!("A function's block must be a brace-delimited `TokenTree::Group`")
+            };
+
+            let sig = pred_func.sig.to_token_stream();
+            parse_quote_spanned! {item_span =>
+                #[requires(#assertion_tokens)]
+                #[pure]
+                // TODO: is `true` the only possibility here? do we require
+                //       all predicates to be `-> bool`?
```
fpoli: Is the comment is ignored by `parse_quote_spanned`? If not or if unsure, place it before the macro call. Anyway, I think it's correct to only have boolean predicates.

prusti-tests/tests/parse/ui/predicate_fail-1.rs
```diff
@@ -0,0 +1,14 @@
+// compile-flags: -Pprint_desugared_specs=true -Pprint_typeckd_specs=true -Pno_verify=true -Phide_uuids=true
+// normalize-stdout-test: "[a-z0-9]{32}" -> "$(NUM_UUID)"
+// normalize-stdout-test: "[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}" -> "$(UUID)"
```
fpoli, line 3: ~~These two lines should not be needed (here an in other files).~~ I found the `.stdout`

prusti-specs/src/lib.rs
dario23: turns out i misunderstood what this parameter is used for; just ignoring it seems like the best idea, given that's how the rest of the codebase deals with this, will change.

fpoli: `#[predicate(123)]` looks like an example (see [here](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros)), so raising a compilation error as you mention in the TODO sounds good. The `pure` and `trusted` macros don't do this check, but I think they should.

fpoli: To return a compilation error you can use `return syn::Error::new(span, "error message").to_compile_error();`. I hope you can obtain the span from `attr`, somehow.

dario23: i'd be happy to address this, but would open a separate PR for that, what do you think?

fpoli: For `pure` and `trusted` yes.

fpoli:
Just to be sure to cover all cases, could you add to `verify/fail` a test that raises a verification error (not parsing or type-checking error)? Like, a postcondition that calls a predicate containing `false`.

prusti-specs/src/lib.rs
```diff
@@ -421,3 +421,37 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+    let item_span = item.span();
+    match item {
+        syn::Item::Fn(pred_func) => {
+            let body_tokens = pred_func.block.to_token_stream();
+            // We've got a function, we know the body must be a braced group.
+            let assertion_tokens =if let Some(TokenTree::Group(group)) = body_tokens.into_iter().next() {
+                group.stream()
+            } else {
+                unreachable!("A function's block must be a brace-delimited `TokenTree::Group`")
+            };
```
Aurel300: Why match like this? Is this to strip the braces? I wonder if you can't simply do:
```suggestion
            let body = pred_func.block;
```
And then `#[requires #body ]`?

prusti-specs/src/lib.rs
Aurel300: Comments are not part of the AST (`syn`'s or Rust's), so these are definitely stripped during lexing of `lib.rs`.

prusti-specs/src/lib.rs
dario23: ah, very nice way to get rid of the braces, yes. didn't think of that before.

prusti-tests/tests/verify/fail/predicates/doesnt-hold.rs
```diff
@@ -0,0 +1,11 @@
+use prusti_contracts::*;
+
+#[predicate]
+fn false_p() -> bool {
+    false
+}
+
+#[ensures(false_p())] //~ ERROR precondition of pure function call might not hold
```
Aurel300: It's good that this raises an error and with a reasonable position, but now that I think about it, the error message is actually a bit confusing. "precondition of pure function" here refers to `false_p`, which is specified without a precondition as far as the user is concerned. Changing this would actually be a bit more involved:

- during macro processing of `#[predicate]`, mark the resulting function with some new prusti-reserved attribute
- after typing, predicates should be identified by the `SpecCollector`
  - `ProcedureSpecifications` should get a new field `predicate: bool` if the attribute was found in the HIR
- `ErrorCtxt` needs a new variant for predicate calls
- `encode_specified_pure_function_call` (in `procedure_encoder`) needs to identify when a pure function call is a call to a predicate (which will now be stored in the specification set)

We can wait until Monday if this is unclear, it also makes me a bit uneasy to introduce a relatively large set of changes to fix one error message…

prusti-specs/src/lib.rs
dario23: https://github.com/viperproject/prusti-dev/pull/391

prusti-tests/tests/verify/fail/predicates/doesnt-hold.rs
dario23: yep, i think i understand the problems you see here, but i agree with the "large set of changes to fix one error message". i'd rather get this feature landed at all first, and do more involved stuff like this subsequently.

dario23:
>     * [ ]  check that the predicate functions are never actually called

do you guys have suggestions where to put that check? i think we talked about somewhere in the spec collector in last monday's weekly, but i haven't seen any obvious locations (like visitors already checking other method calls for example).

edit: what i've had some contact with are [clippy lints](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/mut_mutex_lock.rs). i think the lint interface is at least somewhat reasonable, if not a perfect match, for checking forbidden method calls. on the other hand i haven't seen it used in prusti yet, so i'm assuming there's a way more in line with the existing codebase.

Aurel300:
> do you guys have suggestions where to put that check? i think we talked about somewhere in the spec collector in last monday's weekly, but i haven't seen any obvious locations (like visitors already checking other method calls for example).

Yes, it should be similar to `SpecCollector`, but ideally a separate module. The external specs module is also a visitor that does specifically a `call` match, if that helps. This will be a new type of check indeed, a sort of post-typing consistency check; for now let's just put the check for predicate calls there (make sure to stop checking once you enter Prusti code, i.e. `#[prusti::spec_only]`). We might find that there are checks further down the line (e.g. in the encoder) that actually make sense here, but let's not worry about that for this PR.

prusti-specs/src/lib.rs
```diff
@@ -118,12 +119,22 @@ fn generate_for_requires(attr: TokenStream, item: &untyped::AnyFnItem) -> Genera
...
 
-/// Generate spec items and attributes to typecheck th and later retrieve "ensures" annotations.
+/// Generate spec items and attributes to typecheck and later retrieve "ensures" annotations.
 fn generate_for_ensures(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedResult {
+    common_ensures(attr, item)
+}
+
+/// Generate spec items and attributes to typecheck and later retrieve "ensures_result_eq" annotations.
+fn generate_for_ensures_result_eq(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedResult {
+    let assertion = parse_quote_spanned!{item.span() => result == (#attr)};
```
Aurel300: This is not enough, as it still produces e.g. `result == forall(...)` and expects the preparser to be able to deal with it, which it cannot. The approach here is to parse the attribute as an *expression*, then add the `result == ...` over the parsed tree (without relying on the preparser).

dario23: and by *expression*, you mean a `prusti_specs::specifications:untyped::Assertion`, right?

Aurel300: Yeah, sorry, I was confused due to this being different in VIR. But this raises a good point, I'm not sure you can actually build an assertion that is `AssertionKind::Expr` representing `result == <another assertion>`. I'll need to think about this a bit more…

dario23: put up a WIP of reworking to handle predicates separately all the way through, still some work needed.

prusti-specs/src/lib.rs
```diff
@@ -435,3 +436,61 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+                unreachable!("a function's block must be a brace-delimited `TokenTree::Group`")
+            };
+
+            // TODO: this is very similar to generate_spec_item_fn, should extract that into a common function
+            let mut rewriter = rewriter::AstRewriter::new();
+            let spec_id = rewriter.generate_spec_id();
+            let assertion = handle_result!(rewriter.parse_assertion(spec_id, pred_tokens));
+
+            let item_name = syn::Ident::new(
+                &format!("prusti_{}_item_{}_{}", "pred", pred_func.sig.ident, spec_id),
```
fpoli: "pred" can be inlined in the format string.

dario23:
tests currently failing because the desugaring we generate for `#[predicate]` isn't handled by the viper-encoder properly yet.

prusti-specs/src/lib.rs
```diff
@@ -435,3 +436,65 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+                group.stream()
+            } else {
+                unreachable!("a function's block must be a brace-delimited `TokenTree::Group`")
+            };
+
+            // TODO: this is very similar to generate_spec_item_fn, should extract that into a common function
+            let mut rewriter = rewriter::AstRewriter::new();
+            let spec_id = rewriter.generate_spec_id();
+            let assertion = handle_result!(rewriter.parse_assertion(spec_id, pred_tokens));
+            let predicate = untyped::Predicate::new(assertion);
```
dario23: maybe we don't need this wrapper type, and should just go ahead with the inner assertion here?

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,118 @@
...
+    environment::Environment,
+    utils::has_prusti_attr,
+    PrustiError,
+};
+
+
+pub struct SpecChecker<'tcx> {
+    tcx: TyCtxt<'tcx>,
+
+    predicates: HashMap<DefId, Span>,
```
fpoli: Could you add a `///` comment for the `predicates` field and the `SpecChecker` type?

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,118 @@
...
+// deep visitor, check any calls
+impl<'tcx> Visitor<'tcx> for SpecChecker<'tcx> {
+    type Map = Map<'tcx>;
+
+    fn nested_visit_map(&mut self) -> intravisit::NestedVisitorMap<Self::Map> {
+        intravisit::NestedVisitorMap::All(self.tcx.hir())
+    }
+
+    fn visit_expr(&mut self, ex: &'tcx hir::Expr<'tcx>) {
+        if let hir::ExprKind::Call(ref callee_expr, _) = ex.kind {
```
dario23: not just calls, even just refering to it should be disallowed (i.e. get rid of the outermost if, i think)

Aurel300, line 64: As mentioned in the meeting, this outermost match should not happen to prevent function-pointer passing of predicate values in non-spec code.

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn visit_item(&mut self, item: &'tcx hir::Item<'tcx>) {
+        // collect DefIds for all predicate function items
+        if has_prusti_attr(item.attrs, "pred_spec_id_ref") {
+            let def_id = self.tcx.hir().local_def_id(item.hir_id).to_def_id();
+            self.predicates.insert(def_id, item.span);
+        }
+    }
+
+    fn visit_trait_item(&mut self, _: &'tcx hir::TraitItem<'tcx>) {
+        // nothing here
```
dario23: do we need some default action here to visit sub-items? should also add a test for it.

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    fn visit_impl_item(&mut self, _: &'tcx hir::ImplItem<'tcx>) {
+        // nothing here
```
Aurel300, line 47: As discussed, check if this should indeed remain empty – see if calls to predicates within `impl` blocks are caught.

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn visit_fn(
+        &mut self,
+        fk: intravisit::FnKind<'tcx>,
+        fd: &'tcx hir::FnDecl<'tcx>,
+        b: hir::BodyId,
+        s: Span,
+        id: hir::HirId,
+    ) {
+        // Stop checking inside `prusti::spec_only` functions
+        if has_prusti_attr(fk.attrs(), "spec_only") {
```
fpoli: We have `has_spec_only_attr(fk.attrs())`.

prusti-specs/src/lib.rs
dario23: removed it in 5e6a2469e

prusti-interface/src/specs/checker.rs
dario23: - [x] done in 34cd995b

prusti-interface/src/specs/checker.rs
dario23: 0bab6b18

prusti-interface/src/specs/checker.rs
dario23: seems the ItemLikeVisitor is too shallow here, only visiting toplevel items; rewriting it using `intravisit::Visitor`.

dario23:
tests failing with an ICE trying to typeck the path expression we're currently looking at (because the `qpath_res` method we want to use is a method of the `TypeckResults` struct).

<details>
  <summary>backtrace</summary>

```
> PRUSTI_LOG=debug RUST_BACKTRACE=full ./x.py run --bin prusti-rustc -- --edition=2018 -Pprint_desugared_specs=true -Pprint_typeckd_specs=true -Pno_verify=true -Phide_uuids=true prusti-tests/tests/verify/fail/unsupported/array.rs
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/prusti-rustc --edition=2018 -Pprint_desugared_specs=true -Pprint_typeckd_specs=true -Pno_verify=true -Phide_uuids=true prusti-tests/tests/verify/fail/unsupported/array.rs`
  __          __        __  ___             
 |__)  _\/_  |__) |  | /__`  |   ____\/_  | 
 |      /\   |  \ \__/ .__/  |       /\   | 

{commit 1777beda 2021-02-23 15:54:14 UTC, built on 2021-02-24 09:13:28 UTC}


// From the libc crate



//~ ERROR unsupported type

#![feature(register_tool)]
#![register_tool(prusti)]
#[prelude_import]
use std::prelude::v1::*;
#[macro_use]
extern crate std;
pub const FD_SETSIZE: usize = 1024;
pub struct fd_set {
    #[cfg(target_pointer_width = "64")]
    fds_bits: [i64; FD_SETSIZE / 64],
}
pub fn test(x: fd_set) -> fd_set { x }
fn main() { }
warning: type `fd_set` should have an upper camel case name
 --> prusti-tests/tests/verify/fail/unsupported/array.rs:5:12
  |
5 | pub struct fd_set {
  |            ^^^^^^ help: convert the identifier to upper camel case: `FdSet`
  |
  = note: `#[warn(non_camel_case_types)]` on by default

warning: field is never read: `fds_bits`
 --> prusti-tests/tests/verify/fail/unsupported/array.rs:7:5
  |
7 |     fds_bits: [i64; FD_SETSIZE / 64],
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

[2021-02-24T09:15:50Z DEBUG prusti_interface::specs::checker] ex: Expr { hir_id: HirId { owner: DefId(0:4 ~ array[317d]::fd_set), local_id: 5 }, kind: Path(Resolved(None, Path { span: prusti-tests/tests/verify/fail/unsupported/array.rs:7:21: 7:31 (#0), res: Def(Const, DefId(0:3 ~ array[317d]::FD_SETSIZE)), segments: [PathSegment { ident: FD_SETSIZE#0, hir_id: Some(HirId { owner: DefId(0:4 ~ array[317d]::fd_set), local_id: 4 }), res: Some(Err), args: None, infer_args: true }] })), attrs: ThinVec(None), span: prusti-tests/tests/verify/fail/unsupported/array.rs:7:21: 7:31 (#0) }
error: internal compiler error: compiler/rustc_typeck/src/check/mod.rs:489:9: can't type-check body of DefId(0:4 ~ array[317d]::fd_set)
  --> prusti-tests/tests/verify/fail/unsupported/array.rs:5:1
   |
5  | / pub struct fd_set {
6  | |     #[cfg(target_pointer_width = "64")]
7  | |     fds_bits: [i64; FD_SETSIZE / 64],
8  | |     #[cfg(target_pointer_width = "32")]
9  | |     fds_bits: [i32; FD_SETSIZE / 32],
10 | | }
   | |_^

thread 'rustc' panicked at 'Box<Any>', /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/panic.rs:59:5
stack backtrace:
   0:     0x7f716d61ed20 - std::backtrace_rs::backtrace::libunwind::trace::h9d49145f95eb5894
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x7f716d61ed20 - std::backtrace_rs::backtrace::trace_unsynchronized::hab1d020365bb6864
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f716d61ed20 - std::sys_common::backtrace::_print_fmt::h7659588431e304bd
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x7f716d61ed20 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h09f4a9e3befae3c7
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x7f716d68f46c - core::fmt::write::hf3fdfde304b9a088
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/core/src/fmt/mod.rs:1092:17
   5:     0x7f716d612062 - std::io::Write::write_fmt::h1cb850689c7116f0
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/io/mod.rs:1568:15
   6:     0x7f716d622bc5 - std::sys_common::backtrace::_print::hdbccd5aa093ba544
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/sys_common/backtrace.rs:49:5
   7:     0x7f716d622bc5 - std::sys_common::backtrace::print::hc639c4f320222558
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/sys_common/backtrace.rs:36:9
   8:     0x7f716d622bc5 - std::panicking::default_hook::{{closure}}::hdb012dd7a485bb5d
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/panicking.rs:208:50
   9:     0x7f716d622673 - std::panicking::default_hook::h75facbce77b6ba91
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/panicking.rs:225:9
  10:     0x55d5858b0b36 - <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call::h0f181562d313797a
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/alloc/src/boxed.rs:1560:9
  11:     0x55d5858ce616 - prusti_driver::report_prusti_ice::h51e7f059b3364d13
                               at /home/dario/git/prusti-dev/prusti/src/driver.rs:65:5
  12:     0x55d5858b5e55 - <prusti_driver::ICE_HOOK as core::ops::deref::Deref>::deref::__static_ref_initialize::{{closure}}::h7f205a28f6f20095
                               at /home/dario/git/prusti-dev/prusti/src/driver.rs:48:36
  13:     0x7f716d6234c6 - std::panicking::rust_panic_with_hook::hbcaa5de2cb5e22d5
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/panicking.rs:595:17
  14:     0x7f716e203c26 - std::panicking::begin_panic::{{closure}}::h7e0528aa4525bc6f
  15:     0x7f716e203b19 - std::sys_common::backtrace::__rust_end_short_backtrace::h47154b7e9e4aa5ba
  16:     0x7f716e262422 - std::panicking::begin_panic::h405805d4bcd0e662
  17:     0x7f716e1f6950 - std::panic::panic_any::hca623e320e919f00
  18:     0x7f716e1f2775 - rustc_errors::HandlerInner::span_bug::h011319b94387d4aa
  19:     0x7f716e1f2ab3 - rustc_errors::Handler::span_bug::h5b7840a80ce3da2c
  20:     0x7f716e25bc9e - rustc_middle::ty::context::tls::with_opt::hb3e88bae6683c83e
  21:     0x7f716fe41210 - rustc_middle::util::bug::opt_span_bug_fmt::h10354c7f1e208c55
  22:     0x7f716e25bd0c - rustc_middle::util::bug::span_bug_fmt::ha1c7534cbbc89ed7
  23:     0x7f716f3090f7 - rustc_typeck::check::typeck::ha72d4ca6aef4702c
  24:     0x7f716f48f942 - rustc_query_system::dep_graph::graph::DepGraph<K>::with_task_impl::h61b7b007454981b9
  25:     0x7f716f4aa1cf - rustc_data_structures::stack::ensure_sufficient_stack::h8cf967db44d72d2a
  26:     0x7f716f44bcf4 - rustc_query_system::query::plumbing::force_query_with_job::h0bdaa23f26d9faff
  27:     0x7f716f42df7c - rustc_query_system::query::plumbing::get_query_impl::h8114e6e1b12f1e7f
  28:     0x7f716f47e15d - <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::typeck::hf7dc0094d2d34ff5
  29:     0x55d586453c57 - rustc_middle::ty::query::TyCtxtAt::typeck::h6a0c0bacc200046f
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_middle/src/ty/query/mod.rs:205:17
  30:     0x55d586453c57 - rustc_middle::ty::query::<impl rustc_middle::ty::context::TyCtxt>::typeck::h37dc518bb47e47b9
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_middle/src/ty/query/mod.rs:186:17
  31:     0x55d586453c57 - <prusti_interface::specs::checker::CheckPredicatesVisitor as rustc_hir::intravisit::Visitor>::visit_expr::h0f00a3d0c99a2a65
                               at /home/dario/git/prusti-dev/prusti-interface/src/specs/checker.rs:82:23
  32:     0x55d5868691fb - rustc_hir::intravisit::walk_expr::habc94b6771ed6eac
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:1137:13
  33:     0x55d586453db1 - <prusti_interface::specs::checker::CheckPredicatesVisitor as rustc_hir::intravisit::Visitor>::visit_expr::h0f00a3d0c99a2a65
                               at /home/dario/git/prusti-dev/prusti-interface/src/specs/checker.rs:90:9
  34:     0x55d5868640e8 - rustc_hir::intravisit::walk_body::h77c4693b45ad08bb
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:506:5
  35:     0x55d58695bd04 - rustc_hir::intravisit::Visitor::visit_body::h82874186b4b36545
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:320:9
  36:     0x55d58695c666 - rustc_hir::intravisit::Visitor::visit_nested_body::h46762814b31041d3
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:306:9
  37:     0x55d586854bfb - rustc_hir::intravisit::walk_anon_const::h943879e6402aa07d
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:1100:5
  38:     0x55d58695c2e4 - rustc_hir::intravisit::Visitor::visit_anon_const::h12b97d8380c5d2df
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:366:9
  39:     0x55d58685e9cf - rustc_hir::intravisit::walk_ty::h1a4bca4d8cc45a1e
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:719:13
  40:     0x55d58695d184 - rustc_hir::intravisit::Visitor::visit_ty::hc989e12c34b4d979
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:372:9
  41:     0x55d586859be9 - rustc_hir::intravisit::walk_struct_field::hd9838c2949286bfb
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:1077:5
  42:     0x55d58695c9a4 - rustc_hir::intravisit::Visitor::visit_struct_field::h024a28392aefb4d4
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:427:9
  43:     0x55d586854f8f - rustc_hir::intravisit::walk_struct_def::h30f72c7908d3ffb5
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:1070:5
  44:     0x55d58695ca0a - rustc_hir::intravisit::Visitor::visit_variant_data::h4c63f1b7ec3fcb0d
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:424:9
  45:     0x55d58686f70e - rustc_hir::intravisit::walk_item::ha93ef05d5671ffb5
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:635:13
  46:     0x55d58695bd64 - rustc_hir::intravisit::Visitor::visit_item::h4c60b71834c23f62
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:316:9
  47:     0x55d58695c802 - rustc_hir::intravisit::Visitor::visit_nested_item::h7c0aeddfbd0a0a2b
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:273:9
  48:     0x55d5868604c7 - rustc_hir::intravisit::walk_mod::h2452cc35965b33a5
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:500:9
  49:     0x55d58695d240 - rustc_hir::intravisit::Visitor::visit_mod::hd29a952d582defbb
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:345:9
  50:     0x55d58684fa72 - rustc_hir::intravisit::walk_crate::hc36f32551ceded4f
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_hir/src/intravisit.rs:486:5
  51:     0x55d586453ffc - prusti_interface::specs::checker::SpecChecker::check_predicate_calls::he9fd2170f18e7998
                               at /home/dario/git/prusti-dev/prusti-interface/src/specs/checker.rs:130:9
  52:     0x55d5858c2a71 - <prusti_driver::callbacks::PrustiCompilerCalls as rustc_driver::Callbacks>::after_analysis::{{closure}}::h15848afa265bada3
                               at /home/dario/git/prusti-dev/prusti/src/callbacks.rs:45:13
  53:     0x55d5858cca45 - rustc_interface::passes::QueryContext::enter::{{closure}}::h93ba3f18ce26338a
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_interface/src/passes.rs:753:42
  54:     0x55d5858b01c5 - rustc_middle::ty::context::tls::enter_context::{{closure}}::ha84f3595d7e3ce37
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_middle/src/ty/context.rs:1716:50
  55:     0x55d5858b024d - rustc_middle::ty::context::tls::set_tlv::hd174756e3da4a398
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_middle/src/ty/context.rs:1700:9
  56:     0x55d5858b0196 - rustc_middle::ty::context::tls::enter_context::he6abcb51be8538c1
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_middle/src/ty/context.rs:1716:9
  57:     0x55d5858cc9ec - rustc_interface::passes::QueryContext::enter::h54abc40e43e7be47
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/compiler/rustc_interface/src/passes.rs:753:9
  58:     0x55d5858c1668 - <prusti_driver::callbacks::PrustiCompilerCalls as rustc_driver::Callbacks>::after_analysis::h12a68f8d9297ba9d
                               at /home/dario/git/prusti-dev/prusti/src/callbacks.rs:40:9
  59:     0x7f716fb1781b - rustc_interface::queries::<impl rustc_interface::interface::Compiler>::enter::h65ce91570046485e
  60:     0x7f716fb484e1 - rustc_span::with_source_map::hf064d889e7269127
  61:     0x7f716fb187ba - rustc_interface::interface::create_compiler_and_run::hd1225704c006d4a6
  62:     0x7f716fb169c8 - scoped_tls::ScopedKey<T>::set::h7e1dee359ee5bc4f
  63:     0x7f716fb1708c - std::sys_common::backtrace::__rust_begin_short_backtrace::hfa52e951047f9d06
  64:     0x7f716fb3606a - core::ops::function::FnOnce::call_once{{vtable.shim}}::h77d78bd93c39fd2c
  65:     0x7f716d6336ca - <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::hdc51fe7e73bc86bf
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/alloc/src/boxed.rs:1546:9
  66:     0x7f716d6336ca - <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::he605738a76b56d9d
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/alloc/src/boxed.rs:1546:9
  67:     0x7f716d6336ca - std::sys::unix::thread::Thread::new::thread_start::he44b12fd83e74919
                               at /rustc/fe1bf8e05c39bdcc73fc09e246b7209444e389bc/library/std/src/sys/unix/thread.rs:71:17
  68:     0x7f716bf1eea7 - start_thread
  69:     0x7f716bd04def - clone
  70:                0x0 - <unknown>

note: Prusti or the compiler unexpectedly panicked. This is a bug.

note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new

note: Prusti version: {commit 1777beda 2021-02-23 15:54:14 UTC, built on 2021-02-24 09:13:28 UTC}

query stack during panic:
#0 [typeck] type-checking `fd_set`
end of query stack
error: aborting due to previous error; 2 warnings emitted
```
</details>

i tried updating to the latest nightly rustc version, but the error still persists, will continue debugging. i uploaded the changes necessary for the rustc update as #421 if we want to merge them.

Aurel300:
I was trying to understand why `fd_set` in the failing example wouldn't typecheck – but it's a circular reference because you're actually trying to type the expression `FD_SETSIZE / 64`. I think this is fine to ignore, as any calls in const positions will be rejected by the compiler anyway . Can you only type expressions inside actual functions? It seems like now the visitor is going too deep.

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,160 @@
...
+};
+
+/// Checker visitor for the specifications. Currently checks that `#[predicate]`
+/// functions are never called from non-specification code, but more checks may follow.
+pub struct SpecChecker {
+    /// Map of the `DefID`s to the `Span`s of `#[predicate]` functions found in the first pass.
+    predicates: HashMap<DefId, Span>,
+
+    /// Span of call and definition of predicates illegally called, collected in the second pass.
+    pred_calls: Vec<(Span, Span)>,
```
fpoli: `pred_usages` would be a better name, because of 0bab6b182dbfdf47bec7292b9a83566a0f20e468. Similarly, the comment could say "usage" instead of "call".

dario23: updated in d4e946de.

i switched to many separate commits, mainly because i hope it makes reviewing easier; at least it helps me keep an overview. if you want, we can switch back though, e.g. for nicer git-history or such.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1282,7 +1282,12 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             let procedure = self.env.get_procedure(wrapper_def_id);
             let pure_function_encoder =
                 PureFunctionEncoder::new(self, proc_def_id, procedure.get_mir(), false);
-            let function = if self.is_trusted(proc_def_id) {
+            let function = if let Some(predicate_body) = self.get_predicate_body(proc_def_id) {
+                let predicate_body_encoded = self.encode_assertion(predicate_body, procedure.get_mir(), None, &[], None, false, None, ErrorCtxt::GenericExpression)
+                    .run_if_err(cleanup)?;
+                pure_function_encoder.encode_predicate_function(predicate_body_encoded)
+                    .run_if_err(cleanup)?
+            } else if self.is_trusted(proc_def_id) {
```
fpoli, line 1289: This makes me wonder, what happens if the user specifies both `#[predicate]` and `#[trusted]`? Would it be easy to generate an error message e.g. in the `SpecChecker`?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1282,7 +1282,12 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             let procedure = self.env.get_procedure(wrapper_def_id);
             let pure_function_encoder =
                 PureFunctionEncoder::new(self, proc_def_id, procedure.get_mir(), false);
-            let function = if self.is_trusted(proc_def_id) {
+            let function = if let Some(predicate_body) = self.get_predicate_body(proc_def_id) {
+                let predicate_body_encoded = self.encode_assertion(predicate_body, procedure.get_mir(), None, &[], None, false, None, ErrorCtxt::GenericExpression)
+                    .run_if_err(cleanup)?;
```
fpoli: *  I think that `targets_are_values` should be `true` and that you should provide `target_args` ([here](https://github.com/viperproject/prusti-dev/blob/e207038b94cb8e5499b309de7ec52765b6e2404b/prusti-viper/src/encoder/pure_function_encoder.rs#L289-L293) is an example of how to build them). See [here](https://github.com/viperproject/prusti-dev/blob/e207038b94cb8e5499b309de7ec52765b6e2404b/prusti-viper/src/encoder/spec_encoder.rs#L32-L61) for a description of the arguments.
* I would prefer to move ~~the `get_predicate_body` and~~ the construction of `predicate_body_encoded` into the `encode_predicate_function` function, to keep `encoder.rs` clean.
* This line is very long; consider placing one argument per line.

prusti-interface/src/utils.rs
```diff
@@ -270,7 +270,7 @@ pub fn has_prusti_attr(attrs: &[ast::Attribute], name: &str) -> bool {
     attrs.iter().any(|attr| match &attr.kind {
         ast::AttrKind::Normal(ast::AttrItem {
                                   path: ast::Path { span: _, segments, tokens: _ },
-                                  args: ast::MacArgs::Empty,
+                                  args: _, // ast::MacArgs::Empty,
```
dario23: remove from PR

dario23: rebased onto latest master and undid this change.

dario23: un-removing this; the issue is that `has_prusti_attr` only checked for things like `prusti::spec_only` which don't have values, but i need it to also match `prusti::pred_spec_id_ref = $(UUID)`, so i think this should be `args: _` after all. i'll remove the commented-out previous code of course.

fpoli: If you use `args: _` remember to remove this comment

https://github.com/viperproject/prusti-dev/blob/50d1941a4e6ab78ebab853b48987ccd9b629f880/prusti-interface/src/utils.rs#L268

prusti-interface/src/specs/checker.rs
```diff
@@ -0,0 +1,160 @@
...
+            let def_id = self.tcx.hir().local_def_id(id).to_def_id();
+            self.predicates.insert(def_id, s);
+        }
+
+        intravisit::walk_fn(self, fk, fd, b, s, id);
+    }
+}
+
+// Second predicate checks visitor: check any references to predicate functions
+// from non-specification code
```
Aurel300:
```suggestion
/// Second predicate checks visitor: check any references to predicate functions
/// from non-specification code
```

prusti-specs/src/lib.rs
```diff
@@ -60,6 +60,17 @@ pub fn rewrite_prusti_attributes(
...
     prusti_attributes.extend(extract_prusti_attributes(&mut item));
 
+    // make sure to also update the check in the #[predicate] handling method
+    if prusti_attributes
+        .iter()
+        .any(|(ak, _)| ak == &SpecAttributeKind::Predicate)
+    {
+        return syn::Error::new(
+            item.span(),
+            "`#[predicate]` is incompatible with other prusti attributes",
```
```suggestion
            "`#[predicate]` is incompatible with other Prusti attributes",
```

(Unsure if we are consistent about this.)

prusti-specs/src/lib.rs
```diff
@@ -435,3 +450,75 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+            "the `#[predicate]` attribute does not take parameters"
+        ).to_compile_error();
+    }
+
+    // emit a custom error to the user instead of a parse error
+    let mut item: untyped::AnyFnItem = handle_result!(
+        syn::parse2(tokens)
+            .map_err(|e| syn::Error::new(
+                e.span(),
+                "you can only use `#[predicate]` on function definitions"
```
```suggestion
                "`#[predicate]` can only be used on function definitions"
```

prusti-specs/src/lib.rs
```diff
@@ -435,3 +450,75 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+            ))
+    );
+    let item_span = item.span();
+
+    // check for other attributes -- #[predicate] is incompatible with all others currently
+    // make sure to also update the check in `rewrite_prusti_attributes`
+    let other_prusti_attrs = extract_prusti_attributes(&mut item);
+    if other_prusti_attrs.count() != 0 {
```
Aurel300, line 475:
```suggestion
    if !other_prusti_attrs.is_empty() {
```

prusti-specs/src/lib.rs
```diff
@@ -435,3 +450,75 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+        return syn::Error::new(
+            item.span(),
+            "`#[predicate]` is incompatible with other prusti attributes",
```
```suggestion
            "`#[predicate]` is incompatible with other Prusti attributes",
```

Aurel300:
Super nit-picky changes, otherwise seems ready to merge.

prusti-specs/src/lib.rs
dario23: `extract_prusti_attributes` returns an `impl Iter`, not a `Vec`, which doesn't seem to have an `is_empty` method.

Aurel300:
Thanks. This should also be documented (see the `gh-pages-src` branch).

PR #391, opened 2021-02-05
Check for extra parameters to #[pure], #[trusted] attributes

dario23:
Constructs like `#[pure(123)]` are not allowed, so explicitly refuse
them with a relevant error message.

### DIFF
--- a/prusti-specs/src/lib.rs
+++ b/prusti-specs/src/lib.rs
@@ -211,7 +211,14 @@ fn generate_for_after_expiry_if(attr: TokenStream, item: &untyped::AnyFnItem) ->
 }
 
 /// Generate spec items and attributes to typecheck and later retrieve "pure" annotations.
-fn generate_for_pure(_attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedResult {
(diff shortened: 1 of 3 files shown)


prusti-specs/src/lib.rs
```diff
@@ -211,7 +211,14 @@ fn generate_for_after_expiry_if(attr: TokenStream, item: &untyped::AnyFnItem) ->
 }
 
 /// Generate spec items and attributes to typecheck and later retrieve "pure" annotations.
-fn generate_for_pure(_attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedResult {
+fn generate_for_pure(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedResult {
+    if !attr.is_empty() {
+        return Err(syn::Error::new(
+            attr.span(),
+            "The `#[pure]` attribute does not take parameters"
```
fpoli: Really minor comment: Rust's error messages usually start with a lowercase letter.

fpoli:
Thanks!

prusti-specs/src/lib.rs
dario23: adjusted; also fixed a copy/paste-mistake in the error message.

Issue #392, opened 2021-02-05
Security of dependencies


vakaras:
`./x.py` should download the dependencies via HTTPS and check their fingerprints.

This issue was originally reported on [Zulip](https://prusti.zulipchat.com/#narrow/stream/236614-general/topic/Security.20of.20dependencies).

rw:
I'd also suggest that `./x.py` downloads a release from github, instead of a self-hosted binary.

PR #394, opened 2021-02-06
Update x.py to fetch HTTPS URLs instead of HTTP

rw:
Relates to, but does not solve, https://github.com/viperproject/prusti-dev/issues/392


Aurel300:
Looks good, thanks.

PR #398, opened 2021-02-08
Clean up files in the top folder

fpoli:


CONTRIBUTING.md
```diff
@@ -1,5 +1,7 @@
 # Contributing
```
Aurel300, line 1: I think I would prefer to either just delete this file or keep only the link to the development guide. It's easy to forget to update this.

fpoli: Sure

PR #402, opened 2021-02-12
Cleanup snapshot encoding

Aurel300:
Changes in this PR:

- A new `SnapApp` variant in the VIR. These are patched (in fewer places than before) into the appropriate function applications or similar, e.g. `.val_int` field access or a domain function application.
- The snapshot encoding of types itself is different; snapshots come in a number of varieties (see `enum Snapshot`). ADTs are `Complex` (should probably be renamed), field-less types are `Unit`, etc. Recursive types, boxes, and references should be handled more solidly.
- `MemoryEqEncoder` was replaced with snapshot equality.
- Pure functions take snapshots as arguments instead of `Ref`s.

---

These tests have been disabled (moved to `tests_old`) due to unsupported recursion in pure functions:

- `verify/fail/demos/append-sorted-error-1.rs`
- `verify/fail/demos/append-sorted-error-2.rs`
- `verify/fail/demos/append-sorted-error-4.rs`
- `verify/pass/demos/append-sorted.rs`
- `verify/pass/gitlab-issues/issue-187-1.rs`
- `verify/pass/paper-examples/force_increasing.rs`

To fix these, we will need manual axiomatisation + unfold witnesses, which itself requires multi-step verification to avoid unsoundness. This is outside the scope of this PR.

---

Future work:

- [x] reenable purification of methods
- [ ] implement axiom-based mirrors


vakaras:
> These tests have been disabled (moved to `tests_old`) due to unsupported recursion in pure functions:

Do we have an issue for this?

prusti-common/src/vir/ast/function.rs
```diff
@@ -62,25 +62,36 @@ impl Function {
...
         match typ {
-            Type::Int => "$int$",
-            Type::Bool => "$bool$",
-            Type::TypedRef(ref name) => name,
-            Type::Domain(ref name) => name,
+            Type::Int => "$int$".to_string(),
+            Type::Bool => "$bool$".to_string(),
+            Type::TypedRef(ref name) => name.to_string(),
+            Type::Domain(ref name) => name.to_string(),
+            Type::Snapshot(ref name) => format!("Snap${}", name),
```
vakaras, line 87: I am wondering whether it would not be better to store the snapshot type name as `Snap$<name>` so that we can return a reference here. In case we ever need to extract the `name`, we could subslice, which is a much cheaper operation than creating a new string.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -133,7 +136,7 @@ impl BuiltinEncoder {
             },
         }
     }
-
+    /*
     pub fn encode_builtin_domain(&self, kind: BuiltinDomainKind) -> vir::Domain {
```
vakaras: Why is this commented out?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -130,7 +127,7 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         );
 
         let axiomatized_functions_domain = vir::Domain {
-            name: snapshot::AXIOMATIZED_FUNCTION_DOMAIN_NAME.to_owned(),
+            name: "$MirrorFunctions$".to_string(), // snapshot::AXIOMATIZED_FUNCTION_DOMAIN_NAME.to_owned(),
```
vakaras: Why is this not a constant anymore?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -253,45 +250,27 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
-                axioms: vec![],
-                type_vars: vec![],
-            });
-        }
+        let mut domains = vec![];
+        domains.extend(self.snapshot_encoder.borrow().get_viper_domains());
+        domains.extend(self.mirror_encoder.borrow().get_viper_domains());
 
+        /*
         if config::enable_purification_optimization() {
```
vakaras: Is the purification optimization completely disabled?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -794,9 +753,12 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
                 },
             };
             self.log_vir_program_before_viper(method.to_string());
+            // TODO: ???
+            /*
```
vakaras: This is the call that performs the actual optimization. Why is it commented out?

prusti-viper/src/encoder/mirror_function_encoder.rs
```diff
@@ -4,14 +4,115 @@
...
+        _function: &mut vir::Function,
+    ) {}
+}
+
+// ------------------------------
+// --- OLD ----------------------
+// ------------------------------
+
+/*
+pub fn encode_mirror_of_pure_function(
```
vakaras, line 116: I am confused, why is this commented out?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -449,6 +448,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
             self.cfg_method
                 .add_local_var(&var_name, vir::Type::TypedRef(type_name));
         }
+
+        // TODO: ??? this probably only ensures snapshots are made...
```
vakaras: Yes.

vakaras (changes requested):
Why this PR completely removes the purification optimization?

prusti-common/src/config.rs
```diff
@@ -380,14 +381,18 @@ pub fn optimizations() -> Optimizations {
...
+/// Enable purification optimization for impure functions.
 pub fn enable_purification_optimization() -> bool {
     read_setting("enable_purification_optimization")
 }
 
+/// Enable manual axiomatization of pure functions.
+///
+/// **Note:** this is currently very incomplete and may introduce unsoudnesses.
+pub fn enable_manual_axiomatization() -> bool {
+    read_setting("enable_manual_axiomatization")
```
vakaras, line 393: Is this configuration flag ever used? If not, I would suggest panicking with `unimplemented!` when it is set to `true`. Or even better would be to reenable the manual axiomatization.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -142,7 +147,8 @@ impl BuiltinEncoder {
     }
 
     fn encode_nat_builtin_domain(&self) -> vir::Domain {
-        let nat_domain_name = snapshot::NAT_DOMAIN_NAME;
+        let nat_domain_name = "NatDomain";
```
vakaras, line 150: Is this the only place where this constant is used?

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -163,12 +169,45 @@ impl BuiltinEncoder {
...
-        let domain_name = snapshot::PRIMITIVE_VALID_DOMAIN_NAME;
+        let domain_name = PRIMITIVE_VALID_DOMAIN_NAME;
 
         let mut functions = vec![];
         let mut axioms = vec![];
+
         for t in &[vir::Type::Bool, vir::Type::Int] {
-            let f = snapshot::valid_func_for_type(t);
+            //let f = snapshot::valid_func_for_type(t);
+            let f = {
```
vakaras, line 179: Is this basically an inlined version of `valid_func_for_type`? Could you move it back to a separate function because `encode_primitive_builtin_domain` is already huge?

vakaras (commented):
Looks good. My only big concern is that manual axiomatization is disabled.

prusti-common/src/config.rs
Aurel300: I think it currently causes an encoding of the `Nat` domain only.

vakaras: True. I would add `unimplemented!` [here](https://github.com/viperproject/prusti-dev/blob/afa89f858748fced714b72049e79c2f8442fd715/prusti-viper/src/encoder/encoder.rs#L258) because otherwise it can be very confusing why this flag does nothing.

Issue #410, opened 2021-02-16
internal error: entered unreachable code at prusti-viper/src/encoder/procedure_encoder.rs:2057:37


zgrannan:
For the following code: 

```rust
pub fn max_by_key<A, B: Ord>(a: A, b: A, key: impl Fn(&A) -> B) -> A {
    if key(&a) > key(&b) {
        a
    } else {
        b
    }
}

fn main() {}
```
The output is:

```
~/prusti-reproduce env RUST_BACKTRACE=1 prusti-rustc main.rs
  __          __        __  ___             
 |__)  _\/_  |__) |  | /__`  |   ____\/_  | 
 |      /\   |  \ \__/ .__/  |       /\   | 

{commit b51ec334 2021-02-15 19:43:59 UTC, built on 2021-02-16 06:44:48 UTC}


Verification of 2 items...
thread 'rustc' panicked at 'internal error: entered unreachable code', prusti-viper/src/encoder/procedure_encoder.rs:2057:37
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5fa22fe6f821ac3801d05f624b123dda25fde32c/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/5fa22fe6f821ac3801d05f624b123dda25fde32c/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/5fa22fe6f821ac3801d05f624b123dda25fde32c/library/core/src/panicking.rs:50:5
   3: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode_terminator
   4: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode_statement_at
   5: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode_blocks_group
   6: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode
   7: prusti_viper::encoder::encoder::Encoder::encode_procedure
   8: prusti_viper::encoder::encoder::Encoder::process_encoding_queue
   9: prusti_viper::verifier::Verifier::verify
  10: prusti_driver::verifier::verify
  11: rustc_interface::passes::QueryContext::enter
  12: <prusti_driver::callbacks::PrustiCompilerCalls as rustc_driver::Callbacks>::after_analysis
  13: rustc_interface::queries::<impl rustc_interface::interface::Compiler>::enter
  14: rustc_span::with_source_map
[...]

PR #411, opened 2021-02-16
More descriptive error messages

zgrannan:
When verifying things with Prusti I would sometimes hit errors and have no indication what code was responsible (for example, https://github.com/viperproject/prusti-dev/issues/410). This PR makes the errors somewhat more descriptive by adding the location of the problematic code Prusti encountered. 

Feel free to close the PR if this is better addressed in another way.

### DIFF
--- a/prusti-viper/src/encoder/procedure_encoder.rs
+++ b/prusti-viper/src/encoder/procedure_encoder.rs
@@ -2053,7 +2053,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                                     )?);
                                 }
 
-                                _ => unreachable!()
+                                _ => {
+                                    unimplemented!(
+                                        "Only calls to closures are supported. The term at {:?} is a {:?}, not a closure.",
+                                        term.source_info.span, cl_type.kind()
+                                    )
+                                }
                             }
                         }
 
--- a/prusti-viper/src/encoder/procedure_encoder.rs
+++ b/prusti-viper/src/encoder/procedure_encoder.rs
@@ -2055,7 +2055,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
 
                                 _ => {
                                     cleanup(&self);
-                                    return Err(SpannedEncodingError::internal(
+                                    return Err(SpannedEncodingError::unsupported(
                                         format!("only calls to closures are supported. The term is a {:?}, not a closure.", cl_type.kind()),
                                         term.source_info.span,
                                     ));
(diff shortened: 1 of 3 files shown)


prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2053,7 +2053,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                                     )?);
                                 }
 
-                                _ => unreachable!()
+                                _ => {
+                                    unimplemented!(
+                                        "Only calls to closures are supported. The term at {:?} is a {:?}, not a closure.",
+                                        term.source_info.span, cl_type.kind()
+                                    )
+                                }
```
fpoli: Since you are at it, the even better fix would be to return an internal encoding error. Something like:
```
cleanup(&self);
return Err(SpannedEncodingError::internal(
    format!("only calls to closures are supported. The term is a {:?}, not a closure.", cl_type.kind()),
    term.source_info.span,
));
```

The cleanup is sadly needed because of a hack (see https://github.com/viperproject/prusti-dev/issues/187).

The error message starts with a lowercase character just because it's the convention that `rustc` seems to use.

prusti-viper/src/utils/type_visitor.rs
```diff
@@ -268,7 +268,10 @@ pub fn walk_closure<'tcx, E, V: TypeVisitor<'tcx, Error = E>>(
 ) -> Result<(), E> {
     let cl_substs = substs.as_closure();
     // TODO: when are there bound typevars? can type visitor deal with generics?
-    let fn_sig = cl_substs.sig().no_bound_vars().unwrap();
+    let fn_sig = match cl_substs.sig().no_bound_vars() {
+        Some(t) => t,
+        None => unimplemented!("Bound variables not supported at {:?}", def_id)
+    };
```
fpoli: To avoid the `Some(t) => t`, you could use `no_bound_vars().unwrap_or_else(|| unimplemented!("<message>", ...))` or even `no_bound_vars().expect(&format!("<message>", ...))`.

fpoli:
Could you also add a test in `prusti-tests/tests/verify/fail/unsupported` for the internal error message?

fpoli:
Fixes #410, as it avoids the crash and reports a readable error message.

I realized that the error should be of the kind "unsupported". I changed that.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2053,7 +2053,13 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                                     )?);
                                 }
 
-                                _ => unreachable!()
+                                _ => {
+                                    cleanup(&self);
+                                    return Err(SpannedEncodingError::unsupported(
+                                        format!("only calls to closures are supported. The term is a {:?}, not a closure.", cl_type.kind()),
```
Aurel300, line 2059: Sorry I just realised this now, but this error message is a bit misleading – it sounds as if we didn't support the most basic case of calls to regular functions.

fpoli: True, I didn't think about that interpretation.

PR #423, opened 2021-02-27
Simplify lazy_static invocations

matklad:
just a drive by change, feel free to ignore!


vakaras:
Thank you very much!

matklad:
That's not const-eval, it's just that you can put many things into a single lazy static invocation.  The diff is confusing )

vakaras:
Oh, right. I saw that you deleted many lazy static invocations, but did not notice that you left the first one ☺. Anyway, thanks!

PR #436, opened 2021-03-08
Enable more purification tests in CI.

vakaras:


prusti-tests/tests/verify_overflow/pass/rosetta/Binary_search_shared.rs
```diff
@@ -0,0 +1,155 @@
+//! A version of `Binary_search.rs` that uses shared references.
+
+#![allow(dead_code)]
+use prusti_contracts::*;
```
vakaras, line 4: @tillarnold You may want to add this file to your benchmarks.

PR #438, opened 2021-03-08
Opt-in to rustc_private for `rust-analyzer`

### LINKED ISSUE #437: Opt-in to rustc_private for `rust-analyzer`
i saw various projects adding opt-in commits to signal to rust-analyzer that they're using rustc_private now that https://github.com/rust-analyzer/rust-analyzer/pull/7891 is merged; maybe we want to do that as well?

dario23:
This is suggested to make the dependency on rustc internals explicit in
the crate manifest; see
https://github.com/rust-analyzer/rust-analyzer/pull/7891 for some more
details.

This change adds this hint to all crates that use
`#![feature(rustc_private)]`.

Fixes #437

### DIFF
--- a/analysis/Cargo.toml
+++ b/analysis/Cargo.toml
@@ -11,3 +11,7 @@ serde_json = "1.0"
 
 [dev-dependencies]
 compiletest_rs = "0.5.0"
+
+[package.metadata.rust-analyzer]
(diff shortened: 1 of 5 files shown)


vakaras:
Thank you!

PR #439, opened 2021-03-09
Purify more variables.

vakaras:


vakaras:
@tillarnold Could you please rerun benchmarks with the latest master?

PR #443, opened 2021-03-22
Counterexamples

cedihegi:
**counterexamples for prusti**
If you want to use this branch, you need to compile and assemble my Silicon branch [https://github.com/cedihegi/silicon/tree/counterexample-interface](url) and replace silicon.jar in viper_tools.


prusti-interface/src/environment/mod.rs
```diff
@@ -115,7 +115,8 @@ impl<'tcx> Environment<'tcx> {
         sp: S,
         msg: &str,
         help: &Option<String>,
-        note: &Option<(String, S)>
+        note: &Option<(String, S)>,
+        additional_notes: &Vec<(String, Option<S>)>,
```
Aurel300: Could we keep just the vector and call it `notes`? The compiler receives the same thing for both of these anyway.

prusti-interface/src/prusti_error.rs
```diff
@@ -25,6 +25,7 @@ pub struct PrustiError {
     span: MultiSpan,
     help: Option<String>,
     note: Option<(String, MultiSpan)>,
+    additional_notes: Vec<(String, Option<MultiSpan>)>,
```
Aurel300: Same comment.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
+use std::collections::HashMap;
```
Aurel300: In the context of this PR it is clear what this is, but for posteriority let's make it `counterexample_translation.rs` and likewise `CounterexampleTranslator`.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+use crate::encoder::Encoder;
+use crate::encoder::places::{Local, LocalVariableManager, Place};
+use crate::encoder::counterexample::*;
+
+use rustc_middle::mir;
+use rustc_middle::ty::{self, Ty, AdtKind, AdtDef, TyCtxt};
+use rustc_span::MultiSpan;
+
+
+pub struct Backtranslator<'tcx>{
```
Aurel300:
```suggestion
pub struct Backtranslator<'tcx> {
```
(and similar in other places…)

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+    tcx: TyCtxt<'tcx>,
+    is_pure: bool,
+    disc_info: HashMap<(ProcedureDefId, String), Vec<String>>,
+}
+
+impl<'tcx> Backtranslator<'tcx> {
+    pub fn new(
+        encoder: &Encoder<'_, 'tcx>,
+        def_id: ProcedureDefId,
+        silicon_opt_counterexample: Option<SiliconCounterexample>
```
Aurel300: Maybe this shouldn't even be called when there is no counterexample, i.e. move the `let Some(...)` out?

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+
+pub fn backtranslate<'tcx>(
+    encoder: &Encoder,
+    def_id: ProcedureDefId,
+    silicon_counterexample: Option<SiliconCounterexample>,
+) -> Counterexample {
+    let opt_backtranslator = Backtranslator::new(encoder, def_id, silicon_counterexample);
+    if let Some(backtranslator) = opt_backtranslator {
```
Aurel300:
```suggestion
    if let Some(silicon_counterexample) = opt_silicon_counterexample {
      let backtranslator = Backtranslator::new(encoder, def_id, silicon_counterexample);
```

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                args_to_process.push((rust_name, multi_span, vir_name, typ))
+            }
+        }
+
+        //for the return type:
+        let return_local = Local::from(mir::Local::from_usize(0));
+        //make sure
+        let result_to_process = if local_variable_manager.is_return(return_local){
+            //open question: what would be the right span for return type? None?
+            let rust_name = String::from("result");
```
Aurel300:
```suggestion
            let rust_name = "result".to_string();
```
(and likewise in other places; do you really need a `String` also?)

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+        //most values can not be obtained there because they're folded
+        let last_label: Option<&String> = backtranslator
+            .silicon_counterexample
+            .label_order.last();
+        //optimally this label would just be "old", but as of now mostly the values
+        //are not available at this point
+        let old_label = String::from("l0");
+
+
+
```
Aurel300: Let's not have these big gaps. If it makes sense you can split the function into smaller ones.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                Some((rust_name, vir_name, typ)) => {
+                    let opt_entry = silicon_final_entries.get(&vir_name);
+                    backtranslator.backtranslate_entry(
+                        typ, 
+                        opt_entry, 
+                        vir_name,
+                        silicon_final_entries,
+                    )
+                }
+            }
```
Aurel300: This part is almost identical between the pure and impure cases, it can happen after.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                if let Some(value) = opt_value {
+                    Entry::IntEntry { value }
+                } else {
+                    Entry::UnknownEntry
+                }
+            },
+            ty::TyKind::Char => {
+                let opt_value = self.backtranslate_int(sil_entry);
+                if let Some(value) = opt_value {
+                    let val_t = u8::try_from(value);
```
Aurel300: `Char` is not an 8-bit value; its range is `0..0x10FFFF`, I think.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                if let Some(ModelEntry::RefEntry(name, map)) = sil_entry {
+                    let entry = map.get(&String::from("val_ref"));
+                    let new_vir_name = format!("{}.val_ref", vir_name);
+                    let rec_entry = self.backtranslate_entry(
+                        typ, 
+                        entry, 
+                        new_vir_name,
+                        silicon_ce_entries, 
+                    );
+                    Entry::RefEntry {el: Box::new(rec_entry)}
```
Aurel300:
```suggestion
                    Entry::RefEntry {el: box ec_entry}
```
(and in other places)

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                    let val_t = u8::try_from(value);
+                    if let Ok(v) = val_t {
+                        return Entry::CharEntry{ value: v as char}
+                    } 
+                }
+                Entry::UnknownEntry
+            },
+            ty::TyKind::Ref(_, typ, _) => {
+                if let Some(ModelEntry::RefEntry(name, map)) = sil_entry {
+                    let entry = map.get(&String::from("val_ref"));
```
Aurel300:
```suggestion
                    let entry = map.get("val_ref");
```

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1680,6 +1687,29 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         };
         result
     }
+
+    pub fn add_discriminant_info(&self, enum_id: String, discr_id: String, proc_def_id: ProcedureDefId) {
+        let mut map = self.discriminants_info.borrow_mut();
+        let mut previous_entries = map.get_mut(&(proc_def_id.clone(), enum_id.clone()));
```
Aurel300: No clone needed?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1680,6 +1687,29 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+        match previous_entries {
+            None => {map.insert((proc_def_id, enum_id), vec![discr_id]);},
+            Some(v) => {
+                v.push(discr_id);
+            }
```
```suggestion
            None => map.insert((proc_def_id, enum_id), vec![discr_id]),
            Some(v) => v.push(discr_id),
```

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1680,6 +1687,29 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+        self.discriminants_info.borrow().clone()
+    }
+/*
+    pub fn get_counterexample(
+        &self, 
+        def_id: ProcedureDefId,
+        silicon_counterexample: Option<SiliconCounterexample>
+    ) -> Counterexample {
+        backtranslate(&self, def_id, silicon_counterexample);
+    }*/
```
Aurel300: What is this?

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -155,6 +160,17 @@ impl<'tcx> ErrorManager<'tcx>
...
 
+    pub fn get_def_id(&self, ver_error: &VerificationError) -> Option<&ProcedureDefId> {
+        let opt_pos_id: Option<u64> = match ver_error.pos_id {
+            Some(ref viper_pos_id) => {
+                let parsed = viper_pos_id.parse();
+                parsed.ok()
+            },
+            None => None
+        };
+        opt_pos_id.and_then(|opt_pos_id| self.def_id.get(&opt_pos_id))
```
Aurel300:
```suggestion
        ver_error.pos_id
            .and_then(|id| id.parse().ok())
            .and_then(|id| self.def_id.get(&id)) 
```

prusti-viper/src/encoder/mir_encoder.rs
```diff
@@ -646,10 +646,11 @@ impl<'p, 'v: 'p, 'tcx: 'v> MirEncoder<'p, 'v, 'tcx> {
                         String::from("number"),
                         self.encode_operand_expr_type(operand).with_span(span)?,
                     )];
+                    println!("mir_encoder: {:?}", self.def_id);
```
Aurel300:
```suggestion
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5271,6 +5292,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
             .min()
             .unwrap()
     }
+
+
```
```suggestion
```

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -407,6 +414,8 @@ pub(super) struct PureFunctionBackwardInterpreter<'p, 'v: 'p, 'tcx: 'v> {
     /// when to a undefined function calls. This distinction allows overflow checks to be checked
     /// on the caller side and assumed on the definition side.
     is_encoding_assertion: bool,
+    /// Needed for registering errors
```
Aurel300: (Not a very helpful comment.)

viper/src/silicon_counterexample.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+fn unwrap_model_entry<'a>(
+    env: &'a JNIEnv<'a>,
+    jni: JniUtils<'a>,
+    entry: JObject<'a>,
+    entries: &mut HashMap<String, ModelEntry>
+) -> Option<ModelEntry> {
+    // either check is_instance_of for every possible class or 
+    // get classname and match strings. First one feels saver, maybe more expensive
```
Aurel300: I disagree, you still need a `_` case in such a `match`.

prusti-viper/src/encoder/counterexample.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+impl fmt::Display for Entry{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self{
+            Entry::IntEntry { value } => write!(f, "{}", value),
+            Entry::BoolEntry { value } => write!(f, "{}", value),
+            Entry::CharEntry { value } => write!(f, "'{}' ({:x})", value, *value as i32),
+            Entry::RefEntry { el } => write!(f, "ref({})", indent(el.to_string())),
+            Entry::Enum { super_name, name, named_fields, field_names, field_entries } => {
```
Aurel300: You can probably use [`f.debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) (and similar) for enums, structs, and tuples, then get rid of `indent`.

prusti-viper/src/encoder/counterexample.rs
```diff
@@ -0,0 +1,184 @@
...
+        name: String,
+        field_names: Vec<String>,
+        field_entries: Vec<Entry>
+    },
+    Enum{
+        super_name: String,
+        name: String,
+        named_fields: bool,
+        field_names: Vec<String>,
+        field_entries: Vec<Entry>
```
Aurel300: A single `Vec<(String, Entry)>` should do?

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                                for def in &adt_def.variants{
+                                    let discr = match def.discr {
+                                        ty::VariantDiscr::Relative(d) => d,
+                                        _ => unimplemented!()  
+                                    };
+                                    if discr == discriminant {
+                                        variant = Some(def);
+                                        variant_name = def.ident.name.to_ident_string();
+                                    }
+                                }
```
Aurel300: Yes, this is not good. Is there no API to lookup variant given the discriminant?

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                                    vir_name,
+                                    subst,
+                                    silicon_ce_entries,
+                                );
+                                field_names = result.0;
+                                field_entries = result.1;
+                            }
+                        }
+                        //valid names can not start with integers (?)
+                        let named_fields = field_names.len() > 0 && !field_names[0].parse::<usize>().is_ok();
```
Aurel300: This is not very nice… If we check based on the string value, we can do it when actually printing the `Enum` later, it does not need to be a field.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+                        result_vir_name,
+                        silicon_entries,
+                    )
+                }
+            }
+        };
+        
+        Counterexample::Success {result, args, entries, is_pure: backtranslator.is_pure}
+    } else {
+        Counterexample::Failure(String::from("no counterexample generated"))
```
Aurel300: Is this the only "kind" of failure? Can the result not be an `Option`?

prusti-viper/src/verifier.rs
```diff
@@ -319,8 +320,19 @@ impl<'v, 'tcx> Verifier<'v, 'tcx> {
...
+                
+                //counterexamples:
+                if let Some(id) = error_manager.get_def_id(&verification_error) {
+                    let counterexample = backtranslation::backtranslate(
+                        &self.encoder, 
+                        *id,
+                        verification_error.counterexample
+                    );
+                    counterexample.apply_prusti_error(&mut prusti_error); 
+                }
```
Aurel300: I would have expected the `config` to be checked here?

prusti-interface/src/prusti_error.rs
```diff
@@ -36,6 +37,7 @@ impl PrustiError {
             span,
             help: None,
             note: None,
+            additional_notes: vec![]
```
Aurel300: Please be consistent about the trailing commas, it's easier to edit code if there is always a trailing comma in a multiline call or declaration.

prusti-viper/src/encoder/backtranslation.rs
```diff
@@ -0,0 +1,457 @@
...
+            let vir_name = if !backtranslator.is_pure {
+                local_variable_manager.get_name(return_local)
+            } else {
+                String::from("Result()")
+            };
+            let typ = local_variable_manager.get_type(return_local);
+            Some((rust_name, vir_name, typ))
+        } else {
+            //this case should probably never occur
+            None
```
Aurel300: This is a bit strange indeed, maybe you could just `unreachable!` this.

prusti-viper/src/encoder/counterexample.rs
```diff
@@ -0,0 +1,184 @@
...
+        result: Entry,
+        args: HashMap<(String, MultiSpan), Entry>,
+        entries: HashMap<(String, MultiSpan), Entry>,
+        is_pure: bool,
+    },
+    Failure(String),
+}
+
+impl Counterexample{
+    pub fn apply_prusti_error(&self, prusti_error: &mut PrustiError) {
```
Aurel300:
```suggestion
    pub fn annotate_error(&self, prusti_error: &mut PrustiError) {
```

prusti-viper/src/encoder/encoder.rs
cedihegi: The second pair of curly brackets are totally unnecessary, I agree, but `map.insert()` returns an Option, so there your suggestion does not work because "match arms have incompatible types".

cedihegi:
@Aurel300 thanks for the extensive feedback. First of all, sorry for the obvious stuff, where I left debug stuff or missed spaces and newlines in places that I didn't even modify and other things. I went through all my changes before opening this PR but I still missed a lot of those.
I went through all of your suggestions, and I think I resolved most of them, but I will add comments in the places where I think there is something to discuss.

prusti-viper/src/encoder/backtranslation.rs
cedihegi: I refactored this function quite a bit, I realized it was a huge mess and your comment to split it up into multiple definitely makes sense. This was also resolved in the process.

prusti-viper/src/encoder/encoder.rs
cedihegi: I think they are needed? It doesn't compile without them.

prusti-viper/src/encoder/encoder.rs
cedihegi: Sorry, that's a left-over from an earlier stage.

prusti-viper/src/encoder/counterexample.rs
cedihegi: I tried this, and there is one drawback, it prints everything on one line. I guess we can discuss this, I can also try to clean up the code of my formatting. But having the fields on multiple lines with indentations is pretty useful in my opinion, especially when there are nested structures.

prusti-viper/src/encoder/backtranslation.rs
cedihegi: I remember spending quite a bit of time on this and there is no simple method for that. I will look into it again, something better than what I currently have must definitely be possible.

prusti-viper/src/verifier.rs
cedihegi: I guess that's saver than just relying on Silicon to only produce one when we set the flags. Should I also add this to the counterexample extraction in viper/verifier.rs? Or is that a bad idea because prusti_common is currently not used in that crate?

prusti-viper/src/encoder/backtranslation.rs
cedihegi: I simplified it a bit, but it essentially still does the same. I really don't know how to do this another way. I am very open to suggestions.

prusti-viper/src/encoder/counterexample.rs
```diff
@@ -0,0 +1,169 @@
...
+    Unit,
+    UnknownEntry,
+}
+
+impl fmt::Debug for Entry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Entry::IntEntry { value } => write!(f, "{}", value),
+            Entry::BoolEntry { value } => write!(f, "{}", value),
+            Entry::CharEntry { value } => write!(f, "'{}' ({:x})", value, *value as i32),
```
Aurel300: Reading your evaluation section I realised this can be confusing without the standard `0x` prefix added:

```suggestion
            Entry::CharEntry { value } => write!(f, "'{}' (0x{:x})", value, *value as i32),
```

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -9,8 +9,10 @@ use std::collections::HashMap;
 use rustc_span::source_map::SourceMap;
 use rustc_span::MultiSpan;
 use viper::VerificationError;
+use viper::silicon_counterexample::*;
```
fpoli: Is this import unused?

cedihegi: Oh yes, thank you.

vakaras (commented):
LGTM modulo the failing tests.

PR #446, opened 2021-03-30
Cherry-pick vir! macro from #54

dario23:


prusti-common/src/vir/vir_macro.rs
```diff
@@ -0,0 +1,125 @@
...
+        crate::vir::Expr::BinOp(vir::BinOpKind::And,
+            Box::new(vir!($head)),
+            Box::new(vir!($tail)),
+            crate::vir::Position::default())
+    };
+    ($head: tt || $tail: tt) => {
+        crate::vir::Expr::BinOp(vir::BinOpKind::Or,
+            Box::new(vir!($head)),
+            Box::new(vir!($tail)),
+            crate::vir::Position::default())
```
Aurel300: I think there are functions defined in `Expr` that build particular binary operations with a default position, would make this a bit shorter. There might be similar for `Stmt`s too?

prusti-common/src/vir/vir_macro.rs
```diff
@@ -0,0 +1,125 @@
...
+    };
+    ($antecedent: tt ==> $consequent: tt) => {
+        crate::vir::Expr::BinOp(vir::BinOpKind::Implies,
+            Box::new(vir!($antecedent)),
```
Aurel300: If there is no "shorthand constructor" as described above, use `box xyz` syntax for the boxes.

prusti-common/src/vir/vir_macro.rs
```diff
@@ -0,0 +1,125 @@
...
+            crate::vir::Position::default())
+    };
+    ($lhs: tt {$borrow: expr} --* $rhs: tt) => {
+        crate::vir::Expr::magic_wand(vir!($lhs), vir!($rhs), $borrow)
+    };
+
+    (forall $($name: ident : $type: ident),+ :: {$trigger: tt} $body: tt) => {
+        crate::vir::Expr::ForAll(
+            vec![$(vir_local!($name: $type)),+],
+            vec![crate::vir::Trigger::new(vec![vir!($trigger)])],
```
Aurel300: There can be multiple trigger sets, each with multiple expressions in it. I think covering the former case is more important.
```suggestion
    (forall $($name: ident : $type: ident),+ :: { $($triggers: tt),* } $body: tt) => {
        crate::vir::Expr::ForAll(
            vec![$(vir_local!($name: $type)),+],
            vec![$(crate::vir::Trigger::new(vec![vir!($triggers)])),+],
```

prusti-common/src/vir/vir_macro.rs
```diff
@@ -0,0 +1,125 @@
...
+            Box::new(vir!($body)),
+            crate::vir::Position::default())
+    };
+    ([ $e: expr ]) => { $e.clone() };
+    (( $($tokens: tt)+ )) => { vir!($($tokens)+) }
+}
+
+#[cfg(test)]
+mod tests {
```
Aurel300, line 123: Please give this a more descriptive name; when this fails when running all tests in Prusti, it just says "failed running tests" or something similar.

fpoli: What's the error message? I think that the convention is to always use `mod test { .. }`, because having unique names for each test module sounds difficult to enforce.

dario23: @fpoli i can remove the `s` if you want
@Aurel300  the error seems pretty clear to me; it includes the full module path (except the crate name, but that's not something i can fix with the module name here):

```
failures:
    vir::vir_macro::tests::forall

test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
```

Aurel300: Ok, this is fine then. I remember dealing with failures that just said `tests` instead of the full module path; I guess this happened when the `tests` module was top-level in its crate, since the crate name is not shown.

prusti-common/src/vir/vir_macro.rs
dario23: some `Stmt`s have, some don't .. i don't see why, exactly. tidied up a bit.

dario23:
currently, `vir!` isn't usable from other crates because there's also a module `prusti_common::vir` so import can't find the macro (or the module, sometimes). suggestions/wishes for renaming the macro? `parse_vir!`? `inline_vir!`?

Aurel300:
`vir_quote!` comes to mind.

dario23:
or maybe not, it seems `use prusti_common::vir` makes both the module (i.e. types inside that) and the `vir!` macro work, i just forgot `#[macro_export]` on the definition site.. so i guess the question is: is the nice name worth the ambiguity?

Aurel300:
In that case I prefer just `vir!`, I guess macros and other things are kind of in separate namespaces (i.e. `vir::Expr` will not be shadowed because `vir!` is imported).

Aurel300:
For future reference: big(?) parts of `prusti-viper` can be made nicer with these new macros, but general cleanup is not a goal of Johannes' thesis. We can make this cleanup in a PR at some less busy time.

PR #451, opened 2021-04-06
Adding Support for Floating-Points and Bitwise Operations

lukfried:


prusti-common/src/vir/ast/common.rs
```diff
@@ -144,6 +149,7 @@ pub enum Type {
     /// TypedRef: the first parameter is the name of the predicate that encodes the type
     TypedRef(String),
     Domain(String),
+    Float,
```
vakaras: Unless I am missing something, this should be:
```suggestion
    Float(Float),
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -885,6 +885,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 vir::Type::Bool => BuiltinMethodKind::HavocBool,
                 vir::Type::TypedRef(_) => BuiltinMethodKind::HavocRef,
                 vir::Type::Domain(_) => BuiltinMethodKind::HavocRef,
+                vir::Type::Float => unreachable!("BackendType has no built-in methods?"),
```
vakaras: Here you probably need to have `BuiltinMethodKind::HavocF64` and `BuiltinMethodKind::HavocF32`.

prusti-common/src/vir/ast/common.rs
Aurel300: This and `struct Float` should only exist if Viper has two different float types natively. Does it? Or does it use arbitrary-precision reals? Or both?

lukfried: @Aurel300 As far as I can tell, the BackendType from Viper allows for arbitrary floating-points. `BackendType(s"float${mant}e${exp}", s"(_ FloatingPoint ${exp} ${mant})")` See [src/main/scala/viper/silver/ast/utility/BackendTypeFactories.scala](https://github.com/viperproject/silver/blob/075fd4ba75532b8160429815281fe263b535bac4/src/main/scala/viper/silver/ast/utility/BackendTypeFactories.scala) line 39-73. But Rust only has `f32` and `f64`.

Aurel300: I see. In that case `Float(u32, u32)` would map the native types faithfully, but I guess it's unlikely that we would use a non-standard combination of parameters. So I agree with Vytautas' original code suggestion, although maybe I would rename `enum Float` to `enum FloatSize` or similar.

prusti-common/src/vir/ast/expr.rs
```diff
@@ -80,11 +84,40 @@ pub enum BinOpKind {
...
+pub enum FloatConst {
+    FloatConst64(rustc_apfloat::ieee::IeeeFloat<DoubleS>),
+    FloatConst32(rustc_apfloat::ieee::IeeeFloat<SingleS>),
+}
+
+impl PartialEq for FloatConst {
+    fn eq(&self, other: &FloatConst) -> bool {        
+        match (self, other) {
+            (FloatConst::FloatConst64(f1), FloatConst::FloatConst64(f2)) => f1.to_string() == f2.to_string(),
+            (FloatConst::FloatConst32(f1), FloatConst::FloatConst32(f2)) => f1.to_string() == f2.to_string(),
```
Aurel300: Why `to_string()` for comparisons?

fpoli: Probably because floats are `PartialEq` only.

Aurel300: But this is a `PartialEq` implementation for `FloatConst`. I don't really see how converting to a string helps? If anything, converting to a string could lose information, and I don't expect that the comparison here should do an "equals +/- a small epsilon" kind of comparison.

fpoli: True. I was referring to the `impl Eq for FloatConst {}` below.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1075,6 +1076,8 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             ty::TyKind::Uint(ty::UintTy::U64) => scalar_value.to_u64().unwrap().into(),
             ty::TyKind::Uint(ty::UintTy::U128) => scalar_value.to_u128().unwrap().into(),
             ty::TyKind::Uint(ty::UintTy::Usize) => scalar_value.to_machine_usize(&self.env().tcx()).unwrap().into(),
+            ty::TyKind::Float(ty::FloatTy::F32) => unimplemented!("to_f32 not yet implemented"),
+            ty::TyKind::Float(ty::FloatTy::F64) => scalar_value.to_f64().unwrap().into(),
```
lukfried: @vakaras Compilation fails here with error message

> error[E0277]: the trait bound `prusti_common::vir::Expr: From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::DoubleS>>` is not satisfied
    --> prusti-viper/src/encoder/encoder.rs:1080:83
     |
1080 |             ty::TyKind::Float(ty::FloatTy::F64) => scalar_value.to_f64().unwrap().into(),
     |                                                                                                                            ^^^^ the trait `From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::DoubleS>>` is not implemented for `prusti_common::vir::Expr`
     |
     = help: the following implementations were found:
               <prusti_common::vir::Expr as From<&'a LocalVar>>
               <prusti_common::vir::Expr as From<&'a str>>
               <prusti_common::vir::Expr as From<LocalVar>>
               <prusti_common::vir::Expr as From<bool>>
             and 16 others

However, among the '16 other' implementations is `<prusti_common::vir::Expr as From<rustc_apfloat::ieee::IeeeFloat<rustc_apfloat::ieee::DoubleS>>>` in [prusti-common/src/vir/conversions.rs](https://github.com/viperproject/prusti-dev/pull/451/files#diff-b34dcdcd8d6eeb45501fe2f693881bb0658cfe1899099e091436e845b1678ccf).

So I am at a loss as to what goes wrong.

viper-sys/build.rs
```diff
@@ -200,11 +200,30 @@ fn main() {
...
                 object_getter!(),
             ]),
-
+            
+            java_class!("viper.silver.ast.utility.FloatFactory", vec![
+                constructor!(),
+            ]),
+            java_class!("viper.silver.ast.utility.FloatFactory$", vec![
+                object_getter!(),
+                method!("apply", "(Lscala/Int;Lscala/Int;Lviper/silver/ast/utility/RoundingMode;)Lviper/silver/ast/utility/FloatFactory"),
```
lukfried: @vakaras I did not manage to match the type signature. I think the issue is `Lscala/Int`. I've also tried `Lscala/lang/Int` and `Lscala/lang/Int`. The class it refers to is [here](https://github.com/viperproject/silver/blob/8a7d18d1c9a418ea4fd644f9bed4410572f321dd/src/main/scala/viper/silver/ast/utility/BackendTypeFactories.scala#L45).

prusti-common/src/vir/ast/expr.rs
```diff
@@ -995,6 +1019,9 @@ impl Expr {
             Expr::DomainFuncApp(ref func, _, _) => {
                 &func.return_type
             },
+            Expr::BackendFuncApp(ref func, _, _) => { 
+                unimplemented!("Unsure what comes here")
```
lukfried: @vakaras See unimplemented! message

viper-sys/build.rs
vakaras: You can look up the signatures with `javap`:
```bash
javap -cp viper_tools/backends/silicon.jar -s 'viper.silver.ast.utility.FloatFactory'
```

In this case, it is `(IILscala/Enumeration$Value;)Lviper/silver/ast/utility/FloatFactory;`

prusti-common/src/vir/ast/expr.rs
vakaras: This should return the return type of the function. I guess it should be similar to `DomainFuncApp`.

prusti-common/src/vir/ast/expr.rs
```diff
@@ -1414,6 +1414,7 @@ impl Expr {
                     | Expr::LetExpr(..)
                     | Expr::FuncApp(..)
                     | Expr::DomainFuncApp(..)
+                    | Expr::BackendFuncApp(_, _, _) // unsure what happens here and if this is correct
```
Aurel300: It's better if you write questions like these into comments on your own PR and leave the code without them.

viper/src/ast_factory/expression.rs
```diff
@@ -546,6 +555,25 @@ impl<'a> AstFactory<'a> {
         Expr::new(obj)
     }
 
+    pub fn backend_func_app(
+        &self,
+        backend_function_name: &str,
```
vakaras: @lukfried I think the crash happens because you should pass the actual backend function here (similarly to `domain_func_app`), not just its name. See the signature of [Scala constructor](https://github.com/viperproject/silver/pull/428/files#diff-20523c8dac890ebb6c98a54fbabb02db236d0863632d3cc112c5a837f547ac0fR390).

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5291,8 +5291,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 // encode a closure deep copy or at least a stub
                 debug!("warning: ty::TyKind::Closure not implemented yet");
                 Vec::new()
-            }
-            ty::TyKind::Float(_) => unimplemented!("Unsure what to do here"),
+            }, 
+
+            ty::TyKind::Float(_) => unimplemented!("unsure what comes here"),
```
lukfried: @vakaras See comment. I generally don't understand the purpose of this code, but I've tried adding the `Float` case to to the integer cases above, resulting in the following error:
```
error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_1: Ref(unsupported), Position { line: 2, column: 5, id: 9 }), val_float: Float, Position { line: 2, column: 5, id: 9 }), read)))
```
when verifying the func_signature.rs test added in this commit. I guess it has to do with the `Ref(unsupported)`, but I was unable to track down, where that originates from. Any pointers are appreciated!

vakaras: This commit fixes the fold-unfold error: https://github.com/viperproject/prusti-dev/commit/9399c2b1432b9aaad3d2191eb287a930720e8ab9

I found what needs to be fixed by searching for `"unsupported"`.

Now the error is:
```
[2021-05-28T07:25:42Z ERROR viper::verifier] The verifier returned an unhandled error of type viper.silicon.reporting.Z3InteractionFailed$$anon$2: Interaction with Z3 (instance 08) failed: Unexpected output of Z3. Expected 'success' but found: (error "line 182 column 74: unknown function/constant $SortWrappers.___FloatingPoint_12_52_To$Snap") (<no position>)
thread 'rustc' panicked at 'explicit panic', viper/src/verifier.rs:203:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Is it clear for you how to proceed?

lukfried: I'm actually not quite sure. It could be that the return type of a function does not get encoded. The `$Snap` in the error message should be replaced with the correct type. Or the `$SortWrappers` is out of place, and I forgot to unwrap something.

lukfried: I'm actually not quite sure. It could be that the return type of a function does not get encoded. The `$Snap` in the error message should be replaced with the correct type. Or the `$SortWrappers` is out of place, and I forgot to unwrap something.

viper/src/ast_factory/expression.rs
```diff
@@ -198,29 +197,85 @@ impl<'a> AstFactory<'a> {
...
+
+    pub fn backend_bv64_lit(&self, bits: u64) -> Expr<'a> {    
         let bv_factory_ = ast::utility::BVFactory::with(self.env);
         let bv_factory = ast::utility::BVFactory::new(&bv_factory_, 64).unwrap();
         let from_int = ast::utility::BVFactory::call_from__int(&bv_factory_, bv_factory, self.jni.new_string("toBV64")).unwrap();
         self.backend_func_app(from_int, &[self.int_lit(bits as i64)], self.no_position())
     }
+    
+    // Backend Floating-Points
+    pub fn float_binop(&self, op_code: u8, f_size: u8, left: Expr, right: Expr) -> Expr<'a> {
```
vakaras: Could you please add `FloatOpKind` enum to `viper` crate and use it instead of `u8`?

viper/tests/test_program.rs
```diff
@@ -0,0 +1,46 @@
...
+    static ref VIPER: Viper = Viper::new();
+}
+
+#[test]
+fn test_program() {
+    env_logger::init();
+
+    
+    let verification_context: VerificationContext = VIPER.new_verification_context();
+    let ast = verification_context.new_ast_factory();
```
vakaras: Add:
```rust
let ast_utils = verification_context.new_ast_utils();
```

viper/tests/test_program.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    let program = ast.program(
+        &[],
+        &[],
+        &[foo],
+        &[],
+        &[],
+    );
+
+    let verifier = verification_context.new_verifier(viper::VerificationBackend::Silicon, None);
```
vakaras: Add:
```rust
    let pretty_printed = ast_utils.pretty_print(program);
    println!("Pretty printed:\n{}", pretty_printed);
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2006,6 +2010,15 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                                 )?
                             );
                         }
+                        "core::f32::<impl f32>::is_nan" |
+                        "core::f64::<impl f64>::is_nan" => {
+                            
+                            // I'd like to replace the function call with this VIR Expr
+                            // let expr = vir::Expr::UnaryOp(UnaryOpKind::IsNaN, args, location);
```
vakaras: The code should be similar to [this](https://github.com/viperproject/prusti-dev/blob/c7cb51791623c12c254f5f8f9c1bc40a58585a8a/prusti-viper/src/encoder/procedure_encoder.rs#L2382-L2391), just instead of `inhaled_expr` you need to use your `UnaryOp`.

Aurel300:
@vakaras I assume #567 supersedes this?

vakaras:
> @vakaras I assume #567 supersedes this?

Yes.

PR #454, opened 2021-04-10
Use .into() vir::Expr instances in some more places

dario23:


prusti-common/src/vir/ast/expr.rs
```diff
@@ -339,6 +339,18 @@ impl Expr {
...
+        Expr::Const(Const::Int(i.into()), Position::default())
+    }
+
+    pub fn const_bigint<S: Into<String>>(s: S) -> Self {
+        Expr::Const(Const::BigInt(s.into()), Position::default())
+    }
+
+    pub fn const_bool(b: bool) -> Self {
+        Expr::Const(Const::Bool(b), Position::default())
+    }
```
Aurel300: Aren't there already `Into` implementations for (some of) these?

prusti-viper/src/encoder/encoder.rs
```diff
@@ -1304,7 +1304,7 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
                     ),
                     vir::Position::default(),
                 ),
-                box vir::Expr::Const(vir::Const::Bool(true), vir::Position::default()),
+                box vir!{ true },
```
Aurel300: Just `true.into()` should work.

prusti-common/src/vir/ast/expr.rs
dario23: hm, yes; that makes the PR one commit shorter i guess :-)

prusti-common/src/vir/optimizations/functions/simplifier.rs
```diff
@@ -7,6 +7,7 @@
 //! Function simplifier that simplifies expressions.
 
 use super::super::super::ast::{self, ExprFolder};
+use crate::vir;
```
Aurel300: I guess this is no longer necessary.

PR #455, opened 2021-04-11
Builtins: add array lookup pure function

dario23:


prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -69,6 +77,11 @@ impl BuiltinEncoder {
             BuiltinFunctionKind::Undefined(vir::Type::Bool) => format!("builtin$undef_bool"),
             BuiltinFunctionKind::Undefined(vir::Type::TypedRef(_)) => format!("builtin$undef_ref"),
             BuiltinFunctionKind::Undefined(vir::Type::Domain(_)) => format!("builtin$undef_doman"),
+            BuiltinFunctionKind::ArrayLookupPure { array_ty, array_len, .. } => {
+                // we kind of abuse the TypedRef variant here to store the rust type name as string
```
Aurel300: This is true of a lot of Prusti in general though, no need to comment on it here :)

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -92,6 +105,35 @@ impl BuiltinEncoder {
...
+                    formal_args: vec![
+                        // self,
+                        self_var.clone(),
+                        // idx,
+                        idx_var.clone(),
+                    ],
+                    return_type: return_ty,
+                    pres: vec![
+                        // idx < {len}
+                        vir!([vir::Expr::Local(idx_var, vir::Position::default())]  < [vir::Expr::from(array_len)]),
```
Aurel300:
```suggestion
                        vir!([vir::Expr::local(idx_var)] < [array_len.into()]),
```
Does the RHS with `into()` work?

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -92,6 +105,35 @@ impl BuiltinEncoder {
...
+                        // acc(self, read$())
+                        vir::Expr::predicate_access_predicate(
+                            type_pred,
+                            vir::Expr::Local(self_var, vir::Position::default()),
```
Aurel300:
```suggestion
                            vir::Expr::local(self_var),
```

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -392,6 +405,28 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
+                // let snapshot = self.encoder.encode_snapshot(&self.ty)?;
+                // let return_ty = snapshot.get_type();
+
+                // need the array's corresponding lookup_pure function as well
+                self.encoder.encode_builtin_function_def(
+                    BuiltinFunctionKind::ArrayLookupPure {
+                        array_ty: vir::Type::TypedRef(format!("{}", elem_ty)),
+                        array_len: array_len.try_into().unwrap(),
+                        return_ty: vir::Type::Int,
+                    });
```
Aurel300: For the time being, this is fine, but eventually the encoding of `lookup_pure` should be triggered by actually using it where it is needed.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: removed.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: no, it cannot infer the type.. so i think `vir::Expr::from(array_len)` is the shortest version, i mean `from` is the same as `into` internally anyway.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: yep :-)

prusti-viper/src/encoder/type_encoder.rs
dario23: also should be called lazily by `encode_builtin_function_use` at the top of the function, as [@fpoli noted](https://github.com/viperproject/prusti-dev/pull/463/files/70542500f5013c218a5332fe93f58e79c2ee135f#r614717183). will remove, anyway.

dario23: wait, that's `encode_type_predicate_use` on top there, not `encode_builtin_function_use`. the latter is only called from `pure_function_encoder.rs`, so i think i want to have it called somewhere "manually" for now after all. i'll add a TODO to remove it once we have the `encode_builtin_function_use` calls from other places.

fpoli:
I see there are some fixme/todo/array_todo. Is this PR a work in progress or are the todo items going to be fixed in other PRs?

dario23:
The TODO in `TypeEncoder::encode_value_field` i've solved by returning a proper error; as far as i understand that method should never be called for arrays.

There is a TODO to remove the explicit call to `encode_builtin_function_def`. my plan would be to leave that in and remove it (and the call) when we encode array code using it and can trigger the call from there (indirectly, via `encode_builtin_function_use` i imagine).

So i think we can merge this, modulo further review-comments.

PR #460, opened 2021-04-14
Update toolchain to 2021-04-13

dario23:
The previous version (2021-04-01) had component rustfmt missing, which lead to a multiline warning on each `x.py` call (at least for me).


dario23:
i'll just do the other two tasks of #461 here as well then?

Aurel300:
> i'll just do the other two tasks of #461 here as well then?

Please do. Don't worry about the `prusti-server` crate, this is pending on something external.

fpoli:
The PR looks good to me, thanks! @dario23 are you still working on the upgrades or can I merge?

dario23:
ready to merge from my side

PR #462, opened 2021-04-15
Use #[track_caller] on expect-like functions

dario23:
This way, the call site shows up in the error message (even without
RUST_BACKTRACE), not the definition site.


fpoli:
Nice!

PR #463, opened 2021-04-16
Encode array-using code using PlaceEncoding

dario23:
work in progress, but uploading for @fpoli and @Aurel300 to take a look already

### DIFF
--- a/prusti-viper/src/encoder/builtin_encoder.rs
+++ b/prusti-viper/src/encoder/builtin_encoder.rs
@@ -5,7 +5,8 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use crate::encoder::snapshot;
-use prusti_common::{vir, vir::WithIdentifier};
+use prusti_common::{vir, vir_local, vir::WithIdentifier};
+use rustc_middle::ty;
 
 #[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
 pub enum BuiltinMethodKind {
@@ -20,7 +21,14 @@ pub enum BuiltinFunctionKind {
     Unreachable(vir::Type),
     /// type
     Undefined(vir::Type),
+    /// array lookup pure function, e.g. Array$4$u32$lookup_pure
+    ArrayLookupPure {
+        array_ty: vir::Type,
+        array_len: usize,
+        return_ty: vir::Type,
+    },
 }
+
 #[derive(Clone, Debug, Hash, Eq, PartialEq)]
 pub enum BuiltinDomainKind {
     Nat,
@@ -69,6 +77,11 @@ impl BuiltinEncoder {
             BuiltinFunctionKind::Undefined(vir::Type::Bool) => format!("builtin$undef_bool"),
             BuiltinFunctionKind::Undefined(vir::Type::TypedRef(_)) => format!("builtin$undef_ref"),
             BuiltinFunctionKind::Undefined(vir::Type::Domain(_)) => format!("builtin$undef_doman"),
+            BuiltinFunctionKind::ArrayLookupPure { array_ty, array_len, .. } => {
+                // we kind of abuse the TypedRef variant here to store the rust type name as string
+                let array_ty = if let vir::Type::TypedRef(ty) = array_ty { ty } else { unreachable!() };
+                format!("Array${}${}$lookup_pure", array_len, array_ty)
+            }
         }
     }
 
@@ -91,6 +104,35 @@ impl BuiltinEncoder {
                 pres: vec![],
                 posts: vec![],
                 body: None,
+            },
+            BuiltinFunctionKind::ArrayLookupPure { array_ty, array_len, return_ty } => {
+                let array_ty = if let vir::Type::TypedRef(ty) = array_ty { ty } else { unreachable!() };
+                let type_pred = format!("Array${}${}", array_len, array_ty);
+                let self_var = vir::LocalVar::new("self", vir::Type::TypedRef(type_pred.clone()));
+                let idx_var = vir_local!{ idx: Int };
+
+                vir::Function {
+                    name: fn_name,
+                    formal_args: vec![
+                        // self,
+                        self_var.clone(),
+                        // idx,
+                        idx_var.clone(),
+                    ],
+                    return_type: return_ty,
+                    pres: vec![
+                        // idx < {len}
+                        vir!([vir::Expr::Local(idx_var, vir::Position::default())]  < [vir::Expr::from(array_len)]),
+                        // acc(self, read$())
+                        vir::Expr::predicate_access_predicate(
+                            type_pred,
+                            vir::Expr::Local(self_var, vir::Position::default()),
+                            vir::PermAmount::Read,
+                        )
+                    ],
+                    posts: vec![],
+                    body: None,
+                }
             },
         }
     }
--- a/prusti-viper/src/encoder/mir_encoder.rs
+++ b/prusti-viper/src/encoder/mir_encoder.rs
@@ -17,10 +17,58 @@ use rustc_middle::{mir, ty};
 use rustc_index::vec::{Idx, IndexVec};
 use rustc_span::{Span, DUMMY_SP};
 use log::{trace, debug};
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Display};
 
 pub static PRECONDITION_LABEL: &'static str = "pre";
 pub static WAND_LHS_LABEL: &'static str = "lhs";
+
+/// Result of encoding a place
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum PlaceEncoding {
+    /// Just an expression, the most common case.
+    Expr(vir::Expr),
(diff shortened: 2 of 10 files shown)


prusti-viper/src/encoder/mir_encoder.rs
```diff
@@ -17,11 +17,68 @@ use rustc_middle::{mir, ty};
...
 
+/// Result of encoding a place
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum PlaceEncoding {
+    /// Just an expression, the most common case.
+    Expr(vir::Expr),
+    /// Field access expression
+    FieldAccess { base: Box<PlaceEncoding>, field: vir::Field },
+    /// Array access expression
+    ArrayAccess { base: Box<PlaceEncoding>, index: Box<PlaceEncoding>, typ: vir::Type },
```
fpoli, line 33: Could the index be just `vir::Expr`? I remember we checked whether `x[y[0]]` is encoded as one or two MIR statement during the meeting, but I don't recall the outcome.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -20,7 +21,14 @@ pub enum BuiltinFunctionKind {
     Unreachable(vir::Type),
     /// type
     Undefined(vir::Type),
+    /// array lookup pure function, e.g. Array$4$u32$lookup_pure
+    ArrayLookupPure {
+        array_ty: vir::Type,
+        array_len: usize,
```
fpoli, line 27: Is `array_ty` the type of the array or the type of the elements? In the former case I would expected the array length to be already part of the string contained in `array_ty`, as constructed in the array case of `TypeEncoder::encode_predicate_use`.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1266,7 +1273,10 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
     }
 
     /// Encode the lhs and the rhs of the assignment that create the loan
-    fn encode_loan_places(&self, loan_places: &LoanPlaces<'tcx>) -> (vir::Expr, vir::Expr, bool) {
+    fn encode_loan_places(
+        &self,
+        loan_places: &LoanPlaces<'tcx>
+    ) -> (PlaceEncoding, PlaceEncoding, bool) {
```
fpoli, line 1279: Is every usage of `encode_loan_places` going to call `.expect_expr()` on the returned `PlaceEncoding`? If so, what about calling `.expect_expr()` inside this function, returning `vir::Expr`?

fpoli: Even better, we could add a new `ProcedureEncoder::encode_place` (initially) with the same signature of the *old* `MirEncoder::encode_place`, replacing every `self.mir_encoder.encode_place` call with `self.encode_place`. The conversion from `PlaceEncoding` to `vir::Expr` would need to be implemented just in `ProcedureEncoder::encode_place`.

Once `ProcedureEncoder::encode_place` is in place, we can add a `Vec<vir::Stmt>` to its result to encode the array lookup statements that should preceed the usage of the returned `vir::Expr`.

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -392,6 +405,28 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
+                // let snapshot = self.encoder.encode_snapshot(&self.ty)?;
+                // let return_ty = snapshot.get_type();
+
+                // need the array's corresponding lookup_pure function as well
+                self.encoder.encode_builtin_function_def(
+                    BuiltinFunctionKind::ArrayLookupPure {
+                        array_ty: vir::Type::TypedRef(format!("{}", elem_ty)),
+                        array_len: array_len.try_into().unwrap(),
+                        return_ty: vir::Type::Int,
+                    });
```
fpoli, line 423: (In the end, this shouldn't be needed. `encode_builtin_function_use` is supposed to lazily call `encode_builtin_function_def`.)

prusti-viper/src/encoder/type_encoder.rs
```diff
@@ -392,6 +405,28 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
+
+                vec![
+                    vir::Predicate::new_abstract(vir::Type::TypedRef(predicate_name)),
```
fpoli, line 426: Where is `predicate_name` constructed? I can't find it.

fpoli:
I expected `PlaceEncoding` to be used much less; just in the `ProcedureEncoder::encode_place` suggested [in this comment](https://github.com/viperproject/prusti-dev/pull/463#discussion_r614710930). This way, many less changes should be needed.

dario23:
> I expected `PlaceEncoding` to be used much less; just in the `ProcedureEncoder::encode_place` suggested [in this comment](https://github.com/viperproject/prusti-dev/pull/463#discussion_r614710930). This way, many less changes should be needed.

hm, i will try that.

prusti-viper/src/encoder/type_encoder.rs
dario23: agree, will remove

prusti-viper/src/encoder/procedure_encoder.rs
dario23: what `vir::Expr` would we return for an array access then?

prusti-viper/src/encoder/mir_encoder.rs
dario23: yep, index is just a `vir::Expr`. as the [`ProjectionElem::Index(V)`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.ProjectionElem.html) for `mir::Place` is instantiated with [`V` = `Local`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/type.PlaceElem.html) i think we can rely on that pretty safely.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: renaming to `array_elem_ty` for clarity.

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: Adding `Vec<vir::Stmt>` to the return type of `ProcedureEncoder::encode_place` is needed for that. The statements would allow to store the array lookup to a temporary (VIR) local variable and the `vir::Expr` could then use that temporary variable.

prusti-viper/src/encoder/mir_encoder.rs
```diff
@@ -582,6 +659,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> MirEncoder<'p, 'v, 'tcx> {
     pub fn encode_cast_expr(
         &self,
         operand: &mir::Operand<'tcx>,
+        // src_ty: ty::Ty<'tcx>,
+        // encoded_operand: TODO
```
Aurel300, line 663: What should these be?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2164,18 +2178,26 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                     vir::Expr::not(cond_var.into())
                 };
 
+                let msg_desc =
+                    // rustc_middle/src/mir/mod.rs says BoundsCheck is expected to be handled by
+                    // the caller (?!)
```
Aurel300, line 2183: I guess this is because bounds check are encoded kind of specially, with all the bounds checks for a (potentially nested) projection before the actual projection happens? Does [`fmt_assert_args`](https://github.com/rust-lang/rust/blob/62652865b6029b4776a7c03efa13a37b15c9b953/compiler/rustc_middle/src/mir/mod.rs#L1343) help us here? If not, `"BoundsCheck"` is not good enough as a description for the user, since this message is what Prusti displays in the error when the verifier detects a reachable assertion failure.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5180,15 +5319,23 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                todo!("array deep copy {:?}, {:?}", elem_ty, len)
+            }
 
             ref x => unimplemented!("{:?}", x),
         };
         Ok(stmts)
     }
 
+    /// Assignment with an aggregate on the RHS.
+    /// [dst] = Foo { x: [operand_0], y: [operand_1], .. }; // struct
```
Aurel300, line 5333: Also enums, if you want this comment to be more accurate.

dario23:
closing this PR, as it's all done in separate smaller ones now.

PR #464, opened 2021-04-16
Added a flake for building Prusti with the Nix package manager

jmc-figueira:
This flake allows one to build Prusti through the Nix package manager
(https://nixos.org/), on NixOS or other distros with Nix installed,
as well as macOS.

Caveats:
- Currently the flake will be built from source every time, which takes
  a considerable amount of time. A cache setup through https://app.cachix.org/
  may be used to solve this problem (and can be setup as a CI workflow).
- As of 04/01/2021, rustfmt was not available in the nightly (as per
  https://rust-lang.github.io/rustup-components-history/). The
  `rust-toolchain` file was altered to default to a minimal profile,
  with optional components added explicitly. If this is unacceptable, an
  alternative must be found, as toolchain component unavailability in
  Nix is an error (aborts the build) and not a warning.
- Currently, Viper is simply downloaded from the official Nightly
  download link, as the Python script does. This causes a problem, as
  Nix builds must be hermetic and reproducible, and changes in the Viper
  nightly, which is unversioned, cause the SHA256 to change and abort
  the build. A better solution would be to compile Viper from source
  using a separate Nix expression or to download a version-pinned
  archive (if possible).
- Currently, Prusti is not agnostic to the Rust toolchain instalation
  method and expects `rustup` to be used. This was an issue on runtime
  for `prusti-launch`, which expects to find the Rust sysroot through
  `rustup`. This was altered by allowing an environment variable,
  `PRUSTI_SYSROOT`, to be set, pointing to the Rust sysroot in an
  environment-agnostic manner, which makes it compatible with the way
  the Rust toolchain is installed as a dependency in Nix (through
  https://github.com/oxalica/rust-overlay). The developer documentation
  regarding environment variables should probably be updated to account
  for this
  (https://viperproject.github.io/prusti-dev/dev-guide/development/build.html).
- Documentation should, similarly, be updated to provide instructions
  on installing and setting up a development environment through Nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -43,10 +43,8 @@
               pkgs.makeWrapper
             ];
 
-            copyLibs = true;
-            copyLibsFilter = ''
-              select(.reason == "compiler-artifact" and ((.target.kind | contains(["lib"])) and (.package_id | test("prusti-contracts"))) and .profile.test == false)
-            '';
+            copyTarget = true;
+            compressTarget = false;
 
             override = _: {
               preBuild = ''
@@ -55,7 +53,11 @@
                 export Z3_EXE="${packages.viper}/z3/bin/z3"
                 export ASM_JAR="${packages.ow2_asm}/asm.jar"
               '';
+            };
+            overrideMain = _: {
               postInstall = ''
+                rm $out/bin/test-crates
+
                 for f in $(find $out/bin/ $out/libexec/ -type f -executable); do
                   wrapProgram $f \
                     --set RUST_SYSROOT "${rust}" \
@@ -65,8 +67,11 @@
                     --set Z3_EXE "${packages.viper}/z3/bin/z3"
                 done
 
-                cp -r $out/lib/. $out/bin
-                rm -rf $out/lib/
+                mkdir $out/bin/deps
+                cp $out/target/release/libprusti_contracts.rlib $out/bin
+                cp $out/target/release/deps/libprusti_contracts_internal-* $out/bin/deps
+                rm -rf $out/target
+                rm $out/bin/deps/*.{rlib,rmeta}
               '';
             };
           };
@@ -75,7 +80,7 @@
             name = "viper";
             url = "https://viper.ethz.ch/downloads/ViperToolsNightlyLinux.zip";
             stripRoot = false;
-            hash = "sha256-oV9VgGmzNrLW6n08vXU2rlV6S9Eca3U0L+gNbVxKhEE=";
+            hash = "sha256-82vnyO7QWaLzehnBzPJxuEmdqK0MnWWwQnmdLq28sQc=";
           };
 
           ow2_asm = pkgs.stdenv.mkDerivation rec {
--- a/flake.nix
+++ b/flake.nix
@@ -99,6 +99,53 @@
           };
         };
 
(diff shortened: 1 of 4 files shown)


prusti-launch/src/bin/prusti-rustc.rs
```diff
@@ -36,8 +36,11 @@ fn process(mut args: Vec<String>) -> Result<(), i32> {
     let libjvm_path = prusti_launch::find_libjvm(&java_home)
         .expect("Failed to find JVM library. Check JAVA_HOME");
 
-    let prusti_sysroot = prusti_launch::prusti_sysroot()
-        .expect(&format!("Failed to find Rust's sysroot for Prusti"));
+    let prusti_sysroot = match env::var("PRUSTI_SYSROOT") {
+        Ok(prusti_sysroot) => PathBuf::from(prusti_sysroot),
+        Err(_) => prusti_launch::prusti_sysroot()
+            .expect(&format!("Failed to find Rust's sysroot for Prusti"))
+    };
```
fpoli: Could you move this match into `prusti_sysroot`?

flake.nix
```diff
@@ -0,0 +1,90 @@
...
+              done
+            '';
+          };
+        };
+
+        packages.viper = pkgs.fetchzip {
+          name = "viper";
+          url = "https://viper.ethz.ch/downloads/ViperToolsNightlyLinux.zip";
+          stripRoot = false;
+          hash = "sha256-oV9VgGmzNrLW6n08vXU2rlV6S9Eca3U0L+gNbVxKhEE=";
```
fpoli: `https://viper.ethz.ch/downloads/ViperToolsNightlyLinux.zip` is overwritten often. `http://viper.ethz.ch/downloads/ViperToolsLastReleaseLinux.zip` might a bit more stable, but I haven't tried it. We are still working on a way to refer to an exact old nightly version of Viper.

prusti-launch/src/bin/prusti-rustc.rs
jmc-figueira: Done in https://github.com/viperproject/prusti-dev/pull/464/files#diff-dd855b3156f3186827e7c4436f3c131a7d57f9cbba9616567a0522412c8d3eb5, please check if it is ok.

flake.nix
jmc-figueira: A possible solution suggested by @vakaras in Zulip (https://prusti.zulipchat.com/#narrow/stream/236614-general/topic/Building.20Prusti.20on.20NixOS/near/229701863), at least for Nix, would be to build Viper from source using a separate Nix flake in its repository, which I started working on as well (will reference the relevant pull request when it's ready). It would then be possible to pin the dependency to a specific commit (done automatically by `flake.lock`, in fact), which can be updated as necessary by running `nix flake update`. The only caveat with this approach, as of now, would be the added time required to compile Viper in addition to Prusti. This should be resolved once a cache is hooked up, where it would be possible to have a workflow automatically compiling binaries of both Prusti and Viper on each commit to master, which will be made available to users by the Nix caching system.

vakaras (approved):
LGTM Thanks!

@fpoli Unless you object, I will merge this in.

PR #466, opened 2021-04-21
Add arrays to TypeVisitor

dario23:


dario23:
doing this as it's a self-contained single step towards array support, so as not to accumulate too many changes in a monster PR. hope this simplifies reviewing as well.

fpoli:
From the `test-crates`:
```
thread 'rustc' panicked at 'not implemented: Array(u8, Const { ty: usize, val: Value(Scalar(0x0000000000000010)) })', prusti-viper/src/encoder/memory_eq_encoder.rs:175:22
```

You could add a case to `memory_eq_encoder.rs:175` to report that memory equality between arrays is unsupported.

(@Aurel300 is this code that is going to be removed in the near future?)

fpoli:
> doing this as it's a self-contained single step towards array support, so as not to accumulate too many changes in a monster PR. hope this simplifies reviewing as well.

Sure. Choose whatever is better for you, don't worry about how we can review the code. Having multiple PRs encourages having multiple feedback iterations, which is good for quality but might slow you down a bit (if you ever find yourself waiting on a PR to get merged).

dario23:
i personally like many small PRs much more, so i'll stick to that i guess :-)

fpoli:
It seems fine to move `tests/verify/fail/unsupported/array.rs` to a `tests/verify/pass/arrays` folder.

Aurel300:
> From the `test-crates`:
> 
> ```
> thread 'rustc' panicked at 'not implemented: Array(u8, Const { ty: usize, val: Value(Scalar(0x0000000000000010)) })', prusti-viper/src/encoder/memory_eq_encoder.rs:175:22
> ```
> 
> You could add a case to `memory_eq_encoder.rs:175` to report that memory equality between arrays is unsupported.
> 
> (@Aurel300 is this code that is going to be removed in the near future?)

Yes, `memory_eq_encoder` will be removed relatively soon.

dario23:
> Yes, memory_eq_encoder will be removed relatively soon.

adding such a case anyway, so the tests pass for now. i guess removing it all is not really affected by that addition :-)

PR #467, opened 2021-04-21
Work around AssertKind::description panicing for BoundsCheck

dario23:
AssertKind::BoundsCheck panics on calling AssertKind::display, and the
docs say that the caller is supposed to handle those earlier[0]. So we
use fmt_assert_args instead, which can handle BoundsCheck.

[0] https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/enum.AssertKind.html#impl

### DIFF
--- a/prusti-viper/src/encoder/pure_function_encoder.rs
+++ b/prusti-viper/src/encoder/pure_function_encoder.rs
@@ -932,7 +932,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                 let assert_msg = if let mir::AssertKind::BoundsCheck{ .. } = msg {
                     // Use the debug impl for BoundsCheck, as it is supposed to be handled before
                     // calling display() according to the docs
-                    // TODO: use fmt_assert_args once #BUGNR is merged
+                    // TODO: use fmt_assert_args once https://github.com/rust-lang/rust/pull/84392 is merged
                     format!("{:?}", msg)
                 } else {
                     msg.description().to_string()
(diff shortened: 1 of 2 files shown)


Aurel300:
Oh it's not public… :/

fpoli:
I guess the easiest thing is to create a file in `prusti_interface` (with an appropriate rustc licence header and link to the rustc source) in which we copy-paste the implementation of [`fmt_assert_args`](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_middle/mir/mod.rs.html#1322-1376).

(We could even propose to make that method public with a tiny PR to rustc.)

dario23:
hm, and the only use seems to be in the [mir printing](https://github1s.com/rust-lang/rust/blob/HEAD/compiler/rustc_middle/src/mir/terminator.rs).

i like the idea of doing a rustc PR, but would go with the debug impl (maybe just for the BoundsCheck variant) instead for now, and add a TODO for when/if our rustc-PR is accepted.

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -929,9 +929,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                     vir::Expr::not(cond_val)
                 };
 
+                let assert_msg = if let mir::AssertKind::BoundsCheck{ .. } = msg {
+                    // Use the debug impl for BoundsCheck, as it is supposed to be handled before
+                    // calling display() according to the docs
+                    // TODO: use fmt_assert_args once #BUGNR is merged
```
dario23: will insert the rustc PR number here once i have it.

dario23: alright, i think we could merge this as is for now. alternatively, we could wait ~1 day and hope the rustc side reacts quickly.

fpoli: Let's merge now

PR #470, opened 2021-04-22
Introduce PlaceEncoding machinery

dario23:
This will be used for encoding array accesses. This PR includes just the machinery, no actual array encoding, but makes sure to keep all code working while using the new type.

The reasoning for the `PlaceEncoding` type is
 - we encode array code using an abstract predicate like `Array$4$i32(self: Ref)` and specify behaviour using an abstract pure function like `Array$4$i32$lookup_pure(self: Ref, index: Int): Int`. This way we don't always have a `vir::Expr` to return like for field accesses.
 - We need to encode array accesses differently for pure functions/procedures, so the `MirEncoder` returns a `PlaceEncoding::ArrayAccess` for them, and `PureFunctionEncoder` and `ProcedureEncoder` handle them accordingly
 - `PlaceEncoding::Field` and `PlaceEncoding::Variant` exist just like for `vir::Expr` with the same intent. As MIR handles deref, array indexing, field access all as a list of projections, we need to be able to mix them, so we duplicate them in `PlaceEncoding` and map them back to the respective `vir::Expr` variants when we translate the ArrayAccess.

### DIFF
--- a/prusti-viper/src/encoder/mir_encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir_encoder/mod.rs
@@ -58,6 +58,12 @@ impl PlaceEncoding {
     pub fn try_into_expr(self) -> EncodingResult<vir::Expr> {
         match self {
             PlaceEncoding::Expr(e) => Ok(e),
+            PlaceEncoding::FieldAccess { base, field } => {
+                Ok(base.try_into_expr()?.field(field))
+            },
+            PlaceEncoding::Variant { base, field } => {
+                Ok(vir::Expr::Variant(box base.try_into_expr()?, field, vir::Position::default()))
+            },
             other => Err(EncodingError::internal(format!("PlaceEncoding::try_into_expr called on non-expr '{:?}'", other))),
         }
     }
--- a/prusti-viper/src/encoder/mir_encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir_encoder/mod.rs
@@ -5,6 +5,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 mod downcast_detector;
+mod place_encoding;
 
 use crate::encoder::builtin_encoder::BuiltinFunctionKind;
 use crate::encoder::errors::{
@@ -20,92 +21,14 @@ use rustc_middle::{mir, ty};
 use rustc_index::vec::{Idx, IndexVec};
 use rustc_span::{Span, DUMMY_SP};
 use log::{trace, debug};
-use std::{
-    collections::HashMap,
-    fmt::Display,
-};
+use std::collections::HashMap;
 use prusti_interface::environment::mir_utils::MirPlace;
+
 use downcast_detector::detect_downcasts;
+pub use place_encoding::PlaceEncoding;
 
 pub static PRECONDITION_LABEL: &'static str = "pre";
 pub static WAND_LHS_LABEL: &'static str = "lhs";
-
-/// Result of encoding a place
-#[derive(Debug, PartialEq, Eq, Hash, Clone)]
-pub enum PlaceEncoding {
-    /// Just an expression, the most common case.
-    Expr(vir::Expr),
-    /// Field access expression
-    FieldAccess {
-        base: Box<PlaceEncoding>,
-        field: vir::Field,
-    },
-    /// Array access expression
-    ArrayAccess {
-        base: Box<PlaceEncoding>,
-        index: vir::Expr,
-        // TODO: prusti-common::src::vir::optimizations::purification has fn translate_type
-        array_elem_ty: vir::Type,
-        array_len: usize,
-    },
-    Variant {
-        base: Box<PlaceEncoding>,
-        field: vir::Field,
-    }
-}
-
-impl PlaceEncoding {
-    pub fn try_into_expr(self) -> EncodingResult<vir::Expr> {
-        match self {
-            PlaceEncoding::Expr(e) => Ok(e),
-            PlaceEncoding::FieldAccess { base, field } => {
-                Ok(base.try_into_expr()?.field(field))
-            },
-            PlaceEncoding::Variant { base, field } => {
-                Ok(vir::Expr::Variant(box base.try_into_expr()?, field, vir::Position::default()))
-            },
-            other => Err(EncodingError::internal(format!("PlaceEncoding::try_into_expr called on non-expr '{:?}'", other))),
-        }
-    }
-
-    pub fn field(self, field: vir::Field) -> Self {
-        PlaceEncoding::FieldAccess { base: box self, field }
-    }
-
-    pub fn get_type(&self) -> &vir::Type {
-        match self {
-            PlaceEncoding::Expr(ref e) => e.get_type(),
-            PlaceEncoding::FieldAccess { ref field, .. } => &field.typ,
-            PlaceEncoding::ArrayAccess { ref array_elem_ty, .. } => array_elem_ty,
-            PlaceEncoding::Variant { ref base, ref field } => &field.typ,
-        }
-    }
-
-    pub fn variant(self, index: &str) -> Self {
-        // TODO: somewhat duplicate from vir::Expr::variant()
-        let field_name = format!("enum_{}", index);
-        let field = vir::Field::new(field_name, self.get_type().clone().variant(index));
-
-        PlaceEncoding::Variant { base: box self, field }
-    }
-}
-
-impl From<vir::Expr> for PlaceEncoding {
-    fn from(e: vir::Expr) -> Self {
-        PlaceEncoding::Expr(e)
-    }
-}
-
-impl Display for PlaceEncoding {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            PlaceEncoding::Expr(e) => write!(f, "{}", e),
-            PlaceEncoding::FieldAccess { base, field } => write!(f, "{}.{}", base, field),
-            PlaceEncoding::ArrayAccess { base, index, .. } => write!(f, "{}[{}]", base, index),
-            PlaceEncoding::Variant { base, field } => write!(f, "{}[{}]", base, field),
-        }
-    }
-}
 
 pub trait PlaceEncoder<'v, 'tcx: 'v> {
 
--- a/prusti-viper/src/encoder/mir_encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir_encoder/mod.rs
@@ -20,12 +20,86 @@ use rustc_middle::{mir, ty};
 use rustc_index::vec::{Idx, IndexVec};
 use rustc_span::{Span, DUMMY_SP};
 use log::{trace, debug};
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    fmt::Display,
+};
 use prusti_interface::environment::mir_utils::MirPlace;
 use downcast_detector::detect_downcasts;
 
 pub static PRECONDITION_LABEL: &'static str = "pre";
 pub static WAND_LHS_LABEL: &'static str = "lhs";
+
+/// Result of encoding a place
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum PlaceEncoding {
+    /// Just an expression, the most common case.
+    Expr(vir::Expr),
+    /// Field access expression
+    FieldAccess {
+        base: Box<PlaceEncoding>,
+        field: vir::Field,
+    },
+    /// Array access expression
+    ArrayAccess {
+        base: Box<PlaceEncoding>,
+        index: vir::Expr,
+        // TODO: prusti-common::src::vir::optimizations::purification has fn translate_type
+        array_elem_ty: vir::Type,
+        array_len: usize,
+    },
+    Variant {
+        base: Box<PlaceEncoding>,
+        field: vir::Field,
+    }
+}
+
+impl PlaceEncoding {
+    pub fn try_into_expr(self) -> EncodingResult<vir::Expr> {
+        match self {
+            PlaceEncoding::Expr(e) => Ok(e),
+            other => Err(EncodingError::internal(format!("PlaceEncoding::try_into_expr called on non-expr '{:?}'", other))),
+        }
+    }
+
+    pub fn field(self, field: vir::Field) -> Self {
+        PlaceEncoding::FieldAccess { base: box self, field }
+    }
+
+    pub fn get_type(&self) -> &vir::Type {
+        match self {
+            PlaceEncoding::Expr(ref e) => e.get_type(),
+            PlaceEncoding::FieldAccess { ref field, .. } => &field.typ,
+            PlaceEncoding::ArrayAccess { ref array_elem_ty, .. } => array_elem_ty,
+            PlaceEncoding::Variant { ref base, ref field } => &field.typ,
+        }
+    }
+
+    pub fn variant(self, index: &str) -> Self {
+        // TODO: somewhat duplicate from vir::Expr::variant()
+        let field_name = format!("enum_{}", index);
+        let field = vir::Field::new(field_name, self.get_type().clone().variant(index));
+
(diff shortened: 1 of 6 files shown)


prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -20,13 +20,93 @@ use rustc_middle::{mir, ty};
...
+    }
+}
+
+impl Display for PlaceEncoding {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PlaceEncoding::Expr(e) => write!(f, "{}", e),
+            PlaceEncoding::FieldAccess { base, field } => write!(f, "{}.{}", base, field),
+            PlaceEncoding::ArrayAccess { base, index, .. } => write!(f, "{}[{}]", base, index),
+            PlaceEncoding::Variant { base, field } => write!(f, "{}[{}]", base, field),
```
fpoli: What about `"({} as {})"` to distinguish `Variant` from `ArrayAccess`?

prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -20,13 +20,93 @@ use rustc_middle::{mir, ty};
...
+};
 use prusti_interface::environment::mir_utils::MirPlace;
 use downcast_detector::detect_downcasts;
 
 pub static PRECONDITION_LABEL: &'static str = "pre";
 pub static WAND_LHS_LABEL: &'static str = "lhs";
 
+/// Result of encoding a place
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum PlaceEncoding {
```
fpoli: Could you move `PlaceEncoding` and its `impl` to a new `mir_encoder/place_encoding.rs` file?

prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -205,15 +285,16 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
 
     fn encode_deref(
         &self,
-        encoded_base: vir::Expr,
+        encoded_base: PlaceEncoding,
         base_ty: ty::Ty<'tcx>,
-    ) -> EncodingResult<(vir::Expr, ty::Ty<'tcx>, Option<usize>)> {
+    ) -> EncodingResult<(PlaceEncoding, ty::Ty<'tcx>, Option<usize>)> {
```
fpoli: I find a bit odd to require a `PlaceEncoding` and then immediately call `try_into_expr`. Isn't it better to keep `vir::Expr`? All calls to this function seem to be `.encode_deref(PlaceEncoding::Expr(..), ..)`.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5374,6 +5411,62 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+    fn encode_deref(
+        &self,
+        encoded_base: PlaceEncoding,
+        base_ty: ty::Ty<'tcx>,
+    ) -> EncodingResult<(vir::Expr, Vec<vir::Stmt>, ty::Ty<'tcx>, Option<usize>)> {
+        let (encoded_place, ty, variant_idx) = self.mir_encoder.encode_deref(encoded_base, base_ty)?;
+        let (encoded_expr, encoding_stmts) = self.postprocess_place_encoding(encoded_place)?;
+
+        Ok((encoded_expr, encoding_stmts, ty, variant_idx))
+    }
```
fpoli: Just looking at the code in this PR, I think that this function is not needed. The only usage seems to be `self.encode_deref(PlaceEncoding::Expr(..), ..)` and `MirEncoder::encode_deref` only returns `PlaceEncoding::Expr(..)`. Do you foresee some new use cases?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1287,9 +1288,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
     fn encode_loan_places(&self, loan_places: &LoanPlaces<'tcx>) -> (vir::Expr, vir::Expr, bool) {
         debug!("encode_loan_places '{:?}'", loan_places);
         // will panic if attempting to encode unsupported type
-        let (expiring_base, expiring_ty, _) = self.mir_encoder.encode_place(&loan_places.dest).unwrap();
+        let (expiring_base, pre_stmts, expiring_ty, _) = self.encode_place(&loan_places.dest).unwrap();
+        assert!(pre_stmts.is_empty(), "Unexpected encoding pre-statements: {:?}", pre_stmts);
+
         let encode = |rhs_place| {
-            let (restored, _, _) = self.mir_encoder.encode_place(rhs_place).unwrap();
+            // TODO: pre_stmts here as well..
```
fpoli, line 1295: What does this comment mean?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1287,9 +1288,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            let (restored, pre_stmts, _, _) = self.encode_place(rhs_place).unwrap();
```
fpoli, line 1296: Adding `assert!(pre_stmts.is_empty(), "TODO");` as you did in other places would be very good.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3666,7 +3690,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 if self.mir_encoder.is_reference(arg_ty) {
                     let encoded_arg = self.mir_encoder.encode_local(arg_index)?;
                     let (deref_place, ..) =
```
fpoli, line 3692: This is ignoring a `pre_stms` that you could assert to be empty.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3851,20 +3877,19 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
         let ty = self.locals.get_type(contract.returned_value);
         let encoded_return: vir::Expr = self.encode_prusti_local(contract.returned_value).into();
+        let return_span = self.mir_encoder.get_local_span(contract.returned_value.into());
         let encoded_return_expr = if self.mir_encoder.is_reference(ty) {
-            let return_span = self.mir_encoder.get_local_span(
-                contract.returned_value.into()
-            );
-            let (encoded_deref, ..) = self.mir_encoder
-                .encode_deref(encoded_return, ty)
+            let (encoded_deref, pre_stmts, ..) = self.mir_encoder
```
fpoli: Assert that `pre_stmts` is empty?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4043,7 +4068,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                     continue;
                 }
                 // will panic if attempting to encode unsupported type
-                let (encoded_place, ty, _) = self.mir_encoder.encode_place(&mir_place).unwrap();
+                let (encoded_place, pre_stmts, ty, _) = self.encode_place(&mir_place).unwrap();
+                // TODO: pre_stmts
```
fpoli: Assert that `pre_stmts` is empty?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4074,7 +4100,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                         if let Some(base) = utils::try_pop_deref(self.encoder.env().tcx(), mir_place)
                         {
                             // will panic if attempting to encode unsupported type
-                            let (_, ref_ty, _) = self.mir_encoder.encode_place(&base).unwrap();
+                            let (_, _, ref_ty, _) = self.encode_place(&base).unwrap();
+                            // TODO: pre_stmts
```
fpoli: Assert that `pre_stmts` is empty?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4135,8 +4162,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                                         // that we will lose information about the children of that
                                         // place after the loop and we need to preserve it via local
                                         // variables.
-                                        let (encoded_child, _, _) =
-                                            self.mir_encoder.encode_place(&child_place).unwrap(); // will panic if attempting to encode unsupported type
+                                        let (encoded_child, pre_stmts, _, _) =
+                                            self.encode_place(&child_place).unwrap(); // will panic if attempting to encode unsupported type
+                                        // TODO: pre_stmts
```
fpoli: Assert that `pre_stmts` is empty?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5374,6 +5411,62 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        place: &mir::Place<'tcx>,
+    ) -> EncodingResult<(vir::Expr, Vec<vir::Stmt>, ty::Ty<'tcx>, Option<usize>)> {
+        let (encoded_place, ty, variant_idx) = self.mir_encoder.encode_place(place)?;
+        // TODO: actual encoding of e.g. array access here
+        let (encoded_expr, encoding_stmts) = self.postprocess_place_encoding(encoded_place)?;
+
+        Ok((encoded_expr, encoding_stmts, ty, variant_idx))
+    }
+
+    fn encode_projection(
```
fpoli, line 5426: Just looking at the code in this PR, it seems that this function is not needed.

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -491,6 +491,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionBackwardInterpreter<'p, 'v, 'tcx> {
                     place.local,
                     &place.projection[..],
                 ).with_span(span)?;
+                let encoded_place = encoded_place.try_into_expr().with_span(span)?;
```
fpoli: `try_into_expr` is fine for now, but keep in mind that you'll need to implement for `PureFunctionEncoder` the analogous version of `ProcedureEncoder::postprocess_place_encoding` in order to encode array operations in pure functions and specifications.

prusti-viper/src/encoder/mir_encoder/mod.rs
dario23: so somewhat like the mir printing does? this is from the encoding of `prusti-tests/tests/verify/pass/enums/basic.rs` (file `log/vir_program_before_foldunfold/basic.rs.vir`, line 386 here):
```
    // [mir] _6 = &'_ mut ((_2 as V2).0: u32)
    _6 := builtin$havoc_ref()
    inhale acc(_6.val_ref, write)
    _6.val_ref := mut borrow _2[enum_V2].f$0 // L0
```

so i'd probably make a change to make the last line `_6.val_ref := mut borrow (_2 as enum_V2).f$0` instead? that log/ stuff is never read back anywhere, right? so i'd only have to change what is printed, no parser or anything else?

i'd do that in a separate PR probably, this one feels heavy already, and i quite like the "small PRs, easier/quicker reviews, faster small-step progress" approach.

prusti-viper/src/encoder/mir_encoder/mod.rs
dario23: well, it's called from encode_projection (because `Deref` is a `ProjectionElem` variant), so it seemed to make sense to be able to handle `PlaceEncoding`s there as well, so for
```
fn deref(x: [&mut u32; 2]) {
    *x[0] = 42;
}
```
we would have that combination.

on the other hand, the implementation is not there yet. i think in general there needs to be more handling in some of the places that `try_into_expr` is currently used to be able to support the full set of nesting possibilities.

on the other hand, we might go with the smaller set of changes here and leave this as-is, returning `vir::Expr`, and only change once we actually need it. that way we avoid the wrapping and unwrapping back and forth and are also clearer about the (current) actual requirements/expectations of what the method does.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: that i should be adding an assert here as well. will do together with all the other places it's missing.

prusti-viper/src/encoder/pure_function_encoder.rs
dario23: hm, i guess i should do at least the same wrapping with a local version of `encode_place` and introduce a `postprocess_place_encoding` function in `PureFunctionEncoder` as well; the current way with just `try_into_expr` sprinkled everywhere doesn't seem to be clean nor meant to stay fwiw.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: `encode_projection` is called in `ProcedureEncoder::encode_statement_at`, which has a `prusti_interface::environment::mir_utils::MirPlace` instead of a `rustc_middle::mir::Place`, so we can't call `encode_place` here, nor do i see how to recover/rebuild a `mir::Place` again from that.

prusti-viper/src/encoder/mir_encoder/mod.rs
fpoli: I didn't notice that `field` is a string and not a number, so it should be difficult to confuse the two variants. It's fine as you did.

prusti-viper/src/encoder/mir_encoder/mod.rs
fpoli: I see, I missed the call from `encode_projection`. For the following case, however, wouldn't Prusti generate an internal error instead of an unsupported error because of the `try_into_expr`?
```
fn deref(x: [&mut u32; 2]) {
    *x[0] = 42;
}
```
It might be easier to keep the signature of `encode_deref` as it was and do the conversion from `PlaceEncoding` on the call site. That way, you can generate an "unsupported error" in `encode_projection`.

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: Isn't `ProcedureEncoder::encode_statement_at` calling `MirEncoder` with `self.mir_encoder.encode_projection`? I can't find a call of this `encode_projection`, but I might be missing something because I'm just doing textual search on the diff.

prusti-viper/src/encoder/mir_encoder/mod.rs
dario23: i'll rework this then, so that `encode_place` and `encode_projection` are the only two that return `PlaceEncoding`, and mixed/complicated cases like `encode_deref` will return `EncodingError::unsupported` for now

prusti-viper/src/encoder/procedure_encoder.rs
dario23: removed it, as discussed in the previous comment thread.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: uhm.. that should be [part of my changes](https://github.com/viperproject/prusti-dev/pull/470/files#diff-704bbc6b705baff9ae5a91e768c083923396f2cfd663646531528a9494384a23R1056). didn't show up when text-searching in the diff first, because the diff in `procedure_encoder.rs` was too big so github hid it by default :-/

fpoli: Yes, found it, sorry.

PR #475, opened 2021-04-23
Implement obtaining array length

dario23:


fpoli:
Is this something that you can test with e.g. `fn foo(x: [u32; 4]) { assert!(x.len() == 4); }`?

dario23:
NOTE: tried to add a testcase like this
```
#[ensures(result == 4)]
fn test_len(arr: [i32; 4]) -> usize {
  arr.len()
}
```
but the `len` method is not a method of type array but of type slice, so rustc adds an implicit cast to slice, which is currently not supported (triggers [this error](https://github.com/viperproject/prusti-dev/blob/master/prusti-interface/src/environment/polonius_info.rs#L598), unsize-casting from [i32; 4] to &[i32])

EDIT: haha, beat me to the comment by some seconds :-)

fpoli:
I'll merge after CI finishes.

dario23:
fails on test-crates now that we 1) don't back out of anything array-related early enough anymore from e.g. array length reading (which happens before most things because bounds checks) but 2) still just panic in `postprocess_place_encoding` when anything array-related comes up.. i'll return an `EncodingError::unsupported` for now i guess.

fpoli:
Yes, raising `EncodingError::unsupported` should be enough to silence the test-crates.

fpoli:
The invariant checked by test-crates is that there should be no programs on which Prusti crashes or stops with an internal error. Only user-readable "unsupported feature" errors should be allowed.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5497,7 +5497,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 (expr.field(field), stmts)
             }
             PlaceEncoding::ArrayAccess { base, index, array_elem_ty, array_len } => {
-                panic!("ArrayAccess {{ {:?}, {:?}, {:?}, {:?} }}", base, index, array_elem_ty, array_len);
+                return Err(EncodingError::unsupported(
+                    "postprocessing array access not implemented yet"
```
fpoli: "array operations are not supported yet" is more user friendly. Some people might use the auto-deployed version of Prusti and see this error message before you finish implementing support for arrays.

fpoli:
Something strange is going on. GitHub keep claiming that you are a first-time contributor, while you are clearly not because this is not your first PR. Due to this glitch I have to manually approve every run of CI :thinking:

dario23:
yep, noticed this starting this morning; thought maybe some policy/project setting had changed..

fpoli:
It's https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

Aurel300:
I think the first-time contributor thing is because the e-mail you use when committing (if any) does not match any e-mail you have configured here on GitHub.

dario23:
hm, tried to explicitly commit that last commit using the email address github has stored, as well as adding the other one to my github account, but still get the "requires approval". maybe that only works for the first commit/when creating a PR? or maybe it's something different still.

dario23:
also i don't (yet? some github help page mentioned it might take up to 24h to refresh) show up on [prusti's insights → contributors page](https://github.com/viperproject/prusti-dev/graphs/contributors), but my commits show up in my account as activity/contributions. i guess i'll wait another day and see if anything changes, and/or maybe see if the next PR will change anything.

Aurel300:
The last commit (2173140) shows as coming from your GitHub account, which is good. It was now *actually* your first commit (with this account), so after the approval you should be good for later PRs.

PR #476, opened 2021-04-23
Array access projection

dario23:


dario23:
based on #475 (and incorporating the work in progress changes from there).

sample that i'm testing on:
```rust=
use prusti_contracts::*;

fn main() {}


pub fn first(a: [i32; 4]) -> i32 {
    a[0]
}
```

what works (taken from `log/vir_program_before_foldunfold/`):
```
// [mir] _0 = _1[_2]
_aux_lookup_ret_int := builtin$havoc_ref()
inhale (Array$4$i32$lookup_pure<,Int>(_1, _2)) == (_aux_lookup_ret_int)
_0 := builtin$havoc_ref()
inhale acc(_0.val_int, write)
_0.val_int := copy _aux_lookup_ret_int.val_int
```

what doesn't:
```
error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Acc(Field(Local(_aux_lookup_ret_int: Int, Position { line: 7, column: 5, id: 51 }), val_int: Int, Position { line: 7, column: 5, id: 51 }), read)))
 --> a1.rs:6:1
  |
6 | / pub fn first(a: [i32; 4]) -> i32 {
7 | |     a[0]
8 | | }
  | |_^
```

dario23:
now that i'm looking at this again with a set of somewhat fresher eyes, i think that `_aux_lookup_ret_int` is treated as ref (i.e. i32) on the first line and like a viper `Int` on the second line in the lookup equality, which doesn't match, and so not being able to access the val_int field on an `Int` makes complete sense

dario23:
ok, sample code passes now, still not entirely happy with the code.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,7 +5496,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                    BuiltinFunctionKind::ArrayLookupPure {
+                        array_elem_ty: array_elem_ty.clone(),
+                        array_len,
+                        return_ty: lookup_pure_ret.clone(),
+                    }
+                );
+
+                let array_type = base.get_type().clone();
+                let lookup_res: vir::Expr = self.get_auxiliary_local_var("lookup_ret", array_elem_ty).into();
+                let lookup_res_val_int = lookup_res.clone().field(vir::Field::new("val_int", vir::Type::Int));
```
dario23: this seems more manual than it needs to be, but i currently don't see how to do this better. it boils down to

 - the temporary variable seems to have to be type `Ref`, because the assignment to the final variable generated wants to copy from `.val_int`
 - that seems to mean that we need to pass `.val_int` to `lookup_pure`

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5477,6 +5540,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
     }
 }
 
+// TODO: is there a better way? or even place to put this? type_encoder?
+fn rust_type_name_to_vir_type(rust_type_name: &str) -> EncodingResult<vir::Type> {
```
dario23: i'm really unhappy with this function, and still think i'd rather add the return type of lookup_pure to the `PlaceEncoding::ArrayAccess` variant, as when constructing that we still have access to the rust type, which allows simple generation of both, where now we have to do these name-based hacks..

dario23: so this would be what i'd prefer: https://github.com/dario23/prusti-dev/compare/array-access-projection...dario23:lookup-pure-ret?expand=1

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: Yes, as long as pure functions don't use snapshots for the parameters we have to add the `.val_int`. The field creation with the hardcoded string can be replaced with a call to `Encoder::encode_value_field`.

fpoli: The temporary variable is of Viper type `Ref` because of `array_elem_ty`. By default we encode primitive types as heap-allocated values to simplify the encoding of creation of references. One of the future optimizations  (purification) should optimize cases like this.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,7 +5496,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                        array_elem_ty: array_elem_ty.clone(),
+                        array_len,
+                        return_ty: lookup_pure_ret.clone(),
+                    }
+                );
+
+                let array_type = base.get_type().clone();
+                let lookup_res: vir::Expr = self.get_auxiliary_local_var("lookup_ret", array_elem_ty).into();
+                let lookup_res_val_int = lookup_res.clone().field(vir::Field::new("val_int", vir::Type::Int));
+                let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(*base)?;
```
fpoli, line 5520: I think there is a problem when you have multiple nested `ArrayAccess`, because the `get_auxiliary_local_var` call two lines above is designed to reuse generated variables but you always want a fresh one. Also, when you encode an expression you don't know for sure what's the last time that the expression will be used; it's possible that `postprocess_place_encoding` will be called again perfore that and (incorrectly) reuse the same temporary variable. You could implement something like a new `ProcedureEncoder::generate_fresh_local_var(..)` that uses an internal counter to generate new variable names.

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: I would also prefer to add one more field to avoid this function.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5496,10 +5504,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                );
+
+                let array_type = base.get_type().clone();
+                let lookup_res: vir::Expr = self.cfg_method.add_fresh_local_var(array_elem_ty).into();
+                let lookup_res_val_int = lookup_res.clone().field(val_field.clone());
+                let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(*base)?;
+                stmts.extend(self.encode_havoc_and_allocation(&lookup_res));
+                let lookup_pure_call = vir::Expr::func_app(
+                    lookup_pure,
+                    vec![encoded_base_expr, index.field(val_field)],
```
dario23: NOTE to self: just val_int here, index should always be `usize`

fpoli:
Ready for review

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5496,10 +5504,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                );
+
+                let array_type = base.get_type().clone();
+                let lookup_res: vir::Expr = self.cfg_method.add_fresh_local_var(array_elem_ty).into();
+                let lookup_res_val_int = lookup_res.clone().field(val_field.clone());
+                let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(*base)?;
+                stmts.extend(self.encode_havoc_and_allocation(&lookup_res));
+                let lookup_pure_call = vir::Expr::func_app(
+                    lookup_pure,
+                    vec![encoded_base_expr, index.field(vir::Field::new("val_int", vir::Type::Int))],
```
fpoli, line 5524: In case you need to build thinks like `index.field(vir::Field::new("val_int", vir::Type::Int))` again, `Encoder::encode_value_expr(.., tcx.mk_ty(..))` might be slightly better. (No need to change this PR, snapshots are going to change this anyway.)

fpoli (approved):
LGTM

PR #477, opened 2021-04-24
Add force_matches! macro to simplify if let .. else { unreachable!() } constructs

dario23:


prusti-common/src/vir/optimizations/purification/mod.rs
```diff
@@ -151,26 +152,23 @@ impl<'a> StmtFolder for Purifier<'a> {
...
-                                });
-                            }
-                            "i32" | "usize" | "u32" => {
-                                source = source.field(Field {
-                                    name: "val_int".into(),
-                                    typ: Type::Int,
-                                });
-                            }
-                            x => unreachable!("{}", x),
+                force_matches!(source.get_type(), Type::TypedRef(name) => {
```
fpoli: I would use the `let source_name = force_matches!(..); match source_name.as_str() { .. }` to avoid adding the indentation level.

Cargo.lock
```diff
@@ -1752,6 +1752,7 @@ name = "prusti-specs"
 version = "0.1.0"
 dependencies = [
  "proc-macro2",
+ "prusti-common",
```
fpoli: `prusti-common` dynamically links to the JVM, while `prusti-specs` does not because it's a dependency of procedural macro crates. I don't think we have a crate yet in which to share code between the parser and the rest of Prusti. We could create one (with a good name).

dario23: hm.. prusti-shared? prusti-lib? prusti-basics?

fpoli: @viperproject/prusti-dev suggestions?

fpoli: I'm looking for examples in other projects. `rustc` has `rustc_ty_utils` and `rustc_fs_utils`, so we could mimic them with a `prusti-utils`. The Readme should specify in 1 line that the crate is meant to only contain utilities that don't depend on rustc nor the JVM.

fpoli: Although the best refactoring might be to make the `prusti-common` not depend on the JVM. But that's for the future.

fpoli:
Low priority

PR #482, opened 2021-04-26
WIP: translating array accesses in pure functions

dario23:


dario23:
doing a similar translation as in procedure_encoder, where we wrap encode_place to postprocess `PlaceEncoding::ArrayAccess`.

current problem: we need to return a place, but a `vir::Expr::FuncApp` (i.e. `Array$4$i32$lookup_pure(arr, idx)` is not a place. the rest of the encoding a) happily assumes it has a place and does `.field(val_int)` on it or b) panics because not place.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5496,10 +5517,41 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 let (expr, mut stmts) = self.postprocess_place_encoding(base)?;
                 (expr.field(field), stmts)
             }
-            PlaceEncoding::ArrayAccess { base, index, array_elem_ty, array_len } => {
-                return Err(EncodingError::unsupported(
-                    "array operations are not supported yet"
-                ));
+            PlaceEncoding::ArrayAccess { base, index, array_elem_ty, array_len, lookup_pure_ret, val_field } => {
+                let elem_ty_name = if let vir::Type::TypedRef(ref name) = array_elem_ty { name } else { unreachable!() };
```
fpoli, line 5521: Is this `elem_ty_name` used?

dario23:
superseeded by #505

PR #484, opened 2021-04-27
Implement taking references into arrays

dario23:
This enables code like
```
let b = &a[i];
```
where we don't just read from the array, but remember the reference


fpoli:
Is it possible to add a test (even without specifications or assertions) for this?

fpoli:
Great for the test case! You can check the code coverage of the test suite on Coveralls (e.g. [`procedure_encoder.rs`](https://codecov.io/gh/viperproject/prusti-dev/src/master/prusti-viper/src/encoder/procedure_encoder.rs)). It's measured each time a PR is merged.

PR #489, opened 2021-04-30
Encoding of mutable array indexing

dario23:


dario23:
WIP. current problem: `translate_maybe_borrowed_places` seems to assume that calling `encode_place` again on the same `mir::Place` will yield the exact same encoding, but encoding mutable array access will create new temporary variables (and return one of them), so `expire_borrows` will work on the second encoding of them (which for now i've changed in [`encode_loan_places`](https://github.com/viperproject/prusti-dev/pull/489/files#diff-704bbc6b705baff9ae5a91e768c083923396f2cfd663646531528a9494384a23R1794), but that's not the proper answer).

fpoli:
> WIP. current problem: `translate_maybe_borrowed_places` seems to assume that calling `encode_place` again on the same `mir::Place` will yield the exact same encoding, but encoding mutable array access will create new temporary variables (and return one of them), so `expire_borrows` will work on the second encoding of them (which for now i've changed in [`encode_loan_places`](https://github.com/viperproject/prusti-dev/pull/489/files#diff-704bbc6b705baff9ae5a91e768c083923396f2cfd663646531528a9494384a23R1794), but that's not the proper answer).

I suspect that `translate_maybe_borrowed_place` is an old hack that is no longer needed. From the [test coverage](https://codecov.io/gh/viperproject/prusti-dev/src/master/prusti-viper/src/encoder/procedure_encoder.rs) it seems that this condition always evaluates to `false`, so the `replace_place` is never reached:

https://github.com/viperproject/prusti-dev/blob/b2dc5cf8d0d11f432a9e59238af67e28fc4b0118/prusti-viper/src/encoder/procedure_encoder.rs#L1278

I vote for removing that function (any `translate_maybe_borrowed_place(_, e)` call should become just `e`).

Still, the encoding of this PR has Viper consistency errors. For other reasons, I believe.

fpoli:
> I vote for removing that function (any `translate_maybe_borrowed_place(_, e)` call should become just `e`).

Trying in https://github.com/viperproject/prusti-dev/pull/490

fpoli:
Btw, several `.clone()` are not needed. You can see them with the command
```
git rebase master  # to fix clippy errors
cargo install cargo-diff-tools
./x.py exec cargo-clippy-diff master...HEAD HEAD
```
which only reports clippy warnings for the lines modified by your PR.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,22 +5476,186 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        array_elem_ty: vir::Type,
+        array_len: usize,
+        val_field: vir::Field,
+    ) -> EncodingResult<(vir::Expr, Vec<vir::Stmt>)> {
+        let array_type = base.get_type().clone();
+        let array_type_name = if let vir::Type::TypedRef(ref n) = array_type { n.clone() } else { unreachable!() };
+        let res: vir::Expr = self.cfg_method.add_fresh_local_var(array_elem_ty.clone()).into();
+        let res_val_field = res.clone().field(val_field.clone());
+        let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(base, ArrayAccessKind::Mutable)?;
+        let idx_val_int = index.clone().field(vir::Field::new("val_int", vir::Type::Int));
```
fpoli: `Encoder::encode_value_field` is slightly better than hard-coding `"val_int"`.

Actually, now that I think better at it, it's more serious: this might cause a consistency error if nothing else in Prusti calls `encode_value_field` to generate a `val_int` field. It's rare, but if it happens then the Viper program would not have the`field val_int: Int` declaration that is generated by `Encoder::encode_value_field`.

dario23: agreed, using [`TyCtxt::mk_ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.mk_ty) now and then encoding the freshly-created `usize` type for the `index field`.

dario23:
alright, fixed two issues with consistency checks:

 - had a `vir::Type::TypedRef("Array$3$i32")` that wasn't being "unpacked", so we failed because no predicate of the name "Ref(Array$3$i32)" existed.
 - probably didn't always call enough of the `encode_type_predicate_use` methods, now passing the rust type through the `PlaceEncoding::ArrayAccess` and call the necessary type encoding methods instead of string-formatting shenanigans.

still issues with missing permissions. code as is compiles and generates vir/viper code just fine, if we replace `exhale acc(Array$3$i32(self), write)` with `assert acc(Array$3$i32(self), write)`, but fails to verify the final assertion in this example:
```rust
use prusti_contracts::*;

fn main() {}

fn array_mut(mut a: [i32; 3]) {
    let a0 = a[0];
    let b = &mut a[0];
    assert!(*b == a0);
    *b = 42;
    // b expires here
    assert!(a[0] == 42);
}
```

and looking at it all i'm not sure where the magic wand is applied again, don't see it in the generated viper program (thought expire_borrows should generate something from the vir_before_foldunfold, but it doesn't seem to. is it necessary to package magic wands for them to be auto-applied? we just inhale them currently.

also i somewhat have a different problem with the current implementation, maybe you can comment on that as well. the way we're currently encoding is that taking a (mutable) reference to something is separate from encoding the place that is referenced, i.e. in `encode_place` we don't know that a reference will be taken to the result of the place encoding. so we always return a new (temporary) variable, and all the information we `inhale` about it having the value of a particular `lookup_pure` call and what happens when it expires is done on the temporary variable, not on the actual variable from mir. changing this would require special-casing array accesses in `ProcedureEncoder::encode_assign_ref`. @fpoli @Aurel300 what do you think about that? i feel like you're both leaning towards not making more code so hardcoded/intertwined in procedure encoder..

fpoli:
For future issues, could you attach the Viper file and `log/graphviz_method_before_{viper|foldunfold}/<relevant method>.dot`? It's much quicker to give feedback on the encoding that way.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,67 +5496,226 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        ));
+
+        // exhale Array$4$i32(self)
+        let array_access_pred = vir::Expr::predicate_access_predicate(
+            array_pred.clone(),
+            encoded_base_expr.clone(),
+            vir::PermAmount::Write,
+        );
+
+        stmts.push(vir!{ assert [array_access_pred.clone()] });
```
fpoli: It should be `exhale`, not `assert`.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,67 +5496,226 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        //     ==> Array$3$i32$pure_lookup(self, i)
+        //       == old[lhs](Array$3$i32$pure_lookup(self, i))
+        //   // the given index changed accordingly
+        //   && Array$3$i32$pure_lookup(
+        //     self, unfolding ... in index.val_int) == snap$i32(old[lhs](result))
+
+        // NEEDSWORK: the vir! macro can't do everything we want here..
+        let i_var: vir::Expr = vir_local!{ i: Int }.into();
+
+        let zero_le_i = vir!{ [ vir::Expr::from(0) ] < [ i_var.clone() ] };
```
fpoli: The case `i == 0` should be allowed.

fpoli:
I think we should special case the expiration of a reference that points to an array element. That expiration should just be a magic-wand application statement. Currently, instead of the apply statement the code generates a transfer permission statement, which is wrong (and has the annoying side effect that it triggers the encoding of a `PlaceEncoding`). The relevant code is in `process_expire_borrows.rs`, but I'm not very familiar with it. I'll try to give another look at it later.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5467,67 +5496,226 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        );
+
+        stmts.push(vir::Stmt::Assert(
+            vir!{ [idx_val_int.clone()] < [ vir::Expr::from(array_len) ] },
+            vir::Position::default(),
+        ));
+        stmts.push(vir::Stmt::Assert(
+            vir::Expr::predicate_access_predicate(array_pred, encoded_base_expr, vir::PermAmount::Read),
+            vir::Position::default(),
+        ));
```
fpoli: Why is this assertion generated? It seems redundant to me.

dario23: good catch, thanks, will adjust :-)

prusti-viper/src/encoder/procedure_encoder.rs
dario23: good catch :-)

prusti-viper/src/encoder/procedure_encoder.rs
dario23: yes; this is the change mentioned in my lengthy comment to make it work a bit more, will change back of course.

dario23:
> For future issues, could you attach the Viper file and `log/graphviz_method_before_{viper|foldunfold}/<relevant method>.dot`? It's much quicker to give feedback on the encoding that way.

will do :-)

prusti-viper/src/encoder/procedure_encoder.rs
fpoli: I see, now I understand the exhale/assert in the lengthy comment. Yes, it's probably because the magic wand is never applied.

fpoli:
> also i somewhat have a different problem with the current implementation, maybe you can comment on that as well. the way we're currently encoding is that taking a (mutable) reference to something is separate from encoding the place that is referenced, i.e. in `encode_place` we don't know that a reference will be taken to the result of the place encoding. so we always return a new (temporary) variable, and all the information we `inhale` about it having the value of a particular `lookup_pure` call and what happens when it expires is done on the temporary variable, not on the actual variable from mir.

Mmm, I'm not sure I see the problem. Inhaling stuff using the temporary variables looks fine to me. Could you make a short example (e.g. Rust + (pseudo) Viper) to explain the issue?

>  changing this would require special-casing array accesses in `ProcedureEncoder::encode_assign_ref`. @fpoli @Aurel300 what do you think about that? i feel like you're both leaning towards not making more code so hardcoded/intertwined in procedure encoder..

The expiration needs to be changed (fixed) for sure. How big is this change that you mention? Could you sketch an example in Viper of how the encoding would change?

dario23:
> The expiration needs to be changed (fixed) for sure. How big is this change that you mention? Could you sketch an example in Viper of how the encoding would change?

> Mmm, I'm not sure I see the problem. Inhaling stuff using the temporary variables looks fine to me. Could you make a short example (e.g. Rust + (pseudo) Viper) to explain the issue?

the answer to both of these is the same i think: i'm not suggesting to change the encoding we agreed upon, i'm saying the way i've implemented it in this PR (along the structure  of the existing code) doesn't conform to that encoding, and could cause us problems.

consider this snippet, which encodes taking a mutable reference into an array (taken from [array-ref-mut.rs.m_array_mut.dot.txt](https://github.com/viperproject/prusti-dev/files/6445695/array-ref-mut.rs.m_array_mut.dot.txt), based on [array-ref-mut.rs.txt](https://github.com/viperproject/prusti-dev/files/6445697/array-ref-mut.rs.txt))

```=
    // [mir] _6 = &'_ mut _1[_7]
    __t11 := builtin$havoc_ref()
    inhale acc(i32(__t11), write)
    label l1
    assert (_7.val_int) < (3)
    assert acc(Array$3$i32(_1), write)
    inhale (old[l1](Array$3$i32$lookup_pure<Ref(Array$3$i32), Int,Int>(_1, _7.val_int))) == (__t11.val_int)
    inhale (acc(i32(__t11), write)) Some(L0) --* ((acc(Array$3$i32(_1), write)) && ((forall i: Int  :: (((0) <= (i)) && (((i) < (3)) && (!((i) == (old[l1](_7.val_int)))))) ==> ((Array$3$i32$lookup_pure<Ref(Array$3$i32), Int,Int>(_1, i)) == (old[l1](Array$3$i32$lookup_pure<Ref(Array$3$i32), Int,Int>(_1, i))))) && ((Array$3$i32$lookup_pure<Ref(Array$3$i32), Int,Int>(_1, old[l1](_7.val_int))) == (old[lhs](__t11.val_int)))))
    _6 := builtin$havoc_ref()
    inhale acc(_6.val_ref, write)
    _6.val_ref := mut borrow __t11 // L0
```

This is on the top-level an assignment of a reference, so it gets handled by `ProcedureEncoder::encode_assign_ref`, which at some point [calls `encode_place`](https://github.com/dario23/prusti-dev/blob/array-idx-mut/prusti-viper/src/encoder/procedure_encoder.rs#L4952) to get the encodeing of whatever it is we're referencing, and then goes on to store that in the field corresponding to the kind of reference.

The encoding of the place is what happens in the first part: we create a new temporary variable (`__t11`), and `inhale` what it means that it now has the value at array (`_1`) index `_7.val_int`. Those statements are returned from `encode_place`/`postprocess_place_encoding`. Then `encode_assign_ref` does its part and basically assigns the result into the actual LHS variable `_6`.

I initially that even though we're creating a borrow to a temporary variable, we should be fine, because viper has no concept of expiration and so we won't have problems of "dangling references to temporary variables" or such, like tradinional languages would have here.
**BUT** i now think differently: the magic wand LHS (and all the machinery triggered by that magic wand) refers to the (viper-encoding-)temporary, not the actual variable `_6` rust knows about. in other words, there will never be a point where `__t11` is expired, and therefore no point at which the magic wand will be applied (i'm not sure here, whether prusti calculates much of the expiries itself, or just takes in what rustc/polonius knows).

dario23:
status quo: basic functionality works for mutably referencing into an array, just assigning an array in one statement doesn't generate a loan, so nothing is expired, so doesn't work yet (sample code that doesn't work yet: [array-mut.rs.txt](https://github.com/viperproject/prusti-dev/files/6489786/array-mut.rs.txt))

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4574,6 +4607,89 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        let i_lt_len = vir!{ [ i_var ] < [ vir::Expr::from(at.array_len) ] };
+        let i_ne_idx = vir!{ [ i_var ] != [ old(idx_val_int.clone()) ] };
+        let idx_conditions = vir!{ [zero_le_i] && ([i_lt_len] && [i_ne_idx]) };
+        let lookup_same_as_old = vir!{
+            [at.encode_lookup_pure_call(encoded_array.clone(), i_var.clone())]
+                ==
+            [old(at.encode_lookup_pure_call(encoded_array.clone(), i_var))]
+        };
+        let forall_body = vir!{ [idx_conditions] ==> [lookup_same_as_old] };
+        let all_others_unchanged = vir!{ forall i: Int :: { /* TODO */ } [ forall_body ] };
```
Aurel300: I'm guessing `lookup_pure(array, i)` should be a good enough trigger here?

prusti-common/src/vir/borrows.rs
```diff
@@ -80,11 +80,11 @@ impl Node {
...
-impl fmt::Debug for Node {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self.borrow)
-    }
-}
+// impl fmt::Debug for Node {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//         write!(f, "{:?}", self.borrow)
+//     }
+// }
```
Aurel300: Delete completely or remove the `Debug` derive if you don't need it.

prusti-common/src/vir/to_viper.rs
```diff
@@ -113,7 +113,7 @@ impl<'v> ToViper<'v, viper::Stmt<'v>> for Stmt {
                 ast.inhale(expr.to_viper(ast), fake_position.to_viper(ast))
             }
             Stmt::Exhale(ref expr, ref pos) => {
-                assert!(!pos.is_default());
+                // assert!(!pos.is_default(), "stmt with default pos: {}", self);
```
Aurel300: Do you have statements introduced via `vir!` that don't get their position reassigned to something sensible?

prusti-tests/tests/verify/pass/arrays/array-mut-loop.rs
```diff
@@ -0,0 +1,19 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+fn looping() {
+    let mut a = [0; 3];
+
+    let mut i = 0;
+    let mut cont = true;
+    while cont {
```
Aurel300: Why this strange `cont` boolean instead of `i < 3`?

prusti-tests/tests/verify/pass/arrays/array-mut.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let mut a = [1, 2, 3];
+    a[1] = 23;
+    assert!(a[1] == 23);
```
Aurel300, line 6:
```suggestion
    assert!(a[0] == 1);
    assert!(a[1] == 23);
    assert!(a[2] == 3);
```

prusti-tests/tests/verify/pass/arrays/array-ref-mut.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(nll)]
+use prusti_contracts::*;
+
```
Aurel300:
```suggestion
```

prusti-viper/src/encoder/mir_encoder/place_encoding.rs
```diff
@@ -53,6 +53,26 @@ impl<'tcx> PlaceEncoding<'tcx> {
         }
     }
 
+    /// Ok(e) => the same result try_into_expr would give
+    /// Err(e) => the array base expression
```
Aurel300: Please don't use the error variant for things that are not actually errors. Define a dedicated enum, or see if you can make the uses of this function just do `try_into_array_base`, returning an `Option`, otherwise falling back to `try_into_expr`.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -99,6 +124,8 @@ pub struct ProcedureEncoder<'p, 'v: 'p, 'tcx: 'v> {
     cfg_block_has_been_executed: HashMap<mir::BasicBlock, vir::LocalVar>,
     /// Magic wand generated by a call at a given location with a label used in post.
     magic_wand_at_location: HashMap<mir::Location, (String, vir::Expr, vir::Expr)>,
+    /// magic wands from array accesses
```
Aurel300: Please explain the 4-tuple elements.

prusti-common/src/vir/to_viper.rs
dario23: yes, it seems. is there a reason `vir::Stmt::set_pos` only actually does anything for `Stmt::PackageMagicWand`?

dario23: added `set_default_pos` calls for the statements encoding the expiration of the reference into the array as well, seems to fix it, together with adding `Stmt::Exhale` to the match in `pos`, `set_pos` in `prusti-common/src/vir/ast/stmt.rs`.

prusti-tests/tests/verify/pass/arrays/array-mut-loop.rs
dario23: somehow didn't think that's supported. updated. i can't find the documentation that says what loop conditions are supported/unsupported right now. :-/

prusti-viper/src/encoder/mir_encoder/place_encoding.rs
dario23: tried to keep the duplication with `try_into_expr` in bounds, added a separate `ExprOrArrayBase` enum.

prusti-tests/tests/verify/pass/arrays/array-mut-loop.rs
Aurel300: I think at this point we support basically arbitrary code in the loop guard? We don't support e.g. `for` loops, but IIUC it's because of the iterator there.

prusti-common/src/vir/to_viper.rs
Aurel300: Yeah, that looks a bit strange. I hope that this can only manifest in UI-related tests, if at all.

dario23:
purification-tests failing, but not sure why. the failing testcase passes locally with `PRUSTI_ENABLE_PURIFICATION_OPTIMIZATION=true ./x.py test pass/rosetta/Binary_search_shared.rs`. my guess would be a solver timeout?

fpoli:
> purification-tests failing, but not sure why. the failing testcase passes locally with `PRUSTI_ENABLE_PURIFICATION_OPTIMIZATION=true ./x.py test pass/rosetta/Binary_search_shared.rs`. my guess would be a solver timeout?

Have you tried rebasing on `master`? Update: it still doesn't fail. I tried restarting the CI job.

dario23:
> Update: it still doesn't fail. I tried restarting the CI job.

yes, i'm more and more convinced it's not (only?) a problem with my changes, since the change was just adding a testcase, and the change before (https://github.com/viperproject/prusti-dev/commit/741d433546d71934d4b5af4ef010ffd87619fa34) was changing error messages and first failed, but the force-pushed newer revision https://github.com/viperproject/prusti-dev/commit/43e512fde03b601aeffbee909022c7c5181bea30 succeeded.

dario23:
factored out some stuff not directly related to mutable array indexing in #532, which should be merged first

prusti-common/src/vir/ast/expr.rs
```diff
@@ -1198,7 +1198,7 @@ impl Expr {
         }
         let typaram_substs = match (&target, &replacement) {
             (Expr::Local(tv, _), Expr::Local(rv, _)) => {
-                if tv.typ.is_ref() && rv.typ.is_ref() {
+                if tv.typ.is_ref() && rv.typ.is_ref() && !replacement.local_type().starts_with("Array$") {
```
Aurel300: Generics are messy… What exactly goes wrong if you don't have this extra condition? Does it complain somewhere later that a type is not a place?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1130,160 +1138,262 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                    )?
                 }
             }
-
             ref x => unimplemented!("{:?}", x),
         };
         stmts.extend(encoding_stmts);
-        Ok(stmts
+        Ok(self.set_stmts_default_pos(stmts, stmt.source_info.span))
+
```
```suggestion
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1130,160 +1138,262 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+    fn set_stmts_default_pos(&self, stmts: Vec<vir::Stmt>, default_pos_span: Span) -> Vec<vir::Stmt> {
+        let expr_pos = self
+            .encoder
+            .error_manager()
+            .register(default_pos_span, ErrorCtxt::GenericExpression);
+        let stmt_pos = self
+            .encoder
+            .error_manager()
+            .register(default_pos_span, ErrorCtxt::GenericStatement);
```
Aurel300, line 1179: Does this not register a lot of errors in the error manager that cannot ever be errors?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1130,160 +1138,262 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        // postprocessing if necessary ourselves.
+        if let mir::Rvalue::Ref(_, mir::BorrowKind::Mut { .. }, ref rhs_place) = loan_places.source {
+            let (rhs_place_encoding, ..) = self.mir_encoder.encode_place(rhs_place).unwrap();
+            if let PlaceEncoding::ArrayAccess { .. } = rhs_place_encoding {
+                // encode expiry of the array borrow
+                let (expired_expr, regained_array, wand_rhs) = self.array_magic_wand_at[&loan_places.location].clone();
+
+                // expiring base is something like ref$i32, so we need .val_ref.val_int
+                let deref = self.encoder.encode_value_expr(expiring_base.clone(), expiring_ty);
+                let target_ty = if let ty::TyKind::Ref(_, target_ty, _) = expiring_ty.kind() { target_ty } else { unreachable!() };
```
Aurel300: This doesn't panic but maybe nicer:
```suggestion
                let target_ty = expiring_ty.peel_refs();
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1130,160 +1138,262 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+
+                // we can't refer to the somewhat magic "lhs" label here, because there is no wand,
+                // so there's no LHS. instead, we add a label just before the inhale, and refer to
+                // that
+                let new_lhs_label = self.cfg_method.get_fresh_label_name();
+                stmts.push(
+                    vir::Stmt::Label(new_lhs_label.clone())
+                );
+
+                let wand_rhs_patched_lhs = wand_rhs.fold_expr(
```
Aurel300: Isn't there already a utility function to replace a particular old label somewhere in `vir`? I remember using a similar method in the (still unmerged) closures PR, but I'm not sure if I added it.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1860,7 +1972,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 ..
             } => {
                 // will panic if attempting to encode unsupported type
-                let (encoded_lhs, pre_stmts, _, _) = self.encode_place(lhs).unwrap();
+                // TODO: ignore stmts here as well?
```
Aurel300: Why? Should `pre_stmts` be empty? Can this be asserted?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4003,26 +4119,47 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
         for tree in permissions_forest.get_trees().iter() {
             for (kind, mir_place) in tree.get_permissions().into_iter() {
                 if kind.is_none() {
                     continue;
                 }
                 // will panic if attempting to encode unsupported type
-                let (encoded_place, pre_stmts, ty, _) = self.encode_place(&mir_place).unwrap();
-                assert!(pre_stmts.is_empty(), "Unexpected encoding pre-statements: {:?}", pre_stmts);
-                debug!("kind={:?} mir_place={:?} ty={:?}", kind, mir_place, ty);
+                let (encoded_place, ty, _) = self.mir_encoder.encode_place(&mir_place).unwrap();
```
Aurel300: Add a comment why this uses `mir_encoder` instead of `self`?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4003,26 +4119,47 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                        }
                     }
 
                     // Gives write permission to this node. It must not be a leaf node.
                     PermissionKind::WriteNode => {
-                        let perm = vir::Expr::acc_permission(encoded_place, vir::PermAmount::Write);
-                        permissions.push(perm);
+                        if is_array_access {
+                            let entry = array_pred_perms.entry(encoded_place).or_insert(vir::PermAmount::Write);
+                            *entry= vir::PermAmount::Write;
```
```suggestion
                            array_pred_perms[encoded_place] = vir::PermAmount::Write;
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4463,6 +4624,86 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+
+        let usize_ty = self.encoder.env().tcx().mk_ty(ty::TyKind::Uint(ty::UintTy::Usize));
+        let idx_val_int = index.field(self.encoder.encode_value_field(usize_ty));
+
+        // inhale infos about array contents back
+        let i_var: vir::Expr = vir_local!{ i: Int }.into();
+        let zero_le_i = vir!{ [vir::Expr::from(0)] <= [ i_var ] };
+        let i_lt_len = vir!{ [ i_var ] < [ vir::Expr::from(at.array_len) ] };
+        let i_ne_idx = vir!{ [ i_var ] != [ old(idx_val_int.clone()) ] };
+        let idx_conditions = vir!{ [zero_le_i] && ([i_lt_len] && [i_ne_idx]) };
```
Aurel300, line 4670: (Not super important, but does the `vir!` macro require this kind of parenthesisation?)

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5426,50 +5676,150 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        let at = self.encoder.encode_array_types(array_ty)?;
+
+        let res: vir::Expr = self.cfg_method.add_fresh_local_var(at.elem_ty.clone()).into();
+        let val_field = self.encoder.encode_value_field(at.elem_ty_rs);
+        let res_val_field = res.clone().field(val_field);
+
+        let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(base, ArrayAccessKind::Mutable(None, location))?;
+
+        let usize_ty = self.encoder.env().tcx().mk_ty(ty::TyKind::Uint(ty::UintTy::Usize));
+        let idx_val_int = index.field(self.encoder.encode_value_field(usize_ty));
```
```suggestion
        let idx_val_int = vir::Expr::snap_app(index);
```

prusti-viper/src/encoder/procedure_encoder.rs
dario23: yes it does, sadly, as `&&` is a binary thing (i.e. no repeat in and of itself), and the lhs and rhs are `tt` matchers, so we'd match `[zero_le_i] && [i_lt_len]` and then run into "unexpected token" errors.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: they [removed the `IndexMut` impl for `HashSet`](https://stackoverflow.com/a/30414450), so it seems it's not that simple anymore, but `array_pred_perms.insert(encoded_place, vir::PermAmount::Write)` should work.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: don't think they necessarily are; i think this comment is from when i initially introduced this and wasn't sure if we needed them blindly passed on everywhere. i'll remove the comment.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: indeed, found `Expr::map_old_expr_label`, thanks :-)

prusti-viper/src/encoder/procedure_encoder.rs
dario23: i run into [this assertion](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/foldunfold/requirements.rs#L355) if i try that.. :-/

prusti-viper/src/encoder/procedure_encoder.rs
dario23: not sure, i mostly just moved the code into the separate function.

prusti-viper/src/encoder/procedure_encoder.rs
Aurel300: Ok, then this either needs to be patched manually:
```suggestion
        let idx_val_int = self.encoder.patch_snapshots(vir::Expr::snap_app(index))?;
```
Or just left as is.

prusti-common/src/vir/ast/expr.rs
dario23: not sure anymore. removed it, array testcases still pass. i'll do a full run of the testsuite and see, maybe it's not needed (anymore?) after all.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: hm, manually patching sounds fair enough, i'll go with that.

prusti-common/src/vir/ast/expr.rs
dario23: tests look good :-)

PR #490, opened 2021-04-30
Remove translate_maybe_borrowed_place

fpoli:
`translate_maybe_borrowed_place` is probably an old hack needed to encode discriminants when all places of an enumeration are blocked/uninitialized. PR #459 fixed some issues of that encoding.


fpoli:
This has already been tried in 02aca12f6260a6fe6c6b05cb8bdda4dcc3f883b3, unsuccesfully (a274dc12f2c3af8296505d89f0c06bf6988e9974).

PR #493, opened 2021-05-03
Filter and report Clippy warnings on PRs

fpoli:
A tentative to automatically filter and report Clippy warnings as comments on pull requests.


fpoli:
I'm not sure what caused the OOM on Windows. It may be a glitch of GitHub. I'll ignore it because this PR doesn't change what runs on Windows.

PR #494, opened 2021-05-03
Use AssertKind::fmt_assert_args now that it's public

dario23:
See #467 and rust-lang/rust#84392 for some context.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2108,9 +2108,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.5.2"
+version = "1.5.3"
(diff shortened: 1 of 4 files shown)


dario23:
hm, currently fails because the messages change in the tests.. and looking at the differences, i think we might want to keep using the `Display` impl as far as possible still, as the messages produced by `fmt_assert_args` are not the same as the `Display` impl, e.g.
```
// .fmt_assert_args()
msg: "7:9: 7:14: [Prusti: verification error] assertion might fail with \"\"attempt to divide `{}` by zero\", _4\""
```
vs.
```
// .description()
msg: "assertion might fail with \"attempt to divide by zero\"",
```

PR #495, opened 2021-05-03
Add the Slice builtins/basics

dario23:


fpoli:
Oops, sorry for the unreadable and unrelated automatic comments. I'll try to fix that. There is no need to fix warnings for lines that you didn't modify.

fpoli:
> Oops, sorry for the unreadable and unrelated automatic comments. I'll try to fix that. There is no need to fix warnings for lines that you didn't modify.

Could you rebase on master, just to see if that works now?

dario23:
things on windows still seem broken :-/

also: where would i see any automatic comments? did you delete any previously? or in the job output?

fpoli:
> things on windows still seem broken :-/

Yeah...

> also: where would i see any automatic comments? did you delete any previously? or in the job output?

As comments in the diff of the PR. They now disappeared.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -133,6 +150,70 @@ impl BuiltinEncoder {
                     body: None,
                 }
             },
+            BuiltinFunctionKind::SliceLookupPure { slice_elem_ty, return_ty } => {
+                let elem_ty_name = if let vir::Type::TypedRef(ref ty) = slice_elem_ty { ty } else { unreachable!() };
+                let type_pred = format!("Slice${}", elem_ty_name);
```
fpoli: This does not trigger the encoding of the predicate and will probably generate a consistency error like the one in #489 (the cause might be similar). You should call `Encoder::encode_type_predicate_use`, passing the Rust type of the slice. Either you store that type in the `BuiltinFunctionKind` (adding a `<'tcx>`) or you build a Rust type on the fly with [`TyCtxt::mk_ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.mk_ty).

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -133,6 +150,70 @@ impl BuiltinEncoder {
...
+                            vir::PermAmount::Read,
+                        )
+                    ],
+                    posts: vec![],
+                    body: None,
+                }
+            },
+            BuiltinFunctionKind::SliceLen { slice_elem_ty } => {
+                let elem_ty_name = if let vir::Type::TypedRef(ty_name) = slice_elem_ty { ty_name } else { unreachable!() };
+                let slice_pred = format!("Slice${}", elem_ty_name);
```
fpoli: Same comment as for `SliceLookupPure`

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -133,6 +150,70 @@ impl BuiltinEncoder {
...
+                let self_var = vir::LocalVar::new("self", vir::Type::TypedRef(slice_pred.clone()));
+
+                vir::Function {
+                    name: format!("Slice${}$len", elem_ty_name),
```
fpoli: I think this is `fn_name`.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -133,6 +150,70 @@ impl BuiltinEncoder {
...
+                    ],
+                    return_type: vir::Type::Int,
+                    pres: vec![
+                        vir::Expr::predicate_access_predicate(
+                            slice_pred,
+                            vir::Expr::local(self_var),
+                            vir::PermAmount::Read,
+                        ),
+                    ],
+                    posts: vec![],
```
fpoli: I think for some programs you'll need to add `result >= 0` as postcondition. You can try to test this with a program that contains `assert!(my_slice.len() >= 0);` or something similar.

prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -218,26 +218,40 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
...
+                                index: self.encode_local(*idx)?.into(),
+                                slice_elem_ty: self.encoder().encode_type(elem_ty)?,
+                                lookup_pure_ret: self.encoder().encode_value_type(elem_ty)?,
+                                val_field: self.encoder().encode_value_field(elem_ty),
+                            },
+                            elem_ty,
+                            None,
+                        )
+                    },
+                    _ => unreachable!("index on unsupported type '{:?}'", base_ty),
```
fpoli: Should this be an `EncodingError::unsupported`?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4985,32 +4986,68 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                                ),
+                            ],
+                            vir::Type::Int,
+                            vir::Position::default(),
+                        ),
+                        vir::AssignKind::Copy,
+                    )
+                );
+            },
+            other => unimplemented!("Length operation on unsupported type '{:?}'", other),
```
fpoli: `EncodingError::unsupported`?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5524,6 +5561,42 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                let lookup_res_val_field = lookup_res.clone().field(val_field.clone());
+
+                let (encoded_base_expr, mut stmts) = self.postprocess_place_encoding(*base)?;
+                stmts.extend(self.encode_havoc_and_allocation(&lookup_res));
+
+                let lookup_pure_call = vir::Expr::func_app(
+                    lookup_pure,
+                    vec![
+                        encoded_base_expr,
+                        index.field(vir::Field::new("val_int", vir::Type::Int)),
```
fpoli: If you need this again, `Encoder::encode_value_field` is slightly better than hard-coding `"val_int"`.

fpoli:
The PR looks good, but the comment about `Encoder::encode_type_predicate_use` is important because it might cause consistency errors.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: added #499 to pass the rust array type via `PlaceEncoding`, will rebase once that is accepted.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: that won't work yet, the `core::slice::<impl [i32]>::len` (that's how it's formatted in the MIR) function call doesn't know yet that its result is the same as `Slice$i32$len` -- i don't know how to `#[extern_spec]` that because a) not sure how to specify the module for that `<impl [i32]>`, just ignoring it doesn't work; and b) i have no way to refer to `Slice$i32$len` in `extern_spec` i.e. from rust.

i'll attempt to special-case it in `procedure_encoder` the way `panic`, `Box::new` et al. are.

dario23:
especially re the last commit, not sure if replacing calls to "core::slice::<impl [T]>::len" with `lhs := Slice$<ty>$len(rhs)` like this is the desired way to go, or if there's something more like attaching `#[trusted]` and `#[ensures(result == Slice$<ty>$len(self))]` that i can do here.

prusti-viper/src/encoder/builtin_encoder.rs
dario23: did the special-casing attempt, see last commit.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2071,6 +2071,74 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                            trace!("self_ty: {:?}", self_ty);
+                            trace!("def_id: {:?}", def_id);
+                            trace!("substs: {:?}", substs);
+                            trace!("args: {:?}", args);
+                            trace!("destination: {:?}", destination);
+                            trace!("ty: {:?}", ty);
+
+                            assert!(args.len() == 1, "unexpected args to slice::len(): {:?}", args);
+                            let slice_operand = self.mir_encoder.encode_operand_expr(&args[0])
+                                .with_span(span)?;
```
fpoli: Add `.run_if_err(|| cleanup(&self))?` to call the cleanup closure before returning an error. It's an ugly consequence of the hacky support for generics. (Here and for all the `?` below.)

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2071,6 +2071,74 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+
+                            trace!("{:?} := Slice${:?}$len({:?})", destination, ty, slice_operand);
+
+                            // we need to put a label before, it seems..
```
fpoli: Why? Have you tried without? The `len` function doesn't reborrow, so I suspect that the `encode_transfer_args_permissions` and the label are not needed.

fpoli:
> especially re the last commit, not sure if replacing calls to "core::slice::<impl [T]>::len" with `lhs := Slice$<ty>$len(rhs)` like this is the desired way to go, or if there's something more like attaching `#[trusted]` and `#[ensures(result == Slice$<ty>$len(self))]` that i can do here.

Intercepting "core::slice::<impl [T]>::len" calls as you do looks fine. Defining built-in `#[extern_spec]` is not yet possible as far as I know.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2071,6 +2071,74 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                             }
                         }
 
+                        "core::slice::<impl [T]>::len" => {
```
fpoli, line 2084: Could you move the body of this case to a separate method?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5245,6 +5360,40 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        String,
+        vir::Type,
+        String,
+        vir::Type,
+        vir::Type,
+        ty::Ty<'tcx>,
+    )> {
+        // get rid of ref
+        let slice_ty = match slice_ty.kind() {
+            ty::TyKind::Ref(_, ty, _) => ty,
```
fpoli: If it's known at the call-site of `encode_slice_types` that `slice_ty` is a `TyKind::Ref`, then I would prefer to remove this match and move it to the caller.

prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -218,22 +218,34 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
...
+                                rust_array_ty: base_ty,
+                            },
+                            elem_ty,
+                            None,
+                        )
                     },
-                    elem_ty,
-                    None,
-                )
+                    ty::TyKind::Slice(elem_ty) => {
```
fpoli, line 235: Is it possible to add a test case that executes this branch?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4985,32 +5054,78 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                stmts.extend(
+                    self.encode_copy_value_assign(
+                        encoded_lhs,
+                        len.into(),
+                        dst_ty,
+                        location,
+                    )?
+                );
+            },
+            ty::TyKind::Slice(..) => {
```
fpoli, line 5104: Is it possible to add a test case that executes this branch? I'm not sure how to generate a `mir::Rvalue::Len`. Maybe it's automatically generated by the compiler when it checks the bounds of slice indexes.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4985,32 +5054,78 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                        vir::Expr::func_app(
+                            slice_len,
+                            vec![
+                                encoded_place,
+                            ],
+                            vec![
+                                vir::LocalVar::new(
+                                    String::from("self"),
+                                    vir::Type::TypedRef(ty_name),
+                                ),
```
fpoli: You use this a lot, so it may be better to define a method such that you can just call `vir::LocalVar::new_typed_ref("self", ty_name)`.

fpoli (commented):
It looks good. I commented on the diff.

When adding tests, consider adding some also to `verify/fail`. Otherwise, the tests don't fail when we inhale contradictions by mistake.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: moved the whole block into a separate function, so this should only be needed once on the function call.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: i get a panic saying "thread 'rustc' panicked at 'Location bb0[6] has not yet been encoded', prusti-viper/src/encoder/procedure_encoder.rs:5612:9".

this is with [this testcase](https://github.com/viperproject/prusti-dev/files/6433678/slice-len.rs.txt) and commenting out the label insertion and `encode_transfer_args_permission` call.

[backtrace.txt](https://github.com/viperproject/prusti-dev/files/6433682/backtrace.txt)

i'm not exactly sure, but it seems to me that as this is the last statement in the basic block some permission transfer stuff is being called anyway, to transfer permissions to the next basic block(s), even though we're not doing anything within the statement itself that would require transfering permissions.

dario23: adding the label, but leaving out the `encode_transfer_args_permissions` call, i get a fold/unfold-error:
```
error: [Prusti internal error] generating fold-unfold Viper statements failed (FailedToObtain(Pred(LabelledOld("l0", Field(Local(_6: Ref(ref$Slice$i32), Position { line: 3, column: 1, id: 73 }), val_ref: Ref(Slice$i32), Position { line: 3, column: 1, id: 73 }), Position { line: 3, column: 1, id: 73 }), read)))
 --> prusti-tests/tests/verify/pass/slices/slice-len.rs:3:1
  |
3 | / fn slice_len_gt_0(x: &[i32]) {
4 | |     assert!(x.len() >= 0);
5 | | }
  | |_^
```

dario23: [mir.txt](https://github.com/viperproject/prusti-dev/files/6433699/mir.txt)
[slice-len.rs.vir.txt](https://github.com/viperproject/prusti-dev/files/6433702/slice-len.rs.vir.txt)

fpoli: I see, thanks. It seems required to expire the shared reference that is passed as argument to the MIR function call.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: ack, moved.

prusti-viper/src/encoder/mir_encoder/mod.rs
dario23: hm.. yes and no. i added one, now i get new fold/unfold errors.. will investigate.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: added a testcase, indeed bounds checks are exercising this case.

prusti-viper/src/encoder/mir_encoder/mod.rs
dario23: will do.

fpoli:
For the failing purification tests I suggest to diff the dump of the Viper encoding generated with by the master branch vs the one generated by this PR.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: done.

dario23:
even after rebasing to latest master windows tests fail with the old out of memory error.. :-(

prusti-tests/tests/verify/fail/arrays-slices/out-of-bounds.rs
```diff
@@ -0,0 +1,8 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+
+fn third(s: &[i32]) -> i32 {
+    s[2]  //~ ERROR index out of bounds
```
fpoli: I would add more text from the error message to check that Prusti says that the statement *may* fail with an index out of bounds.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2201,6 +2223,70 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            BuiltinFunctionKind::SliceLen {
+                slice_ty_pred: st.slice_pred.clone(),
+                elem_ty_pred: st.elem_pred,
+            }
+        );
+
+        let rhs = vir::Expr::func_app(
+            slice_len_name,
+            vec![
+                slice_operand.clone(),
```
fpoli: See the Clippy warning(s) reported in Github's diff.

fpoli (commented):
It looks good to me, just a few Clippy warnings to fix and a test annotation to change.

prusti-tests/tests/verify/fail/arrays-slices/out-of-bounds.rs
dario23: updated the error message.. it's a bit ugly, because i have to include the double quotes now, as the full message is `assertion might fail with ""index out of bounds: the length is {} but the index is {}", move _3, _2"`, but for testing it's clearer now i think.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: adressed, i really like that way of showing the warnings :-)

prusti-tests/tests/verify/fail/arrays-slices/out-of-bounds.rs
fpoli: Uh, didn't you fix the "ugliness"? Is the fix outside of this PR?

Aurel300: Also it's a little strange that the message says "assertion" when there is no assertion as far as the user is concerned.

dario23: @fpoli what we fixed was a panic/`unreachable!()` when displaying an `AssertKind::BoundsCheck`, the message is now what `fmt_assert_args` prints.
@Aurel300 i think i have to take back what i said in today's meeting; the "assertion might fail with" part comes from [prusti-viper/src/encoder/errors/error_manager.rs:273](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/errors/error_manager.rs#L273), so we can only really influence what is inside the quotes without extra effort. viper only reports an assertion failure as well, it obviously cannot tell that this is inside a bounds check.

Aurel300: Ok, I see. However, in the file you link, the `ErrorCtxt::AssertTerminator(...)` is actually something Prusti adds when registering the assertion failure with the error manager. Can we easily (i.e. hopefully without string matching…) detect that a particular assert in MIR corresponds to a bound check and if so, register it with a different `ErrorCtxt`?
This might be a lot of work for a tiny annoyance, but it's worth checking.

fpoli: I think changing "assertion might fail with" to "statement might fail with" would be an improvement for all `AssertTerminator`'s usages.

dario23: @Aurel300 hm, good idea, added that :-)
@fpoli i guess that general change is something for a different PR

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -52,6 +52,8 @@ pub enum ErrorCtxt {
     /// A Viper `assert false` that encodes the failure (panic) of an `assert` Rust terminator
     /// Arguments: the message of the Rust assertion
     AssertTerminator(String),
+    /// A Viper `assert false` in the context of a bounds check (if we were able to detect it)
```
Aurel300: > "if we were able to detect it"

Do you know of any cases when the bounds check is not detected? If not, let's be more positive in the comment and leave out the last part.

dario23: hm, you're right, i don't know of any cases we're missing, as long as rustc has the correct `AssertKind` on them.. removed the last part.

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -473,6 +475,13 @@ impl<'tcx> ErrorManager<'tcx>
                     .set_help("The implemented method's postcondition should imply the trait's postcondition.")
             }
 
+            ("assert.failed:assertion.false", ErrorCtxt::BoundsCheckAssert) => {
+                PrustiError::verification(
+                    format!("the array or slice index may be out of bounds"),
```
fpoli:
```suggestion
                    "the array or slice index may be out of bounds".to_string(),
```

fpoli:
CI fails due to the `cleanup` function that has been removed from `master`; a rebase is needed.

PR #497, opened 2021-05-04
Fix git command used to filter Clippy warnings

fpoli:


fpoli:
* [x] Error messages are reported correctly.
* [x] Warning messages are reported correctly.

PR #499, opened 2021-05-04
Pass rust type through in PlaceEncoding

dario23:
This makes it easier to encode the array predicates and lookup functions
properly where we need them, and also reduces the number of things we
need to put into the `PlaceEncoding` struct for array (and later slice) accesses.


prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5245,6 +5245,27 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+
+        // types
+        let array_ty = self.encoder.encode_type(array_ty)?;
+        let elem_ty = self.encoder.encode_type(elem_ty_rs)?;
+        let elem_value_ty = self.encoder.encode_value_type(elem_ty_rs)?;
+
+        let array_len = self.encoder.const_eval_intlike(&len.val)?
+            .to_u64().unwrap().try_into().unwrap();
+
+        Ok((array_pred, array_ty, elem_pred, elem_ty, elem_value_ty, elem_ty_rs, array_len))
```
fpoli, line 5266: Could you define a `struct EncodedArrayTypes { array_pred: ... }` private to `procedure_encoder.rs` to avoid returning the large tuple?

fpoli: I'm going to merge this because overall it's an improvement. You can add the `EncodedArrayTypes` in a new PR.

dario23: oh, thought about that as well while writing it, would have added in this PR if it was still open :-) but new PR is fine as well.

PR #500, opened 2021-05-05
Add EncodedArrayTypes struct for result of encode_array_types

dario23:
The struct is local to procedure_encoder, just meant to not have to type
out very lengthy tuples.

### DIFF
--- a/prusti-viper/src/encoder/procedure_encoder.rs
+++ b/prusti-viper/src/encoder/procedure_encoder.rs
@@ -68,6 +68,16 @@ use crate::encoder::errors::EncodingErrorKind;
 use crate::encoder::snapshot;
 use std::convert::TryInto;
 
+struct EncodedArrayTypes<'tcx> {
+    array_pred: String,
+    array_ty: vir::Type,
+    elem_pred: String,
+    elem_ty: vir::Type,
+    elem_value_ty: vir::Type,
+    elem_ty_rs: ty::Ty<'tcx>,
+    array_len: usize,
+}
+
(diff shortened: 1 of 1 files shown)


dario23:
hm, is something up with the quick-tests? fails with `fatal: couldn't find remote ref array-ty-struct`

fpoli:
> hm, is something up with the quick-tests? fails with `fatal: couldn't find remote ref array-ty-struct`

Yes. Could you add `continue-on-error: true` after [this line](https://github.com/viperproject/prusti-dev/blob/53096dad43e9ce7c9c45d3f4b1c6b062b57534e5/.github/workflows/test.yml#L45) (same indentation as `run:`)?

dario23:
> Yes. Could you add `continue-on-error: true` after [this line](https://github.com/viperproject/prusti-dev/blob/53096dad43e9ce7c9c45d3f4b1c6b062b57534e5/.github/workflows/test.yml#L45) (same indentation as `run:`)?

seems to have done the trick for quick-tests, yes. now just the windows-tests remaining :-)

should i leave this commit in, or do you want that to be a separate PR (not by me, necessarily)?

fpoli:
> should i leave this commit in, or do you want that to be a separate PR (not by me, necessarily)?

The proper fix should be https://github.com/viperproject/prusti-dev/pull/500, but I cannot easily test it. After I merge that you could try to rebase on master and remove the `continue-on-error`.

fpoli:
> seems to have done the trick for quick-tests, yes. now just the windows-tests remaining :-)

Oh, I thought I fixed that issue on windows :(

It's a bug in rustup: https://github.com/rust-lang/rustup/issues/2748

PR #504, opened 2021-05-07
Allow interpolating type in vir_local! macro

dario23:
This makes the macro usable in a lot more cases, where the variable name
is a string literal, but the type is some computed expression.

### DIFF
--- a/prusti-common/src/vir/vir_macro.rs
+++ b/prusti-common/src/vir/vir_macro.rs
@@ -2,11 +2,12 @@
 macro_rules! vir_type {
     (Int) => {$crate::vir::Type::Int};
     (Bool) => {$crate::vir::Type::Bool};
+    ({$ty:expr}) => { $ty }
 }
 
 #[macro_export]
 macro_rules! vir_local {
-    ($name: ident : $type: ident) => {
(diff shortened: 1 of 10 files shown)


fpoli:
My only concern is that `var: [type_foo]` looks a bit like an array of Foos.

dario23:
> My only concern is that `var: [type_foo]` looks a bit like an array of Foos.

yes, thought of that as well. i feel like the more interpolation-y delimiters are curly braces, but went for consistency with how expression interpolation in the `vir!` macro is handled. i'm inclined to change to curly braces though, now that you mentioned it as well.

fpoli:
> > My only concern is that `var: [type_foo]` looks a bit like an array of Foos.
> 
> yes, thought of that as well. i feel like the more interpolation-y delimiters are curly braces, but went for consistency with how expression interpolation in the `vir!` macro is handled. i'm inclined to change to curly braces though, now that you mentioned it as well.

I suggest to change this PR to curly braces, and to add the `vir!` change to a todo-list of "things that would have been nice to implement given more time". At the end of the project you could open an enhancement issue with such a todo-list.

Actually, you can even create an issue now and keep adding stuff to its description.

PR #181, opened 2020-09-23
Ghost Type Encodings

aakp10:
This PR adds encoding definitions for Ghost Types.

PR #505, opened 2021-05-07
Array snapshot encoding and specs

dario23:


prusti-viper/src/encoder/snapshot/encoder.rs
```diff
@@ -537,6 +557,120 @@ impl SnapshotEncoder {
...
+                    );
+                    let read_call = vir::Expr::domain_func_app(
+                        constructor.clone(),
+                        vec![
+                            cons_call,
+                            vir_local!{ idx: Int }.into(),
+                        ],
+                    );
+
+                    let seq_lookup: vir::Expr = todo!("there is no vir::Expr to express `data[idx]` viper-Seq lookup :-(");
```
dario23: this is currently my problem. i want to say
```
forall data: Seq<Snap$...>, index: Int :: {read(cons(data), index)}
      read(cons(data), index) == data[index]
}
```
but as that forall body has to be a `vir::Expr`, i can't really express `data[index]` (i.e. indexing on a viper `Seq`). do i just add `Seq` and `SeqIndex` to the `vir::Expr` enum? don't think that's the proper solution, given how long we spent discussing whether to add `ArrayAccess` to it. @fpoli @Aurel300

fpoli: PR #181 defines a `Expr::ContainerOp` and a `ContainerOpKind` with various operations, but not indexing on a Viper `Seq`. To me the organization of that PR looks like a good plan that you can follow: you could add a `ContainerOpKind` with just a `SeqIndex` operation (but don't spent time adding unneeded operations). I think @Aurel300 is more familiar than me with that work. `ArrayAccess` is more a Rust concept, that's why we were reluctant with adding it to VIR.

Aurel300: I agree. VIR is both a subset and a superset of Viper ASTs. It is a subset because we haven't had the need to use all of Viper's AST yet. Adding variants which map pre-existing Viper nodes is fine. On the other hand, it is a superset because we sometimes want to express Rust/Prusti specifics. Here the decision to add new things is not always obvious, as we don't want to add ad hoc variants to solve tiny problems if they could be solved with the existing variants and better supporting code.

dario23:
still has the first four commits of #532 that need rebasing out once that is merged, but otherwise ready for review i guess.

i can successfully translate [selection_sort.rs](https://github.com/viperproject/prusti-dev/files/6603070/selection_sort.rs.txt) to [selection_sort.rs.vpr](https://github.com/viperproject/prusti-dev/files/6603078/selection_sort.rs.vpr.txt), but not actually verify it. i'm currently not sure if the problem is in the encoding or the invariants.

dario23:
hm, also some errors when creating snapshots of arrays.. i'll look into that.

dario23:
current status: `lookup_pure(a, i)` behaves differently from `read(snap(a), i)`; i'm starting to doubt the current `snap$Array` function implementation, as that's not all that different from calling `lookup_pure` in terms of preserving originally heap-dependent values purely/independent of the heap, the way snapshots do for other types (which have an explicit representation, not just information about them inhaled, i.e. inside the viper state):
```
function snap$__$TY$__Array$3$i32$Snap$Array$3$i32(self: Ref): Snap$Array$3$i32
  requires acc(Array$3$i32(self), read$())
{
  cons$Snap$Array$3$i32$__$TY$__Seq$Int$Snap$Array$3$i32(Seq(lookup_pure__$TY$__Array$3$i32$$int$$$int$(self, 0), lookup_pure__$TY$__Array$3$i32$$int$$$int$(self, 1), lookup_pure__$TY$__Array$3$i32$$int$$$int$(self, 2)))
}
```

dario23:
using `lookup_pure(a, i)` instead of `read(snap(Seq(lookup_pure(a, 0), ..., lookup_pure(a, N)), i)` in specification code makes it work, including the now verifiable selection sort example :+1:

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4075,9 +4078,14 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
     ) -> vir::Expr {
-        // this label is generated and inserted in encode_loop_invariant_exhale_stmts at the point
+        // this label is inserted in encode_loop_invariant_exhale_stmts at the point
         // where the other "preserve equality" assignments are made
-        let old_label = format!("loop_preserve_values_label_{:?}", loop_head);
+        use std::collections::hash_map::Entry::*;
+        let old_label = match self .array_loop_old_label.entry(loop_head) {
+            Occupied(lbl) => lbl.into_mut(),
+            Vacant(v) => v.insert(self.cfg_method.get_fresh_label_name()),
+        };
```
Aurel300: Is this not just:
```suggestion
        let old_label = self.array_loop_old_label.entry(loop_head)
            .or_insert_with(|| self.cfg_method.get_fresh_label_name());
```
?

dario23: that's what i first had as well, but it moves self into the closure :-/

Aurel300: Oh, that… Oh well, we can leave this. Nitpick:
```suggestion
        use std::collections::hash_map::Entry::*;
        let old_label = match self.array_loop_old_label.entry(loop_head) {
            Occupied(lbl) => lbl.into_mut(),
            Vacant(v) => v.insert(self.cfg_method.get_fresh_label_name()),
        };
```

prusti-common/src/vir/ast/expr_transformers.rs
```diff
@@ -256,6 +260,26 @@ pub trait ExprFolder: Sized {
...
+        Expr::ContainerOp(
+            kind,
+            self.fold_boxed(l),
+            self.fold_boxed(r),
+            p,
+        )
+    }
+
+    fn fold_seq(&mut self, t: Type, elems: Vec<Expr>, p: Position) -> Expr {
+        Expr::Seq(t, elems.into_iter().map(|e| self.fold(e)).collect(), p)
```
Aurel300, line 280: I forget, does this work in Rust?
```suggestion
        Expr::Seq(t, elems.into_iter().map(self.fold).collect(), p)
```

prusti-common/src/vir/ast/expr_transformers.rs
```diff
@@ -256,6 +260,26 @@ pub trait ExprFolder: Sized {
...
+    }
+
```
Aurel300:
```suggestion
```

prusti-common/src/vir/ast/expr_transformers.rs
dario23: it seems it doesn't :-(

prusti-common/src/vir/ast/expr_transformers.rs
```diff
@@ -695,6 +740,35 @@ pub trait FallibleExprFolder: Sized {
...
+        ty: Type,
+        elems: Vec<Expr>,
+        p: Position,
+    ) -> Result<Expr, Self::Error> {
+        let mut folded = Vec::with_capacity(elems.len());
+        for e in elems {
+            folded.push(self.fallible_fold(e)?);
+        }
+
+        Ok(Expr::Seq(ty, folded, p))
```
Aurel300:
```suggestion
        Ok(Expr::Seq(
            ty,
            elems.into_iter()
                .map(|e| self.fallible_fold(e))
                .collect::<Result<_>>()?,
            p,
        ))
```

prusti-common/src/vir/optimizations/methods/purifier.rs
```diff
@@ -422,6 +422,7 @@ impl ast::StmtFolder for VarPurifier {
                 ast::Type::TypedRef(_) => "builtin$havoc_ref",
                 ast::Type::Domain(_) => unreachable!(),
                 ast::Type::Snapshot(_) => unreachable!(),
+                ast::Type::Seq(_) => unreachable!(),
```
Aurel300:
```suggestion
                ast::Type::Domain(_)
                | ast::Type::Snapshot(_)
                | ast::Type::Seq(_) => unreachable!(),
```

prusti-common/src/vir/ast/expr_transformers.rs
dario23: oh, neat, didn't know collect had `Control.Monad.sequence` built-in :D

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -19,9 +19,6 @@ use prusti_common::{
     vir_local,
 };
 
-const LOOKUP_PURE_NAME: &str = "lookup_pure";
```
Aurel300, line 22: Are these stale changes or are we abandoning the approach with common names for `lookup_pure` etc for some reason?

prusti-viper/src/encoder/foldunfold/footprint.rs
```diff
@@ -68,6 +68,27 @@ impl ExprFootprintGetter for vir::Expr {
...
+                    .collect::<Vec<_>>();
+
+                let mut res = HashSet::new();
+                for ep in elem_perms {
+                    res.extend(ep);
+                }
+                res
+            }
+
+
```
Aurel300:
```suggestion
```

prusti-viper/src/encoder/snapshot/encoder.rs
```diff
@@ -388,12 +401,43 @@ impl SnapshotEncoder {
...
+            read
+        } else {
+            return Err(EncodingError::internal(
+                format!("called encode_array_idx on non-array-type {:?}", array_ty)
+            ));
+        };
+
+        Ok(read_func.apply(
+            vec![
+                vir::Expr::snap_app(array),
```
Aurel300: Given we are in snapshot encoder, can we eagerly expand this `snap_app`? Or is this one of those things that will cause a double borrow?

prusti-viper/src/encoder/snapshot/encoder.rs
```diff
@@ -586,6 +630,162 @@ impl SnapshotEncoder {
                 self.encode_complex(encoder, variants, predicate_name)
             }
 
+            ty::TyKind::Array(elem_ty, ..) => {
+                let elem_snap_ty = self.encode_type(encoder, elem_ty)?;
+                let at = encoder.encode_array_types(ty)?;
```
Aurel300: "at" again.

prusti-viper/src/encoder/snapshot/encoder.rs
```diff
@@ -586,6 +630,162 @@ impl SnapshotEncoder {
...
+                    let rhs_arg = vir_local!{ _r_data: {seq_type.clone()} };
+
+                    let lhs_call = Expr::domain_func_app(
+                        cons.clone(),
+                        vec![lhs_arg.clone().into()],
+                    );
+                    let rhs_call = Expr::domain_func_app(
+                        cons.clone(),
+                        vec![rhs_arg.clone().into()],
+                    );
```
Aurel300: IIRC I also added `.apply` to domain functions? (Also some cases further down.)

prusti-viper/src/encoder/snapshot/encoder.rs
dario23: hm, could work if we called `self.patch_snapshots_expr` directly instead of `encoder.patch_snapshots`. i'll test and report back.

prusti-viper/src/encoder/snapshot/encoder.rs
dario23: ah, yes.. i even used it (two reviews above) :laughing:

prusti-viper/src/encoder/snapshot/encoder.rs
dario23: maybe `self.snap_app` rather

prusti-common/src/vir/ast/expr_transformers.rs
fpoli: Probably because `self.fold(e)` has two parameters: `Self::fold(self, e)`

prusti-viper/src/encoder/snapshot/encoder.rs
dario23: hm.. doesn't seem to work, but not because double borrows. getting problems with (i think) double applying snapshots(?)
```
error: [Prusti internal error] consistency error: Consistency error: Function snap$__$TY$__Array$3$u32$Snap$Array$3$u32 with formal arguments List(self: Ref) cannot be applied to provided arguments List(cons$Snap$Array$3$u32$__$TY$__Seq$$int$$Snap$Array$3$u32(Seq(0, 1, 2))). (@0.0)
```

FWIW my testcase is `PRUSTI_LOG=prusti_viper::encoder::pure_function_encoder=trace,prusti_viper::encoder::snapshot::encoder=trace ./x.py run --bin prusti-rustc -- --edition 2018 -Pdump_viper_program=true prusti-tests/tests/verify/pass/arrays/array-snap.rs` if you want to have a look yourself.

i think that "duplicate" thing is becase we have a `cons :: Seq[Int] -> Snapshot(Array$...)` and `snap :: TypedRef(Array$...) -> Snapshot(Array$...)`, and even though both declare `Snapshot(Array$...)` as their return type, doing `snap(cons(..))` is a problem.

now that i think about it, we may want to get rid of one of those two, because we're doing the "pure function encoder called from procedure encoder for specifications" case with just `lookup_pure(a, i)` instead anyway, so what's currently called `cons` could just be `snap` for arrays.

Aurel300: Hm. `snap_app` (both the `Expr` variant and the `fn` in snapshot encoder) shouldn't do anything when given an already-snapshot thing. What is the input going into `encode_array_idx` here?

dario23:
```
array=Local(_1: Ref(Array$3$u32), Position { line: 0, column: 0, id: 0 }), array.ty=TypedRef("Array$3$u32")
```

dario23: as the pure function encoder works backwards, it doesn't get the snapshoted thing yet, that happens only later by substituting in the state.

Aurel300: Maybe just omit the `snap_app` and see if the patcher fixes this?

dario23: seems to work, anyway.

prusti-viper/src/encoder/foldunfold/footprint.rs
```diff
@@ -68,6 +68,26 @@ impl ExprFootprintGetter for vir::Expr {
...
+                &right.get_footprint(predicates),
+            ),
+
+            vir::Expr::Seq(_, elems, _) => {
+                let elem_perms = elems
+                    .iter()
+                    .map(|e| {
+                        e.get_footprint(predicates)
+                    })
+                    .collect::<Vec<_>>();
```
Aurel300: I see there is [`FromIterator` on `HashSet`](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3CT%3E-7) – does it work to just `.collect:<HashSet<_>>()` here?

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -495,8 +495,63 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionBackwardInterpreter<'p, 'v, 'tcx> {
...
+                    )
+                } else {
+                    self.encoder.encode_snapshot_array_idx(
+                        rust_array_ty,
+                        postprocessed_base,
+                        idx_val_int,
+                    )?
+                }
+            }
+            PlaceEncoding::SliceAccess { .. } => todo!("slice access"),
```
Aurel300: Use an unsupported encoding error here.

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -495,8 +495,63 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionBackwardInterpreter<'p, 'v, 'tcx> {
...
+            }
+
+            &mir::Operand::Constant(_) => None,
+        })
+    }
+
+    fn postprocess_place_encoding(
+        &self,
+        place_encoding: PlaceEncoding<'tcx>,
+        ) -> EncodingResult<vir::Expr> {
```
Aurel300:
```suggestion
    ) -> EncodingResult<vir::Expr> {
```

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -1259,6 +1345,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
                     }
 
+                    &mir::Rvalue::Len(ref place) => {
+                        let place_ty = self.encode_place(place).with_span(span)?.1;
+                        match place_ty.kind() {
+                            ty::TyKind::Array(..) => {
+                                let array_types = self.encoder.encode_array_types(place_ty).with_span(span)?;
+                                state.substitute_value(&opt_lhs_value_place.unwrap(), array_types.array_len.into());
+                            }
+                            _ => unimplemented!("len of '{:?}'", place_ty),
```
Aurel300: Can we do a proper error here as well?

prusti-viper/src/encoder/foldunfold/footprint.rs
dario23: there is, but for `T`, where `e.get_footprint` returns a `Vec<T>`. i guess i could get rid of the intermediate `elem_perms` vec, but i haven't found a simple way to merge all `HashSet`s from the iter without the temporary variables. 
i guess i could use `fold`, what do you think abou
```
            vir::Expr::Seq(_, elems, _) => {
                elems
                    .iter()
                    .map(|e| {
                        e.get_footprint(predicates)
                    })
                    .fold(HashSet::new(), |hs, fp| {
                        hs.extend(fp); hs
                    })
            }
```

Aurel300: I see, that's unfortunate. The current version works then.

dario23:
alright, after monday's meeting some more reworking of this PR. we now always encode using snapshots in PureFunctionEncoder, adding an equality "axiom" as postcondition to `snap$Array`. we can verify the snippet we previously couldn't with a pure function with postcondition, which i've [added as a testcase](https://github.com/viperproject/prusti-dev/pull/505/files/#diff-df4d71e8e7ee66c50d15c4917d1b27e2e93401b7f4dfa5f20837f42a0384aa35) as well.

the selection_sort example encodes without panic, but doesn't currently verify; i'm not exactly sure why, but i've removed it from the PR for now. i'd suggest we merge this PR soon and i add stuff like the selection sort example or the mergesort i've been trying to verify later on in separate PRs.

Aurel300:
Looks good. You should address the Clippy complaints. The "test on crates" error may also be worth looking into.

PR #514, opened 2021-05-12
Replace run_if_err with local struct with drop impl

dario23:
This seems a little more idiomatic rust, as it's impossible to forget
adding a `.run_if_err(cleanup)` to possibly-failing calls now.


prusti-viper/src/encoder/encoder.rs
```diff
@@ -1200,15 +1200,20 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+        // FIXME: this is a hack to support generics. See issue #187.
+        struct CleanupTyMapStack<'a, 'tcx> {
+            tymap_stack: &'a std::cell::RefCell<Vec<HashMap<ty::Ty<'tcx>, ty::Ty<'tcx>>>>,
+        }
+        impl<'a, 'tcx> Drop for CleanupTyMapStack<'a, 'tcx> {
+            fn drop(&mut self) {
+                self.tymap_stack.borrow_mut().pop();
+            }
+        }
+        let _cleanup_token = CleanupTyMapStack { tymap_stack: &self.typaram_repl };
```
Aurel300: Is it guaranteed by Rust that `Drop` will only run after everything else in this function runs? I.e. could the `Drop` run right here, since `_cleanup_token` is not used anywhere else?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1935,11 +1935,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
                         tymap_stack.push(tymap);
                     }
-                    let cleanup = |this: &ProcedureEncoder| {
-                        // FIXME: this is a hack to support generics. See issue #187.
-                        let mut tymap_stack = this.encoder.typaram_repl.borrow_mut();
-                        tymap_stack.pop();
-                    };
+
+                    // FIXME: this is a hack to support generics. See issue #187.
+                    struct CleanupTyMapStack<'a, 'tcx> {
```
Aurel300: It would be a bit nicer if this was defined only once.

prusti-viper/src/encoder/encoder.rs
dario23: yes. see e.g. [this thread](https://users.rust-lang.org/t/do-variables-shorten-the-scope-lifetime/5754). rustc uses this mechanism internally for e.g. profiling or span enter/leave; `git grep _[a-z]` on the rustc source yields quite some results.

fpoli: I had the same feeling, but it seems fine: https://rust-unofficial.github.io/patterns/idioms/dtor-finally.html.

I like that we no longer have to remember to use `run_if_err`, which was error-prone.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: true. i think it would make sense then to have some method like `enter` or `push` that enters the new "tymap stack frame" and returns the cleanup token, but i'll leave that for some other time i guess and mark this PR as Draft.

prusti-viper/src/encoder/encoder.rs
dario23: [the rust reference](https://doc.rust-lang.org/reference/destructors.html#drop-scopes) seems to be the most official thing i found on the subject; which also only talks about blocks and scopes, not other analyses (e.g. NLL).

fpoli: I would add `#[must_use]` to the struct; it doesn't hurt.

prusti-viper/src/encoder/procedure_encoder.rs
dario23: found some time today to implement this with a `push_temp_tymap` method on `Encoder`, so we only have the struct and drop impl once. what do you think?

Aurel300: I guess it makes sense that the API is on `Encoder`. Maybe the name is not as self-explanatory as it could be but then neither is `tymap` itself. This is an improvement overall, if CI passes I'm happy to merge this.

PR #518, opened 2021-05-13
Remove unused ProcedureEncoder::magic_wand_apply_post map

dario23:


fpoli:
If you remember the use case, @vakaras, I can revert.

vakaras:
I suspect that this is some dead code that was introduced during the compiler upgrade.

Issue #444, opened 2021-03-28
Optimize out redundant statements


vakaras:
Currently, in the Viper encoding generated by Prusti we end-up with statements like `inhale true`, `exhale true`, `assert true`. We should remove them before passing to Viper.

fpoli:
This optimization is supposed to do that. The `inhale true` case is missing.

https://github.com/viperproject/prusti-dev/blob/9546a218875a209f9d5d3262b70640b86bab363e/prusti-common/src/vir/optimizations/methods/assert_remover.rs#L15-L25

PR #522, opened 2021-05-16
Added optimization for "inhale true" case

Maxwell-Yang-2001:
Added optimization for missing case "inhale true": https://github.com/viperproject/prusti-dev/issues/444
@vakaras


vakaras (approved):
LGTM. Thanks!

PR #523, opened 2021-05-16
WIP: svirpti to prusti-dev

Maxwell-Yang-2001:
Copied vir-generate and created stub vir crate.

@vakaras


vakaras:
So far looks good. Could you now add [`build.rs`](https://github.com/vakaras/svirpti/blob/main/vir/build.rs)?

Just adding the file should cause a compilation error. You should be able to fix it by removing [these three lines](https://github.com/vakaras/svirpti/blob/045fc1749bf4d7a279ae487539c847454aa5c894/vir/build.rs#L15-L17).

To check whether `build.rs` works, add the following line to `vir/src/lib.rs`:
```rust
include!(concat!(env!("OUT_DIR"), "/vir.rs"));
```

Maxwell-Yang-2001:
Added build.rs and consequently components, since the latter is required for build.rs to work. 3 lines to smt, low and high are removed, so no need to copy those over yet. @vakaras

fpoli:
What about naming the crate `vir-gen` instead of `vir-generate` for consistency with `jni-gen`?

vir-generate/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "svirpti-vir-generate"
```
fpoli: This should correspond to the name of the folder

vakaras: I agree.

vir/build.rs
```diff
@@ -0,0 +1,24 @@
...
+    println!("cargo:rerun-if-changed=defs");
+    for entry in walkdir::WalkDir::new("defs") {
+        let entry = entry.unwrap();
+        println!("cargo:rerun-if-changed={}", entry.path().display());
+    }
+
+    let root = quote! { components =>
+    };
+    let tokens = define_vir(root, Path::new("defs/root.rs"));
+    // let definition = parse_schema(Path::new("src/schema.dl"), Path::new("src/derived.dl"));
```
vakaras: Please remove the commented out lines.

vir/defs/components/declaration/rsmt.rs
```diff
@@ -0,0 +1,17 @@
+trait Interface {
```
vakaras: This file is not needed for Prusti and should be removed.

vir/defs/components/expression/rsmt.rs
```diff
@@ -0,0 +1,226 @@
+trait Interface {
```
vakaras: This file is not needed for Prusti. Please delete.

vir/defs/components/model/mod.rs
```diff
@@ -0,0 +1,48 @@
+trait Interface {
```
vakaras: This file is not needed for Prusti. Please delete.

vir/defs/components/sort/rsmt.rs
```diff
@@ -0,0 +1,24 @@
+trait Interface {
```
vakaras: This file is not needed for Prusti, please delete.

vakaras (changes requested):
Looks good so far. I added some comments.

vakaras:
> What about naming the crate `vir-gen` instead of `vir-generate` for consistency with `jni-gen`?

Sounds good to me. @Maxwell-Yang-2001 Could you please do that?

Maxwell-Yang-2001:
Thanks for the comments, both of you! @vakaras I have addressed your feedback in the latest commit.

vakaras (approved):
Looks good. Thanks!

vakaras:
The next step (as a new PR) would be to define a polymorphic VIR. To define it, you need to [add a line to `build.rs`](https://github.com/vakaras/svirpti/blob/main/vir/build.rs#L15):

```rust
pub mod polymorphic;
```

and create folder `polymorphic` in [`defs`](https://github.com/vakaras/svirpti/tree/main/vir/defs) that contains a definition of the polymorphic VIR. As a starting point, I would suggest copying [the current VIR](https://github.com/viperproject/prusti-dev/tree/master/prusti-common/src/vir/ast) (just nodes, no methods).

vakaras:
The second step could be refactoring the [`Expr` enum](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/vir/ast/expr.rs#L16-L57) into an enum of structs:
```rust

pub enum Expr {
    Local(Local),
    // ...
}

/// A local var
struct Local {
    variable: LocalVar,
    position: Position,
}
// ...
```

PR #525, opened 2021-05-19
WIP: vir ast copying and Expr enum refactoring

Maxwell-Yang-2001:
Copied files from [prusti-common/src/vir/ast](https://github.com/viperproject/prusti-dev/tree/master/prusti-common/src/vir/ast) to under vir crate in a newly created [polymorphic directory](https://github.com/Maxwell-Yang-2001/prusti-dev/tree/polymorphic-vir/vir/defs/polymorphic). Removed any impl or defined traits, but still had links to prusti-common as of now for [here](https://github.com/Maxwell-Yang-2001/prusti-dev/blob/polymorphic-vir/vir/defs/polymorphic/expr.rs#L8) and [here](https://github.com/Maxwell-Yang-2001/prusti-dev/blob/polymorphic-vir/vir/defs/polymorphic/stmt.rs#L8) (which before moving was using contents under the same crate), since only part of ast in vir is copied.

Refactored the enum `Expr` in [`expr.rs`](https://github.com/Maxwell-Yang-2001/prusti-dev/blob/polymorphic-vir/vir/defs/polymorphic/expr.rs) into several structs, currently located in the [same file](https://github.com/Maxwell-Yang-2001/prusti-dev/blob/polymorphic-vir/vir/defs/polymorphic/expr.rs#L95-L239) (could be moved to a directory structure if needed). Naming of struct fields are based on [comments](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/vir/ast/expr.rs#L17-L56) and [usages on this enum](https://github.com/viperproject/prusti-dev/blob/master/prusti-common/src/vir/ast/expr.rs#L103-L219).

@vakaras thanks in advance!


vir/Cargo.toml
```diff
@@ -10,9 +10,11 @@ serde = { version = "1.0", features = ["derive"] }
 rsmt2 = { git = "https://github.com/vakaras/rsmt2", branch = "labels" }
 thiserror = "1.0.23"
 derivative = "2.1.3"
+regex = "1.5"
 syn = { version = "1.0", features = ["full", "fold", "parsing", "derive"] }
 quote = "1.0"
 proc-macro2 = { version = "1.0", features = ["span-locations"] }
+prusti-common = { path = "../prusti-common" }
```
vakaras: We need to get rid of this dependency because `prusti-common` will depend on `vir`.

vir/defs/polymorphic/stmt.rs
```diff
@@ -0,0 +1,78 @@
+// © 2019, ETH Zurich
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use super::*;
+use prusti_common::vir::borrows::{Borrow, DAG as ReborrowingDAG};
```
vakaras: Please also copy `borrows` so that we can remove dependency on `prusti_common`.

Maxwell-Yang-2001: I see. Where would be a good place to put `borrows`? In [`defs`](https://github.com/Maxwell-Yang-2001/prusti-dev/tree/polymorphic-vir/vir/defs) (Same level as the `polymorphic` directory)?

vakaras:
Looks good so far. The goal of this PR would be to have two VIR “layers”:

1. `polymorphic` – a version of VIR that supports generics.
2. `legacy` – the current version of VIR moved without any modifications from `prusti-common`. We need it to so that we can migrate the Prusti codebase gradually.

To enable gradual migration, we also need a way to convert `polymorphic` VIR into `legacy`. I think it would be best to implement [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) trait for this.

I would suggest pursuing the goal in the following steps (I am writing from the top of my head, so likely missing something):

* [ ] Remove dependency on `prusti-common`.
* [ ] Finish cleaning up the enums (for example, `Stmt` should be converted in the same way as `Expr`; need to check whether we have also other enums to convert).
* [ ] Design and implement support for generics (replace `vir/defs/polymorphic/typaram.rs` with something proper). This will require at least adding a new type for generic arguments and figuring out how to do type substitution.
* [ ] Move `VIR` from `prusti-common` to `legacy`. I think it will be easiest if `prusti-common` reexports `legacy` as `VIR` so that we can avoid changing code all other the place.
* [ ] Implement conversion from `polymorphic` to `legacy`.

vir/defs/polymorphic/stmt.rs
vakaras: Let's put it inside `vir/src`. Then, the following import should work:

```rust
use crate::borrows::{Borrow, DAG as ReborrowingDAG};
```

Maxwell-Yang-2001: And this `borrows.rs` also relies on part of `polymorphic`. By being placed outside of `polymorphic` (so without being generated as part of `vir.rs`), how would it import things from `vir.rs`? I think the main confusion for me is how `vir.rs` would be located relative to other files or directories in the `vir` crate.

vakaras: Oh, I missed that `borrows.rs` depends on VIR. In that case, it needs to be inside `defs/polymorphic`.

vakaras: > I think the main confusion for me is how vir.rs would be located relative to other files or directories in the vir crate.

The generated `vir.rs` is put into the `target` directory. It is included by the include macro in `vir/src/lib.rs`, so you can think as all code from `vir.rs` is inlined at that place.

prusti-common/src/vir/mod.rs
```diff
@@ -4,21 +4,7 @@
...
-pub use self::to_viper::*;
-pub use self::program::*;
-
-mod ast;
-pub mod borrows;
-mod cfg;
-mod conversions;
-pub mod fixes;
-pub mod optimizations;
-mod to_viper;
```
vakaras: I think that it is better to move only the AST/CFG nodes and leave all other code here. For example, I think modules `to_viper`, `optimizations`, and `fixes` should stay.

vir/Cargo.toml
```diff
@@ -10,9 +10,16 @@ serde = { version = "1.0", features = ["derive"] }
...
 derivative = "2.1.3"
+regex = "1.5"
 syn = { version = "1.0", features = ["full", "fold", "parsing", "derive"] }
 quote = "1.0"
 proc-macro2 = { version = "1.0", features = ["span-locations"] }
+uuid = { version = "0.8", features = ["v4"] }
+log = { version = "0.4", features = ["release_max_level_info"] }
+config = "0.11"
+itertools = "0.10.0"
+viper = { path = "../viper" }
```
vakaras: `vir` should not depend on `viper`.

vir/build.rs
```diff
@@ -13,6 +13,11 @@ fn main() {
     }
 
     let root = quote! { components =>
+        pub mod legacy;
+        pub mod polymorphic;
+        pub mod config;
```
vakaras: `config` should not be here because it has nothing to do with VIR.

vir/build.rs
```diff
@@ -13,6 +13,11 @@ fn main() {
...
+        pub mod utils;
```
vakaras: `utils` probably also can be left in their original place for now.

vir/build.rs
```diff
@@ -13,6 +13,11 @@ fn main() {
...
+        pub mod report;
```
vakaras: `report` should also be left in the original place.

vakaras:
@Maxwell-Yang-2001 Could you please:

* [ ] Rebase your code to target the `master` branch
* [ ] Make sure that the code compiles.

PR #532, opened 2021-06-03
Refactor array types encoding

dario23:
- Moved to `Encoder` so it can be used from pure function encoder as well.
 - Moved encoding of `lookup_pure` calls onto the `EncodedArrayTypes` and `EncodedSliceTypes` structs to avoid repeating it everywhere.

### DIFF
--- a/prusti-viper/src/encoder/array_encoder.rs
+++ b/prusti-viper/src/encoder/array_encoder.rs
@@ -19,6 +19,10 @@ use prusti_common::{
     vir_local,
 };
 
+const LOOKUP_PURE_NAME: &str = "lookup_pure";
+const SLICE_LEN_NAME: &str = "Slice$len";
+
+
 /// The result of `ArrayEncoder::encode_array_types`. Contains types, type predicates and length of the given array type.
 #[derive(Clone, PartialEq, Eq, Hash, Debug)]
 pub struct EncodedArrayTypes<'tcx> {
@@ -34,14 +38,12 @@ pub struct EncodedArrayTypes<'tcx> {
     pub elem_ty_rs: ty::Ty<'tcx>,
     /// The length of the array, e.g. 3
     pub array_len: usize,
-    /// The name of the lookup_pure function for this instance, e.g. "Array$3$i32$lookup_pure"
-    pub lookup_pure_name: String,
 }
 
 impl<'tcx> EncodedArrayTypes<'tcx> {
     pub fn encode_lookup_pure_call(&self, array: vir::Expr, idx: vir::Expr) -> vir::Expr {
         vir::Expr::func_app(
-            self.lookup_pure_name.clone(),
+            LOOKUP_PURE_NAME.to_owned(),
             vec![
                 array,
                 idx,
@@ -69,16 +71,12 @@ pub struct EncodedSliceTypes<'tcx> {
     pub elem_value_ty: vir::Type,
     /// The non-encoded element type as passed by rustc
     pub elem_ty_rs: ty::Ty<'tcx>,
-    /// The name of the `lookup_pure` function for this instance, e.g. "Array$3$i32$lookup_pure"
-    pub lookup_pure_name: String,
-    /// The name of the slice length function for this instance, e.g. "Slice$i32$len"
-    pub slice_len_name: String,
 }
 
 impl<'tcx> EncodedSliceTypes<'tcx> {
     pub fn encode_lookup_pure_call(&self, slice: vir::Expr, idx: vir::Expr) -> vir::Expr {
         vir::Expr::func_app(
-            self.lookup_pure_name.clone(),
+            LOOKUP_PURE_NAME.to_owned(),
             vec![
                 slice,
                 idx,
@@ -94,7 +92,7 @@ impl<'tcx> EncodedSliceTypes<'tcx> {
 
     pub fn encode_slice_len_call(&self, slice: vir::Expr) -> vir::Expr {
         vir::Expr::func_app(
-            self.slice_len_name.clone(),
+            SLICE_LEN_NAME.to_owned(),
             vec![
                 slice,
             ],
@@ -124,16 +122,10 @@ impl<'p, 'v: 'p, 'tcx: 'v> ArrayTypesEncoder<'tcx> {
 
     /// Encode types, type predicates and builtin lookup functions required for handling slices of
     /// type `slice_ty`.
-    ///
-    /// Optionally pass a pre-encoded `vir::Type` for the element snapshot type. This is useful
-    /// when calling `encode_slice_types` from e.g. inside the `SnapshotEncoder` where `Encoder`'s
-    /// `snapshot_encoder` field is already borrowed, so `encoder.encode_snapshot_type` would
-    /// result in a double borrow (i.e. panic).
     pub fn encode_slice_types(
         &mut self,
         encoder: &'p Encoder<'v, 'tcx>,
         slice_ty_rs: ty::Ty<'tcx>,
-        pre_encoded_elem_snap_ty: Option<vir::Type>,
     ) -> EncodingResult<EncodedSliceTypes<'tcx>> {
         if let Some(cached) = self.slice_types_cache.get(&slice_ty_rs) {
             return Ok(cached.clone());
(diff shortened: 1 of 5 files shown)


Aurel300:
I would rather keep making `Encoder` lighter and dumber over time. Could you put all the types and methods into a new `ArrayEncoder`? It can also do simple caching like the snapshot encoder.

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub struct EncodedArrayTypes<'tcx> {
+    pub array_pred: String,
+    pub array_ty: vir::Type,
+    pub elem_ty: vir::Type,
+    pub elem_value_ty: vir::Type,
+    pub elem_ty_rs: ty::Ty<'tcx>,
+    pub array_len: usize,
+    pub lookup_pure_name: String,
```
Aurel300: I wonder: can `lookup_pure` (and others) always just have a constant name? E.g. for snapshots the snapshot function is called `snap$` for any (non-primitive etc) type. Prusti later does mangling based on the types found in the signature so the function name becomes unique again.

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub struct EncodedSliceTypes<'tcx> {
+    pub slice_pred: String,
+    pub slice_ty: vir::Type,
+    pub elem_ty: vir::Type,
+    pub elem_value_ty: vir::Type,
+    pub elem_ty_rs: ty::Ty<'tcx>,
+    pub lookup_pure_name: String,
+    pub slice_len_name: String,
```
Aurel300: (same comment, can these always be named the same)

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+    builtin_encoder::BuiltinFunctionKind,
+};
+use prusti_common::{
+    vir,
+    vir_local,
+};
+
+
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub struct EncodedArrayTypes<'tcx> {
```
Aurel300, line 28: At least a short comment on what these represent would be useful. (same for the slice type)

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    /// Encode types, type predicates and builtin lookup functions required for handling slices of
+    /// type `slice_ty`.
+    ///
+    /// Optionally pass a pre-encoded `vir::Type` for the element snapshot type. This is useful
+    /// when calling `encode_slice_types` from e.g. inside the `SnapshotEncoder` where `Encoder`'s
+    /// `snapshot_encoder` field is already borrowed, so `encoder.encode_snapshot_type` would
+    /// result in a double borrow (i.e. panic).
```
Aurel300: I find this `pre_encoded_elem_snap_ty` API a bit messy. How does `SnapshotEncoder` call this?

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+        let slice_ty = encoder.encode_type(slice_ty_rs)?;
+        let elem_ty = encoder.encode_type(elem_ty_rs)?;
+
+        let elem_value_ty = if let Some(pre_encoded) = pre_encoded_elem_snap_ty {
+            pre_encoded
+        } else {
+            encoder.encode_snapshot_type(elem_ty_rs)?
+        };
+
+        // lookup_pure
```
Aurel300, line 145:
```suggestion
```

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+        );
+
+        let slice_len_name = encoder.encode_builtin_function_use(
+            BuiltinFunctionKind::SliceLen {
+                slice_ty_pred: slice_pred.clone(),
+                elem_ty_pred: elem_pred,
+            }
+        );
+
+        let encoded = EncodedSliceTypes{
```
Aurel300:
```suggestion
        let encoded = EncodedSliceTypes {
```

prusti-viper/src/encoder/array_encoder.rs
```diff
@@ -0,0 +1,237 @@
...
+        let lookup_pure_name = encoder.encode_builtin_function_use(
+            BuiltinFunctionKind::ArrayLookupPure {
+                array_ty_pred: array_pred.clone(),
+                elem_ty_pred: elem_pred,
+                array_len,
+                return_ty: elem_value_ty.clone(),
+            }
+        );
+
+        let encoded = EncodedArrayTypes{
```
Aurel300:
```suggestion
        let encoded = EncodedArrayTypes {
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2224,25 +2205,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
         let slice_ty_ref = self.mir_encoder.get_operand_ty(&args[0]);
         let slice_ty = if let ty::TyKind::Ref(_, slice_ty, _) = slice_ty_ref.kind() { slice_ty } else { unreachable!() };
-        let st = self.encode_slice_types(slice_ty).with_span(span)?;
-        let slice_len_name = self.encoder.encode_builtin_function_use(
-            BuiltinFunctionKind::SliceLen {
-                slice_ty_pred: st.slice_pred.clone(),
-                elem_ty_pred: st.elem_pred,
-            }
-        );
+        let st = self.encoder.encode_slice_types(slice_ty, None).with_span(span)?;
```
Aurel300: Please replace `st` and `at` with a more descriptive name. We know what they are now but … :)

prusti-viper/src/encoder/array_encoder.rs
dario23: i had the same thought some time ago, but i think it's at least not as simple as just dropping them from the name. currently we have
```
function Array$5$i32$lookup_pure__$TY$__Array$5$i32$$int$$$int$(self: Ref, idx: Int): Int
  requires acc(Array$5$i32(self), read$())
  requires idx < 5
```
so the different `lookup_pure` instances are actually different because they mention their respective type predicate and length in their preconditions. we might be able to get rid of the preconditions and sort-of inline them as assertions whenever we use them, but at that point i'm not sure it's worth it, so i dropped the idea at the time.

prusti-viper/src/encoder/array_encoder.rs
dario23: it is messy indeed. one alternative i thought about is to drop that field from the struct, but we need it in `encode_lookup_pure_call`.

SnapshotEncoder is (sometimes/mostly) called through [`Encoder::encode_snapshot_type`](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/encoder.rs#L881), which does `        self.snapshot_encoder.borrow_mut().encode_type(self, ty)`. The `SnapshotEncoder::encode_type` call goes through pretty much the whole snapshot machinery for the type, building the required functions and axioms and whatnot, and in the end only returning the type of the snapshot. If anything inside that machinery calls `encode_array_types`, then that would arrive at the same `self.snapshot_encoder.borrow_mut()` again, making prusti panic. So instead, `SnapshotEncoder` would just call `self.encode_type` on the array element type first, and use the result of that, without going through `Encoder::encode_snapshot_type` thus avoiding the borrowing problem.

I'm not sure if we can make that `borrow_mut` more fine-grained by moving the `Rc` into `SnapshotEncoder` (so we'd only need a shared reference for the calls from `Encoder`, or if that would just move the problem elsewhere. i hope it would drastically shorten the duration (and split it up by its fields) of the `SnapshotEncoder` are borrowed).

dario23: hm, now that i think about it.. we could make it a little less messy by dropping the element snapshot type from the `EncodedArrayTypes` struct and requiring the user to pass it to `EncodedArrayTypes::encode_lookup_pure_call`. That way, we can still pass whatever is appropriate/possible in the current situation, but we don't need the admittedly ugly API to thread this additional (not really intuitive) parameter through.

prusti-viper/src/encoder/array_encoder.rs
Aurel300: The functions would still be different, they would all just share the same name. In the example you have, everything after `__$TY$__` is not really part of the name, it is added by name mangling later.

dario23: ah, that sounds like a plan, yes. i'll change that.

prusti-viper/src/encoder/array_encoder.rs
dario23: we still need it to call `encode_builtin_function_use`, which i think should be called in `encode_array_types`, as we don't have access to `encoder` anymore in `encode_lookup_pure_call` (which only has the `EncodedArrayTypes` as `self`).

i have removed all the weird "passing in `elem_value_ty`" stuff for now, and just call `encoder.encode_snapshot_type`. for how we'll do that from within `SnapshotEncoder` i'll worry later. one step at a time i guess. let's get this PR and #489 merged and take it from there.

PR #533, opened 2021-06-04
Converted unimplemented panic to error message

ani003:
Unimplemented panic triggered by `vec!` converted to error message


Aurel300:
Looks good. "unsizing" might not be an operation that is familiar to Rust developers, but it's a start. I'm not sure we can be more specific with the message because this could also happen if the programmer literally wrote the cast in their code.

Please add a test case into `prusti-tests/tests/verify/fail`. See the files in that directory for an example of the `//~` syntax to declare that an error is expected on a given line.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1149,7 +1149,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
 
             mir::StatementKind::Assign(box (ref lhs, ref rhs)) => {
                 // FIXME: the following line will panic if attempting to encode unsupported types.
```
```suggestion
```

Aurel300:
Thanks!

PR #537, opened 2021-06-06
Change predicate syntax to allow nested implications

dario23:
Constructs like Prusti's implication `==>` that aren't valid Rust were
not possible in previous style `#[predicate]` functions, because rustc
parsed the functions already for e.g. name resolution, before we were
able to rewrite them. With the new function-like proc-macro aproach, we
avoid that problem.


prusti-tests/tests/parse/ui/no-attr-params.rs
```diff
@@ -13,7 +13,9 @@ fn pure_err() -> bool {
 #[trusted(dont write things here)]
 fn trusted_err() {}
 
-#[predicate(dont put anything here)]
-fn pred_err() {}
+predicate! {
+    #[dont_put_anything_here]
+    fn pred_err() {}
+}
```
Aurel300: I think this test can just be deleted now.

prusti-specs/src/lib.rs
```diff
@@ -480,64 +480,55 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
     }
 }
 
-pub fn predicate(attr: TokenStream, tokens: TokenStream) -> TokenStream {
-    if !attr.is_empty() {
-        return syn::Error::new(
-            attr.span(),
-            "the `#[predicate]` attribute does not take parameters"
-        ).to_compile_error();
+// for syn::parse the predicate thingy
```
```suggestion
```

prusti-tests/tests/parse/ui/no-attr-params.rs
Aurel300: Sorry, I should have been clearer – I meant the highlighted lines, the remainder of the file is still good.

dario23: ah, fixed.

dario23:
test that fails is `[run-pass] pass/quick/enums.rs`, which i haven't changed anything about. `git bisect` thinks it's the commit merging the new viper-tools handling (#553). is there a significantly different viper/z3 version in that?

PR #538, opened 2021-06-06
Implement enough of array slicing to call .len() on arrays

dario23:


dario23:
the `all-tests` tests currently fail on
```rust
    for slot in cpuset.bits.iter_mut() { //~ ERROR cast statements that create loans are not supported
        *slot = 0;
    }
```
which makes sense, since the cast is an unsize-cast just like i need for array length calls (`cpuset.bits` is a fixed-size array). the example still fails, with an error that is unrelated to my change as far as i can tell: `error: [Prusti internal error] unregistered verification error: [unfold.failed:insufficient.permission] Unfolding m_Option$_beg_$ref$u64$_end_Some(_10.enum_Some) might fail. There might be insufficient permission to access m_Option$_beg_$ref$u64$_end_Some(_10.enum_Some) (<no position>)`.

i think that error is from the iterator being unsupported, and i don't have the time to add that support currently. do i just update the expected error (and filename, consequently)? the part with arrays we have in `prusti-tests/tests/verify/pass/arrays/fd_set.rs` as well, so i might just delete the testcase as well?

fpoli:
Surely you shouldn't add support for iterators; that's not trivial. You can either disable the test temporary, or throw an `EncodingError::unsupported(..)` at some point during the encoding if you can easily identify the cause.

The tests in `prusti-tests/tests/verify/fail/unsupported` are meant to check that Prusti produces readable error messages of the "unsupported feature" kind, without crashing or generating internal errors. However, the "[unfold.failed:insufficient.permission]" is a verification error, which is weird because it seems to signal that no encoding error was raised. I wonder where the "cast statements that create loans are not supported" error ended up.

prusti-interface/src/environment/mir_utils/split_aggregate_assignment.rs
```diff
@@ -54,7 +54,9 @@ impl<'tcx> SplitAggregateAssignment<'tcx> for mir::Statement<'tcx> {
                     .collect()
             }
             mir::Rvalue::Use(_) |
-            mir::Rvalue::Ref(_, _, _) =>
+            mir::Rvalue::Ref(_, _, _) |
+            // slice creation is ok™
```
Aurel300:
```suggestion
            // slice creation is ok
```

prusti-interface/src/environment/polonius_info.rs
```diff
@@ -593,7 +593,22 @@ fn get_borrowed_places<'a, 'tcx: 'a>(
...
+                &mir::Rvalue::Cast(ref kind, ref operand, ref ty) => {
+                    // slice creation involves an unsize pointer cast like [i32; 3] -> &[i32]
+                    if kind == &mir::CastKind::Pointer(ty::adjustment::PointerCast::Unsize) {
+                        if let ty::TyKind::Ref(_, ref ref_ty, _) = ty.kind() {
+                            if let ty::TyKind::Slice(..) = ref_ty.kind() {
+                                trace!("slice: operand={:?}, ty={:?}", operand, ty);
+                                return Ok(match operand {
+                                    mir::Operand::Copy(ref place) |
+                                        mir::Operand::Move(ref place) => vec![place],
+                                        mir::Operand::Constant(_) => vec![],
```
Aurel300:
```suggestion
                                    mir::Operand::Copy(ref place)
                                    | mir::Operand::Move(ref place) => vec![place],
                                    mir::Operand::Constant(_) => vec![],
```

dario23:
updated the early unsupported error for iterators to only match on `Iterator::next`, which is actually the only required one to implement `Iterator`, i.e. the one all the others build on. i think the problems we currently have with it don't come up when creating an iterator, just during the iteration. also the other method `Iterator::into_iter` i had there isn't actually a method on iterator, but on a separate trait `IntoIterator`. if the tests go through i'd keep the filter on just the `Iterator::next`, don't need to be more restrictive than necessary i guess.

PR #539, opened 2021-06-07
Fixes for panics in encode_copy2

ani003:
This PR fixes the panics in `encode_copy2` by either performing the right action or returning an error


Aurel300:
Thanks!

PR #545, opened 2021-06-09
Replaced unreachable panic with error

ani003:
Inexplicable unreachable panic triggered by this:
```
fn main() {
    println!("Hi");
    let mut t = [0];
    t[0] = 1;
}
```

Replaced panic by a generic error message

### DIFF
--- a/prusti-interface/src/environment/borrowck/regions.rs
+++ b/prusti-interface/src/environment/borrowck/regions.rs
@@ -84,9 +84,28 @@ impl PlaceRegions {
                         not supported".to_string()
                     ));
                 }
-                x => {
+                mir::ProjectionElem::Index(_) => {
                     return Err(PlaceRegionsError::Unsupported(
-                        format!("determining the region of projection {:?} is not supported", x)
+                        "determining the region of array indexing is \
+                        not supported".to_string()
+                    ));
+                }
+                mir::ProjectionElem::ConstantIndex{..} => {
+                    return Err(PlaceRegionsError::Unsupported(
+                        "determining the region of constant indexing is \
+                        not supported".to_string()
+                    ));
+                }
+                mir::ProjectionElem::Subslice{..} => {
+                    return Err(PlaceRegionsError::Unsupported(
+                        "determining the region of a subslice is \
+                        not supported".to_string()
+                    ));
+                }
+                mir::ProjectionElem::Downcast(_, _) => {
+                    return Err(PlaceRegionsError::Unsupported(
+                        "determining the region of a downcast is \
+                        not supported".to_string()
                     ));
                 }
             })
--- a/prusti-interface/src/environment/borrowck/regions.rs
+++ b/prusti-interface/src/environment/borrowck/regions.rs
@@ -79,31 +79,31 @@ impl PlaceRegions {
             .map(|elem| match elem {
                 mir::ProjectionElem::Field(f, _) => Ok(f.index()),
                 mir::ProjectionElem::Deref => {
-                    return Err(PlaceRegionsError::Unsupported(
+                    Err(PlaceRegionsError::Unsupported(
                         "determining the region of a dereferentiation is \
                         not supported".to_string()
                     ));
                 }
                 mir::ProjectionElem::Index(_) => {
-                    return Err(PlaceRegionsError::Unsupported(
+                    Err(PlaceRegionsError::Unsupported(
                         "determining the region of array indexing is \
                         not supported".to_string()
                     ));
                 }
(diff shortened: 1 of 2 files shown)


fpoli:
Could you add a regression test in `prusti-tests/tests/verify/fail/unsupported`?

ani003:
The simplest program I found causing this panic is the one above. The strange thing is that if we remove the print statement, the program is verified successfully. Hence I did not add it as a test.

fpoli:
Isn't it possible to add a test with an expected error message for the `println` line? The goal is to test that Prusti doesn't crash on such an example. It's fine to have in `prusti-tests/tests/verify/fail` programs that don't verify due to usages of unsupported statements, even if by removing the unsupported statements the program would verify.

dario23:
> The strange thing is that if we remove the print statement, the program is verified successfully.

ah, was already wondering why this didn't come up with any of my array testcases :-)

ani003:
> Isn't it possible to add a test with an expected error message for the `println` line? 

Again, the strange thing is that the error propagates back to the line `t[0] = 1`. I will be pushing the test shortly.

Aurel300:
Courtesy of @ani003 the desugaring of the program looks like this:

```rust
#![feature(register_tool)]
#![register_tool(prusti)]
#[prelude_import]
use ::std::prelude::rust_2015::*;
#[macro_use]
extern crate std;
fn main() {
    {
        ::std::io::_print(::core::fmt::Arguments::new_v1(&["Hi\n"],
                                                         &match () {
                                                              () => [],
                                                          }));
    };
    let mut t = [0];
    t[0] = 1;
}
```

No matter where the `println!` statement is (before the declaration, between the declaration and access, or after the array access), the crash happened at `t[0] = 1`.

We won't get a good answer without having a close look at the MIR/CFG here. Clearly the presence of the print statement shuffles *something* around, but the desugared source code does not reflect this.

dario23:
> > The strange thing is that if we remove the print statement, the program is verified successfully.
> 
> ah, was already wondering why this didn't come up with any of my array testcases :-)

ah, i think `println!()` in general isn't really supported, just that some recent changes (likely array/slice-related) made it not outright reject anything containing arrays (take a look at the MIR generated for a single `println!()` invocation).

just a `println!()` in a function yields an `unsupported feature` error as expected, but as far as i can tell later in the code than the error. i added `trace` in `prusti-interface/src/environment/borrowck/regions.rs` and the order of places for which `PlaceRegions::translate_place` is called is not deterministic accros invocations on the same code (not sure what that means), but at some point we hit the `_13[_14]`, which is the array index expression, and panic on it.

what's interesting is that for the case without `println!()`, only `translate_place(_0)` is called, which i think is the return value. i'm guessing that some machinery is never triggered, because there are no loans in the whole program in the non-`println!()` case.

dario23:
> some machinery is never triggered

that would be [this loop](https://github.com/viperproject/prusti-dev/blob/master/prusti-interface/src/environment/polonius_info.rs#L458) from what i can tell. i'm not exactly sure what it does and why it does it the way it does, in too much detail, but i think we're fine without what it does for simple array accesses like in the example (which don't need any expiry info). returning a dummy value for `mir::ProjectionElem::Index(..)` makes it pass this stage and run until emitting the "cast statements that create loans are not supported" error that `println!()` alone would as well.

prusti-interface/src/environment/borrowck/regions.rs
```diff
@@ -84,7 +84,11 @@ impl PlaceRegions {
                         not supported".to_string()
                     ));
                 }
-                x => unreachable!("{:?}", x),
+                x => {
+                    return Err(PlaceRegionsError::Unsupported(
+                        format!("determining the region of projection {:?} is not supported", x)
+                    ));
+                }
```
fpoli: So far we avoided dumping internal representations of the program in the "unsupported" error messages, to ensure that the messages are not too intimidating for the users. ("internal" error messages, by contrast, can be as ugly as we want.)

So, I think we should remove the `{:?}`. The message `"determining the region of a projection is not supported"` would be fine, but a bit too vague. The best thing would be to complete the match and add a separate error message for `Index`, `Downcast`, `ConstantIndex`, `Subslice` (e.g. `"determining the region of a downcast/array indexing/... is not supported"`).

ani003: Updated and pushed

Aurel300:
@ani003 Please resolve the clippy warnings (you can see them in the [files view](https://github.com/viperproject/prusti-dev/pull/545/files)).

PR #547, opened 2021-06-10
Add extensible test suite for cargo-prusti

### LINKED ISSUE #3: Test Prusti on a collection of crates
We would need a way to run Prusti (`cargo-prusti`) on multiple test crates. This would allow to have regression tests for issues like #2 and #528.

Ideally, the test suite should check that the position and message of the reported errors matches the expected one, as `compiletest-rs` does.

### LINKED ISSUE #528: cargo-prusti ignores specification
I tried running `cargo-prusti` on the following crate:

`Cargo.toml`:
```
[package]
name = "example"
version = "0.1.0"
edition = "2018"

[dependencies]
prusti-contracts = { git = "https://github.com/viperproject/prusti-dev.git" }
```

`src/main.rs`:
```
use prusti_contracts::*;

#[requires(x > 123)]
fn test(x: i32) {
    assert!(x > 123);
}

fn main() {
    assert!(true);
}
```

However, the assertion in the `test` function does not verify and Prusti reports an error. I checked the Viper dump and the `x > 123` precondition is not assumed at the beginning of the function, but it should be. I suspect that for some reason all specifications are ignored while running `cargo-prusti`. Maybe this is related to the `cfg(prusti)` compilation flag.

This issue also shows that `cargo-prusti` is not properly tested. We really need #3.

fpoli:
This PR uses the `cargo_test_support` crate from the test suite of [rust-lang/cargo](https://github.com/rust-lang/cargo) to test the output of `cargo-prusti`.

Closes #3 and fixes #528.


prusti-tests/tests/cargotest.rs
```diff
@@ -0,0 +1,152 @@
...
+    let mut test_builder = project_builder.build().process(cargo_prusti_path());
+    let opt_expected_stdout = fs::read_to_string(project_path.join("output.stdout")).ok();
+    let opt_expected_stderr = fs::read_to_string(project_path.join("output.stderr")).ok();
+    if let Some(ref expected_stdout) = opt_expected_stdout {
+        test_builder.with_stdout(expected_stdout);
+    }
+    if let Some(ref expected_stderr) = opt_expected_stderr {
+        test_builder.with_status(101).with_stderr(expected_stderr);
+    }
+
```
Aurel300, line 154: Shouldn't we also clean up the symlinks?

fpoli: They are created in a temporary folder (e.g. `prusti-dev/target/debug/tmp/cit/t2/`) which is automatically deleted before running the test a second time.

fpoli: The symlinks are created during the `test_builder.run()` below. The statements above are just declarations.

fpoli:
Regarding `cargo_verify/failing_crate`, running `PRUSTI_PRINT_DESUGARED_SPECS=true cargo-prusti` outputs the following, which doesn't actually contain desugared specifications:

```
#![feature(register_tool)]
#![register_tool(prusti)]
#[prelude_import]
use std::prelude::rust_2018::*;
#[macro_use]
extern crate std;
mod utils {
    use prusti_contracts::*;
    pub fn requires_large_number(x: i32) -> i32 { x }
}
use utils::requires_large_number;
fn bad_client() { requires_large_number(10); }
fn main() { if !true { ::core::panicking::panic("assertion failed: true") }; }
```

prusti-tests/tests/cargo_verify/failing_crate/output.stderr
```diff
@@ -0,0 +1,69 @@
...
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
+   Compiling [..]
```
Aurel300: Surely the number of these lines is not pre-determined?

fpoli: It's one for each dependency. I should add the `Cargo.lock` file to keep it reproducible.

fpoli: The order of `Compiling <dependency name>` is not deterministic because dependencies form a DAG, not a chain. That's why I had to add `[..]` in various places.

Aurel300: I see. This is kind of a nitpick then, but I still think it would be cleaner to download and build dependencies separately, then only compare the output of the main run. Unfortunately, this is [not supported by `cargo`](https://github.com/rust-lang/cargo/issues/2644), although that thread has various workarounds. For example, one can run `cargo build` with an "empty" project first, then add the real source code…

fpoli: `[cargo fetch;] cargo build; cargo-prusti` works fine and it's actually useful to test that the crate compiles without Prusti; I'll use that. Still, I have no clue how `cargo` manages to switch to the other implementation of `prusti-contracts` without recompiling it when running `cargo-prusti`.

prusti-tests/tests/cargotest.rs
```diff
@@ -0,0 +1,170 @@
...
+        eprintln!("Symlink dir {:?} -> {}", crate_name, prusti_dev_path.join(crate_name).display());
+        project_builder = project_builder.symlink_dir(
+            prusti_dev_path.join(crate_name).as_path(),
+            &Path::new(crate_name)
+        );
+    }
+
+    // Fetch dependencies
+    let project = project_builder.build();
+    project.process("cargo").arg("build").run();
```
LDY1998, line 142: Hi @fpoli , I am trying to add a test case for `#[extern_spec]` with cargo-prusti, the attribute `#[extern_spec]` requires a rewrite for all the functions in the declared module but it will error out if we build the project with `cargo` first:
```rust
// Report an error because function is without body
#[extern_spec]
mod std {
    mod mem {
        use prusti_contracts::*;

        #[ensures(*a == old(*b) && *b == old(*a))]
        pub fn swap(a: &mut i32, b: &mut i32);
        // pub fn swap<T: std::cmp::PartialEq + Copy>(a: &mut T, b: &mut T);
    }
}
```

Wonder if we actually need this line or is it for compiling the dependencies only? Also, do you run cargo-prusti on the test case to see what the actual output is? I am having problem matching my expect output with the actual one

Aurel300: @LDY1998 As we discussed in the meetings, this is something that you need to solve. The reason `cargo build` is executed first is to compile all dependencies and to make sure that both the dependencies and the main crate itself can be built with the regular, non-Prusti compiler. If a regular `cargo build` cannot build a crate that uses Prusti, it is a problem.

`cargo prusti` [is executed](https://github.com/viperproject/prusti-dev/blob/b58ced8dfd14ef30582b503d517167ccd771eaff/prusti-tests/tests/cargotest.rs#L137) too, and the stderr and stdout are compared.

fpoli: Exactly as Aurel wrote: the `cargo build` is needed to test that the crate compiles with the regular compiler, then `cargo-prusti` runs too. To reproduce the test it should be enough to run the cargo commands in the crate of the test, creating the symlinks for `prusti-contracts` and similar creates as the testsuite does.

LDY1998: Sorry if I misunderstand it, the syntax for `#[extern_spec]` cannot be compiled with cargo since all the fns we specify there don't have body, it will produce an error and fail when we run cargo on the crate. So do we want to make this syntax work? Or is it really necessary to run `cargo build` since `cargo-prusti` will check the crate to see if it can be compiled?

Aurel300: Yes, it is necessary that any crates using Prusti can also compile with the regular Rust compiler.

Please double check what code you are giving the compiler (e.g. with [`print_desugared_specs`](https://viperproject.github.io/prusti-dev/dev-guide/config/arguments.html)). `#[extern_spec]` on its own compiles to valid Rust code. You need to make sure that this processing still happens.

LDY1998: I think I got it now:

`cargo build` will do the path resolution of the attribute, but the attribute is not called to actually make the rewrite happens, it just makes sure that the attribute is somewhere defined in crate/dependencies.

As the rewrite is not triggered, the syntax for `#[extern_spec]` will be a function without body, hence it will error out so that `cargo` cannot actually build it. 

The `print_desugared_specs` won't be able to trigger here since we don't even reach prusti, the process is stopped when we are building with `cargo`. 

The only way to get current syntax pass `cargo` is to have the user write some stub body for the function (which is not what we want) like:
```rust
#[extern_spec]
mod std {
    mod mem {
        use prusti_contracts::*;

        #[ensures(*a == old(*b) && *b == old(*a))]
        pub fn swap(a: &mut i32, b: &mut i32) {
            unimplemented!();
        }
        // pub fn swap<T: std::cmp::PartialEq + Copy>(a: &mut T, b: &mut T);
    }
}
```
, or we must change the attribute to a procedural macro:
```rust
extern_spec!(
mod std {
    mod mem {
        use prusti_contracts::*;

        #[ensures(*a == old(*b) && *b == old(*a))]
        pub fn swap(a: &mut i32, b: &mut i32)
    }
}

)
```

PR #553, opened 2021-06-14
Migrate to GitHub Viper releases

### LINKED ISSUE #339: Unexpected output
when checking a program with prusti-dev, i obtain the following unexpected output:

```bash
    Checking formal_test v0.1.0 (/Users/user/Downloads/TMP/formal_test)
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/user/Downloads/TMP/prusti-dev2/viper_tools/backends/ch.qos.logback.logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/user/Downloads/TMP/prusti-dev2/viper_tools/backends/silicon.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
    Finished dev [unoptimized + debuginfo] target(s) in 5.47s
```

fpoli:
Viper releases migrated to GitHub - Thanks @ArquintL!

This PR adds a `viper-toolchain` file that contains the version to be downloaded by `./x/py setup`. It also fixes #339.

Please take kare of the following:
- [x] Fix UI tests to remove any `SLF4J` output.
- [ ] Extend the auto-update script to update `viper-toolchain` to the latest nightly Viper release that is available on GitHub.


ArquintL:
As already mentioned to @vakaras: The auto-update could use the following command to get the URL to the latest stable or nightly build (whichever is newer) and store it in the `viper-toolchain` file (instead of just the tag name):
```
curl -s https://api.github.com/repos/viperproject/viper-ide/releases | \
 jq ".[0].assets | map(select(.name == \"ViperToolsLinux.zip\")) | .[0].browser_download_url"
```

However, having a nightly release seems less than ideal for your use case. One option would be to store the silver, carbon, and silicon commits and only trigger a nightly release when either of them have changed. Do you have any opinions?

fpoli:
> As already mentioned to @vakaras: The auto-update could use the following command to get the URL to the latest stable or nightly build (whichever is newer) and store it in the `viper-toolchain` file (instead of just the tag name).
> However, having a nightly release seems less than ideal for your use case. One option would be to store the silver, carbon, and silicon commits and only trigger a nightly release when either of them have changed. Do you have any opinions?

* I don't have a strong opinion regarding storing the tag vs the full url in the `viper-toolchain` file. Storing just the version is more `rust-toolchain`-style, but it requires hardcoding part of the url in `x.py`.
* Downloading a pre-compiled zip seems easier because we don't have to add a Java compilation step to our setup and we can delegate the task of making sure that Viper compiles to someone else (debugging failures, finding matching silicon+carbon+viper versions...) :wink: I agree that it's strange to download a JAR that also contains the Viper server, but it's surely better than the current zip on which we depend.
* If we really end up doing it, using git submodules is in my opinion better than storing the silver, carbon and silicon commits in files.

ArquintL:
> * I don't have a strong opinion regarding storing the tag vs the full url in the `viper-toolchain` file. Storing just the version is more `rust-toolchain`-style, but it requires hardcoding part of the url in `x.py`.
> * Downloading a pre-compiled zip seems easier because we don't have to add a Java compilation step to our setup and we can delegate the task of making sure that Viper compiles to someone else (debugging failures, finding matching silicon+carbon+viper versions...) 😉 I agree that it's strange to download a JAR that also contains the Viper server, but it's surely better than the current zip on which we depend.
> * If we really end up doing it, using git submodules is in my opinion better than storing the silver, carbon and silicon commits in files.

This was more like thinking out loud. Currently the ViperServer repo triggers a nightly release every morning which would result in many PRs on your end even though the ZIP file remains unchanged. Storing and checking the commit hashes could be something on my end such that there only is a new Viper Tools release when something indeed has changed

> Missing? :-)

Fixed

fpoli:
> the ViperServer repo triggers a nightly release every morning which would result in many PRs on your end even though the ZIP file remains unchanged

Our update-dependencies bot runs twice per month, so it shouldn't be a problem.

vakaras (approved):
LGTM

PR #556, opened 2021-06-14
Update predicate syntax in docs

dario23:


dario23:
Merge after #537 is merged.

PR #560, opened 2021-06-16
Add unsupported error function to TypeVisitor

ani003:
This PR adds `unsupported` function to `TypeVisitor` trait. Also fixes the panic caused [here](https://github.com/viperproject/prusti-dev/blob/02732f22ec0a9afb5438bbd9626309c87125cf87/prusti-viper/src/utils/type_visitor.rs#L286)

### DIFF
--- a/prusti-viper/src/utils/type_visitor.rs
+++ b/prusti-viper/src/utils/type_visitor.rs
@@ -17,6 +17,8 @@ pub trait TypeVisitor<'tcx>: Sized {
 
     fn tcx(&self) -> TyCtxt<'tcx>;
 
+    fn unsupported<S: ToString>(&self, _msg: S) -> Self::Error;
+
     fn visit_ty(&mut self, ty: Ty<'tcx>) -> Result<(), Self::Error> {
         trace!("visit_ty({:?})", ty);
         self.visit_sty(ty.kind())?;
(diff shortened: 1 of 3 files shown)


prusti-viper/src/utils/type_visitor.rs
```diff
@@ -275,15 +277,22 @@ pub fn walk_raw_ptr<'tcx, E, V: TypeVisitor<'tcx, Error = E>>(
...
+        match cl_substs.sig().no_bound_vars() {
+            None => {
+                return Err(visitor.unsupported(
+                    "higher-ranked lifetimes and types are not supported"
+                ))
+            }
+
+            Some(x) => x
+        };
+
```
Aurel300:
```suggestion
        cl_substs.sig().no_bound_vars()
            .ok_or_else(|| visitor.unsupported("higher-ranked lifetimes and types are not supported"));
```

ani003: @Aurel300 this change is throwing an error while building. It seems that a `?` is missing at the end of line 287
```suggestion
        cl_substs.sig()
                 .no_bound_vars()
                 .ok_or_else(|| visitor.unsupported("higher-ranked lifetimes and types are not supported"))?; 
```

Aurel300:
Please add a testcase.

PR #561, opened 2021-06-16
Improve snapshots of argumentless enums

### LINKED ISSUE #562: Regression: enum test fails with new viper-tools
this snippet from the quick tests enum.rs file seems to make prusti panic reproducibly, while verifying fine previously:
```
pub enum E3<T> {
    V1(T),
    V2(T),
}

pub fn test3<T>(e: &mut E3<T>) -> &mut T {
    match *e {
        E3::V1(ref mut b) => b,
        E3::V2(ref mut b) => b,
    }
}
```
`git bisect` seems to think it's related to the merge of #553, but that by itself got merged OK, and testing this locally using the exact viper tools version used in the PR verification build still fails, so i'm not sure what it actually is.

Aurel300:
- Fix var-less quantifiers in the injectivity and constructor axioms.
- Omit injectivity axioms for argumentless variants.

Closes #562

### DIFF
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -452,6 +452,7 @@ impl Expr {
     }
 
     pub fn forall(vars: Vec<LocalVar>, triggers: Vec<Trigger>, body: Expr) -> Self {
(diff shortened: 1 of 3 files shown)


vakaras (approved):
LGTM

PR #564, opened 2021-06-17
Add snapshot encoding for slices

dario23:


dario23:
working: taking snapshots of slices, encoding of pure functions using slices, including bounds checks/`.len()` calls.
in progress: using `sorted(slc)` as a precondition and actually get anything from that information; e.g.
```rust
use prusti_contracts::*;

fn main() {}


predicate! {
    fn sorted(s: &[i32]) -> bool {
        forall(|i: usize, j: usize| (0 <= i && i < j && j < s.len()) ==> s[i] <= s[j])
    }
}

#[requires(sorted(s))]
fn requires_sorted(s: &[i32]) {
    if s.len() > 3 {
        assert!(s[0] <= s[3]);
    }
}
```
similar to the issues that we discussed, i can manually fix the generated viper file to inhale a forall over `lookup_pure` calls instead, making it verify correctly, but manually inhaling the equivalent `read$Snap$..` info isn't enough.

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -1345,9 +1361,17 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
+                                    place_ty,
+                                    self.encode_place(place).with_span(span)?.0,
+                                ).with_span(span)?;
+
+                                state.substitute_value(&opt_lhs_value_place.unwrap(), snap_len);
+                            }
                             _ => {
                                 return Err(SpannedEncodingError::unsupported(
-                                    "checking the length of anything but arrays in pure code is not implemented yet",
+                                    "checking the length of anything but arrays and slices in pure code is not implemented yet",
```
Aurel300: I guess based on [`Rvalue::Len` documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.Rvalue.html#variant.Len), the variant should never be emitted for anything other than arrays or slices?

dario23: hm, true. so go with `SpannedEncodingError::internal` then? or full-on `unreachable!()`?

Aurel300: `unreachable!` should be fine, with an appropriate comment.

prusti-tests/tests/verify/pass/arrays/slice-sorted.rs
```diff
@@ -0,0 +1,24 @@
...
+        assert!(s[0] <= s[3]);
+    }
+}
+
+/*
+#[requires(sorted(&a[0..2]))]
+fn requires_sorted_start(a: [i32; 3]) {
+    assert!(a[0] <= a[1]);
+}
+*/
```
Aurel300, line 24: Is this meant to be commented out?

dario23: yes; the `&a[0..2]` part doesn't work yet (only the unsize-cast that turns a full array into a slice like happens for `arr.len()`).

now that i think about it, i guess implementing full slicing (at least of arrays) makes sense within this PR.

prusti-tests/tests/verify/pass/arrays/slice-split.rs
```diff
@@ -0,0 +1,19 @@
+use prusti_contracts::*;
+
+fn main() {
+    let a = [0; 13];
+    let (s0, s1) = split_at(&a, 7);
+
+    assert!(s0.len() == 7);
+    assert!(s1.len() == 6);
```
Aurel300, line 8: Given the postconditions, you should also be able to prove assertions about concrete values (e.g. `assert!(s0[2] == 0);`), is that right? If so, please add an assertion like that.

prusti-tests/tests/verify/pass/arrays/slice-sorted.rs
dario23: alright, within snapshots we can now handle slicing of both arrays and slices :-)

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -1345,15 +1406,62 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
-                                ));
+                            ty::TyKind::Slice(..) => {
+                                let snap_len = self.encoder.encode_snapshot_slice_len(
+                                    place_ty,
+                                    self.encode_place(place).with_span(span)?.0,
+                                ).with_span(span)?;
+
+                                state.substitute_value(&opt_lhs_value_place.unwrap(), snap_len);
                             }
+                            _ => span_bug!(span, "length should only be requested on arrays or slices"),
```
Aurel300, line 1417: Does a `span_bug!` turn into a panic? If so, can we use an `internal` error instead?

dario23: i think it does, just like `unreachable!()` which we thought about earlier. what made me think more towards `span_bug!` instead of an internal error is that as far as we understand it, if we encounter anything else than array or slice here, it's a bug in `rustc` more than in prusti.

Aurel300: Ah, you're right, it's the `Len` thing. Never mind then, `span_bug` to panic makes sense.

PR #566, opened 2021-06-18
Verify selectionsort

dario23:


prusti-tests/tests/verify/pass/arrays/selection_sort.rs
```diff
@@ -0,0 +1,56 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+
+// TODO: replace 10 with a.len()
```
Aurel300: Is this something that will be part of this PR?

dario23: that's a relict from me writing the test before #538 was merged, and now that i checked, it'll also need more slice snapshot support for using it in the invariants (unsize cast, `.len()` call, .. -> #564). For the non-pure part (i.e. the loop conditions) we can use `.len()` already, so i'll change that.

PR #568, opened 2021-06-22
Verify mergesort's merge function on static arrays

dario23:
Slices aren't far enough along yet, but the main (difficult) loop of
mergesort can be verified for two static pre-sorted sub-arrays already

### DIFF
--- a/prusti-tests/tests/verify/pass/arrays/merge.rs
+++ b/prusti-tests/tests/verify/pass/arrays/merge.rs
@@ -1,0 +1,60 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+
+predicate! {
+    fn sorted3(a: [i32; 3]) -> bool {
+        forall(|i: usize, j: usize| (0 <= i && i < j && j < 3) ==> a[i] <= a[j])
+    }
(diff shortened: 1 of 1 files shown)


prusti-tests/tests/verify/pass/arrays/merge.rs
```diff
@@ -0,0 +1,60 @@
+use prusti_contracts::*;
+
+fn main() {}
```
Aurel300: I guess the postcondition of `merge` is not strong enough to verify e.g. `merge([1, 3, 5], [2, 4, 6])[2] == 3`, and we would need at least (multi)set conversions to do this?

What about a client test like this:
```suggestion
fn main() {
    let x = [1, 3, 5];
    let y = [2, 4, 6];
    let m = merge(x, y);
    assert!(m[1] <= m[4]);
}
```

dario23: re `merge(..)[2] == 3`, i think so too. doesn't verify currently, and not specifying the output multiset seems like the obvious culprit.

added the suggested change in main.

PR #589, opened 2021-07-07
Update dependencies (nightly-2021-07-15)

viper-admin:
* [x] Update rustc version to `nightly-2021-07-15`.
* [x] Manualy update outdated dependencies (see the list below).
* [x] Manualy run `cargo update`.

<details><summary>List of direct outdated dependencies:</summary>

```
$ cargo outdated --root-deps-only --workspace

prusti
================
Name        Project  Compat  Latest  Kind    Platform
----        -------  ------  ------  ----    --------
env_logger  0.8.4    ---     0.9.0   Normal  ---

viper
================
Name        Project  Compat  Latest  Kind         Platform
----        -------  ------  ------  ----         --------
env_logger  0.8.4    ---     0.9.0   Development  ---

viper-sys
================
Name        Project  Compat  Latest  Kind    Platform
----        -------  ------  ------  ----    --------
env_logger  0.8.4    ---     0.9.0   Normal  ---

prusti-server
================
Name        Project  Compat  Latest  Kind    Platform
----        -------  ------  ------  ----    --------
env_logger  0.8.4    ---     0.9.0   Normal  ---
futures     0.1.31   ---     0.3.15  Normal  ---
reqwest     0.9.24   ---     0.11.4  Normal  ---
tokio       0.1.22   ---     1.8.1   Normal  ---
warp        0.1.23   ---     0.3.1   Normal  ---

prusti-launch
================
Name  Project  Compat  Latest  Kind    Platform
----  -------  ------  ------  ----    --------
nix   0.21.0   ---     0.22.0  Normal  cfg(unix)

test-crates
================
Name        Project  Compat  Latest  Kind    Platform
----        -------  ------  ------  ----    --------
env_logger  0.8.4    ---     0.9.0   Normal  ---

vir
================
Name       Project  Compat  Latest  Kind    Platform
----       -------  ------  ------  ----    --------
thiserror  1.0.25   ---     1.0.26  Normal  ---

systest
================
Name        Project  Compat  Latest  Kind   Platform
----        -------  ------  ------  ----   --------
env_logger  0.8.4    ---     0.9.0   Build  ---
```
</details>

@fpoli could you take care of this?

PR #590, opened 2021-07-07
Update dependencies.

vakaras:


Aurel300:
Is this instead of the scheduled #589?

vakaras:
> Is this instead of the scheduled #589?

No. For some reason, the scheduled job failed to open a PR last two times, so I tried to create my own PR. The CI is failing with Clippy crash, so it seems that we need to find the version of the Rust compiler that works.

PR #591, opened 2021-07-07
Remove polonius hack non optimized

vakaras:


prusti-tests/tests/verify/fail/unsupported/static_const.rs
```diff
@@ -0,0 +1,8 @@
+const fn foo() -> usize {
+    1
+}
+
+fn main() { //~ ERROR: [Prusti internal error] tried to encode a projection that accesses the field 0 of a variant without first downcasting its enumeration std::option::Option<std::fmt::Arguments
```
Aurel300: This thing is unsupported because of `assert_eq!`, not because of the const, right?

vakaras: Yes, you are right.

PR #592, opened 2021-07-09
Fix panic caused by constant tuples

ani003:
This PR fixes panic caused by the below program by matching the case with the proper branch
```
const C: (i32, i32) = (0, 0);

fn main() {
    let _ = C.0;
}
```


Aurel300:
Add a test case, please.

PR #595, opened 2021-07-09
Return unsupported error for variadic functions

ani003:
This PR fixes the panic caused by external calls to variadic functions.


Aurel300:
We should test this somehow. What happens if you put this into a testcase file?

```rust
unsafe extern "C" fn func(mut args: ...) {}
```

PR #596, opened 2021-07-12
Verify each method separately.

vakaras:
This PR is preparation for reenabling purification.

### DIFF
--- a/prusti-common/src/vir/ast/expr.rs
+++ b/prusti-common/src/vir/ast/expr.rs
@@ -1631,6 +1631,18 @@ impl Expr {
         let mut patcher = TypePatcher { substs: substs };
         patcher.fold(self)
     }
+
+    /// Is this expression a constant?
+    pub fn is_constant(&self) -> bool {
+        match self {
+            Expr::Const(_, _) => true,
+            Expr::UnaryOp(_, box subexpr, _) => subexpr.is_constant(),
+            Expr::BinOp(_, box subexpr1, box subexpr2, _) => {
+                subexpr1.is_constant() && subexpr2.is_constant()
+            }
+            _ => false,
+        }
+    }
 }
 
 impl PartialEq for Expr {
--- a/prusti-common/src/vir/ast/function.rs
+++ b/prusti-common/src/vir/ast/function.rs
@@ -101,3 +101,12 @@ impl WithIdentifier for Function {
         compute_identifier(&self.name, &self.formal_args, &self.return_type)
     }
 }
+
+#[derive(PartialEq, Eq, Hash, Clone)]
+pub struct FunctionIdentifier(String);
+
+impl From<String> for FunctionIdentifier {
+    fn from(string: String) -> Self {
+        Self(string)
+    }
+}
--- a/prusti-common/src/vir/ast/function.rs
+++ b/prusti-common/src/vir/ast/function.rs
@@ -72,6 +72,15 @@ impl Function {
             Position::default(),
         )
     }
+
+    /// Does the function has a body that does not depend neither on
+    /// function parameters nor on the heap?
+    pub fn has_constant_body(&self) -> bool {
+        match self.body {
+            Some(ref expr) => expr.is_constant(),
+            None => false,
+        }
+    }
 }
 
 pub fn compute_identifier(name: &str, formal_args: &[LocalVar], return_type: &Type) -> String {
--- a/prusti-common/src/vir/ast/function.rs
+++ b/prusti-common/src/vir/ast/function.rs
@@ -102,7 +102,7 @@ impl WithIdentifier for Function {
     }
 }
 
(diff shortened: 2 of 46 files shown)


prusti-viper/src/encoder/encoder.rs
```diff
@@ -229,8 +231,9 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         self.error_manager.borrow_mut()
     }
 
-    pub fn get_viper_program(&self) -> vir::Program {
+    pub fn finalize_viper_program(&self, name: String) -> vir::Program {
         vir::Program {
+            name,
             domains: self.get_used_viper_domains(),
             fields: self.get_used_viper_fields(),
             builtin_methods: self.get_used_builtin_methods(),
```
Aurel300: Is this going to add more and more (potentially unused) domains and fields to each program as we progress through the processing queue?

vakaras: I have created this PR just to see whether the test suite goes through. I will fix that eventually.

prusti-viper/src/encoder/counterexample_translation.rs
```diff
@@ -137,7 +137,7 @@ impl<'ce, 'tcx> CounterexampleTranslator<'ce, 'tcx> {
             def_id,
             silicon_counterexample,
             tcx: encoder.env().tcx(),
-            is_pure: encoder.is_pure(def_id),
+            is_pure: false, // No verified functions are pure. encoder.is_pure(def_id),
```
Aurel300, line 140: A lot more can be simplified in counterexamples once this is the case. Please add a `TODO` at least.

prusti-common/src/vir/ast/expr.rs
```diff
@@ -1631,6 +1631,18 @@ impl Expr {
...
+
+    /// Is this expression a constant?
+    pub fn is_constant(&self) -> bool {
+        match self {
+            Expr::Const(_, _) => true,
+            Expr::UnaryOp(_, box subexpr, _) => subexpr.is_constant(),
+            Expr::BinOp(_, box subexpr1, box subexpr2, _) => {
+                subexpr1.is_constant() && subexpr2.is_constant()
+            }
+            _ => false,
```
Aurel300, line 1643: What is this for? You might want to add (domain) function applications with only constant arguments. One example that comes to mind is the "unit domain" produced as the snapshot for data-less types (which is a 0-arity domain function `unit()` or something).

(Now I see this is actually just moved from older code. Still might be useful depending on where it is being called.)

prusti-common/src/vir/ast/function.rs
```diff
@@ -72,6 +72,15 @@ impl Function {
             Position::default(),
         )
     }
+
+    /// Does the function has a body that does not depend neither on
+    /// function parameters nor on the heap?
```
Aurel300:
```suggestion
    /// Does the function have a body that depends neither on
    /// function parameters nor on the heap?
```

prusti-interface/src/prusti_error.rs
```diff
@@ -100,6 +130,7 @@ impl PrustiError {
 
     /// Report the encoding error using the compiler's interface
     pub fn emit(self, env: &Environment) {
+        assert!(!self.is_disabled);
```
Aurel300, line 140: Why not add the logic for skipping error emission here into `emit`, instead of requiring the caller to use `emit` or `cancel`?

prusti-tests/tests/verify/fail/equality/pure-post-1.rs
```diff
@@ -1,3 +1,5 @@
+// ignore-test: issue https://github.com/viperproject/prusti-dev/issues/599
```
Aurel300: Is this a new case of that issue?

prusti-viper/src/encoder/counterexample_translation.rs
vakaras: Added. Thanks!

prusti-common/src/vir/ast/expr.rs
vakaras: Good point. However, I would prefer in this PR to touch VIR as little as possible.

prusti-common/src/vir/ast/function.rs
vakaras: Thanks!

prusti-interface/src/prusti_error.rs
vakaras: My intention was to allow disabling errors only in places where the caller expects them. Otherwise, we may end up with bugs that silently swallow errors.

prusti-tests/tests/verify/fail/equality/pure-post-1.rs
vakaras: Works now.

PR #600, opened 2021-07-13
Only report verification success if there were no encoding errors.

vakaras:


prusti/src/verifier.rs
```diff
@@ -53,10 +53,14 @@ pub fn verify<'tcx>(
 
         match verification_result {
             VerificationResult::Success => {
-                user::message(format!(
-                    "Successful verification of {} items",
-                    verification_task.procedures.len()
-                ));
+                if env.has_errors() {
+                    user::message("Verification result is inconclusive.");
```
Aurel300, line 57: Isn't this a bit vague? We should at least say "because errors were encountered during encoding" or similar…

vakaras: Good point. I will fix it.

PR #603, opened 2021-07-14
encoding of slice creation/indexing

dario23:


dario23:
status: slicing basically works, e.g. this testcase passes:

```
use prusti_contracts::*;

fn main() {}

fn slice(a: [i32; 3]) {
    let s = &a[1..3];
    assert!(s.len() == 2);
    assert!(a[1] == s[0]);
    assert!(a[2] == s[1]);
}
```

slicing twice doesn't work, i think there is something left to do when the first slice/reference expires. the current error is

```
thread 'rustc' panicked at 'assertion failed: perm_amount.is_valid_for_specs()', prusti-viper/src/encoder/foldunfold/mod.rs:890:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

fwiw: this is my slice-twice testcase:

```
use prusti_contracts::*;

fn main() {}

fn slice(a: [i32; 3]) {
    let s = &a[1..3];
    assert!(s.len() == 2);

    let s = &a[0..2];
    assert!(a[2] == s[1]);
}
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2540,27 +2538,54 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         let (encoded_lhs, encode_stmts, lhs_ty, _) = self.encode_place(
             &destination.as_ref().unwrap().0,
             ArrayAccessKind::Mutable(None, location),
-        ).with_span(span)?;
+        )?;
         stmts.extend(encode_stmts);
-        assert!(lhs_ty.is_slice(), "non-slice LHS type '{:?}' not supported yet", lhs_ty);
+        if !lhs_ty.is_slice() { return Err(EncodingError::unsupported(
+            format!("Non-slice LHS type '{:?}' not supported yet", lhs_ty)
+        ))}
```
```suggestion
        if !lhs_ty.is_slice() {
            return Err(EncodingError::unsupported(
                format!("Non-slice LHS type '{:?}' not supported yet", lhs_ty)
            ));
        }
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2569,66 +2594,55 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        // construction of the fields..
+        let usize_ty = self.encoder.env().tcx().mk_ty(ty::TyKind::Uint(ty::UintTy::Usize));
+        let start = match &*idx_ident {
+            "std::ops::Range" |
+            "std::ops::RangeFrom" => self.encoder.encode_struct_field_value(encoded_idx.clone(), "start", usize_ty)?,
+            // RangeInclusive is wierdly differnet to all of the other Range*s in that the struct fields are private
+            // and it is created with a new() fn and start/end are accessed with getter fns
+            // See https://github.com/rust-lang/rust/issues/67371 for why this is the case...
+            "std::ops::RangeInclusive" => return Err(
+                EncodingError::unsupported("RangeInclusive currently not supported".to_string())
```
Aurel300: `RangeInclusive` might not be familiar to the user if they just used the shorthand syntax.

```suggestion
                EncodingError::unsupported("slicing with RangeInclusive (e.g. [x..=y]) currently not supported".to_string())
```

(+ same comment below)

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2569,66 +2594,55 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         self.slice_created_at.insert(location, encoded_lhs);
 
         // TODO: what do we actually do here? this seems a littly hacky.
```
Aurel300: Unsure the comment is still relevant. Is it just that we rely on "sdtlib internals"? If so, let's say that in the comment.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2444,12 +2511,202 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        let idx_ty = self.mir_encoder.get_operand_ty(&args[1]);
+        let idx_ident = self.encoder.env().tcx().def_path_str(idx_ty.ty_adt_def().unwrap().did);
+        trace!("ident: {}", idx_ident);
+
+        self.slice_created_at.insert(location, encoded_lhs);
+
+        // TODO: what do we actually do here? this seems a littly hacky.
+        // there's fields like _5.f$start.val_int on `encoded_idx`, it just feels hacky to
+        // manually re-do them here when we probably just encoded the type and the
+        // construction of the fields..
```
JonasAlaif:
```suggestion
        // TODO: there's fields like _5.f$start.val_int on `encoded_idx`, it just feels hacky to
        // manually re-do and hardcode them here when we probably just encoded the type
        // and the construction of the fields.
```

JonasAlaif: So yes, the relying on "sdtlib internals" is the hacky thing imo. Though if these are unlikely to change then maybe it's ok to hardcode, and just remove this comment?

prusti-viper/src/encoder/procedure_encoder.rs
JonasAlaif: See below

prusti-viper/src/encoder/procedure_encoder.rs
Aurel300: Yeah I guess for now this is ok. I wonder if eventually we can "decouple" from this somehow, but slice indexing seems to be very deeply embedded in the stdlib. There is the [`Index`](https://doc.rust-lang.org/std/ops/trait.Index.html#impl-Index%3CI%3E-1) implementation for slices, which itself requires a [`SliceIndex`](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html) implementation. The concrete `Range*` types are mentioned [there](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html#impl-SliceIndex%3Cstr%3E). Maybe a long time down the line we could extern spec these trait implementations in `prusti-stdlib` somehow, instead of hard-coding the `Range*` types here.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2444,12 +2511,204 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            "std::ops::RangeFull" | "core::ops::RangeFull" |
+            "std::ops::RangeToInclusive" | "core::ops::RangeToInclusive" => vir::Expr::from(0),
+            _ => unreachable!("{}", idx_ident)
+        };
+        let end = match &*idx_ident {
+            "std::ops::Range" | "core::ops::Range" |
+            "std::ops::RangeTo" | "core::ops::RangeTo" =>
+                self.encoder.encode_struct_field_value(encoded_idx.clone(), "end", usize_ty)?,
+            "std::ops::RangeInclusive" | "core::ops::RangeInclusive" => return Err(
+                EncodingError::unsupported("RangeInclusive currently not supported".to_string())
```
JonasAlaif:
```suggestion
                EncodingError::unsupported("slicing with RangeInclusive (e.g. [x..=y]) currently not supported".to_string())
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2444,12 +2511,204 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            "Pure function call location {:?} has label {}",
+            location, label
+        );
+        self.label_after_location.insert(location, label);
+
+        // plan
+        //
+        // [x] inhale Array$lookup_pure == Slice$lookup_pure with quantifier from start to end
+        // [?] label, encode_transfer_permissions?
+        // [ ] look at the other range types (at least those from the stdlib?
```
JonasAlaif:
```suggestion
        // [x] look at the other range types (at least those from the stdlib?
```

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -900,28 +900,49 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
+                                // construction of the fields..
+                                // Also, duplication with procedure_encoder.rs
+                                let usize_ty = self.encoder.env().tcx().mk_ty(ty::TyKind::Uint(ty::UintTy::Usize));
+                                let start = match &*idx_ident {
+                                    "std::ops::Range" | "core::ops::Range" |
+                                    "std::ops::RangeFrom" | "core::ops::RangeFrom" =>
+                                        self.encoder.encode_struct_field_value(encoded_idx.clone(), "start", usize_ty).with_span(span)?,
+                                    // See procedure_encoder.rs
+                                    "std::ops::RangeInclusive" | "core::ops::RangeInclusive" => return Err(
+                                        SpannedEncodingError::unsupported("RangeInclusive currently not supported".to_string(), span)
```
JonasAlaif:
```suggestion
                                        SpannedEncodingError::unsupported("slicing with RangeInclusive (e.g. [x..=y]) currently not supported".to_string(), span)
```

prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -900,28 +900,49 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
+                                    "std::ops::RangeFull" | "core::ops::RangeFull" |
+                                    "std::ops::RangeToInclusive" | "core::ops::RangeToInclusive" => vir::Expr::from(0),
+                                    _ => unreachable!("{}", idx_ident)
+                                };
+                                let end = match &*idx_ident {
+                                    "std::ops::Range" | "core::ops::Range" |
+                                    "std::ops::RangeTo" | "core::ops::RangeTo" =>
+                                        self.encoder.encode_struct_field_value(encoded_idx.clone(), "end", usize_ty).with_span(span)?,
+                                    "std::ops::RangeInclusive" | "core::ops::RangeInclusive" => return Err(
+                                        SpannedEncodingError::unsupported("RangeInclusive currently not supported".to_string(), span)
```
JonasAlaif:
```suggestion
                                        SpannedEncodingError::unsupported("slicing with RangeInclusive (e.g. [x..=y]) currently not supported".to_string(), span)
```

prusti-viper/src/encoder/procedure_encoder.rs
Aurel300: Actually, if there is a "checklist" left in the comments, that should be removed.

JonasAlaif:
It would be nice to have a Merge test for Slice just like the one for Array, but this can be done in a separate PR imo. This should be merged once all checks pass

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2453,12 +2520,205 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            "Pure function call location {:?} has label {}",
+            location, label
+        );
+        self.label_after_location.insert(location, label);
+
+        // plan
+        //
+        // [x] inhale Array$lookup_pure == Slice$lookup_pure with quantifier from start to end
+        // [?] label, encode_transfer_permissions?
+        // [x] look at the other range types (at least those from the stdlib?
```
JonasAlaif:
```suggestion
```

prusti-viper/src/encoder/mir/pure_functions/interpreter.rs
```diff
@@ -514,29 +514,49 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
...
+                                    _ => unreachable!("{}", idx_ident)
+                                };
+                                let end = match &*idx_ident {
+                                    "std::ops::Range" | "core::ops::Range" |
+                                    "std::ops::RangeTo" | "core::ops::RangeTo" =>
+                                        self.encoder.encode_struct_field_value(encoded_idx.clone(), "end", usize_ty).with_span(span)?,
+                                    "std::ops::RangeInclusive" | "core::ops::RangeInclusive" => return Err(
+                                        SpannedEncodingError::unsupported("slicing with RangeInclusive (e.g. [x..=y]) currently not supported".to_string(), span)
+                                    ),
+                                    "std::ops::RangeToInclusive" => {
```
Pointerbender:
```suggestion
                                    "std::ops::RangeToInclusive" | "core::ops::RangeToInclusive" => {
```

JonasAlaif:
Closes #734. Should be merged once checks pass since it includes a few important fixes

Pointerbender (changes requested):
The `//~^ ERROR generating fold-unfold Viper statements failed` lines should both be removed from the test case when its moved to `pass`, otherwise it'll fail the test without errors :) Thanks for fixing this bug!

Edit: this comment was intended for commit 7c50a4a9e8be22cefcff057b7aa81ba017ea9a69 but somehow ended up on the main PR. It's about file `prusti-tests/tests/verify/pass/slices/toy-index-impl.rs` (renamed from `prusti-tests/tests/verify_partial/fail/issues/issue-709-13.rs`).

JonasAlaif:
> The `//~^ ERROR generating fold-unfold Viper statements failed` lines should both be removed from the test case when its moved to `pass`, otherwise it'll fail the test without errors :) Thanks for fixing this bug!
> 
> Edit: this comment was intended for commit [7c50a4a](https://github.com/viperproject/prusti-dev/commit/7c50a4a9e8be22cefcff057b7aa81ba017ea9a69) but somehow ended up on the main PR. It's about file `prusti-tests/tests/verify/pass/slices/toy-index-impl.rs` (renamed from `prusti-tests/tests/verify_partial/fail/issues/issue-709-13.rs`).

Thanks. I thought I had removed these lines but seems that it didn't make it into the commit. I didn't actually realise that the `//~^` syntax is more than a comment though - useful to know!

Pointerbender:
> I didn't actually realise that the `//~^` syntax is more than a comment though - useful to know!

The rustc development guide explains it in full detail, it has some more cool tricks :)  https://rustc-dev-guide.rust-lang.org/tests/adding.html#error-annotations

Pointerbender:
Hi, I noticed the test suite was failing and had some free time on my hands, so decided to see if I can help with debugging a bit :)

Currently the [test suite fails](https://github.com/viperproject/prusti-dev/actions/runs/1407889309) due to an error on all 3 operating systems at the exact same spot (in test `tests/verify_overflow/pass/rosetta/Selection_sort.rs`): `java.lang.OutOfMemoryError: Java heap space`.

However, I don't think this is because of this particular test, because:

* If I run it with `x.py test Selection_sort.rs` locally then it finishes without errors.
* If I run it with `JAVA_HEAP_SIZE=128 x.py test Selection_sort` (128mb heap for the JVM), then it also finishes correctly. The [default heap size](https://github.com/viperproject/prusti-dev/blob/master/viper/src/viper.rs#L31) that Prusti uses is 512mb, so this is another indication that this test does not consume all available memory.
* In an [earlier CI run](https://github.com/viperproject/prusti-dev/runs/3719621630?check_suite_focus=true#step:7:1128) from September 27th the same test was completing okay (the entire run completed in in 51m 59s time). The test has not been modified since.
* The error originates in the JVM, suggesting that the root cause may lie inside Viper (or its backend) and not in Prusti.
* The test suite on the `more-slices` branch adds just a couple of tests, but the total running time of the CI jumped up significantly. The [last successful CI](https://github.com/viperproject/prusti-dev/actions/runs/1398276141) on the `master` branch took `1h 16m 29s` on October 29th and the [last failing CI](https://github.com/viperproject/prusti-dev/actions/runs/1407889309) on the `more-slices` branch failed after `1h 38m 10s` on November 1st. That's a ~29% percent increase in overall running time for an additional 3 basic verification tests. I don't think the Rust code changes are the root cause here, since the out-of-memory error happens in the JVM.
* The difference in CI running time between the `master` and `more-slices` branches can't be explained on the Rust side by adding just 3 additional verification tests. I'm starting to suspect that the Java garbage collector is having a hard time cleaning old objects from previous test runs. Since the test suite keeps a single JVM around to run every test on, all these old Java objects start to accumulate in memory. Near the very end, I suspect the garbage collector pauses the JVM very often in search of objects to free (explaining the 29% increase in running time) and then at some point it fails with a `java.lang.OutOfMemoryError`.
* When [pulling the latest commits](https://github.com/viperproject/prusti-dev/pull/603/commits/5163985052b28918845998a71fc097207e106a3a) from `master` to `more-slices`, this added 26 additional verification test cases. Together with the other 3 additional test cases, this may have been the tipping point where the JVM could no longer survive with 512mb of heap memory.

So the good news is, I don't think the problem is on the Rust side :) The bad news is, I think Viper may have a memory leak or a circular memory reference somewhere, preventing the Java garbage collector from doing its work. As a work-around, temporarily [increasing the JVM heap size](https://github.com/viperproject/prusti-dev/blob/master/viper/src/viper.rs#L31) should do the trick (and also speed up the test suite a bit if my theory proves correct, because the garbage collector should then pause less often), but for a permanent solution it's probably preferable to identify and solve the memory issue in the JVM.

I hope this helps!

fpoli:
It's possible that the `prusti-server` is preventing the JVM from garbage collecting old Viper AST instances. You can check that by monitoring the heap usage with [VisualVM](https://visualvm.github.io/) while the test suite runs.

To let the GC run, I think that the implementation of `verify`

https://github.com/viperproject/prusti-dev/blob/adfda1e34336dd1fdf54067b4fd248ab592020fc/viper/src/verifier.rs#L113-L303

should be wrapped in a `with_local_frame` call, like here

https://github.com/viperproject/prusti-dev/blob/b9c5b83d930ef0f648347a930658b4986a0ae774/jni-gen/systest/tests/jvm_builtin_classes.rs#L36

Documentation of `with_local_frame`: https://docs.rs/jni/0.19.0/jni/struct.JNIEnv.html#method.with_local_frame

Pointerbender:
> should be wrapped in a with_local_frame call, like here

That sounds promising! Is 16mb of capacity enough for the test run in this case or should this number be tweaked as well?

fpoli:
> > should be wrapped in a with_local_frame call, like here
> 
> That sounds promising! Is 16mb of capacity enough for the test run in this case or should this number be tweaked as well?

My understanding is that the capacity (16 in that test) is a lowerbound on the number of local references can be created ([documentation](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#push_local_frame)). The JVM should be able to extend the capacity if needed. [This test](https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-binding/core/rust/integration_tests/tests/local_frame_memory_leak_jvm.rs) seems to confirm that exceeding the capacity shouldn't be a problem.

Pointerbender:
Screenshot_2021-11-02_15-03-25

This screenshot of VisualVM is of a test run with the `env.with_local_frame(16, || {})` work-around applied in `viper/src/verifier.rs`, but the same problem still persists. The graph reveals a couple of interesting findings though:

* After every test the lower bound on the allocated heap size keeps increasing.
* Around 2.20 pm the lower bound on the allocated heap size does not increase for about 6-7 minutes. This is during the test run of `verify/pass/rosetta/Knights_tour.rs` which takes longer than other tests. This seems to imply that during the tests, the lower bound of the allocated heap size is not increasing.
* As suspected, the garbage collector gets very busy after a while (starting between 2:40 and 2:45 pm) and CPU usage goes through the roof (up to 81 live Java threads at some point, too - this seems high). This period corresponds with the running of test case `verify_overflow/pass/rosetta/Selection_sort.rs`. The available heap space is already almost running out at this point.
* Just before 3pm the allocated heap size shrinks, this moment corresponds with the out-of-memory error. What's interesting here is that still 200mb of heap memory stays allocated, despite the out-of-memory error.

Since there is a gradual build-up of the allocated heap memory, and the fix in `pub fn verify(&self, program: Program) -> VerificationResult` in `viper/src/verifier.rs` did not resolve it, there may be a different or more than one obstacle for the GC.

fpoli:
Wow, thanks for the charts! If you take heap dumps at different points in time it should be possible to inspect the diff in VisualVM, and for each object it should be possible to follow the referents until an object is marked as "JNI local". JNI locals are Java objects created from Rust, and they cannot be touched from the garbage collector. If we identify them, we can check if they are created within a `with_local_frame`.

fpoli:
Uh, maybe it's because we don't have a `with_local_frame` around the implementation of this `verify`. So, `viper_program`s don't get garbage collected.

https://github.com/viperproject/prusti-dev/blob/6af18c454cfe16693b18d9eb07e1f1e8b28524c1/prusti-server/src/verifier_runner.rs#L59-L65

We could add a `with_local_frame` method to the two `VerificationContext` types to make it possible.

Pointerbender:
I have another update :) With an additional `with_local_frame` (implemented through `prusti-common/src/verification_context.rs` and `viper/src/verification_context.rs`) wrapped around the function body of `pub fn verify(&self, program: Program) -> viper::VerificationResult` in `prusti-server/src/verifier_runner.rs`, the resulting graph looks strikingly similar (the flat line is at the end is because I came back to the computer later, it had already finished at that point):

Screenshot_2021-11-02_20-06-13

The out-of-memory error keeps happening at the same place. Then the next thing I tried is to update `verify()` in `prusti-server/src/verifier_runner.rs` to:

```rust
    pub fn verify(&self, program: Program) -> viper::VerificationResult {
        return viper::VerificationResult::Success;
    }
```

and profile the test suite by running `x.py test pass` (note that these are less test cases than the full run `x.py test`, but otherwise it'll trip on the `ui` and `fail` tests :smile: ). This yielded an interesting graph, nice and smooth showing the build-up of the irreclaimable memory:

Screenshot_2021-11-02_20-31-18

Just short of 100mb this time, but I think that's because I'm not running the full test suite. The pattern of increased memory usage after every test is still present.

I'm not sure what's going on now :)

Pointerbender:
> showing the build-up of the irreclaimable memory:

Actually, it looks like I jumped to a conclusion too fast here. If I run the test suite as `JAVA_HEAP_SIZE=75 x.py test pass` then the JVM does manage to reclaim this memory (where the total "heap size" goes down is where I performed a heap dump, the first "used heap" reduction was initiated by the JVM itself):

Screenshot_2021-11-02_21-06-44

Pointerbender:
I spend a little extra time last night adding local frames around every interaction between the Verifier and the JVM through JNI (even making sure that `Verfier.verifier_instance` in `viper/src/verifier.rs` is garbage collected when the `Verifier` is dropped in Rust). This did not seem to have any effect and the OOM error is still happening at the same moment.

Looking at the heap dumps of the JVM, 70% or more of the memory is taken by Java `byte[]` types. Are these held on to anywhere by Prusti or Viper?

fpoli:
> Then the next thing I tried is to update `verify()` in `prusti-server/src/verifier_runner.rs` to:
> 
> ```rust
>     pub fn verify(&self, program: Program) -> viper::VerificationResult {
>         return viper::VerificationResult::Success;
>     }
> ```
> 
> and profile the test suite by running `x.py test pass` (note that these are less test cases than the full run `x.py test`, but otherwise it'll trip on the `ui` and `fail` tests smile ).

That's a nice experiment! I tried that with `PRUSTI_LOG=trace`, which should log every JNI call that is made, and it seems that the JVM is not invoked at all (as expected) when processing a verification request. So, the memory usage might come from the interaction with VisualVM.

> I spend a little extra time last night adding local frames around every interaction between the Verifier and the JVM through JNI (even making sure that `Verfier.verifier_instance` in `viper/src/verifier.rs` is garbage collected when the `Verifier` is dropped in Rust). This did not seem to have any effect and the OOM error is still happening at the same moment.
> 
> Looking at the heap dumps of the JVM, 70% or more of the memory is taken by Java `byte[]` types. Are these held on to anywhere by Prusti or Viper?

The prusti-server doesn't drop the verifier between verification requests. If that's what blocks the GC, it's a bug that we can report in Viper. In VisualVM you can go to heapdump > instances > references to see which other instances reference a given instance, so you can walk the graph backwards until you find what's preventing the garbage collection.

fpoli:
I gave a look. After wrapping in a `with_local_frame` the `verify` in `prusti-server/src/verifier_runner.rs:57`, each time `fn main() { assert!(true); }` gets verified the JVM creates (among other things) 2 instances of `viper.silicon.state.term.Less`. Those instances doesn't get garbage-collected, because they are referenced by the verifier and they don't get freed by the inner [`reset`](https://github.com/viperproject/silicon/blob/16fd1f6412ab1b3b09d5e2acea637e47b6f4154a/src/main/scala/Silicon.scala#L140-L148) calls. It's a memory leak in Silicon; I opened https://github.com/viperproject/silicon/issues/579.

The short-term solution for Prusti is to create a verifier instance for each verification request. This seems to have a tiny overhead (0.04 s). PR https://github.com/viperproject/prusti-dev/pull/745 does that.

Pointerbender:
Woohoo, @fpoli fixed the out-of-memory issue just now! If master is merged again into this PR, then the tests should pass this time around.

JonasAlaif:
Awesome! Thanks for sorting this out @Pointerbender and @fpoli.
> If master is merged again into this PR, then the tests should pass this time around.

I'll do that.

Pointerbender:
Ah bummer, looks like there is another failing test case after all:

```
test [compile-fail] fail/index-vec.rs ... 
error: tests/verify_overflow/fail/index-vec.rs:6: unexpected error: '6:5: 6:9: [Prusti: unsupported feature] Non-slice LHS type '&u32' not supported yet'
```

The test case is:

```rust
use prusti_contracts::*;

#[pure]
fn get_third(v: &Vec<u32>) -> u32 {
    v[2] //~ ERROR Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions
}

fn main(){}
```

JonasAlaif:
When I run the test locally I get:
```
error: [Prusti: unsupported feature] Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions
 --> prusti-tests/tests/verify_overflow/fail/index-vec.rs:6:5
  |
6 |     v[2] //~ ERROR Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions
  |     ^^^^

error: [Prusti: unsupported feature] Non-slice LHS type '&u32' not supported yet
 --> prusti-tests/tests/verify_overflow/fail/index-vec.rs:6:5
  |
6 |     v[2] //~ ERROR Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions
  |     ^^^^

Verification failed
error: aborting due to 2 previous errors; 1 warning emitted
```
So it seems that the issue is with the test now generating two errors rather than just one. This could just be due to merging in all of the new stuff from master. Any idea @fpoli?
The error is technically valid since the `v[2]` is translated to `<Vec<u32> as Index<usize>>::index(move _3, const 2_usize)` in MIR, which correctly goes to Prusti's [`encode_sequence_index_call`](https://github.com/dario23/prusti-dev/blob/4bc5bba6358146c27d990e53a70c2b3690be0564/prusti-viper/src/encoder/procedure_encoder.rs#L2535-L2726) for encoding. However, this function is currently only designed to handle `<T as Index<std::ops::Range*<usize>>>::index`. This should probably be improved - ideally by defining `#[extern_spec]` for `Index:: index` rather than hardcoding it all in [`encode_sequence_index_call`](https://github.com/dario23/prusti-dev/blob/4bc5bba6358146c27d990e53a70c2b3690be0564/prusti-viper/src/encoder/procedure_encoder.rs#L2535-L2726).

Pointerbender:
Since both the error `Non-slice LHS type '&u32' not supported yet` and `Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions` are intended and correct errors, could a solution be to update the test case to expect both errors? For example:

```rust
use prusti_contracts::*;

#[pure]
fn get_third(v: &Vec<u32>) -> u32 {
    v[2]
    //~^ ERROR Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions
    //~^^ ERROR Non-slice LHS type '&u32' not supported yet
}

fn main(){}
```

Or should the error message be updated into a single error message instead?

> However, this function is currently only designed to handle `<T as Index<std::ops::Range*<usize>>>::index`. This should probably be improved - ideally by defining `#[extern_spec]` for `Index::index` rather than hardcoding it all in `encode_sequence_index_call`.

This also sounds promising, would that be part of the `more-slices` PR as well?

> This could just be due to merging in all of the new stuff from master. Any idea @fpoli?

The new error `Non-slice LHS type '&u32' not supported yet` does not exist on the `master` branch yet. The other error message `Using usize as index/range type for &std::vec::Vec<u32> is not currently supported in pure functions` is new on `master` since October 6th and was introduced through PR #699 by @zgrannan . This error indeed entered the `more-slices` branch when the `master` branch was merged into it on October 27th through merge-commit 5163985052b28918845998a71fc097207e106a3a.

JonasAlaif:
That seems like the best solution for now.
> This also sounds promising, would that be part of the more-slices PR as well?

Sadly, not yet. But taking this approach seems to me as a good way to remove a lot of bloat from [procedure_encoder.rs](https://github.com/viperproject/prusti-dev/blob/28f4988708575f3bcf7e5e75abfcc1dea80c1f71/prusti-viper/src/encoder/procedure_encoder.rs#L2124-L2331) and possibly elsewhere. At the moment `#[extern_spec]` doesn't support the features needed to specify a contract for `Index::index`, but this should hopefully change with #753.

PR #604, opened 2021-07-14
Implement existential quantifiers

Aurel300:
Extracted from the closure branch and cleaned up a little.

Remaining TODO:

- [x] What to do about the quantifier fixer @vakaras ? It was not used at all in the closure branch, so there was no implementation for `exists`.
- [x] Add test cases?
- [x] Document in the user guide.

PR #607, opened 2021-07-16
 Implement existential quantifiers

vakaras:
Supersedes #604


prusti-tests/tests/parse/ui/composite.rs
```diff
@@ -53,4 +53,25 @@ fn test15() {}
 #[requires(forall(|b: i32| b == 10) ==> true ==> forall(|a: u32, b: u32| a == 5))]
 fn test16() {}
 
+#[requires(true && exists(|a: i32| a == 5))]
+fn test17() {}
+
+#[requires(true && exists(|a: i32| a == 5))]
+fn test18() {}
```
Aurel300: This one is the same as the previous one?

vakaras: Yes.

vakaras:
@Aurel300 @fpoli 
Cargo also seems to have a bug: https://stackoverflow.com/questions/68283463/cargo-fails-to-download-dependencies-when-running-in-github-actions#comment120888659_68283463

PR #608, opened 2021-07-16
Document existentials.

vakaras:


user-guide/src/syntax.md
```diff
@@ -70,9 +71,11 @@ forall(|x: isize, y: isize| ...)
 The syntax of quantifiers is:
 
 ```plain
-forall(|<bound variable>: <bound variable type>, ...| <filter> ==> <expression>)
+<quantifier_type>(|<bound variable>: <bound variable type>, ...| <filter> ==> <expression>)
```
Aurel300, line 74: `<filter> ==>` does not make so much sense for existentials. Maybe just show two different versions?

vakaras: A fair point.

PR #609, opened 2021-07-16
Testing on ci (PR for Testing only)

ani003:
This PR is opened to run `test-crates` on CI to determine the number of crates which do not panic. Previous test had 70 crates which did not panic. Will update the current number after running the tests


ani003:
Current count of crates not panicking: 76

PR #549, opened 2021-06-10
Refactor Parser into Pure Recursive Descent Parser

JustinHuPrime:
This pull request:

1. rewrites the preparser to use a recursive descent parser instead of the hybrid between shift-reduce and recursive descent that the old preparser used
2. documents the syntax parsed by the preparser in `prusti-specs/src/specifications/mod.rs` as an EBNF

These changes are preparation to rewrite the preparser to accept re-entry from Rust into Prusti syntax.

This work is being done as part of a research internship with @Aurel300 and @alexanderjsummers.

PR #620, opened 2021-07-19
Implementation of reentrance into Prusti syntax from Rust

JustinHuPrime:
This pull request allows for Prusti syntax to be nested within Rust syntax. As a result, the following changes were made:

- The preparser now produces a Rust expression when asked to parse a Prusti expression.
- The rewriter now has functions to desugar Prusti AST nodes into Rust expressions - this is used by the preparser.
- The rewriter no longer needs to deal with JSON - there is now only one function per assertion.
- The spec collector no longer needs to associate Rust expressions with their assertions - instead, it only need to associate assertions with their implementation functions and asserted-about functions.
- TODO: something about changes to the verifier.

This PR builds on the work in #549, and is being done as part of a research internship with @Aurel300 and @alexanderjsummers.


prusti-contracts/src/lib.rs
```diff
@@ -87,4 +87,29 @@ pub fn old<T>(arg: T) -> T {
     arg
 }
 
+#[allow(unused_variables)]
+pub fn implication(lhs: bool, rhs: bool) -> bool {
+    true
+}
+
+#[allow(unused_variables, unused_mut)]
+pub fn entailment<F1, F2>(mut f: F1, req: F2) -> bool {
```
Aurel300: Even if we keep this underconstrained signature for taking in closure-like arguments, the `mut` doesn't do anything and can be removed (+ `unused_mut`).

prusti-interface/src/specs/mod.rs
```diff
@@ -54,170 +40,63 @@ pub struct SpecCollector<'tcx> {
     tcx: TyCtxt<'tcx>,
     extern_resolver: ExternSpecResolver<'tcx>,
 
-    /// Collected assertions before deserialisation.
-    spec_items: Vec<SpecItem>,
+    /// Map between specification ids and their corresponding function
```
Aurel300: `corresponding function` here is inteded to mean "the Rust function that contains the typed body of the specification", right? At a first glance this might be misunderstood as `corresponding function` being the function to which the specs are attached. Maybe:
```suggestion
    /// Map between specification ids and their corresponding typed expressions.
```

prusti-interface/src/specs/mod.rs
```diff
@@ -299,56 +164,15 @@ impl<'tcx> intravisit::Visitor<'tcx> for SpecCollector<'tcx> {
...
 
-            // Collect loop invariant
-            if spec_type == SpecType::Invariant {
-                self.loop_specs
-                    .entry(local_id)
-                    .or_insert(vec![])
-                    .push(spec_id);
+            // Collect loop specifications            
+            if has_prusti_attr(attrs, "loop_body_invariant_spec") {
+                // TODO: surely this isn't all we need to do for loops?
```
Aurel300: Is this comment still relevant? This might indeed be all that needs to happen here. Loop specs are then inserted into `def_spec` in the original `determine_loop_specs` function. These are later picked up again in the `ProcedureEncoder`.

prusti-tests/tests/parse/pass/basic_prusti_reentry.rs
```diff
@@ -1,6 +1,5 @@
 use prusti_contracts::*;
 
-// ignore-test: prusti-expr! not yet implemented
 #[requires(x == 1 || prusti_expr!(y == 0 ==> z == 0))]
```
Aurel300: This should also not have `prusti_expr!` at all?

prusti-contracts/src/lib.rs
JustinHuPrime: Doesn't this allow user closures which might mutate their environment to be specified with an entailment?

prusti-tests/tests/parse/pass/basic_prusti_reentry.rs
JustinHuPrime: We haven't yet implemented disjunctions as a native prusti operator

prusti-interface/src/specs/mod.rs
JustinHuPrime: Addressed by 1bbbdfb4b539f253b77aef890ba056f8df9197e9

prusti-interface/src/specs/mod.rs
JustinHuPrime: Addressed by 1bbbdfb4b539f253b77aef890ba056f8df9197e9

prusti-tests/tests/parse/pass/basic_prusti_reentry.rs
Aurel300: Sure, but isn't the plan for re-entrancy to work based on parentheses rather than this explicit macro?

prusti-contracts/src/lib.rs
Aurel300: No, that would be the difference between e.g. an `Fn(...)` type and an `FnMut(...)` trait constraint. Here we are not specifying either of those types, instead using a more generic `F1` type parameter. The `mut` modifier on the argument simply means that `f` can be reassigned inside the body of this `entailment` function, but the body of the function is irrelevant—we will match on the `entailment` call later on in the encoder instead.

prusti-specs/src/rewriter.rs
```diff
@@ -116,70 +96,105 @@ impl AstRewriter {
             &format!("prusti_{}_item_{}_{}", spec_type, item.sig().ident, spec_id),
             item_span,
         );
-        let mut statements = TokenStream::new();
-        assertion.encode_type_check(&mut statements);
         let spec_id_str = spec_id.to_string();
-        let assertion_json = crate::specifications::json::to_json_string(&assertion);
 
-        let mut spec_item: syn::ItemFn = parse_quote_spanned! {item_span=>
+        let mut spec_item: syn::ItemFn = parse_quote_spanned! {item_span =>
```
Aurel300: For the record this weird spacing is suggested [here](https://docs.rs/quote/1.0.9/quote/macro.quote_spanned.html#syntax) but I don't find the justification very good. I haven't been consistent about the spacing and probably wrote the space before `=>` more often than not.

prusti-specs/src/rewriter.rs
```diff
@@ -116,70 +96,105 @@ impl AstRewriter {
...
-            fn #item_name() {
-                #statements
+            fn #item_name() -> bool {
+                #expr
             }
         };
         spec_item.sig.generics = item.sig().generics.clone();
         spec_item.sig.inputs = item.sig().inputs.clone();
-        if spec_type == SpecItemType::Postcondition {
+        if spec_type != SpecItemType::Precondition {
```
Aurel300: Does this actually make sense for a predicate?

prusti-specs/src/rewriter.rs
```diff
@@ -116,70 +96,105 @@ impl AstRewriter {
...
+        self.generate_spec_item_fn(
+            SpecItemType::Pledge,
+            spec_id,
+            Parser::from_token_stream(tokens).extract_pledge_expr()?,
+            &item
+        )
+    }
+
+    /// Parse a loop invariant into a Rust expression
+    pub fn process_loop(
```
Aurel300:
```suggestion
    pub fn process_loop_invariant(
```
?

prusti-specs/src/rewriter.rs
```diff
@@ -116,70 +96,105 @@ impl AstRewriter {
...
+                || -> bool {
+                    #expr
                 };
             }
-        }
+        })
     }
 
-    /// Generate statements for checking a closure specification.
-    /// TODO: arguments, result (types are typically not known yet after parsing...)
```
Aurel300, line 168: Please keep the TODO.

prusti-specs/src/rewriter.rs
```diff
@@ -209,6 +223,159 @@ impl AstRewriter {
...
+    if conjuncts.len() == 1 {
+        conjuncts.pop().unwrap()
+    } else {
+        let last = conjuncts.pop().unwrap();
+        let rest = translate_conjunction(conjuncts);
+        parse_quote_spanned! {
+            rest.span().join(last.span()).unwrap() =>
+            #rest && #last
+        }
+    }
```
Aurel300: Seems like a job for `Iterator::reduce`?

prusti-specs/src/rewriter.rs
```diff
@@ -209,6 +223,159 @@ impl AstRewriter {
...
+            #ident: #typ
+        }
+    } else {
+        let (ident, typ) = args.pop().unwrap();
+        let rest = args_to_tokens(args);
+        quote_spanned! {
+            rest.span().join(typ.span()).unwrap() =>
+            #rest, #ident: #typ
+        }
+    }
```
Aurel300: Also `Iterator::reduce` (and `map`).

prusti-specs/src/rewriter.rs
```diff
@@ -209,6 +223,159 @@ impl AstRewriter {
...
+    let post_conjuncts = translate_conjunction(posts);
+    parse_quote_spanned! {
+        // TODO: get the right span
+        Span::call_site() =>
+        entailment(#closure, |#arg_tokens| {
+            {
+                #pre_conjuncts
+            } &&
+            {
+                #post_conjuncts
```
Aurel300: These (pre and post) should probably be separate arguments to `entailment` under separate closures. `post` should be able to use `result` (or `cl_result`), `pre` should not.

prusti-specs/src/rewriter.rs
```diff
@@ -209,6 +223,159 @@ impl AstRewriter {
...
+            #expr,
+        }
+    } else {
+        let last = exprs.pop().unwrap();
+        let rest = tuple_to_tokens(exprs);
+        quote_spanned! {
+            rest.span().join(last.span()).unwrap() =>
+            #rest #last,
+        }
+    }
```
Aurel300, line 338: `reduce`?

prusti-specs/src/rewriter.rs
JustinHuPrime: Fixed in 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-specs/src/rewriter.rs
JustinHuPrime: Fixed by 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-specs/src/rewriter.rs
JustinHuPrime: No, it doesn't. Fixed in 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-specs/src/rewriter.rs
JustinHuPrime: Right. Fixed by 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-tests/tests/parse/pass/basic_prusti_reentry.rs
JustinHuPrime: Yes. My bad. Fixed by 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-contracts/src/lib.rs
JustinHuPrime: Oops. Fixed in 681ccfe0b2aa35aa9ed35a05b55490ba69ad34b9

prusti-specs/src/rewriter.rs
JustinHuPrime: Fixed in 6268917f30bbc875485f34bde92f5bd18cbde277

prusti-specs/src/rewriter.rs
JustinHuPrime: Fixed in 6268917f30bbc875485f34bde92f5bd18cbde277

prusti-specs/src/rewriter.rs
JustinHuPrime: Yes it does. Fixed in 6268917f30bbc875485f34bde92f5bd18cbde277

prusti-specs/src/rewriter.rs
JustinHuPrime: Partially fixed by 33723fd23eaa1171543ec8351422a3dcf5e66c56

I'm not quite sure how to get the type of the result here.

JustinHuPrime: Does something like 8dc5deb602a9a11d28486b914bf81b1185b64e30 work?

Aurel300: No, `: T` like that would just trigger a typing error, it does not refer to the `T` in `entailment` (which you can remove again). Actually it turns out we can't know the result type without having the user declare it explicitly. I've made changes to the parser in the closure branch but I guess these are not on `master` yet. You can assume that the type is either known and is passed to this method somehow, or just put `i32` for now, which is what the behaviour was on `master` IIRC.

JustinHuPrime: I've made it an `i32` for now in 81ebcf180f7230702581aafa798071b09b8bf5ee.

vakaras:
@Aurel300 What needs to be done to merge this PR? Rebasing and checking that all tests pass, or something more?

Aurel300:
@vakaras There's still quite a bit:

- rebase
- fix the current f/u errors (which I believe are because the `SpecEncoder` used to replace args + return places from spec. closures to make sense in the method but now this code is disabled)
- implement all of our specification constructs

vakaras:
I see. Thank for an update.

Aurel300:
Superseded by #831.

PR #622, opened 2021-07-21
Change the default of check_overflows to true.

vakaras:
TODOs:

* [x] Update documentation.


Aurel300:
Given that this is the default it might make more sense if the majority of our test suite was using the flag. I'm guessing the majority of `verify` test cases should also pass int `verify_overflow` and maybe should be moved there?

vakaras:
> Given that this is the default it might make more sense if the majority of our test suite was using the flag. I'm guessing the majority of `verify` test cases should also pass int `verify_overflow` and maybe should be moved there?

A good point.

PR #636, opened 2021-07-29
[WIP]Export spec macro

LDY1998:
1. Make fn/struct/impl/mod specifications exportable
2. Add import impl
3. Integrate export with external spec, now by default all `#[extern_spec]` are exported
4. Add test case
5. Refactor contracts to make sure specs can pass `cargo build`
6. Provide a custom Emitter impl to handle potential export error specifically


prusti-contracts-impl/src/lib.rs
```diff
@@ -51,11 +51,24 @@ pub fn refine_trait_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream
...
+pub fn export_spec(_attr: TokenStream, _tokens: TokenStream) -> TokenStream {
+    let callsite_span = Span::call_site();
+    (quote_spanned!(callsite_span=> )).into()
+}
+
+#[proc_macro]
+pub fn prusti_use(_tokens: TokenStream) -> TokenStream {
+    let callsite_span = Span::call_site();
+    (quote_spanned!(callsite_span=> )).into()
+}
```
Aurel300:
```suggestion
}

```

prusti-contracts-internal/src/lib.rs
```diff
@@ -57,3 +57,13 @@ pub fn extern_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
...
+
+#[proc_macro_attribute]
+pub fn export_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
+    prusti_specs::export_spec(attr.into(), tokens.into()).into()
+}
+
+#[proc_macro]
+pub fn prusti_use(tokens: TokenStream) -> TokenStream {
+    prusti_specs::prusti_use(tokens.into()).into()
+}
```
Aurel300:
```suggestion
}

```

prusti-contracts/src/lib.rs
```diff
@@ -35,6 +35,12 @@ mod private {
     /// A macro for defining a predicate using prusti expression syntax instead
     /// of just Rust expressions.
     pub use prusti_contracts_impl::predicate;
+
+    /// A macro for specifying specs to export
+    pub use prusti_contracts_impl::export_spec;
+
+    /// A macro for specifying specs to import
```
Aurel300: Please make the comments the same in the two versions (here and L82). We should eventually get rid of the needless duplication but still…

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,336 @@
+use proc_macro2::{TokenStream, TokenTree};
+use quote::{quote, quote_spanned, ToTokens};
+use syn::spanned::Spanned;
+
+// generate a hash of path
```
Aurel300: Please use triple slashes `///` for comments that are actually meant to document what a method does.

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,336 @@
...
+fn hash_path(path: syn::Path) -> u64 {
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
+    println!("path to hash: {:?} \n", &path);
```
Aurel300: I know this is still WIP but this should eventually become `debug!` (or just removed entirely).

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,336 @@
...
+    let mut hasher = DefaultHasher::new();
+    path.hash(&mut hasher);
+    hasher.finish()
+}
+
+// generate a macro identifier with given path in this format:
+// {last_segment in path}_{hash(path)}
```
Aurel300: Why only the last one? Is this to make the identifier still semi-readable but also not too long?

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+            };
+        }
+    }
+}
+
+// helper macro for getting the value of attribute with given name in an item
+// attribute must be in this format: #[name(value)]
+// if attribute is not present then path is not changed
+// @param: $path_mut_ref must be a mutable reference of Path, i.e. &mut syn::Path / mut syn::Path
+macro_rules! get_prusti_path_in_attr {
```
Aurel300: This is extremely ugly… Please just make this a function that returns a value, rather than accepting an argument that is filled in.

prusti-specs/src/export_spec_rewriter.rs
LDY1998: Yes, that's what I intend to do, putting up the full path makes it too long and it's not that useful (the info is in hash anyway)

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// expand the module path recorded when traversing AST
+fn expand_mod_path_in_macro(
+    path: &syn::Path,
+    mut tokens: TokenStream,
+    uses_same_level: &TokenStream,
+) -> TokenStream {
+    match &path.segments.empty_or_trailing() {
+        true => tokens,
```
Aurel300: `match X { true => ...` is just `if X ...`! (Same in later cases)

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+        false => {
+            for seg in path.segments.to_owned().into_iter().rev() {
```
Aurel300: I don't understand why there are multiple places which seem to depend on `empty_or_trailing`… Are you sure you didn't want to just check empty or similar? Is the behaviour actually different if the path is empty here?

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// generate a macro identifier with given path in this format:
+// {last_segment in path}_{hash(path)}
+pub fn generate_macro_ident(path: syn::Path) -> syn::Ident {
+    let ident = path.segments.last().unwrap().ident.to_owned();
+    syn::Ident::new(&format!("{}_{}_{}", "prusti_export", ident, hash_path(path)), ident.span())
+}
+
+// generate macro for export based on given identifier and item
```
Aurel300: This mostly repeats the name of the function, but doesn't really say what it does. In these macro-generating functions you might want to provide concrete examples. If it is easier, you can choose to write this in a single, large comment somewhere at the beginning of the file.

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+    };
+    match item {
+        syn::Item::Fn(item_fn) => {
+            rewrite_fn(
+                path,
+                mod_path,
+                item_fn,
+                &mut macros_tokens,
+                &TokenStream::new(),
+            );
```
Aurel300: Make these functions return a `TokenStream` instead of taking a mutable reference to the one you create here? The `match` block can be wrapped in `Ok(...)` and it can be the last expression in the method so the `return` is not necessary. All the `{ ... }` braces in this `match` can be omitted.

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+        syn::Item::Struct(item_struct) => {
+            rewrite_struct(
+                item_struct,
+                mod_path,
+                &mut macros_tokens,
+                &TokenStream::new(),
+            );
+        }
+        _ => {
+            unimplemented!();
```
Aurel300, line 206: Is this panic actually reachable if the user uses Prusti badly? If yes, there should be an explanation here. If not, this should be `unreachable!()` instead.

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+}
+
+fn parse_valid_export_type(ty_path: &mut syn::Path) {
+    for seg in &mut ty_path.segments {
+        if let syn::PathArguments::AngleBracketed(_) = &mut seg.arguments {
+            seg.arguments = syn::PathArguments::None;
+        }
+    }
+}
```
Aurel300:
```suggestion
}

```

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+    match attr.is_empty() {
+        true => Err(syn::Error::new(span, "export path must be non-empty")),
+        false => match attr.to_string().trim() == "::" {
+            true => Ok(syn::Path {
+                leading_colon: None,
+                segments: syn::punctuated::Punctuated::new(),
+            }),
+            false => syn::parse2(attr).map_err(|_e| syn::Error::new(span, "invalid export path")),
+        },
+    }
```
Aurel300:
```suggestion
    let span = attr.span();
    if attr.is_empty() {
        return Err(syn::Error::new(span, "export path must be non-empty"));
    }
    if attr.to_string().trim() == "::" {
        Ok(syn::Path {
            leading_colon: None,
            segments: syn::punctuated::Punctuated::new(),
        })
    } else {
        syn::parse2(attr).map_err(|_| syn::Error::new(span, "invalid export path"))
    }
```

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+    };
+}
+
+// expand the module path recorded when traversing AST
+fn expand_mod_path_in_macro(
+    path: &syn::Path,
+    mut tokens: TokenStream,
+    uses_same_level: &TokenStream,
```
Aurel300, line 177: The variable name here is not great. It sounds like "uses same level" would be a boolean, but I guess these are `use`s on the same level (as what?)?

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,334 @@
...
+                rewrite_impl(&path, mod_path, item_impl, macros_tokens, &uses)?;
+            }
+            syn::Item::Mod(item_mod) => {
+                rewrite_mod(&path, mod_path, item_mod, macros_tokens, &uses)?;
+            }
+            syn::Item::Struct(item_struct) => {
+                rewrite_struct(item_struct, mod_path, macros_tokens, &uses);
+            }
+            syn::Item::Use(item_use) => {
+                uses.extend(quote!(#item_use));
```
Aurel300, line 263: Why exactly is this stateful and order-dependent? Why do we collect `use`s and only add the ones we have seen so far into subsequent `impl`s, `mod`s, etc?

prusti-specs/src/lib.rs
```diff
@@ -445,12 +446,56 @@ pub fn refine_trait_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream
...
+        segments: syn::punctuated::Punctuated::new(),
+    };
+    let mut mod_path_for_hashing = syn::Path {
+        leading_colon: None,
+        segments: syn::punctuated::Punctuated::new(),
+    };
+
+    let mut seg_iter = path.segments.clone().into_iter();
+
+    if let Some(crate_seg) = seg_iter.next() {
```
Aurel300, line 482: Why not `first` instead of cloning and creating an iterator?

prusti-tests/tests/cargo_verify/export-crate-tests/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(test)]
```
Aurel300: Is this file necessary?

prusti-specs/src/export_spec_rewriter.rs
Aurel300: Please go through the file diff here on github and make sure every file you added or modified is still ending with a linebreak. GitHub highlights when it doesn't with a red symbol.

prusti-specs/src/export_spec_rewriter.rs
LDY1998: Yes, this is the `uses` (import) they have at the same module level

prusti-specs/src/export_spec_rewriter.rs
LDY1998: I think it's because how rust deal with `uses`, we can have duplicate `use` in different levels but not the same level:
```rust
use prusti_contracts::*;
mod any {
  //this is valid
  use prusti_contracts::*;

  //this is error
  use prusti_contracts::*;
}

```

Therefore the `uses` is different in each level so that I think we will have to retain that structure to avoid potential conflicts, so I intend to collect `uses` at one level and pass that to the children (they will need it like they need the `mod_path` for later populating the macro)

prusti-specs/src/lib.rs
LDY1998: Oh because at here if our path do have the first and the last segment then I still need to get the segments after the first one, because first one is the crate: `crate_name::mod::fn`.
The path to generate hash should be `mod::fn`, therefore I will need to have the iterator step through the first one anyway

prusti-contracts-impl/src/lib.rs
```diff
@@ -51,11 +51,24 @@ pub fn refine_trait_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream
 }
 
 #[proc_macro_attribute]
-pub fn extern_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream {
-    tokens
+pub fn extern_spec(_attr: TokenStream, _tokens: TokenStream) -> TokenStream {
+    let callsite_span = Span::call_site();
+    (quote_spanned!(callsite_span=> )).into()
```
Aurel300:
```suggestion
    TokenStream::new()
```
I don't think the span does anything when there are no tokens. Please also change `export_spec` and `prusti_use`.

prusti-contracts-internal/src/lib.rs
```diff
@@ -57,3 +57,13 @@ pub fn extern_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
 pub fn predicate(tokens: TokenStream) -> TokenStream {
     prusti_specs::predicate(tokens.into()).into()
 }
+
+#[proc_macro_attribute]
+pub fn export_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
+    prusti_specs::export_spec(attr.into(), tokens.into()).into()
```
Aurel300, line 63: Why the `into`s? The types should work out:
```suggestion
    prusti_specs::export_spec(attr, tokens)
```
Also in `prusti_use` below.

prusti-interface/src/emitter/export_error_emitter.rs
```diff
@@ -0,0 +1,142 @@
...
+    annotate_snippet_emitter_writer::AnnotateSnippetEmitterWriter,
+    emitter::{ColorConfig, Destination, Emitter, EmitterWriter, HumanReadableErrorType},
+    json::JsonEmitter,
+    registry::Registry,
+    Diagnostic, DiagnosticId, Handler,
+};
+use rustc_span::source_map::SourceMap;
+use rustc_session::config;
+
+
```
Aurel300, line 14: Please document at least briefly (using `///` comments) what this type does.

prusti-interface/src/emitter/export_error_emitter.rs
```diff
@@ -0,0 +1,142 @@
...
+        emitter_dest: Option<Box<dyn Write + std::marker::Send>>,
+    ) -> Self {
+        Self {
+            inner: build_inner_emitter(&sopts, registry, source_map, emitter_dest),
+        }
+    }
+
+    fn custom_emit(&mut self, diag: &Diagnostic) {
+        if let Some(DiagnosticId::Error(err_code)) = &diag.code {
+            if &err_code == &"E0428"
```
Aurel300: Does this work?
```suggestion
            if err_code == "E0428"
```

prusti-interface/src/emitter/export_error_emitter.rs
```diff
@@ -0,0 +1,142 @@
...
+                && diag
+                    .message
+                    .iter()
+                    .any(|(err_msg, _)| err_msg.contains("prusti_export_"))
+            {
+                let mut diag = diag.to_owned();
+                for (msg, _) in diag.message.iter_mut() {
+                    let msg_vec: Vec<&str> = msg.split("_").collect();
+                    let len = msg_vec.len();
+                    if len >= 4 {
```
Aurel300: This seems like a filter that is extremely specific to the error message. Why split on `_`? Why does it need to have 4 or more elements? If there is no other way, please add a comment here to show an example of the error you are filtering out.

prusti-interface/src/emitter/export_error_emitter.rs
```diff
@@ -0,0 +1,142 @@
...
+            } else {
+                self.inner.as_mut().emit_diagnostic(diag);
+            }
+        } else {
+            self.inner.as_mut().emit_diagnostic(diag);
+        }
+    }
+}
+
+fn build_inner_emitter(
```
Aurel300, line 78: Add a comment here that this code comes from `rustc`, and ideally link to the exact lines on GitHub (select the lines on GitHub, then choose "Copy permalink").

prusti-interface/src/emitter/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+fn build_handler_with_emitter(
+    sopts: config::Options,
+    registry: rustc_errors::registry::Registry,
+    dest: Option<Box<dyn Write + std::marker::Send>>,
+    source_map: Lrc<SourceMap>,
+    flags: HandlerFlags
+) -> Handler {
+    Handler::with_emitter_and_flags(
+        box ExportErrorEmitter::new(sopts, registry, source_map, dest),
+        flags
```
Aurel300:
```suggestion
        flags,
```

prusti-interface/src/emitter/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    box callback
+}
+
+fn build_handler_with_emitter(
+    sopts: config::Options,
+    registry: rustc_errors::registry::Registry,
+    dest: Option<Box<dyn Write + std::marker::Send>>,
+    source_map: Lrc<SourceMap>,
+    flags: HandlerFlags
```
Aurel300:
```suggestion
    flags: HandlerFlags,
```

prusti-interface/src/emitter/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    registry: rustc_errors::registry::Registry,
+    dest: Option<Box<dyn Write + std::marker::Send>>,
+) -> Box<dyn FnOnce(&mut ParseSess) + std::marker::Send> {
+    let callback = |sess: &mut ParseSess| {
+        let flags = sopts.debugging_opts.diagnostic_handler_flags(true);
+        sess.span_diagnostic =
+            build_handler_with_emitter(sopts, registry, dest, sess.clone_source_map(), flags);
+    };
+
+    box callback
```
Aurel300:
```suggestion
    box |sess: &mut ParseSess| {
        let flags = sopts.debugging_opts.diagnostic_handler_flags(true);
        sess.span_diagnostic =
            build_handler_with_emitter(sopts, registry, dest, sess.clone_source_map(), flags);
    }
```

Aurel300:
Please also address the clippy warnings.

LDY1998:
> Please also address the clippy warnings.

My local clippy run fail and report this: 
```shell
error[E0514]: found crate `cfg_if` compiled by an incompatible version of rustc
  --> /home/t530/.cargo/registry/src/github.com-1ecc6299db9ec823/crc32fast-1.2.1/src/lib.rs:33:1
   |
33 | extern crate cfg_if;
   | ^^^^^^^^^^^^^^^^^^^^
   |
   = help: please recompile that crate using this compiler (rustc 1.56.0-nightly (4e282795d 2021-07-31))
   = note: the following crate versions were found:
           crate `cfg_if` compiled by rustc 1.55.0-nightly (74ef0c3e4 2021-07-16): /home/t530/dev/prusti-dev/target/debug/deps/libcfg_if-621b92c8c202d759.rmeta

error: build failed

```

prusti-specs/src/export_spec_rewriter.rs
```diff
@@ -0,0 +1,488 @@
...
+                }
+                bounds.push_value(input.parse::<syn::TypeParamBound>()?);
+                if input.peek(Token![where]) || input.peek(Token![=]) || input.peek(Token![;]) {
+                    break;
+                }
+                bounds.push_punct(input.parse::<Token![+]>()?);
+            }
+        }
+        generics.where_clause = input.parse()?;
+        Ok(Self(generics))
```
Aurel300: Can this whole thing not just be a `parse` with a `Punctuated<WherePredicate, Token![,]>` result?

LDY1998: Yeah I think this is related to what kind of trait bound syntax we want to support for the attribute. I was trying to support all valid syntax for impl with generics bounds (this is a little bit overkill). I read the src of impl parse again and looks like we might not even need the punctuated, just let generics and where parse handle it would be fine

PR #638, opened 2021-08-01
Update dependencies (nightly-2021-08-01)

viper-admin:
* [x] Update rustc version to `nightly-2021-08-01`.
* [x] Manualy update outdated dependencies (see the list below).
* [x] Manualy run `cargo update`.

<details><summary>List of direct outdated dependencies:</summary>

```
$ cargo outdated --root-deps-only --workspace

analysis
================
Name        Project  Compat  Latest   Kind    Platform
----        -------  ------  ------   ----    --------
serde       1.0.126  ---     1.0.127  Normal  ---
serde_json  1.0.64   ---     1.0.66   Normal  ---

prusti-common
================
Name   Project  Compat  Latest   Kind    Platform
----   -------  ------  ------   ----    --------
serde  1.0.126  ---     1.0.127  Normal  ---

viper
================
Name   Project  Compat  Latest   Kind    Platform
----   -------  ------  ------   ----    --------
serde  1.0.126  ---     1.0.127  Normal  ---

prusti-contracts
================
Name      Project  Compat  Latest  Kind         Platform
----      -------  ------  ------  ----         --------
trybuild  1.0.42   ---     1.0.43  Development  ---

prusti-contracts-impl
================
Name         Project  Compat  Latest  Kind    Platform
----         -------  ------  ------  ----    --------
proc-macro2  1.0.27   ---     1.0.28  Normal  ---

prusti-specs
================
Name         Project  Compat  Latest   Kind    Platform
----         -------  ------  ------   ----    --------
proc-macro2  1.0.27   ---     1.0.28   Normal  ---
serde        1.0.126  ---     1.0.127  Normal  ---
serde_json   1.0.64   ---     1.0.66   Normal  ---
syn          1.0.73   ---     1.0.74   Normal  ---

prusti-contracts-internal
================
Name         Project  Compat  Latest  Kind    Platform
----         -------  ------  ------  ----    --------
proc-macro2  1.0.27   ---     1.0.28  Normal  ---

prusti-interface
================
Name   Project  Compat  Latest   Kind    Platform
----   -------  ------  ------   ----    --------
serde  1.0.126  ---     1.0.127  Normal  ---

prusti-viper
================
Name       Project  Compat  Latest   Kind    Platform
----       -------  ------  ------   ----    --------
backtrace  0.3.60   ---     0.3.61   Normal  ---
serde      1.0.126  ---     1.0.127  Normal  ---

prusti-server
================
Name     Project  Compat  Latest   Kind    Platform
[...]

PR #641, opened 2021-08-05
Interpret numbers as decimal

fpoli:
Apparently, `$(( .. ))` interprets numbers starting with `0` as octals 🤯

See https://stackoverflow.com/questions/21049822/value-too-great-for-base-error-token-is-09

This should fix cases like https://github.com/viperproject/prusti-dev/pull/638.


vakaras:
> Apparently, `$(( .. ))` interprets numbers starting with `0` as octals exploding_head

Makes sense… ☺

PR #648, opened 2021-08-18
Trivial encoding of strings

### LINKED ISSUE #624: Presence of string literals results in an error
This code:
```
fn main() {
   "Hello world";
}
```
results in an error:
```
error: [Prusti: unsupported feature] the encoding of this reference copy has not been implemented
 --> test.rs:2:4
  |
2 |    "Hello world";
  |    ^^^^^^^^^^^^^
```
Similar code, i.e `print!("Hello World")` also triggers this error

zgrannan:
This PR adds just enough functionality for Prusti to handle string literals without crashing. However, string literals are not actually encoded.

Should resolve https://github.com/viperproject/prusti-dev/issues/624


prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1446,12 +1446,18 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-                (expiring_base, restored, false, stmts)
+                (expiring_base, Some(restored), false, stmts)
             }
 
             mir::Rvalue::Use(mir::Operand::Constant(box mir::Constant { literal, .. })) => {
                 let (ty, val) = mir_constantkind_to_ty_val(literal);
-                let restored = self.encoder.encode_const_expr(ty, &val)?;
+
+                let restored : Option<vir::Expr> = match ty.kind() {
+                    ty::TyKind::Ref(_, inner, _) if inner.is_str() => None,
```
vakaras: I think you need to check if the reference is `'static`. Otherwise, this is probably unsound.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4951,16 +4973,23 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-                            .encode_const_expr(ty, &val)
-                            .with_span(span)?;
-                        // Initialize value of lhs
-                        stmts.push(vir::Stmt::Assign(
-                            lhs.clone().field(field),
-                            const_val,
-                            vir::AssignKind::Copy,
-                        ));
+                        let is_str = match ty.kind() {
+                            ty::TyKind::Ref(_, inner, _) => inner.is_str(),
```
vakaras: This should check that the string is `'static`.

prusti-viper/src/encoder/procedure_encoder.rs
zgrannan: It looks like this is already in the case branch for a literal. If it's literal, shouldn't it be `'static` in principle?

In fact it's not in practice, I see that sometimes the region is `'_#3r` instead of `ReStatic`; not sure how this could happen.

But anyways I think the check here needs only to be checking for a string. Regardless of the region of the string literal, currently we dont encode a VIR constant expression for it.

prusti-viper/src/encoder/procedure_encoder.rs
zgrannan: I think the same thing applies here. If we already know that it's a constant, shouldn't it certainly be `'static`?

prusti-viper/src/encoder/procedure_encoder.rs
vakaras: Yes, I think you are right. Could you please add a comment here and below explaining that?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1446,12 +1446,20 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                (expiring_base, Some(restored), false, stmts)
             }
 
             mir::Rvalue::Use(mir::Operand::Constant(box mir::Constant { literal, .. })) => {
                 let (ty, val) = mir_constantkind_to_ty_val(literal);
-                let restored = self.encoder.encode_const_expr(ty, &val)?;
+
+                // Encoding of string literals is not yet supported, so do not return
+                // an expression in restored here.
+                let restored : Option<vir::Expr> = match ty.kind() {
```
```suggestion
                let restored: Option<vir::Expr> = match ty.kind() {
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1620,13 +1628,13 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
             .with_span(span)?.unwrap();
         let (expiring, restored, is_mut, mut stmts) = self.encode_loan_places(&loan_places)
             .with_span(span)?;
-        let borrowed_places = vec![restored.clone()];
+        let borrowed_places = restored.clone().into_iter().collect(); // rvec![restored.clone()];
```
```suggestion
        let borrowed_places = restored.clone().into_iter().collect();
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4950,16 +4958,27 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-                            vir::AssignKind::Copy,
-                        ));
+
+                        // Encoding of string literals is not yet supported, so do not encode
+                        // an assignment if the RHS is a string
+                        let is_str = match ty.kind() {
+                            ty::TyKind::Ref(_, inner, _) => inner.is_str(),
+                            _ => {
+                                false
+                            }
```
```suggestion
                            _ => false,
```

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1446,12 +1446,20 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
 
-                (expiring_base, restored, false, stmts)
+                (expiring_base, Some(restored), false, stmts)
             }
 
             mir::Rvalue::Use(mir::Operand::Constant(box mir::Constant { literal, .. })) => {
                 let (ty, val) = mir_constantkind_to_ty_val(literal);
-                let restored = self.encoder.encode_const_expr(ty, &val)?;
+
+                // Encoding of string literals is not yet supported, so do not return
```
Aurel300: Please add `TODO:` in the comments, so the future work is easier to spot later.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1662,7 +1670,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         if is_mut {
             stmts.extend(self.encode_transfer_permissions(
                 lhs_place.clone(),
-                rhs_place,
+                rhs_place.unwrap(),
```
Aurel300, line 1682: `is_mut` implies `rhs_place.is_some()`?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5583,17 +5602,22 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
             match vir_assign_kind {
                 vir::AssignKind::Copy => {
                     if field.typ.is_ref() {
-                        // TODO: Inhale the predicate rooted at dst_field
-                        return Err(SpannedEncodingError::unsupported(
-                            "the encoding of this reference copy has not \
-                            been implemented",
-                            self.mir_encoder.get_span_of_location(location),
-                        ));
+                        if field.typed_ref_name() == Some("str".to_string()) {
```
Aurel300: Any way to access the MIR type here? If not, perhaps this equality check should be something implemented in `vir::Type`, similar to `is_ref`?

prusti-viper/src/encoder/procedure_encoder.rs
zgrannan: Yes. The tuple returned by `encode_loan_places` only returns `None` for `restored` when the `source` is a string literal (as it can't be encoded); and in the case of the string literal, `is_mut` is `False`. https://github.com/viperproject/prusti-dev/blob/71b66481f11fbc6eb1038a46d7e7271f5edcb8c6/prusti-viper/src/encoder/procedure_encoder.rs#L1458

prusti-viper/src/encoder/procedure_encoder.rs
zgrannan: I refactored this to instead just pass a flag to this enabling reference copies.

The flag right now is true for references to literals, as per Zulipchat https://prusti.zulipchat.com/#narrow/stream/288456-prusti-developers/topic/Supporting.20String.20Literals/near/250032719

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5573,6 +5596,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         field: vir::Field,
         location: mir::Location,
         vir_assign_kind: vir::AssignKind,
+        can_copy_reference: bool
```
fpoli: Could you describe in the documentation? Like, "it's unsound if ...".

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4949,17 +4960,27 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-                            vir::AssignKind::Copy,
-                        ));
+
+                        // TODO Encoding of string literals is not yet supported,
+                        // so do not encode an assignment if the RHS is a string
+                        let is_str = match ty.kind() {
+                            ty::TyKind::Ref(_, inner, _) => inner.is_str(),
+                            _ => false,
+                        };
+                        if !is_str {
```
fpoli: (For another PR) Could this if-body be skipped for any type that `TypeEncoder::encode_predicate_def` encodes with a bodyless predicate, instead of just for `str`?

zgrannan: I think that would be okay.

vakaras (approved):
LGTM. Once the test pass, we can merge it.

shell.nix
```diff
@@ -0,0 +1,18 @@
+{}:
+
```
vakaras: Why is this file needed?

zgrannan: Oh this is not needed!

vakaras:
Thanks!

PR #650, opened 2021-08-19
Unsupported error for reference type fields

ani003:
Unsupported error is returned for reference type fields. Additionally, `is_reference` function has been moved from `mir_encoder` to `utils`. `struct-with-ref` test has been removed from `pass/equality` as it is not supported.


prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -134,6 +134,11 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
                         };
                         let field = &variant_def.fields[field.index()];
                         let field_ty = field.ty(tcx, subst);
+                        if utils::is_reference(field_ty) {
+                            return Err(EncodingError::unsupported(
+                                "reference type fields are not supported"
```
Aurel300:
```suggestion
                                "access to reference-typed fields is not supported",
```

PR #652, opened 2021-08-23
Verify binary search on slices

dario23:
NOTE: this cleverly uses the fact that full slicing isn't supported yet,
but "assuming we have a slice" (i.e. passing as function parameter) is.

### DIFF
--- a/prusti-tests/tests/verify/pass/binary_search.rs
+++ b/prusti-tests/tests/verify/pass/binary_search.rs
@@ -1,0 +1,79 @@
+use prusti_contracts::*;
+
+
+fn main() {}
+
+predicate! {
+    fn sorted(s: &[i32]) -> bool {
+        forall(|i: usize, j: usize| (0 <= i && i < j && j < s.len()) ==> s[i] <= s[j])
+    }
+}
+
+enum UsizeOption {
+    Some(usize),
+    None,
+}
+
+impl UsizeOption {
+    #[pure]
+    fn is_some(&self) -> bool {
+        match self {
+            UsizeOption::Some(_) => true,
+            UsizeOption::None => false,
+        }
+    }
+    #[pure]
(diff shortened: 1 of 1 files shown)


prusti-tests/tests/verify/pass/binary_search.rs
```diff
@@ -0,0 +1,79 @@
...
+        match self {
+            UsizeOption::Some(n) => *n,
+            UsizeOption::None => unreachable!(),
+        }
+    }
+}
+
+
+#[requires(sorted(s))]
+#[requires(s.len() < 18446744073709551615)] // usize limits
```
Aurel300, line 41: Can you omit this if you place the test into `verify_overflow` (which is now the default behaviour)?

dario23: ran with `-Pcheck_overflows=true` to try, but it seems we can't. i think the missing piece of information is that `len()` returns a usize semantically, so the result cannot be bigger than `usize::MAX`, but the `Slice$len` viper pure function currently returns a viper `Int` and only specifies `result >= 0`, no upper bound, and we have kind-of shortcut out the creation of an intermediate `Ref(usize)`-typed value that would carry this information over.

the more i think about it the more complicated it gets though, because this literal value here is `usize::MAX` on x86-64, but not necessarily everywhere else, so while the exact value doesn't matter too much here, this example might fail on a machine with 32bit-`usize` (or not, if no one else relies on the bounds too much i guess?).. i.e. i'm not even sure anymore what value to put here. i guess it's universally safe for a value less than or equal to any actual machine's `usize::MAX`, but it also sounds nice to use a machine's concrete `usize::MAX` so the verification is done for all possible lengths that slices can have on this particular machine.. probably a problem for another day though, so **tl;dr:** check_overflows doesn't help, what value do we want to put here?

PR #654, opened 2021-08-27
Polymorphic vir clean-up

Maxwell-Yang-2001:
Changes:

- **polymorphic_vir** in encoders is now renamed **vir** (and use statements to legacy vir are removed), since all depdencies to legacy vir have been removed.
- **polymorphic_vir_macro** (which used to be a polymorphic equivalent of **vir_macro**) has been moved to become the default **vir_macro**, since all usages of previous vir_macro have been removed. There is also a change to the macro definition (`vir!` separates into `vir_expr! `and `vir_stmt!`, to avoid importing naming collisions with the vir crate).
- All **Stmt** related traits (`StmtWalker`, `StmtFolder`, `FallibleStmtFolder`) have been updated according to the new **Stmt** definition (each enum variant having its own struct).
- Similar to the previous line, all **Expr** related traits (`ExprWalker`, `ExprFolder`, `FallibleExprFolder`) have been updated according to the new **Expr** definition (each enum variant having its own struct).


vakaras (approved):
LGTM Thanks!

PR #656, opened 2021-08-30
Enable clippy for the entire repository

vakaras:


.github/workflows/test.yml
```diff
@@ -74,16 +74,11 @@ jobs:
       - name: Build with cargo
         run: python x.py build --all --verbose
       - name: Check and report Clippy errors
-        env:
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
```
fpoli, line 78: The `fetch-depth: 0` above can be removed too.

PR #657, opened 2021-08-30
Remove prefix from field name

zgrannan:
Rust allows raw identifiers in struct definitions, to encode fields that clash with reserved keywords. For example:

```rust
struct Foo {
    bar: i32,
    r#type: i32
}
```

However, the `#` is not valid in Viper, so referencing the field causes an error

`error: [Prusti internal error] consistency error in check_bar_fn: Consistency error: f$r#type is not a valid identifier`

This PR changes the encoding of the struct field to remove the `r#` prefix.

### DIFF
--- a/prusti-tests/tests/verify/pass/quick/struct.rs
+++ b/prusti-tests/tests/verify/pass/quick/struct.rs
@@ -14,5 +14,10 @@ fn check_bar_fn(foo: &Foo) {
 fn check_rtype_fn(foo: &Foo) {
 }
 
+#[pure]
+fn rtype_same(foo1: &Foo, foo2: &Foo) -> bool {
+    foo1.r#type == foo2.r#type
+}
+
 fn main(){
 }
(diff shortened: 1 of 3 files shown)


zgrannan:
This doesn't work currently when snapshots are involved.

Aurel300:
> This doesn't work currently when snapshots are involved.

https://github.com/viperproject/prusti-dev/blob/306ca0165cae9e9687378d9b39e6c4f5e6b8de90/prusti-viper/src/encoder/snapshot/encoder.rs#L688

https://github.com/viperproject/prusti-dev/blob/306ca0165cae9e9687378d9b39e6c4f5e6b8de90/prusti-viper/src/encoder/snapshot/encoder.rs#L724

zgrannan:
@Aurel300 Thanks for pointing out the place to make the change. This is ready to merge now (the failing checks seem to be due to a network failure, not sure how to re-run them)

PR #658, opened 2021-08-31
Improve error message for types missing structural equality

dario23:
Previously, the error was along the lines of "cannot call impure
function PartialEq::eq from specifications/pure code", which is
confusing for users, because equality is OK most of the time, and the
problem is comparing types without structural equality in
specifications, as missing structural equality likely means we can't
capture what makes two instances equal/different in snapshots.

also, maybe slices should be declared to have structural equality, but
that would likely require more changes as slices mostly occur behind
references, so some reference types would have it, while others don't..

### DIFF
--- a/prusti-tests/tests/verify/ui/no-structural-eq.rs
+++ b/prusti-tests/tests/verify/ui/no-structural-eq.rs
@@ -1,0 +1,9 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[requires(a[0..2] == [0, 0])]
+pub fn foo(a: [i32; 5]) {
+    assert!(a[0] == 0);
+    assert!(a[1] == 0);
+}
--- a/prusti-tests/tests/verify/ui/no-structural-eq.stderr
+++ b/prusti-tests/tests/verify/ui/no-structural-eq.stderr
@@ -1,0 +1,8 @@
+error: [Prusti: invalid specification] cannot compare values of type &[i32] in specifications, as it has no structural equality
+ --> $DIR/no-structural-eq.rs:5:12
+  |
+5 | #[requires(a[0..2] == [0, 0])]
+  |            ^^^^^^^^^^^^^^^^^
+
+error: aborting due to previous error
+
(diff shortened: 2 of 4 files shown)


prusti-viper/src/encoder/pure_function_encoder.rs
```diff
@@ -882,6 +882,22 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                                 state
                             }
 
+                            "std::cmp::PartialEq::eq"
+                            | "std::cmp::PartialEq::ne" => {
+                                // we only get here if `has_structural_eq_impl` returned false
+                                // above
```
Aurel300: Can we keep this in the same match arm but instead of an `if` guard, we return the error `if !has_structural_eq_impl...`?

Aurel300:
Improving the error message is good; we are planning to add a feature to do a snapshot comparison directly, regardless of `(Partial)Eq` implementation or its purity.

prusti-viper/src/encoder/pure_function_encoder.rs
dario23: sounds good, refactored to have all in the same match case.

Aurel300:
Unfortunately the test case now fails on trying to encode the `[0, 0]` constant…

dario23:
looking at this again, i think the problem is worse than than even. yes the  encoding of `[0, 0]` fails, because it's actually of type `&[i32; 2]`, since it's passed like that to the `eq` call. this might be somewhat fixable in `prusti-viper/src/encoder/encoder.rs:encode_const_expr`, probably needs keeping track of the implicit dereference as well, and whatever else i'm not thinking of right now. what  (i think) **really** prevents this from working even then is that the call is `<[i32] as PartialEq<[i32; 2]>>::eq(..)`, i.e. we're comparing a `&[i32]` to a `&[i32; 2]`, and rust lets us, because there is a corresponding trait impl, but i think prusti currently doesn't handle comparing different types well.

so, backtracking once more, not sure if that's a problem that necessarily breaks our necks for this PR, all i wanted was an instance of something without structural equality, to compare to, in order to trigger this error. however, comparing the match blocks in `encode_const_expr` and `has_structural_eq_impl` (both in `prusti-viper/src/encoder/encoder.rs`), the things that are const-encodable all have structural equality (which makes sense, thinking about it). soo.. i'm not sure this PR is really needed anymore, since the issue can't be triggered, at least currently. intending to close this, what do you think?

Aurel300:
I have a long-term solution in mind for supporting "pure" equality in specs (not necessarily based on `StructuralEq`). For now the error message complaining about an impure `PartialEq` implementation will have to do, so yes, I'll close this. Thanks for the PR though!

Issue #627, opened 2021-07-23
Predicates cannot be exported


enjhnsn2:
Hello, I'm trying to use Prusti to verify some code that spans multiple files, and having a bit of trouble with predicates. Specifically, there seems to be no way to make a predicate accessible from another file. This has lead to me having to duplicate all my predicates in each file, which is quite inconvenient. 
For example, if I define a predice `nonnegative` in file `spec.rs`:
```
predicate! {
    fn nonnegative(num: i64) -> bool {
        num >= 0
    }
}
```
I cannot use it in my `algorithm.rs` file 

```
#[requires(nonnegative(a) && nonnegative(b))]
#[ensures(nonnegative(result))]
fn multiply(a: i64, b: i64) -> i64{
    return a * b;
}
```
as it errors out with:
```
error[E0603]: function `safe` is private
```

I tried to add the visibility marker inside the macro: 
```
predicate! {
    pub fn nonnegative(num: i64) -> bool {
        num >= 0
    }
}
```
but this fails with:
```
`predicate!` can only be used on function definitions. it supports no attributes.
```

Is there an established way to export predicates?
If not, how hard would it be to add one?

dario23:
i think there's no established way yet, but it should be fairly straightforward to add the public-nes as an attribute to the generated function if it's there in the original code.

if you want to have a go at it, the following might be useful:

 - parsing is done in [prusti-specs/src/lib.rs](https://github.com/viperproject/prusti-dev/blob/master/prusti-specs/src/lib.rs#L491) by implementing the `syn::Parse` trait for a struct representing a predicate function signature+body.
 - you should be able to add an `Option<syn::Visibility>` to that struct and corresponding code to the `syn::Parse` impl block
 - the rewriting generates two functions named `spec_fn`, which contains the specification snippets generated from the predicate body, and `cleaned_fn`, which is the original function's signature but with an empty body. you'll want to attach the visibility to the second one.
 - bonus points for adding a testcase. i think you can have a single-file testcase like this:
```
use prusti_contracts::*;

fn main() {}

mod foo {
    use super::predicate;
    predicate! {
        pub fn ok() -> bool {
            true
        }
    }
}

#[requires(foo::ok())]
fn test_pub_pred() {}
```

enjhnsn2:
Cool, I'll take a swing at implementing this.

enjhnsn2:
Fixed by https://github.com/viperproject/prusti-dev/pull/670

PR #670, opened 2021-09-13
Allow visibility specifiers on predicates

emlaufer:
This allows predicates to be marked as public so they can be used across module boundaries, see issue #627 . It enables programs like the following:

```
use prusti_contracts::*;

fn main() {}

mod foo {
    use super::predicate;
    predicate! {
        pub fn ok() -> bool {
            true
        }
    }
}

#[requires(foo::ok())]
fn test_pub_pred() {}
```

### DIFF
--- a/prusti-specs/src/lib.rs
+++ b/prusti-specs/src/lib.rs
@@ -478,18 +478,20 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
 
 #[derive(Debug)]
 struct PredicateFn {
+    visibility: Option<syn::Visibility>,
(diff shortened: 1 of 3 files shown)


dario23:
thanks for your PR :-)

i'm not actually a maintainer (so i can't approve the CI run, maybe @Aurel300 or @vakaras can help here?), but this looks pretty much exactly like the changes i imagined in #627, so if i didn't miss anything there, i think this should be good to go pretty much immediately.

vakaras (approved):
LGTM.

vakaras:
Thank you for fixing it!

PR #677, opened 2021-09-21
Test more crates

fpoli:
Let's see how long it takes.


fpoli:
We already have 8 regressions :cry: Better to add these tests soon.

PR #680, opened 2021-09-22
Reuse the build cache in test-crates

fpoli:
Let's see if this gives any speedup. Before this, `test-crates` takes about 49 minutes.


fpoli:
Nope

PR #682, opened 2021-09-24
Enable two-state functions. Fixes #681.

### LINKED ISSUE #681: Using old(struct) in predicate arguments
Hey, 
I'm trying to define a predicate that enforces that a data structure has been safely updated by comparing the old and new versions of the structure, but I'm getting some verification errors. 
For example, if I have a struct Point:
```
struct Point {
    pub x: u32, 
    pub y: u32, 
}
```
I can define a function that updates it and a contract to ensure it was updated properly, like so:
```
#[ensures(old(point.x) + 1 == point.x && old(point.y) + 1 == point.y)]
fn diagonal_shift(point: &mut Point){
    point.x += 1;
    point.y += 1;
}
```
However, if the data structure and its invariants were more complex, I would want to encapsulate the safety invariant in a predicate like so:
```
predicate! {
    fn safely_translated(old_point: &Point, point: &Point) -> bool {
        old_point.x + 1 == point.x && 
        old_point.y + 1 == point.y
    }
}

#[ensures(safely_translated(old(point), point))]
fn diagonal_shift(point: &mut Point){
    point.x += 1;
    point.y += 1;
}
```
However, this causes the error `[Prusti: invalid specification] the type of the old expression is invalid`

So, a few questions:
Is there any standard way to cleanly encapsulate step-wise data structure invariants like this?
I don't really know how `old` interacts with struct field accesses, but would it be difficult to implement the ability to pass the old and new struct to a predicate like this? If not, I would be willing to implement it if you could give me a pointers. 
If it is difficult / fundamentally conflicts with how viper handles structures, do you know of any workaround that is more clean then just reproducing whole set of step-wise in invariants at every function that enforces them?

vakaras:


prusti-interface/src/environment/mod.rs
```diff
@@ -305,7 +305,19 @@ impl<'tcx> Environment<'tcx> {
...
+                    // Shared references are copy.
+                    true
+                }
+                ty::TyKind::Array(_, _) | ty::TyKind::Slice(_) => {
+                    // Arrays and slices are not copy.
+                    false
+                }
+                _ => {
+                    self.type_implements_trait(ty, copy_trait_def_id)
+                }
```
Aurel300, line 322: Why this weird check? Is our notion of `copy` different from types that implement Rust's `Copy` trait?

vakaras: Our implementation of `type_implements_trait` is wrong (it, for example, reports shared references as non-Copy). I should add a comment explaining that.

Issue #665, opened 2021-09-07
Caching builds in CI


fpoli:
It would be useful to cache builds in CI. The proper way seems to be what's described here: https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-caching.

fpoli:
Closed by https://github.com/viperproject/prusti-dev/pull/685

PR #685, opened 2021-09-27
Caching CI Builds

gavinleroy:
This PR was opened to address issue #665 regarding caching CI builds.

From material online, it seems to me that the [`Swatinemt/rust-cache`](https://github.com/marketplace/actions/rust-cache) action is frequently used for Rust projects and is the best option. I've used `rust-cache@v1.3.0` as it is a stable release, whereas `rust-cache@v1` points to the master branch. 

One potential problem I see here is that the [action job](https://github.com/marketplace/actions/rust-cache#cache-details) is part of the key. Currently, if a key is specified this is *in addition to* the automatic keys, however, in the future there may be an API option for overriding the key, in which case the GitHub Action job could be removed from the key. To my understanding, this means that between different jobs (e.g. building and testing) there will be a cache miss when theoretically they could use the same cached build. 

A few simple timings I gathered:

| OS          | Build Cache Miss | Build Cache Hit |
| ------------ | ----------------------- | --------------------- |
| Ubuntu    | 11m 19s               | 4m 13s              |
| Windows | 23m 48s               | 7m 11s              |
| Mac         | 17m 31s              | 5m 31s               |

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,8 +28,6 @@ jobs:
           java-version: '15'
       - name: Set up the environment
         run: python x.py setup
-      - name: Cache cargo
-        uses: Swatinem/rust-cache@v1.3.0
       - name: Build with cargo --release
         run: python x.py build --release --all --verbose
       - name: Run cargo tests --release
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,8 +1,8 @@
 name: Deploy
 
-on:
-  push:
-    branches: 'master'
+on: [push, workflow_dispatch]
+  # push:
+  #   branches: 'master'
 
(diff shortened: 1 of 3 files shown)


fpoli:
Nice! Thanks

fpoli:
My only preference is to leave `.github/workflows/deploy.yml` without caching, just to have less moving parts in the build that releases Prusti. Since we use nightly `rustc`, it's possible to get weird bugs from time to time. The deploy workflow is not called when checking PRs, so it shouldn't be noticeable.

gavinleroy:
Yes, I agree, caching the build in `deploy.yml` is indeed unnecessary.

I also just added caching to the `test-on-crates.yml` which is also run on pull requests.

fpoli:
Great, thanks!

PR #686, opened 2021-09-27
Reenable datatype invariants

zgrannan:


prusti-interface/src/specs/mod.rs
```diff
@@ -65,7 +65,13 @@ pub struct SpecCollector<'a, 'tcx: 'a> {
 
     /// Resolved specifications.
     procedure_specs: HashMap<LocalDefId, ProcedureSpecRef>,
+
     loop_specs: HashMap<LocalDefId, Vec<SpecificationId>>,
+
+    // Struct invariants
+    // The keys are the struct ID definition, the first entry in each tuple
+    // is the id of the procedure implementing the spec
+    struct_specs: HashMap<LocalDefId, Vec<(LocalDefId, SpecificationId)>>
```
Aurel300: Could this be a proper `struct` similar to `ProcedureSpecRef`?

prusti-interface/src/specs/mod.rs
```diff
@@ -335,6 +351,8 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for SpecCollector<'a, 'tcx> {
                     SpecType::Postcondition
                 } else if fn_name.starts_with("prusti_pred_item_") {
                     SpecType::Predicate
+                } else if fn_name.starts_with("prusti_struct_invariant_") {
+                    SpecType::Invariant
```
Aurel300: Let's have different `SpecType`s for loop invariants and type invariants.

prusti-interface/src/specs/mod.rs
```diff
@@ -343,12 +361,23 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for SpecCollector<'a, 'tcx> {
             let spec_item = SpecItem {spec_id, spec_type, specification};
             self.spec_items.push(spec_item);
 
-            // Collect loop invariant
+            // Collect invariants
             if spec_type == SpecType::Invariant {
-                self.loop_specs
+                if is_loop_invariant {
+                  self.loop_specs
```
Aurel300: The indentation is wrong, but also this branching should be replaced by the two separate `SpecType`s mentioned above.

prusti-interface/src/specs/mod.rs
```diff
@@ -375,3 +404,26 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for SpecCollector<'a, 'tcx> {
         }
     }
 }
+
+fn get_struct_id_from_impl_node(node: rustc_hir::Node) -> Option<DefId> {
+  let item_kind = if let rustc_hir::Node::Item(item) = node {
```
Aurel300: The indentation is wrong here (2 spaces instead of 4, inconsistent in some branches). Also, although the `Option` "try" operators are neat, I think it's much clearer to just express this as 4 nested `if let` conditionals.

prusti-tests/tests/verify/fail/invariants/invariants-basic-2.rs
```diff
@@ -0,0 +1,25 @@
+extern crate prusti_contracts;
```
Aurel300: Can all the invariants tests be placed into `verify_overflow`? Also have you looked if there are any `invariant`-related tests in `prusti-tests/tests_old`?

prusti-specs/src/specifications/common.rs
```diff
@@ -315,6 +315,24 @@ impl<EID, ET, AT> LoopSpecification<EID, ET, AT> {
...
+pub struct StructSpecification<EID, ET, AT> {
+    pub invariant: Vec<(ET, Assertion<EID, ET, AT>)>
+}
+
+impl<EID, ET, AT> StructSpecification<EID, ET, AT> {
+    pub fn new(invariant: Vec<(ET, Assertion<EID, ET, AT>)>) -> Self {
+        Self { invariant }
+    }
+    pub fn empty() -> Self {
+        Self::new(Vec::new())
```
Aurel300:
```suggestion
        Self::new(vec![])
```

Also is this function used?

prusti-tests/tests/verify/fail/invariants/invariants-basic-2.rs
zgrannan: I've moved the tests from test-old into the `tests/verify_overflow`

vakaras:
@zgrannan I thought a bit about the possible implementation of type invariants and I think that the cleanest way of integrating them into the refactored system would be by renaming the new [`TypeDecl`](https://github.com/viperproject/prusti-dev/blob/7f2e8545e901429e5971f8f83234a3e8c3cbd68b/vir/defs/high/ast/type_decl.rs#L5-L23) into `TypeDeclKind` and adding a new `TypeDecl`:
```rust
struct TypeDecl {
  kind: TypeDeclKind,
  user_invariant: Box<Expression>,
}
```

Let me know if you have any questions or need help.

Aurel300:
@vakaras Is the new `TypeDecl` a misnomer then? If these types are meant to be passed around similarly to `ty::Ty` or (the old) `vir::Type`, they are not *declarations*.

fpoli:
Why the `Box`? It seems unnecessary.

vakaras:
> @vakaras <https://github.com/vakaras> Is the new TypeDecl a misnomer
> then? If these types are meant to be passed around similarly to ty::Ty or
> (the old) vir::Type, they are not *declarations*.
>
It is a declaration. The thing passed around is `vir::high::Type`.

>

vakaras:
> Why the Box? It seems unnecessary.
>
Yes, mixed up code.

>

fpoli:
> Why the Box? It seems unnecessary.
> Yes, mixed up code.
> […](#)

If it turns out to be needed to cut a mutual recursion with the expression definition then I prefer it over `Box<TypeDecl>`.

JonasAlaif:
So I've been thinking about this a little, particularly with adding "dependent parameters" expressed as const generics. The classic use case with a Type Alias is the following:
```rust
#[invariant(_ < s.len())]
type Index<const s: &[u32]> = usize;
fn foo(s: &[i32], i: Index<{s}>) -> i32 { s[i] }
```
Porting this from my work on Type Aliases to Datatypes wouldn't be so difficult, and it seems to be a super nice solution to the `invariants-func.rs` example:
```rust
#[invariant(self.invariants_enabled ==> self.value <= 100)]
struct Percentage {
    invariants_enabled: bool,
    value: u8,
}
fn print_percentage(percentage: &Percentage) { }

#[requires(percentage.invariants_enabled)]
fn adjust_percentage(percentage: &mut Percentage) {
    percentage.value += 10; // temporarily break invariant

    percentage.invariants_enabled = false;
    print_percentage(percentage);
    percentage.invariants_enabled = true;

    percentage.value -= 10; // restore invariant
}
```
Would instead be:
```rust
#[invariant(invariants_enabled ==> self.value <= 100)]
struct Percentage<const invariants_enabled: bool> {
    value: u8,
}
fn print_percentage(percentage: &Percentage<{false}>) { }

fn adjust_percentage(percentage: &mut Percentage<{true}>) {
    percentage.value += 10; // temporarily break invariant

    print_percentage(percentage);

    percentage.value -= 10; // restore invariant
}
```
However, my current solution requires the whole thing be wrapped in a `prusti!{}` macro to find all of the `{true}` and `{false}` arguments and hide them from the type system - even though the current Rust const generics support `bool` types. So an even nicer solution would be to try to write this as 'valid' Rust:
```rust
#![feature(const_generics_defaults)]

#[invariant(invariants_enabled ==> self.value <= 100)]
struct Percentage<const invariants_enabled: bool = true> {
    value: u8,
}
fn print_percentage(percentage: &Percentage<false>) {}

fn adjust_percentage(percentage: &mut Percentage) {
    percentage.value += 10; // temporarily break invariant

    print_percentage(percentage as &Percentage<false>);

    percentage.value -= 10; // restore invariant
}
```
The only snag is that we cannot `percentage as &Percentage<false>` - I don't know const generics well enough; the only thing I could come up with is creating a new instance: `&Percentage::<false> { value: percentage.value }`, but that kinda defeats the purpose.

Edit: @vakaras suggested `&Percentage::<false> { ..percentage }` which actually seems to me as a viable option. For us to support this the `invariant` macro would have to not panic when processing const generics `struct S<const x: bool = true> {}` and Prusti would have to include support for const generics.

zgrannan:
@vakaras
I recall a few weeks ago this was blocked because the datatype invariants encoding uses: https://github.com/viperproject/prusti-dev/blob/2bce0c81854db2e955dd1a595bfc263f83c0e092/prusti-viper/src/encoder/encoder.rs#L625-L637 which returns a `vir::Expr`, but the type invariant encoder: https://github.com/viperproject/prusti-dev/blob/2bce0c81854db2e955dd1a595bfc263f83c0e092/prusti-viper/src/encoder/mir/types/encoder.rs#L525 needs the body to be `vir::Expression`, and there is no way to go from polymorphic vir to high vir.

Is this still the case?

vakaras:
> Is this still the case?

Yes. And I doubt that https://github.com/viperproject/prusti-dev/pull/756 will be enough to fix that.

Aurel300:
I rebased this. I think the [last comment](https://github.com/viperproject/prusti-dev/pull/686#issuecomment-964799664) still *kind of* applies. More specifically, I see:

- `mir::types::MirTypeEncoderInterface::encode_type_invariant_def_poly` into which I rebased the implementation in this PR. This function returns a `vir_poly` function and is never used.
- `high::types::HighTypeEncoderInterface::encode_type_invariant_use/def` which is currently a stub. The first one is actually used from the procedure encoder. However, even though these are in `high/types`, they also return a `String` and a `vir_poly::FunctionIdentifier`, respectively.

So… should I just remove the `encode_type_invariant_def_poly` and move it into `high`?

vakaras:
> I rebased this. I think the [last comment](https://github.com/viperproject/prusti-dev/pull/686#issuecomment-964799664) still _kind of_ applies. More specifically, I see:
> 
>     * `mir::types::MirTypeEncoderInterface::encode_type_invariant_def_poly` into which I rebased the implementation in this PR. This function returns a `vir_poly` function and is never used.
> 
>     * `high::types::HighTypeEncoderInterface::encode_type_invariant_use/def` which is currently a stub. The first one is actually used from the procedure encoder. However, even though these are in `high/types`, they also return a `String` and a `vir_poly::FunctionIdentifier`, respectively.
> 
> 
> So… should I just remove the `encode_type_invariant_def_poly` and move it into `high`?

Seems like that would be the easiest option.

JonasAlaif:
It seems that after the rebase, things don't work at all atm (i.e. things behave as if there was no invariant annotation)

Aurel300:
@JonasAlaif I do have some WIP for this locally, but I want to wait until after #899, because invariants with generics were not nice.

Aurel300:
bors r+

(@vakaras I'll add the trusted flag shortly)

PR #687, opened 2021-09-27
Encode statements in the block even if there is a panic

zgrannan:
The following function would cause Prusti to crash:

```rust
fn prepare_client(client_state: &AnyClientState) -> Result<u32, u64> {
    let r = match client_state {
        AnyClientState::Tendermint(cs) => cs,
        _ =>  return Err(0)
    };
    panic!()
}
```

I think the reason is that the behaviour of `encode_block_statements` is to not encode the statements if the block ends in `panic!()`. However, that doesn't seem sound, considering it is possible for the control flow to return explicitly with a `return` statement (in fact the `panic!()` may never be reachable).

This removes the panic check and always encodes the statements instead.

### DIFF
--- a/prusti-interface/src/environment/procedure.rs
+++ b/prusti-interface/src/environment/procedure.rs
@@ -151,36 +151,6 @@ impl<'tcx> Procedure<'tcx> {
         self.reachable_basic_blocks.contains(&bbi)
     }
 
-    pub fn is_panic_block(&self, bbi: BasicBlockIndex) -> bool {
-        if let TerminatorKind::Call {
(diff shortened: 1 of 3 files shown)


fpoli:
Nice! I think that the `if !is_panic_block { ... }` used to have an else branch generating an `assert false` statement to report the panic, but for some reason it got lost. The check was there because Prusti was crashing when encoding the statements that prepare the string of the panic message. However, with your PR https://github.com/viperproject/prusti-dev/pull/648 Prusti no longer crashes, so there is no need to keep that check.

PR #692, opened 2021-10-02
Fixed missing support for `#![no_std]` crates in some places

Pointerbender:
Hi! I'm currently kicking the tires on Prusti in a `#![no_std]` crate. I ran into an error message `[Prusti: invalid specification] use of impure function "core::cmp::PartialEq::eq" in pure code is not allowed` when trying to perform a structural equality check in a Prusti contract. I started digging around the Prusti codebase a bit once I realized this error was not happening without the `#![no_std]` attribute applied at the crate level. There I noticed that in some places the codebase does anticipate certain methods being in both the Rust standard library and the Rust core library, but it was missing a check for this in other places. I did a quick audit of the code base to see if there were any other places and found some other cases, too. I fixed all the occurrences that I could find and added a regression test for  `core::cmp::PartialEq::{eq,ne}`.

By the way, thank you for building such a marvelous tool! I'm enjoying it a great deal and I'm heavily impressed with all the cool features it already offers. This is my first PR for Prusti!

### DIFF
--- a/prusti-specs/src/extern_spec_rewriter.rs
+++ b/prusti-specs/src/extern_spec_rewriter.rs
@@ -231,7 +231,7 @@ pub fn generate_new_struct(item: &syn::ItemImpl) -> syn::Result<syn::ItemStruct>
     // Add `PhantomData` markers for each type parameter to silence errors
     // about unused type parameters.
     for param in generics.params.iter() {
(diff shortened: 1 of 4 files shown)


vakaras (approved):
Thank you!

PR #695, opened 2021-10-04
Make it easier to read the generated VIR code

fpoli:
The generated code is piped through `rustfmt` and it's then included using `#[path = "../gen/vir_gen.rs"]`, which is recognized by CLion. Is it also understood by rust-analyzer?


vakaras:
> Is it also understood by rust-analyzer?

I have not tried, but also I do not see the reason why it should not work.

vakaras (approved):
LGTM

PR #698, opened 2021-10-05
Clean up expiring borrows

fpoli:
The code that processes expiring borrows seems unnecessarily complicated. In this PR I'm rewriting its methods that don't use `self` into functions, I'm replacing various `&mut self` with `&self` and I'm moving many `debug!` messages to `trace!`.

### DIFF
--- a/prusti-viper/src/encoder/foldunfold/mod.rs
+++ b/prusti-viper/src/encoder/foldunfold/mod.rs
@@ -358,142 +358,10 @@ impl<'p, 'v: 'p, 'tcx: 'v> FoldUnfold<'p, 'v, 'tcx> {
         );
         nearest_block.unwrap().0
     }
-
-    /// Restore `Write` permissions that were converted to `Read` due to borrowing.
-    fn restore_write_permissions(
-        &self,
-        borrow: vir::borrows::Borrow,
-        pctxt: &mut PathCtxt,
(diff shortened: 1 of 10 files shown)


fpoli:
Uh, why isn't CI running on the latest commit?

prusti-viper/src/encoder/foldunfold/mod.rs
```diff
@@ -1112,44 +985,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> vir::CfgReplacer<PathCtxt<'p>, ActionVec> for FoldUnf
...
-                "[exit] prepend_join(..{}): {}",
-                &bcs.len(),
-                branch_actions_vec
-                    .iter()
-                    .map(|v| format!("[{}]", v.iter().to_sorted_multiline_string()))
-                    .to_string()
-            );
-            Ok((branch_actions_vec, merged_pctxt))
-        }
+        prepend_join(bcs)
```
Aurel300, line 988: Why should this `prepend_join` exist if it just delegates directly to the `prepend_join` below?

fpoli: Because `process_expiring_borrows.rs` needs to call `prepend_join`, but I didn't want to pass to it the mutable reference required to call this method. An alternative solution is to use `&self` instead of `&mut self` in the trait definition.

PR #699, opened 2021-10-05
Display an error message if an unsupported index type is used.

zgrannan:
Previously, the following pure function would cause Prusti to crash:

```rust
#[pure]
fn get_third(v: &Vec<u32>) -> u32 {
    v[2]
}
```

with the error:

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', prusti-viper/src/encoder/pure_functions/interpreter.rs:513:71
```

Now, it reports an error:
```
[Prusti: unsupported feature] Using usize as index/range type for &std::vec::Vec<u32> is not supported yet
```


fpoli:
What about adding "in pure functions" to the error message? Otherwise it seems to apply to impure code as well.

zgrannan:
Done!

PR #703, opened 2021-10-07
Split out the type encoder

vakaras:


prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -67,106 +75,87 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
-                    "inference types are not supported"
+                    "inference types are not supported",
                 ));
             }
 
             ty::TyKind::Error(_) => {
-                return Err(EncodingError::unsupported(
-                    "error type is not supported"
-                ));
+                return Err(EncodingError::unsupported("error type is not supported"));
```
Aurel300, line 158: I think there is now a utility function to give the readable name for a `Ty`, somewhere in `utils`?

vakaras: Unfortunately, I cannot find it. Do you remember where it could be?

Aurel300: This is the one I meant: https://github.com/viperproject/prusti-dev/blob/3dcefa903c4c3024850651179232f907dffd07ed/prusti-viper/src/utils/mod.rs#L13

PR #708, opened 2021-10-11
Fix #707.

### LINKED ISSUE #707: Consistency error: expected the same type, but got Ref and Snap$m_A$_beg_$_end_ (@0.0)
Hi :wave: I've ran into a different consistency error, which may or may not be related to #693. There are a couple of differences between this case and #693:

1. The error message is different: `Consistency error: expected the same type, but got Ref and Snap$m_A$_beg_$_end_ (@0.0)`
2. In the below example, the consistency error appears when calling a pure function from a non-pure function (#693 is about calling a pure function from a pure function).

An annotated example that reproduces the error message is as follows:

```rust
use prusti_contracts::*;

#[derive(Clone,Copy)]
pub struct A {
    _inner: usize,
}

#[derive(Clone,Copy)]
pub struct B {
    inner: A,
}

impl B {
    #[pure] // <-- removing this attribute makes the below error go away
    // #[trusted] // <-- adding this attribute doesn't help
    pub fn get(&self) -> A {
        self.inner
    }
}

// This method triggers the Prusti error:
// "error: [Prusti internal error] consistency error in test: Consistency error:
// expected the same type, but got Ref and Snap$m_A$_beg_$_end_ (@0.0)"
// The error only shows up when returning an ADT, there is no error when
// the return type is a primitive.
pub fn test(b: &B) -> A {
    b.get()
}
```

The reason I think it might be related, is because it looks like it's also tripping over the snapshot encoding of `A` (`Snap$m_A$_beg_$_end_`), which is similar to #693, even though the error message is different.

I'll try to see if I can narrow down where this is coming from in the Viper program and report back!

vakaras:


prusti-tests/tests/verify_overflow/pass/issues/issue-707.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn get(&self) -> A {
+        self.inner
+    }
+}
+
+// This method triggers the Prusti error:
+// "error: [Prusti internal error] consistency error in test: Consistency error:
+// expected the same type, but got Ref and Snap$m_A$_beg_$_end_ (@0.0)"
+// The error only shows up when returning an ADT, there is no error when
+// the return type is a primitive.
```
Aurel300, line 25: Should we really keep these comments here? I know they are from the issue as written, but they might be misleading in the future.

Issue #709, opened 2021-10-11
[Arrays] Consistency error: expected the same type, but got Snap$m_A$_beg_$_end_ and Ref (@16.23)


Pointerbender:
Hi again! I've found another `Consistency error: expected the same type, but got Snap$m_A$_beg_$_end_ and Ref` error and it's similar to #707. This time it's in the context of arrays. This program triggers two consistency errors:

```rust
#[derive(Clone, Copy)]
pub struct A(usize);

#[derive(Clone, Copy)]
pub struct B {
    inner: [A; 16],
}

impl B {
    pub fn new(a: A) -> Self {
        Self { inner: [a; 16] }
        //            ^^^^^^^
        // error: [Prusti internal error] consistency error in B::new:
        // Consistency error: expected the same type, but got Snap$m_A$_beg_$_end_ and Ref
    }
}

pub fn test() {
    let mut b = B::new(A(0));
    b.inner[2] = A(1);
    //     ^^^
    // error: [Prusti internal error] consistency error in test:
    // Consistency error: expected the same type, but got Snap$m_A$_beg_$_end_ and Ref

}

pub fn main() {}
```

The first one is due to an `inhale` statement where the left-hand side is snapshot encoded, but the right-hand side is not:

```
function lookup_pure__$TY$__Array$16$m_A$_beg_$_end_$$int$$Snap$m_A$_beg_$_end_(self: Ref, idx: Int): Snap$m_A$_beg_$_end_
      requires acc(Array$16$m_A$_beg_$_end_(self), read$())
      requires 0 <= idx
      requires idx < 16

method m_new() returns (_0: Ref)
      --- SNIP ---
      var _2: Ref
      var _3: Ref
      --- SNIP ---
      _2 := builtin$havoc_ref()
      inhale acc(Array$16$m_A$_beg_$_end_(_2), write)
      inhale lookup_pure__$TY$__Array$16$m_A$_beg_$_end_$$int$$Snap$m_A$_beg_$_end_(_2, 0) == _3
      --- SNIP ---
```

The second one has the same problem, but performs a modifying lookup through a temporary variable:

```
field f$inner: Ref

function lookup_pure__$TY$__Array$16$m_A$_beg_$_end_$$int$$Snap$m_A$_beg_$_end_(self: Ref, idx: Int): Snap$m_A$_beg_$_end_
      requires acc(Array$16$m_A$_beg_$_end_(self), read$())
      requires 0 <= idx
      requires idx < 16

[...]

PR #710, opened 2021-10-11
Issue #709. Fix and extra regression test case for `ProcedureEncoder::encode_assign_array_repeat_initializer()`

Pointerbender:
I've created a fix and regression test case for `ProcedureEncoder::encode_assign_array_repeat_initializer()`, where it failed to initialize arrays holding ADTs. I'm currently not sure about a couple of things, such as:

* Should I check whether the right-hand size implements the `Copy` trait before applying the snapshot encoding? Or would it never end up here in the first place due to that a "pure" array lookup is already happening? (i.e. it would fail before it gets to the lookup if the return type is not `Copy`).
* In this case, it seems to work without generating additional pre-statements. However, I can already give one corner case where this breaks. In this counter-example Viper complains about the missing `UnitDomain` (I guess it's missing because the pre-statements where not properly generated with my fix?)
```rust
#[derive(Clone, Copy)]
pub struct C([(); 16]);

impl C {
    #[pure]
    /// Initialization of an array holding units `()`.
    pub fn new() -> Self {
        Self([(); 16])
    }
}
```

### DIFF
--- a/prusti-tests/tests/verify/pass/arrays/repeat-expr-initialization.rs
+++ b/prusti-tests/tests/verify/pass/arrays/repeat-expr-initialization.rs
@@ -4,6 +4,23 @@
 //   initializers like [1, 2, 3]
 // - bool as element type
 
+extern crate prusti_contracts;
+use prusti_contracts::*;
+
+#[derive(Clone, Copy)]
+pub struct A(usize);
+
+#[derive(Clone, Copy)]
+pub struct B([A; 16]);
+
+impl B {
+    #[pure]
+    /// Initialization of an array holding ADTs.
+    pub fn new() -> Self {
+        Self([A(0); 16])
+    }
+}
+
 fn main() {
     let a = [false; 3];
 
--- a/prusti-tests/tests/verify/pass/arrays/repeat-expr-initialization.rs
+++ b/prusti-tests/tests/verify/pass/arrays/repeat-expr-initialization.rs
@@ -4,23 +4,6 @@
 //   initializers like [1, 2, 3]
 // - bool as element type
 
-extern crate prusti_contracts;
-use prusti_contracts::*;
-
-#[derive(Clone, Copy)]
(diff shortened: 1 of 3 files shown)


prusti-tests/tests/verify/pass/arrays/repeat-expr-initialization.rs
```diff
@@ -4,6 +4,23 @@
 //   initializers like [1, 2, 3]
 // - bool as element type
 
+extern crate prusti_contracts;
```
fpoli: I would place the test in `verify/pass/issues/issue-709.rs`, to make it clear that it'a regression test for that issue.

vakaras (commented):
Thanks! The PR looks good to me. I am just wondering whether scattering `is_snapshot() || is_domain()` in multiple places is the cleanest way of going forward. I need to think about this.

vakaras:
Anyway, once the test file is in the right folder, we can merge this in.

Pointerbender:
Thanks guys! I'm planning to update the PR with the feedback later today.

Pointerbender:
The PR is now updated and the CI is running :+1:

Aurel300:
It might help if we changed `is_snapshot` to consider any domain that is produced by the snapshot encoder as a snapshot. E.g. an instance of a `UnitDomain` really is a snapshot…

vakaras:
Thank you @Pointerbender for fixing this!

vakaras:
> It might help if we changed `is_snapshot` to consider any domain that is produced by the snapshot encoder as a snapshot. E.g. an instance of a `UnitDomain` really is a snapshot…

I am planning to rewrite all this stuff soonish and move domains and snapshots to different VIR layers because, at least to me, domains are a lower level construct used to implement snapshots and it does not make much sense to mix them.

Pointerbender:
Thanks everyone!

PR #712, opened 2021-10-12
Refactor type encoder

vakaras:


prusti-viper/src/encoder/errors/with_span.rs
```diff
@@ -12,6 +12,8 @@ use log::trace;
 /// `Result<T, SpannedEncodingError>`.
 pub trait WithSpan<T> {
     fn with_span<S: Into<MultiSpan>>(self, span: S) -> Result<T, SpannedEncodingError>;
+    // FIXME: Make the method names consistent.
+    fn set_span_with<S: Into<MultiSpan>>(self, span_callback: impl Fn() -> S) -> Result<T, SpannedEncodingError>;
```
fpoli, line 16: `with_span_fn`? `with_span_cb`?

vakaras: `with` suffix is a convention in Rust to insert a missing value: https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with

PR #719, opened 2021-10-16
Issue #718 regression tests

Pointerbender:
The test run for this PR is currently failing, because `x.py test` is not recognizing the error shown by `prusti-rustc` (see #718). This is PR is to show how to reproduce the problem. Once the test suite can recognize the error, these tests can both become regression tests :-)

Issue #683, opened 2021-09-25
Fold-unfold error when a pure function returns a non-pure value


vakaras:
Test: `prusti-tests/tests/verify/fail/issues/issue-416.rs`.

fpoli:
I would disable that test rather than making the test expect an internal error. No program in the test suite should ever expect an internal error IMO.

vakaras:
> No program in the test suite should ever expect an internal error IMO.

Not sure I agree. This corresponds to `UnexpectedOutput` errors we have in Viper and I find them useful to catch things that were “fixed by accident”. For example, things like functions taking arbitrary `Copy` arguments, which we had semi-working for 3 months until I actually noticed that.

fpoli:
Sure, but the proper error should report to the user that pure functions can only return a pure value, not an internal fold/unfold error.

vakaras:
True.

JonasAlaif:
Why is the file called `issue-416.rs` when this is 683? (i.e. can/should I rename it?)

fpoli:
The file name is correct; it's a regression test for https://github.com/viperproject/prusti-dev/issues/416.

@vakaras Do you agree that `issue-416.rs` should be supported? I have no idea what's a *non-pure* value.

Also, why don't we report "unsafe code is unsupported" errors for that test? I think we should.

Answered those questions, if the test doesn't raise an internal error anymore I think we can remove the comment about 683 and close this issue.

vakaras:
> @vakaras Do you agree that `issue-416.rs` should be supported? I have no idea what's a _non-pure_ value.

`Vec::get_unchecked` is a non-pure method, so the example should be rejected at least for that reason.

> Also, why don't we report "unsafe code is unsupported" errors for that test? I think we should.

Simply because nobody had time to add such checks.

fpoli:
> > Also, why don't we report "unsafe code is unsupported" errors for that test? I think we should.
> 
> Simply because nobody had time to add such checks.

On a second thought, I think we don't have such checks because Prusti assumes that all called functions are verified and that it's the verification of `Vec::get_unchecked` that should report the "unsafe code is unsupported" error. This is related to the default precondition `true` that we use for external functions. I opened https://github.com/viperproject/prusti-dev/issues/910 to track that, so that this issue can be closed by https://github.com/viperproject/prusti-dev/pull/895.

PR #720, opened 2021-10-16
Issue #709 additional regression tests

Pointerbender:
Adding some additional regression tests for the snapshot encoding of arrays and slices per [#709](https://github.com/viperproject/prusti-dev/issues/709#issuecomment-943297838). Most regression tests are currently failing and are placed in the `prusti-tests/tests/verify/fail/issues` directory.


fpoli:
Thanks for preparing these minimal examples and the one in #719! Any internal error (e.g. consistency errors, fold-unfold errors) is definitely a bug and a PR like this is really nice to make sure that the issue is reproducible. However, it's not something that I would merge because the tests currently expect the issue to *happen*, while in practice we don't care if the error message changes due to e.g. some refactoring or a compiler upgrade. In other words, what are the tests testing? Personally, I would leave the PR open until the issue is fixed, or we could move these tests to a different folder that is not normally run by CI, or we could add these programs as attachments to the issue #709.

vakaras:
> Thanks for preparing these minimal examples and the one in #719! Any internal error (e.g. consistency errors, fold-unfold errors) is definitely a bug and a PR like this is really nice to make sure that the issue is reproducible. However, it's not something that I would merge because the tests currently expect the issue to _happen_, while in practice we don't care if the error message changes due to e.g. some refactoring or a compiler upgrade. In other words, what are the tests testing? Personally, I would leave the PR open until the issue is fixed, or we could move these tests to a different folder that is not normally run by CI, or we could add these programs as attachments to the issue #709.

@fpoli I have a different opinion on this: while getting internal errors is not ideal, crashes would be significantly worse and they can be reintroduced by refactorings if we do not have tests covering such edge cases. Therefore, I would suggest merging this PR.

fpoli:
Testing that Prusti doesn't crash is useful, I agree, but then I would place the tests in `tests/core_proof` (if it still works), so that the content of the error messages is not checked:
https://github.com/viperproject/prusti-dev/blob/7f2e8545e901429e5971f8f83234a3e8c3cbd68b/prusti-tests/tests/compiletest.rs#L181-L183

vakaras:
> Testing that Prusti doesn't crash is useful, I agree, but then I would place the tests in `tests/core_proof` (if it still works), so that the content of the error messages is not checked:
> 
> https://github.com/viperproject/prusti-dev/blob/7f2e8545e901429e5971f8f83234a3e8c3cbd68b/prusti-tests/tests/compiletest.rs#L181-L183

In this case, the examples have specifications, so I am not sure whether running with panics disabled will help. Also, we can easily disable the checking of the actual error message by leaving only `//~ ERROR` without any text following it.

fpoli:
I see, so part of the examples are fully supported and other parts are only checked not to crash Prusti. Removing the message from `//~ ERROR` wouldn't be enough; in some cases no error message should be reported at all if the issue is fixed.

The principle that I'm trying to apply is that a test is "good" if "good" code changes won't cause the test to fail. Fixing #709 should be a good code change, but it would break the tests in this PR because the tests expect unexpected errors to happen. This means that they will require adjustments from time to time, which is additional maintenance.

This is the second time that the principle would be broken (first being https://github.com/viperproject/prusti-dev/issues/683), out of 600+ good tests that we currently have. I think that this principle helped *a lot* so far. When the test suite fails or CI is red, I trust that it's for a good reason; I don't need to open every single file to find out if the test is checking something that it's not supposed to check.

vakaras:
> Fixing #709 should be a good code change, but it would break the tests in this PR because the tests expect unexpected errors to happen. This means that they will require adjustments from time to time, which is additional maintenance.

I do not think that this will cause additional maintenance because if you fix the issue, you should add a regression test, so you will have to change the test suite anyway. Having a test in the test suite has an additional benefit that in case the bug is fixed by “unrelated work”, we will notice that.

> When the test suite fails or CI is red, I trust that it's for a good reason; I don't need to open every single file to find out if the test is checking something that it's not supposed to check.

We could achieve this by moving all these tests to a separate CI job so that when something fails, you can understand what kind of problem is causing that without looking into every single file. Would that be an acceptable compromise for you?

fpoli:
> I do not think that this will cause additional maintenance because if you fix the issue, you should add a regression test, so you will have to change the test suite anyway. Having a test in the test suite has an additional benefit that in case the bug is fixed by “unrelated work”, we will notice that.

The maintenance that I'm referring to are the test-adjustments needed for the changes that don't fix the issue.

> We could achieve this by moving all these tests to a separate CI job so that when something fails, you can understand what kind of problem is causing that without looking into every single file. Would that be an acceptable compromise for you?

Yes, that's one of the alternatives I suggested above.

fpoli:
So, what about a `tests/verify/fail/partially_supported`, or a `tests/verify_partial` folder? To me, a special folders to mark these tests is enough, even if in the end they run in the same CI job.

vakaras:
> So, what about a `tests/verify/fail/partially_supported`, or a `tests/verify_partial` folder? To me, a special folders to mark these tests is enough, even if in the end they run in the same CI job.

I would vote for `tests/verify_partial`.

Pointerbender:
> I would vote for tests/verify_partial.

Thank you both for working out this alternative. Shall I update this PR and PR #719 already to move the test files into the new folder? :) Where the file names will be following the pattern `prusti-tests/tests/verify_partial/issues/issue-*.rs` if I'm understanding correctly? i.e. no intermediary `fail` folder, since these partially verified programs are always expected to fail compilation. If these programs would no longer fail verification at some point, then these can be moved to `prusti-tests/tests/verify/pass/` instead.

fpoli:
Yes, please :) If we also include the `fail` folder (`prusti-tests/tests/verify_partial/fail/issues/issue-*.rs`) then the changes are minimal; you just need to copy-paste these lines to add the new test folder: https://github.com/viperproject/prusti-dev/blob/7f2e8545e901429e5971f8f83234a3e8c3cbd68b/prusti-tests/tests/compiletest.rs#L177-L179

I agree that there should be no `pass` folder, but `ui` could be used the same way as `fail`.

Pointerbender:
Good point about the `ui/` folder. I will also add the `fail/` folder in that case and include the changes to `prusti-dev/prusti-tests/tests/compiletest.rs` in this PR. Thanks!

fpoli:
Thanks to you :)

Pointerbender:
The test cases have been moved into `prusti-tests/tests/verify_partial/fail/issues/` and the test suite has been updated to run these tests as well (with overflow checks enabled). The CI passed and it is now ready to merge :)

fpoli:
Final nit :) Could you add a newline at the end of every new file when missing and replace `pub fn main() {}` with `fn main() {}`?

Pointerbender:
Sure thing! Shall I see if I can add a rule for the extra newline at the end to `rustfmt` as well? (maybe in a separate PR if this causes any regressions in other files outside of this PR)? Merely out of curiosity: why is the `pub` omitted in `fn main() {}`? I had noticed it before in other files, but now I'm even more curious about the underlying reason :) Thanks!

Pointerbender:
I just pushed the requested fixes :)

fpoli:
> Shall I see if I can add a rule for the extra newline at the end to rustfmt as well? (maybe in a separate PR if this causes any regressions in other files outside of this PR)?

Uh, yes, thanks! I didn't know that it can be checked by rustfmt. A new PR sounds good.

> Merely out of curiosity: why is the pub omitted in fn main() {}? I had noticed it before in other files, but now I'm even more curious about the underlying reason :) Thanks!

If another crate uses as a library the crate in which you define the `main`, they can import and call the main function (see [here](https://users.rust-lang.org/t/what-does-pub-main-do/56005)). However, that's rarely used in practice. I think it's more common to use separate `lib.rs` and`main.rs` files if a crate needs to compile both to a library and to a binary. I suggested to remove the `pub` from `main` just as a convention.

Pointerbender:
Thanks for explaining!

> I didn't know that it can be checked by rustfmt.

I assumed it would be possible, but there is [no rule](https://rust-lang.github.io/rustfmt/?version=v1.4.37&search=newline) where rustfmt lets us configure this. However, it seems that this behavior is already enforced by rustfmt [by default](https://github.com/rust-lang/rls/issues/282), it may be that it's just not running on the tests folder. I will dig a little deeper :+1: 

Thanks for the merge!

vakaras:
> it may be that it's just not running on the tests folder.

Yes, rustfmt is currently not running on the tests folder. We probably should change this. The crates/modules on which rustfmt is executed are specified here: https://github.com/viperproject/prusti-dev/blob/master/x.py#L23-L53

PR #415, opened 2021-02-18
Update dependencies

Aurel300:
- Bump Rust edition to 2018 in all crates
- Update to latest `warp`, `tokio`, `futures`, and `reqwest` in `prusti-server`

Locally, I get a strange linker error due to that last change. Let's see what CI says.

PR #724, opened 2021-10-21
Refactor VerificationService

fpoli:
This refactoring prepares `VerificationService` to be asynchronous (in a future PR that makes `VerificationService::verify` return a `Future`, upgrading to `reqwest >= 0.10`).

The Prusti server has been designed to handle concurrent verification requests, but so far we used it sequentially because there were concurrency issues (now fixed) in Viper.

@vakaras I changed the service to take a single program, not a vector. It makes it easier to limit the number of concurrent verification requests.

The diff is negative :tada:

### DIFF
--- a/prusti-common/src/verification_service.rs
+++ b/prusti-common/src/verification_service.rs
@@ -5,13 +5,12 @@ use viper::{self, VerificationBackend};
 use crate::vir::Program;
 
 pub trait VerificationService {
-    fn verify(&self, request: VerificationRequest) -> viper::ProgramVerificationResult;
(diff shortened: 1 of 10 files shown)


vakaras (approved):
LGTM

Aurel300:
Are the changes to `prusti-server` introduced in https://github.com/viperproject/prusti-dev/pull/415 necessary anymore? Or did you only update a subset of the outdated dependencies?

fpoli:
I didn't update the dependencies. I think we still need #415 in the end.

fpoli:
Silicon doesn't seem to be thread-safe yet: https://github.com/viperproject/silicon/issues/578

PR #726, opened 2021-10-23
Enable rustfmt on prusti-tests and fixed formatting in tests

Pointerbender:
In #720 it was discovered that rustfmt is not enabled on the `prusti-tests` folder. This PR fixes that and also fixes the formatting for the tests contained within (only 2 test files were affected). Running `x.py fmt-all` will successfully ensure newlines at the end of all formatted test files.

### DIFF
--- a/prusti-tests/.rustfmt.toml
+++ b/prusti-tests/.rustfmt.toml
@@ -1,0 +1,5 @@
+# Allow longer line widths, so that error messages don't get cut off, because this could break a test case.
+# Note: rustfmt is very picky about trailing comments and might move a comment to the next line anyway :'-(
+max_width = 1000
+wrap_comments = false
+comment_width = 1000
--- a/prusti-tests/.rustfmt.toml
+++ b/prusti-tests/.rustfmt.toml
@@ -1,5 +1,0 @@
-# Allow longer line widths, so that error messages don't get cut off, because this could break a test case.
-# Note: rustfmt is very picky about trailing comments and might move a comment to the next line anyway :'-(
-max_width = 1000
-wrap_comments = false
-comment_width = 1000
--- a/prusti-tests/tests/cargotest.rs
+++ b/prusti-tests/tests/cargotest.rs
@@ -5,8 +5,10 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use cargo_test_support::{cargo_test, project, symlink_supported};
-use std::path::{Path, PathBuf};
-use std::fs;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 fn cargo_prusti_path() -> PathBuf {
     let target_directory = if cfg!(debug_assertions) {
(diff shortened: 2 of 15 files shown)


vakaras:
> only 2 test files were affected

This sounds super suspicious. Did you check whether rustfmt was actually executed on all Rust files in the `prusti-tests` folder and its subfolders and not just on Rust files that are directly in the `prusti-tests` folder?

Pointerbender:
Good call, it turns out that the files are not formatted if these are not reachable through the Rust module path. I found a way to make it run on the tests, but `rustfmt` is moving the `//~` comments around and this is breaking some integration test assertions. I could not find a way to prevent `rustfmt` from doing this. I tried the `max_width`, `wrap_comments` and `comment_width` settings, but these settings are not respected by `rustfmt` when the trailing comments appear at the opening of a block (e.g. a function: `fn a() { //~ ERROR` the comment is moved to the next line). Also this will probably mess up quite a few `ui` tests, because the line numbers will change. For now I will only enable it on the `verify_partial` folder first to see if the new approach is making sense.

Pointerbender:
A little extra background information on what I changed and investigated:

`cargo fmt` is now running on the `prusti-tests` crate, but because of how the crate module resolution works, this only triggers on 2 files (`cargotest.rs` and `compiletest.rs`). `rustfmt` is now also running on `prusti-tests/tests/verify_partial/**/*.rs`. I added support for glob patterns in the `RUSTFMT_PATHS` array in `x.py` to make this work.

This yielded a lot more corrections (as expected) in just the `verify_partial` folder, which are also fixed in this PR. For the other test files, I think it may be better to do it in separate partial PRs, because there is quite a bit of technical debt hidden away in the formatting of all the test files.

The 2 main problems when reformatting all the tests are:

* If it's an `ui` test, then the corresponding existing `.stdout` and `.stderr` files might not match with the line numbers anymore.
* `rustfmt` will move around comments containing `//~ ERROR` and `//~^ ERROR` which will make the assertion point to the wrong line, this has to be fixed manually by adding an extra `^` after the `//~` for every line it is moved down by `rustfmt` (also see the [`rustc` guide](https://rustc-dev-guide.rust-lang.org/tests/adding.html#error-annotation-examples) for how this works).

Apparently this is also a hassle for the Rust integration tests, where the Rust team takes an approach to just [ignore the entire test folder](https://github.com/rust-lang/rust/blob/master/rustfmt.toml#L14-L16) when formatting. Based on my attempts to convince `rustfmt` not to change any line numbers (through the `.rustfmt.toml` settings), it looks like `rustfmt` currently does not support this properly. This would be in line with the `rustfmt` philosophy, which [discourages](https://users.rust-lang.org/t/rustfmt-and-alignment-of-trailing-comments/39121/2) trailing comments in general.

If the CI passes on all the changes and if this approach is acceptable, then this PR is ready for review & merge.

vakaras (approved):
LGTM

vakaras:
Thanks for improving this!

>     * If it's an `ui` test, then the corresponding existing `.stdout` and `.stderr` files might not match with the line numbers anymore.

I think that it probably makes sense to just ignore all UI tests. The reason is that in UI tests we often want to check that the parser is doing the correct thing and for that use strangely formatted code.

>     * `rustfmt` will move around comments containing `//~ ERROR` and `//~^ ERROR` which will make the assertion point to the wrong line, this has to be fixed manually by adding an extra `^` after the `//~` for every line it is moved down by `rustfmt` (also see the [`rustc` guide](https://rustc-dev-guide.rust-lang.org/tests/adding.html#error-annotation-examples) for how this works).

This feels like quite a bit of work if we cannot do this automatically.

PR #727, opened 2021-10-25
Add snapshot VIR

vakaras:


vir/defs/snapshot/ast/type_decl.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[display(fmt = "{} {{\n{}}}\n", name, "display::foreach!(\"{},\n\", fields)")]
+pub struct Product {
+    pub name: String,
+    /// Fields.
+    pub fields: Vec<FieldDecl>,
+}
+
+#[display(fmt = "{}", name)]
+pub struct Sum {
```
fpoli, line 34: What is a "Sum" snapshot? An enumeration or a union type?

vakaras: Yes.

PR #728, opened 2021-10-26
Refactor analysis

fpoli:
* Deduplicate the reaching definition analysis, which was implemented both in `analysis` and in `prusti-interface`. Excluding the new tests, overall this PR removes 180 LOC.
* Refactor `analysis` crate to allow each analysis to define its own constructor.
* Add a definitely initialized analysis that doesn't uninitialize places of Copy types when they are moved around. (I'll likely need this.)


fpoli:
@vakaras I removed the "root loan" column from the "dump_borrowck_info". Ping me if it's something that you would like to keep.

vakaras:
> @vakaras I removed the "root loan" column from the "dump_borrowck_info". Ping me if it's something that you would like to keep.

That should be fine with me.

PR #730, opened 2021-10-27
Handle DomainFuncApp in the fold-unfold algorithm

### LINKED ISSUE #364: Overlapping shared references in pure functions are not supported
Prusti generates incorrect encoding for the following program:

```rust
use prusti_contracts::*;

struct Child {}

struct Tree {
    child: Child,
}

#[pure]
fn use_both(_: &Tree, _: &Child) -> bool {true}

#[pure]
fn test(tree: &Tree) -> bool {
    use_both(tree, &tree.child)
}
```

The problem is that `test` is encoded as follows:
```javascript
function test(_1: Ref): Bool
  requires acc(Tree(_1), read$())
  requires true
  ensures true
{
  (unfolding acc(Tree(_1), read$()) in use_both(_1, _1.child))
}
```
which fails because we cannot have permissions to both `_1` and `_1.child` at the same time.

cc @tillarnold Purification should fix this bug.

fpoli:
The impure encoding of pure functions currently generates domain function applications that need unfolding expressions in the arguments. The `DomainFuncApp` case was not implemented in the fold-unfold algorithm, and the algorithm was not recursing on the function arguments.

Closes #364.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-364-1.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-364-1.rs
@@ -1,0 +1,17 @@
+use prusti_contracts::*;
+
+struct Child {}
+
+struct Tree {
+    child: Child,
+}
+
+#[pure]
+fn use_both(_: &Tree, _: &Child) -> bool {true}
+
+#[pure]
+fn test(tree: &Tree) -> bool {
+    use_both(tree, &tree.child)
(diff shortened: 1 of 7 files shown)


fpoli:
There is a lot of duplication between the `vir::Expr::DomainFuncApp` and `vir::Expr::FuncApp` cases. I think we already considered merging them.

Issue #737, opened 2021-10-29
Failing magic wand generation


fpoli:
Prusti fails to generate the magic wand at the end of the `next` method:
```rust
use prusti_contracts::*;

struct Node {
    next: Option<Box<Node>>,
}

impl Node {
    fn next(&mut self) -> &mut Node {
        match self.next {
            None => { unreachable!() },
            Some(ref mut node) => node,
        }
    }
}

#[trusted]
fn main() {}
```

fpoli:
The cause might be that `old[l3](_8.val_ref)` is not recognized to be a suffix of `old[l3](_8.val_ref.val_ref)`.

fpoli:
Similar program:
```rust
use prusti_contracts::*;

struct Node {
    elem: i32,
    next: Option<Box<Node>>,
}

pub struct Stack {
    head: Option<Box<Node>>,
}

impl Stack {
    #[pure]
    #[trusted]
    fn len(&self) -> i32 {
        unimplemented!()
    }

    #[requires(0 < self.len())]
    fn top_mut(&mut self) -> &mut i32 {
        match self.head {
            None => { unreachable!() },
            Some(ref mut node) => &mut node.elem,
        }
    }
}

#[trusted]
fn main() {}
```

fpoli:
PR #739 fixes the former program, but not the latter.

fpoli:
The issue is in the separation of duties between the fold-unfold algorithm and the code that handles the expiration of borrows. Currently, the fold-unfold algorithm considers expressions like `f.g` and `old[..](f.g)` to be always different, thus the algorithm expects to see a ghost statement that explicitly moves the permissions between `f.g` and `old[..](f.g)` when it's necessary to do so. The code that handles the expiration of references does that: when the two loans

```
let f = &mut y;
label L1
let x = &mut f.g;
```

expire, it generates VIR statements similar to

```
transfer P(x) --> Q(old[L1](f.g));
transfer Q(old[L1](f)) --> R(old[..](y));
```

[...]

PR #739, opened 2021-11-01
Fix TransferPerm between old(..) expressions

fpoli:
This PR fixes one of the programs in issue #737. This is a short-term patch. The proper fix would be to make aliasing a first-class concept in the fold-unfold algorithm (e.g tracking sets of surely-aliasing expressions), such that `TransferPerm` statements (and the reborrowing DAG) are no longer needed. However, I need to find time for that.


fpoli:
The `old_has_prefix`, `old_has_proper_prefix` and `old_replace_place` functions that I added could become `Expr` methods, but (1) I'm not aware of any other code that might use them and (2) I'm actually unsure how to add code to the VIR 2.0 definitions, given the on-going refactoring.

PR #741, opened 2021-11-02
Handle failures of cargo-outdated

fpoli:


.github/workflows/update-deps.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           UPDATE_NUMBER=$(( 10#$(date +%m) * 2 + (10#$(date +%d) / 15) - 2 ))
           DEVELOPER="${DEVELOPERS[ $UPDATE_NUMBER % ${#DEVELOPERS[@]} ]}"
           echo "The assigned developer is $DEVELOPER"
-          OUTDATED_DEPENDENCIES="$(cargo outdated --root-deps-only --workspace)"
+          OUTDATED_DEPENDENCIES="$(cargo outdated --root-deps-only --workspace 2>&1 || true)"
```
Aurel300, line 31: Maybe instead of `true` we should `echo "cargo outdated failed to execute"`?

fpoli: Currently stderr is informative enough (https://github.com/viperproject/prusti-dev/pull/742) but yes, an explicit message would be better.

Issue #744, opened 2021-11-03
Memory leak in Silicon


fpoli:
Calling `Silicon::verify` multiple times on the same verifier instance leaks memory (https://github.com/viperproject/silicon/issues/579). Until that's fixed, the `prusti-server` should create a verifier instance for each verification request.

PR #745, opened 2021-11-03
Avoid reusing verifiers (workaround #744)

fpoli:
Thanks to @Pointerbender for debugging this in https://github.com/viperproject/prusti-dev/pull/603!

### DIFF
--- a/prusti-common/src/verification_context.rs
+++ b/prusti-common/src/verification_context.rs
@@ -54,7 +54,7 @@ impl<'v> VerificationContext<'v> {
     pub fn new_viper_verifier(
         &self,
         backend_config: &ViperBackendConfig,
-    ) -> viper::Verifier<viper::state::Started> {
+    ) -> viper::Verifier {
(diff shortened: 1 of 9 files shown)


JonasAlaif:
So if I understand correctly, most of these changes should be reverted once #744 is fixed. Is it a good idea to create a draft PR doing this, while you still remember which changes to undo @fpoli?

fpoli:
> So if I understand correctly, most of these changes should be reverted once #744 is fixed. Is it a good idea to create a draft PR doing this, while you still remember which changes to undo @fpoli?

Not all the changes should be reverted; just a few. However, I believe that https://github.com/viperproject/silicon/issues/579 will not be fixed anytime soon. It's not an easy bug to debug, fix nor test, and I think we were the only Viper frontend that was reusing verifier instances. So, I think it's better to assume that the correct way of interacting with Viper is to create a new verifier each time. The overhead of doing so seems to be 0.044s on my laptop. A revert-changes draft PR would quickly become outdated.

PR #746, opened 2021-11-03
JNI local frame fixes for `viper/src/verifier.rs`

Pointerbender:
I added some additional local frames around the JNI interaction in `viper/src/verifier.rs`.

I guess the most significant one is the popping of the local frame that holds `self.verifier_instance`. This is a `JObject` (which implements the `Copy` trait). Due to the type state pattern used in `Verifier`, it is a little hard to place the local frame boundaries, so I had to resort to `push_local_frame` and `pop_local_frame` statements. This works, but looks prone to error from a code maintenance perspective. I'm not sure how to make this look nicer, without removing the type state pattern altogether, but that seemed a bit out-of-scope for this PR.

### DIFF
--- a/viper/src/verifier.rs
+++ b/viper/src/verifier.rs
@@ -37,28 +37,41 @@ impl<'a, VerifierState> Verifier<'a, VerifierState> {
         report_path: Option<PathBuf>,
     ) -> Verifier<'a, state::Uninitialized> {
         let jni = JniUtils::new(env);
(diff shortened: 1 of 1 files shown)


fpoli:
Oops, this has been closed automatically when merging the other PR.

Pointerbender:
No worries, I will make a new one to master instead :)

PR #762, opened 2021-11-14
update `prusti-contracts` not to depend on `std`, `prusti-specs` and `quote`

Pointerbender:
Currently `prusti-contracts` depends on the Rust standard library, and the crates `prusti-specs` and `quote`. This slows down cargo build times and also prevents building crates for `#![no_std]` targets. See also this [Zulip thread](https://prusti.zulipchat.com/#narrow/stream/236614-general/topic/Remove.20dependencies.20from.20.60prusti-contracts-impl.60). This PR fixes that.

When removing the standard library from the `prusti-contracts` crate, some other test binaries under `prusti-tests` that were annotated with `#![no_std]` started to fail. This is because Rust does not yet fully support Rust `#![no_std]` binaries, and `prusti-contracts` was giving the binaries the missing symbols needed to compile okay. Now with the dependency on the standard library removed, that made those tests fail. I solved this by removing the `#![no_std]` annotation for these test binaries (and by refactoring the test case where the intention was to test methods from the `core` crate explicitly).

I added a new regression test under `prusti-tests/tests/parse/ui/nostd.rs` that will detect if the dependency on the standard library is ever reintroduced).


Pointerbender:
There are 2 `cargo_verify` test cases failing atm: `test_failing_crate` and `test_prusti_toml`. I'm not sure how to best fix these. What is the intended purpose of the `cargo_verify` test suite? Does it test with Prusti enabled or without Prusti?

fpoli:
> What is the intended purpose of the `cargo_verify` test suite? Does it test with Prusti enabled or without Prusti?

Both of them. `cargo_verify` is meant to test `cargo-prusti` and the `prusti_contracts` crate. It checks on a set of crates that (1) `cargo build` succeeds and (2) `cargo prusti` produces the expected output.

(We should document this in prusti-tests' readme.)

fpoli:
> This is because Rust does not yet fully support Rust #![no_std] binaries

That's useful to know. Do you have a reference to a `rustc` issue or something similar?

prusti-launch/src/bin/cargo-prusti.rs
```diff
@@ -35,8 +35,22 @@ where
     let prusti_log_dir =
         std::env::var("PRUSTI_LOG_DIR").unwrap_or_else(|_| format!("{}/log", cargo_target));
 
+    // If the target crate depends on the `prusti-contracts` crate, then enable its `prusti` feature.
```
fpoli: Why should the `prusti` feature enabled (only) in that case?

fpoli: When running `cargo-prusti` or `prusti-rustc` the feature is currently always enabled:

https://github.com/viperproject/prusti-dev/blob/3fbee9b0be249c051fa9f0b9ca8ff3be7327a324/prusti-launch/src/bin/prusti-rustc.rs#L122

fpoli:
From a quick look it seems the `cargo_verify` tests fails because their output (expected to match `output.stderr` or `output.stdout`) changed. In one case a new macro appeared in the debug output. If you added it, it seems fine to me to update the expected output. In the other case `cargo prusti` seems to compile dependencies that have not already been compiled by the previous `cargo build` run. If this behavior is correct, we could update the expected output. Since the compilation order of the dependencies is not deterministic, we should use some [special pattern](https://github.com/rust-lang/cargo/blob/3a3a071cc85f0858f2014985e3c3529ae8671297/crates/cargo-test-support/src/compare.rs#L3-L18) from [`cargo-test-support`](https://github.com/rust-lang/cargo/tree/master/crates/cargo-test-support), which we use to compare the actual with the expected output.

fpoli:
To fix the failing `test_failing_crate`:
- Add `.arg("--quiet")` to the builder https://github.com/viperproject/prusti-dev/blob/c804a3a561b66c57ea895476771b4c7a026dc5ab/prusti-tests/tests/cargotest.rs#L163
- Update all expected outputs by removing any "Downloaded/Compiling" line. Our tests shouldn't depend on them.

Regarding the failing `test_prusti_toml`, it's a mistery to me why the `force_matches_macro` (including comments) gets printed as part of the desugaring. (Any clue @Aurel300?)

This PR could also add a new `cargo_verify/test_no_std` library crate to avoid future regressions. It's what the `nostd.rs` test was trying to test.

prusti-tests/tests/verify/pass/equality/nostd.rs
```diff
@@ -2,8 +2,6 @@
 //! [Prusti: invalid specification] use of impure function "core::cmp::PartialEq::eq" in pure code is not allowed
 //! [Prusti: invalid specification] use of impure function "core::cmp::PartialEq::ne" in pure code is not allowed
 
-#![no_std]
```
fpoli, line 5: What about renaming this file to `core_eq.rs`, since we can't test `#![no_std]`?

prusti-launch/src/bin/cargo-prusti.rs
Pointerbender: So this is a tricky one :smiley: Indeed `prusti-rustc` always enables the `prusti` feature. However, `prusti-rustc` is called with certain linking arguments for external crates. Since `prusti-contracts-internal` is now an optional dependency (which in turn explicitly depends on the `prusti-specs` crate and implicitly on the `prusti` crate), the compiler will fail due to error `error[E0433]: failed to resolve: use of undeclared crate or module 'prusti'` if `prusti-rustc` sees the `prusti` feature, because cargo does not add a link flag to this crate when invoking `prusti-rustc`.

Thus, the code here is to tell cargo to properly link the external crates on which `prusti-rustc` depends when the `prusti` feature is enabled.

fpoli: Oh, I see. Good catch! What about always enabling `--features=prusti-contracts/prusti`? Have you found issues with that? I'm trying to find an alternative that doesn't try to parse the `Cargo.toml`, because of whitespaces issues like with `prusti-contracts<space><space>=`.

Pointerbender:
> > This is because Rust does not yet fully support Rust #![no_std] binaries
> 
> That's useful to know. Do you have a reference to a `rustc` issue or something similar?

I found this an enlightening read: https://os.phil-opp.com/freestanding-rust-binary/ The good stuff is at the end "_Note that this is just a minimal example of a freestanding Rust binary. This binary expects various things, for example that a stack is initialized when the \_start function is called. So for any real use of such a binary, more steps are required._" The [rust book](https://doc.rust-lang.org/1.7.0/book/no-stdlib.html) only says that `#![no_std]` for binaries only works on nightly (but I think they meant to say here only for particular targets, not all targets can support this on nightly yet I believe).

> > Both of them. cargo_verify is meant to test cargo-prusti and the prusti_contracts crate. It checks on a set of crates that (1) cargo build succeeds and (2) cargo prusti produces the expected output.
> 
> (We should document this in prusti-tests' readme.)

I will add it to the readme as part of this PR :+1: 

> Regarding the failing test_prusti_toml, it's a mistery to me why the force_matches_macro (including comments) gets printed as part of the desugaring.

I'm also not sure about this.

> This PR could also add a new cargo_verify/test_no_std library crate to avoid future regressions. It's what the nostd.rs test was trying to test.

Great idea, I will add it.

Thanks!

prusti-launch/src/bin/cargo-prusti.rs
Pointerbender: > What about always enabling --features=prusti-contracts/prusti? Have you found issues with that?

I've encountered two issues with this:

1) Cargo build times are significantly longer, not just the building of the source code, also `cargo doc`, `cargo clean` etc.
2) When turning on the `prusti-contracts/prusti` feature, this creates a dependency on the Rust standard library. This breaks compilation of `#![no_std]` crates when building for targets that don't support the Rust standard library.

> I'm trying to find an alternative that doesn't try to parse the Cargo.toml, because of whitespaces issues like with `prusti-contracts<space><space>=`.

This one I could work-around in the code, but I realized there are more pitfalls with my approach :smile: It's possible to specify [target specific dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies) in `Cargo.toml`, for example `[target.'cfg(windows)'.dependencies]`. The current implemented check would evaluate to true even if just 1 target depends on `prusti-contracts` but not the others. If the non-existing feature flag is enabled for targets which don't depend on `prusti-contracts`, this will result in an error.

Another corner case which seems relevant for issue #2, is that the current approach will not enable the prusti contracts for dependencies that are not a direct dependency specified in `Cargo.toml` (i.e. dependencies of dependencies).

Is there a way to make `prusti-rustc` always include the correct linking arguments, even if the cargo feature is not explicitly turned on? If that is possible, that might be a better solution.

Pointerbender:
> Regarding the failing test_prusti_toml, it's a mistery to me why the force_matches_macro (including comments) gets printed as part of the desugaring.

I tried to take a deeper look, but I am still puzzled by this. If I build the crate from the command line (by temporarily updating the path to the `prusti-contracts` dependency, i.e. `prusti-contracts = { path = "../../../../prusti-contracts", features = ["prusti"] }`), then the `force_matches` macro (from the `prusti-utils` crate) does not appear in the standard output. Neither in debug and release modes, in case it's somehow caused by some weird crate-inlining optimization that cargo does in the background during release builds.

`prusti_toml` is the only crate that has a `.stdout` file btw, if we can't crack this one, one option to consider could be to remove the `.stdout` file but keep the `.stderr` file (although not ideal, I admit!)

Pointerbender:
Actually, I may have another work-around. If in these lines we replace

https://github.com/viperproject/prusti-dev/blob/95e8fa6e2d89f8b927c9b413eb8d285873651146/prusti-tests/tests/cargotest.rs#L166-L168

with

```rust
    if let Some(ref expected_stdout) = opt_expected_stdout {
        test_builder.with_stdout_contains(expected_stdout);
    }
```

then the test passes okay, while still partially verifying the standard output, at least.

Pointerbender:
The CI tests are passing with the latest commit, but there are still 2 unresolved items for this PR:

1. add a `cargo_verify/test_no_std` library crate which should verify that `prusti-contracts` does not include the Rust standard library when the `prusti` feature is disabled.
2. improve the linking strategy for the `prusti` crates through `cargo-prusti` and/or `prusti-rustc`. No "best" approach has been chosen on this topic yet. Current ideas are modifying the `cargo-prusti` binary to scan for the `prusti-contracts` library (has [some known issues](https://github.com/viperproject/prusti-dev/pull/762#discussion_r749114396)) or update `prusti-rustc` to always link the required `prusti` crates, even if `cargo` does not have the `prusti` feature explicitly enabled. I'm not sure how this would look like in practice, probably something like passing several `-L` flags to `rustc` from `prusti-rustc`, but this requires linking to `.rlib` files if I'm not mistaken, which have non-deterministic file names due to the build hash being part of their filenames. Any suggestions are welcome for this part :)

prusti-tests/tests/cargotest.rs
```diff
@@ -161,10 +161,11 @@ fn test_local_project<T: Into<PathBuf>>(project_name: T) {
 
     // Set the expected exit status, stdout and stderr
     let mut test_builder = project.process(cargo_prusti_path());
+    test_builder.arg("--quiet");
```
fpoli, line 164:
```suggestion
    let mut test_builder = project.process(cargo_prusti_path()).arg("--quiet");
```
(If it compiles)

prusti-tests/tests/cargotest.rs
```diff
@@ -161,10 +161,11 @@ fn test_local_project<T: Into<PathBuf>>(project_name: T) {
...
     let opt_expected_stdout = fs::read_to_string(project_path.join("output.stdout")).ok();
     let opt_expected_stderr = fs::read_to_string(project_path.join("output.stderr")).ok();
     if let Some(ref expected_stdout) = opt_expected_stdout {
-        test_builder.with_stdout(expected_stdout);
+        test_builder.with_stdout_contains(expected_stdout);
```
fpoli, line 170:
```suggestion
        // In some cases, Prusti outputs more macro definitions than needed.
        // See: https://github.com/viperproject/prusti-dev/pull/762
        test_builder.with_stdout_contains(expected_stdout);
```

prusti-tests/tests/cargotest.rs
Pointerbender: This was the first thing I tried, but unfortunately that didn't compile :) The issue is that `.arg()` returns a `&mut Execs`, which is a borrow of temporary variable `project.process(cargo_prusti_path())`, which then doesn't live "long enough".

prusti-launch/src/bin/cargo-prusti.rs
fpoli: > Cargo build times are significantly longer, not just the building of the source code, also cargo doc, cargo clean etc.

Do you know if it's just because there are more dependencies to compile, or because the build cache gets invalidated due to the new flag? ([`cargo build -Ztimings`](https://internals.rust-lang.org/t/exploring-crate-graph-build-times-with-cargo-build-ztimings/10975/22) might be useful to debug build times.)

> When turning on the `prusti-contracts/prusti` feature, this creates a dependency on the Rust standard library. This breaks compilation of `#![no_std]` crates when building for targets that don't support the Rust standard library.

Removing any dependency on the standard library from `prusti_contracts` looks difficult. We could ask the developers to add a target that supports the standard library just for the Prusti verification. Would that be too restrictive?

> Is there a way to make prusti-rustc always include the correct linking arguments, even if the cargo feature is not explicitly turned on? If that is possible, that might be a better solution.

That looks difficult because of the random-looking filenames.

Pointerbender: > Do you know if it's just because there are more dependencies to compile, or because the build cache gets invalidated due to the new flag?

Without the `prusti` feature for `prustic-contracts`, it compiles 7 dependencies in total for a crate where `prusti-contracts` is the only dependency. When the `prusti` feature is enabled, this number goes up to 65 dependencies. 

> Removing any dependency on the standard library from prusti_contracts looks difficult. We could ask the developers to add a target that supports the standard library just for the Prusti verification. Would that be too restrictive?

Another alternative I see here, is that we always enable `prusti` feature by default for `prusti-contracts`, but it can be disabled in the host crate's `Cargo.toml` like so:

```toml
[dependencies]
prusti-contracts = { path = "path/to/prusti-dev/prusti-contracts", default-features = false }
```

and then only explicitly enable it when needed, e.g.:

```shell
cargo-prusti --features prusti-contracts/prusti
```

This still gives the performance benefits when not building with prusti (e.g. when building a crate obtained from crates.io) and avoids pulling in the Rust standard library, but requires slightly more management on the end-user side in order to obtain these benefits.

This doesn't fully address cross-crate verification using Prusti (also see #2), but this can be worked-around by the host crate's `Cargo.toml`:

```toml
[depedencies]
prusti-contracts = { path = "path/to/prusti-dev/prusti-contracts", default-features = false }

[features]
default = []
prusti = ["prusti-contracts/prusti"]
```

and then another crate depending on this crate, can enable its `prusti` feature for cross-crate verification (this again needs a little extra management by the crate maintainer and would probably need to be documented somewhere).

I was hoping for a zero-configuration solution that always works(tm), although that seems non-trivial to achieve :) This also seems particularly tricky to implement in just the `cargo_verify/test_no_std` library crate, because the `prusti` feature should be off when running `cargo build`, but it should be on when running `cargo-prusti` and I'm not sure about what is the cleanest way to implement this. Ideally without changes to `cargotest.rs`, but I don't see another way currently.

> That looks difficult because of the random-looking filenames.

One way to do this might be to do a glob pattern search in the prusti target directory for `prusti-contracts*.rlib` and sort the artifacts by last modification time. And then only link the last touched library file from `prusti-rustc`, because this should be the most recently built library file. Likewise for other libraries that might need to be linked. Would this work for all intended usages of Prusti? Or couldn't we assume that these library files are present on the file systems in all thinkable use cases of Prusti (now or in the future)? Also there might be exotic systems out there that do not set the last modification time of a file (e.g. when the [`S_NOCMTIME`](https://serverfault.com/a/906970) kernel flag is set, see [here](https://www.kernel.org/doc/html/v4.14/filesystems/index.html#c.file_update_time)). Although, these are most likely optimized production systems and not build servers, so I'm not sure if this will be an issue in practice (quite a footgun still for someone that wants to verify their crate using Prusti on such a server :P ).

Pointerbender:
Just pushed one more commit, where I removed the `proc-macro2` dependency from the `prusti-contracts-impl` crate. Instead, it suffices to only depend on the `proc-macro` crate, which is shipped with the Rust compiler. This removed an additional 4 dependencies from the `prusti-contracts-impl` crate :smile:

prusti-launch/src/bin/cargo-prusti.rs
fpoli: I would avoid trying the `prusti-contracts*.rlib` glob pattern search or checking the modification times. So far we don't depend on those cargo internals.

> Without the `prusti` feature for `prustic-contracts`, it compiles 7 dependencies in total for a crate where `prusti-contracts` is the only dependency. When the `prusti` feature is enabled, this number goes up to 65 dependencies.

There is still something that I don't understand. Why having `enable_prusti_contracts` always `true` in `cargo-prusti` would cause more dependencies to be compiled? If Prusti is running and if a crate depend on `prusti_contracts`, I expect that those extra dependencies that gets compiled are necessary to run `prusti_contracts_internal`. I don't see how we can avoid them. If Prusti is not running or if the crate does not depend on `prusti_contracts`, then I expect that the dependencies of `prusti_contracts_internal` shouldn't be compiled anyway.

Pointerbender: > If Prusti is running and if a crate depend on prusti_contracts, I expect that those extra dependencies that gets compiled are necessary to run prusti_contracts_internal. I don't see how we can avoid them.

Agreed :)

> If Prusti is not running or if the crate does not depend on prusti_contracts, then I expect that the dependencies of prusti_contracts_internal don't get compiled anyway.

This is indeed the effect I'm trying to achieve. If the `prusti` feature is enabled for the `prusti-contracts` dependency, then all 65 dependencies will be compiled regardless if the build is invoked by `cargo build` (the non-Prusti version) or `cargo-prusti`. It also affects the evaluation of the macros in `prusti-contracts`. Unfortunately, the vanilla `cargo` is not able to automatically detect that compiling the additional dependencies or evaluating the macros from `prusti-contracts-internal` are not needed, when the `prusti` feature is enabled by default in `prusti-contracts`. One work-around would be as suggested in my previous comment by specifying `default-features =  false` for the `prusti-contracts` dependency in the crate of a Prusti end-user. However, this puts an additional burden on the end-user to get this right. So from the perspective of a better end-user experience, I'm still searching for a "nicer" way (although, if we at some point indeed exhaust all options, it is a viable alternative).

I have one more idea I want to try today. It's possible to produce different kinds of libraries through setting the [`crate-type` ](https://doc.rust-lang.org/reference/linkage.html) in `Cargo.toml`. I'm hoping that one of these options will produce a linkable artifact which has a deterministic file name, in order to avoid extra trickery in the `cargo-prusti` internals. This artifact will then be linked through `prusti-rustc` and we can leave `cargo-prusti` untouched if I can get it to work. If this works, then we can leave the `prusti` feature disabled by default in `Cargo.toml` for `prusti-contracts`, but then it will still properly compile `prusti-contracts-internal` when passing through `prusti-rustc`. Hopefully I can show more later today :)

fpoli: > This is indeed the effect I'm trying to achieve.

Thanks! Sorry for asking trivial questions; I'm trying to evaluate all options without actually trying them because I don't have much spare time :)

> If the prusti feature is enabled for the prusti-contracts dependency, then all 65 dependencies will be compiled regardless if the build is invoked by cargo build (the non-Prusti version) or cargo-prusti

From the cargo documentation "by default, all features are disabled unless explicitly enabled." So, how is the prusti feature enabled by default in `prusti-contracts/Cargo.toml`? I don't see any `default = ["prusti"]` in it.

Pointerbender: No worries! Happy to explain in further depth, because this also strengthens my own understanding of what is going on :)

> So, how is the prusti feature enabled by default in prusti-contracts/Cargo.toml?

Before this PR, `prusti-contracts-internal` was not an optional dependency, hence it was always included in the linked libraries, including its 65 dependencies:

https://github.com/viperproject/prusti-dev/blob/75eec658bfc8fe716d8f465d30bf4d6eb5b12012/prusti-contracts/Cargo.toml#L7-L9

In the current state of this PR, there is indeed no `default = ["prusti"]` in `prusti-contracts/Cargo.toml`. Instead, the enabling of the `prusti` feature is injected through `cargo-prusti`:

https://github.com/viperproject/prusti-dev/blob/dcb95e391ca0ea67513de9c68c5e58dcd0af4e54/prusti-launch/src/bin/cargo-prusti.rs#L38-L52

> What about always enabling `--features=prusti-contracts/prusti`? Have you found issues with that?

I just realized I may have misinterpreted this earlier question and did not give a proper full explanation :) One piece of information that was missing in my explanation is that this will break the build if the crate does not depend on `prusti-contracts` with `` error: Package `my_crate` does not have a dependency named `prusti-contracts` ``. This is quite often the case, for example when running Prusti on crates from crates.io as described in one of the Prusti papers. So the above linked code will do a check if the crate depends on `prusti-contracts`, and if so, then inject the `--features=prusti-contracts/prusti` snippet in the call to `cargo`.

However, we also established that this approach in `cargo-prusti` is not ideal for these reasons:

1. whitespaces issues like with `prusti-contracts<space><space>=`;
2. the `prusti-contracts` dependency may be conditional on the target platform;
3. it won't work when (in the future) we run Prusti cross-crate, where we also want to evaluate the specifications of dependencies (of dependencies, and so on).

The solution that I'm working towards, is that the `prusti` feature of `prusti-contracts` is never enabled in any `Cargo.toml` nor in `cargo-prusti`. Instead, the `--cfg=feature=\"prusti\"` flag is injected by `prusti-rustc` (this is already there today, like you linked earlier). This will correctly trigger the macro expansion from `prusti-contracts-internal`. But, because `prusti-contracts-internal` is an optional dependency, cargo does not link the `prusti-contracts-internal` crate into the `prusti-rustc` call. This creates another compilation error, because it can't find `prusti-contracts-internal`. To address this last problem, I will attempt to update `prusti-rustc` to always explicitly link `prusti-contracts` and `prusti-contracts-internal`, even if the crate has no dependency on `prusti-contracts`.

fpoli: > Instead, the enabling of the `prusti` feature is injected through `cargo-prusti`:

Thanks, I thought you meant that it's enabled by cargo, but it's cargo-prusti.

> will break the build if the crate does not depend on `prusti-contracts`

Aha! Yes, I was missing that. Now I fully understand the messages.

> To address this last problem, I will attempt to update `prusti-rustc` to always explicitly link `prusti-contracts` and `prusti-contracts-internal`, even if the crate has no dependency on `prusti-contracts`

Even if you manage to generate the correct `rustc` arguments to link `prusti-contracts` and `prusti-contracts-internal`, what guarantees that they have been compiled?

Pointerbender: > Even if you manage to generate the correct rustc arguments to link prusti-contracts and prusti-contracts-internal, what guarantees that they have been compiled?

It is currently an assumption that someone will have run the command `x.py build` first :) I believe it is currently already the case on the `master` branch that we have to run `x.py build` before being able to run `x.py test`. For example, this fails:

```shell
$ x.py clean
$ x.py setup
$ x.py test
--- SNIP ---
     Running tests/cargotest.rs (target/debug/deps/cargotest-c8ee7120bcf841fd)

running 6 tests
test simple_assert_false ... FAILED
test simple_assert_true ... FAILED
test test_failing_crate ... FAILED
test test_prusti_toml ... FAILED
test test_prusti_toml_fail ... FAILED
test test_symlinks ... ok
```

All the tests failed on the same error `thread 'main' panicked at 'failed to execute prusti-driver ("/path/to/prusti-dev/target/debug/prusti-driver")', prusti-launch/src/bin/prusti-rustc.rs:135:29`. If I run `x.py build` before `x.py test`, this error does not happen.

> Would this work for all intended usages of Prusti? Or couldn't we assume that these library files are present on the file systems in all thinkable use cases of Prusti (now or in the future)?

Although, come to think of it, this is not a hard enough guarantee when Prusti is shipped in binary form (e.g. for the Prusti Assistant or through [Github releases](https://github.com/viperproject/prusti-dev/releases)). Would it be possible to include the linked libraries as separate files with the binary distributions of Prusti? Are there any other corner cases where this might be a problem?

Pointerbender: Was just digging little deeper and found out that there is already a `libprusti_contracts.rlib` shipped with Prusti and used in:

https://github.com/viperproject/prusti-dev/blob/697ee88f78fc2b8de9ded763b362e9722836b0e8/prusti-launch/src/bin/prusti-rustc.rs#L111-L119

Right now it's just missing the `.rlib` for `prusti-contracts-internal` (a `proc-macro` crate). I'm going to investigate further if we can bring that in the same way as `prusti-rustc` already does for `prusti-contracts`.

fpoli: > I believe it is currently already the case on the `master` branch that we have to run `x.py build` before being able to run `x.py test`

To test Prusti yes, but I was thinking at use cases like VSCode.

> Would it be possible to include the linked libraries as separate files with the binary distributions of Prusti? Are there any other corner cases where this might be a problem?

Yes, good point! Including the libraries in the release is certainly possible. Actually, we already release some of them (I forgot about it), ~~but I think they are never used by Prusti~~. That way, the library needs to be chosen when we prepare the release and that makes the glob pattern approach more robust (e.g. we don't have to worry about weird file systems or cargo version; we control them).

The only question is what happens if the source code of the `prusti_contracts`/`prusti_contracts_impl` dependency used by the crate to be verified doesn't match the library in the release. Like, what if we add or remove a derive macro? How bad is the cargo error message (if any)? Ideally, Prusti should give a readable error message, but if the cargo error is not too bad we could postpone the readable error to a future PR.

Pointerbender: @vakaras I saw in the git history that you've worked on this section in the past, might you have any tips how to best explicitly link the `prusti-contracts-internal` proc-macro crate from within `prusti-rustc`? :) In this PR, I've made `prusti-contracts-internal` an optional dependency of `prusti-contracts`. `prusti-contracts-internal` is still linked correctly though:

```shell
$ rustc -Z ls target/debug/libprusti_contracts.rlib
Crate info:
name prusti_contracts-410d1d3be9361555
hash 910f03a0d2f44dca stable_crate_id StableCrateId(10652995453112551957)
proc_macro false
=External Dependencies=
1 core-0e3656b1fda5fd7b hash 90110cb6a5230b89 host_hash None kind Explicit
2 compiler_builtins-eecefd843a0dbc02 hash 7951e33a8234605c host_hash None kind Explicit
3 rustc_std_workspace_core-2a6a2797f7a73818 hash df1b6930d614cd5e host_hash None kind Explicit
4 prusti_contracts_internal-4e86e2685f664ef2 hash 214fce9f03f0ad0d host_hash None kind MacrosOnly
```

And when I run `prusti-rustc` I get an error: `` use of undeclared crate or module `prusti_contracts_internal` ``.

Things I've tried so far:

* Add a `crate-type = ["rlib"]` in `prusti-contracts-internal/Cargo.toml`, but then cargo complains about `` cannot mix `proc-macro` crate type with others ``.
* I considered adding an extra intermediary crate between `prusti-contracts` and `prusti-contracts-internal` to work-around the above error message, but I'm worried that this intermediary crate will then again depend on `prusti-contracts-internal` and get the same error. I still have to try this though.

Pointerbender: > Including the libraries in the release is certainly possible. Actually, we already release some of them (I forgot about it), but I think they are never used by Prusti.

Looks like one of those is used already. `prusti-rustc` uses `libprusti_contracts.rlib` currently.

> The only question is what happens if the source code of the `prusti_contracts`/`prusti_contracts_impl` dependency used by the crate to be verified doesn't match the library in the release. Like, what if we add or remove a derive macro? How bad is the cargo error message (if any)? Ideally, Prusti should give a readable error message, but if the cargo error is not too bad we could postpone the readable error to a future PR.

When removing a macro, we could replace the token stream with a `compile_error!("macro X has been removed/renamed in version Y of Prusti");` [statement](https://doc.rust-lang.org/std/macro.compile_error.html). I don't foresee problems with adding a new macro. If the end-user is on an older version of `prusti-contracts-impl`, that new macro won't exist yet and they won't be able to call it yet (barring any name collisions with other imports when new Prusti macros are introduced, but the Rust compiler will give a decent error message in that case).

In the case the versions mismatch, I don't expect any linking errors. Because the distributed `.rlib` file is always on the same version as the Prusti binaries.

Pointerbender: When doing clean builds of Prusti:

```
x.py clean
x.py build
x.py build --release
```

I see the following libraries for `prusti-contracts-internal`:

```
target/debug/deps/libprusti_contracts_internal-4e86e2685f664ef2.so
target/debug/libprusti_contracts_internal.d
target/debug/libprusti_contracts_internal.so
target/release/deps/libprusti_contracts_internal-a2ff594f86b4a9d7.so
target/release/deps/libprusti_contracts_internal-0f18580e1dd0de70.so
target/release/libprusti_contracts_internal.d
target/release/libprusti_contracts_internal.so
```

In the debug output, there is only 1 `.so` file in the `deps/` folder, but for the release build, it has 2 `.so` files in its `deps/` folder. Is there a specific reason for that? I would expect only 1 `.so` file in both builds, but I'm venturing into new territory here so I'm not sure if that's intentional or not :)

Pointerbender: Quick update: I managed to find a relatively simple fix for the linking of `prusti-contracts-internal`. I still have to fix a test case on Windows for `cargo_verify/test_no_std` and then I'll do a write-up of what I found out :) This might take me a couple of more commits before the Windows test is fixed, as I don't have access to a Windows machine atm.

Pointerbender:
I'm having a lot of difficulty creating a `#![no_std]` binary on Windows (on Linux and Mac OS it works fine). Would it be acceptable to ignore the `test_no_std` test on Windows? (If so, then I think this PR is ready for reviewing/merging.) Unless there would be Windows specific logic in the `prusti-contracts` crate, it should in theory be warning enough if either the Linux of Mac OS test fails the canary test of accidentally pulling in the standard library.

On the other platforms, the tests are passing okay. After a lot of research and learning, the fix for linking the `prusti-contracts` crate turned out to be relatively straightforward :) The key insights here were:

1. `prusti-rustc` was already passing the correct linking arguments to `prusti-contracts` and `prusti-contracts-internal` since before this PR.
2. However, it was appending these linking arguments instead of replacing existing linking arguments (if any). When a crate already depends on `prusti-contracts`, this led to `prusti-contracts` being linked twice. This sometimes creates ambiguity which could cause a compilation to fail (e.g. for some `cargo_verify` test cases). The fix implemented here is to replace the existing linking arguments to `prusti-contracts` if they exist, otherwise append the linking arguments to the distributed libraries.
3. Additionally, `prusti-contracts` should not be explicitly linked by `prusti-rustc` when it is passed the `prusti-contracts` dependency from `cargo-prusti` (this will happen for any crate that depends on `prusti-contracts`).
4. Last, the `prusti` feature should only be enabled by `prusti-rustc` when NOT compiling `prusti-contracts`. This works, because the distributed `libprusti-contracts.rlib` library is compiled by `cargo build` and not by `cargo-prusti`.

In the end I did not have to make any changes to `cargo-prusti`, only the above additional logic was added to `prusti-rustc`.

Pointerbender:
p.s. in case anyone wants to take a stab at getting `#![no_std]` Windows binaries to work with Prusti in the future, I ran into the problem that `prusti-driver`/`rustc` was unable to link symbols like `memcpy` and `memset` (and a bunch of others) and that `libc` is not supported on Windows in combination with `#![no_std]` (see [here](https://github.com/rust-lang/libc/issues/1203), [here](https://users.rust-lang.org/t/linker-error-when-building-no-std-windows-dll-that-uses-libc/23976/2) and [here](https://github.com/rust-lang/libc/blob/72cf1803b693f3e2787f2f8a547615110ba13cea/src/windows/mod.rs#L257-L261), this was never properly fixed until now). As a work-around, I tried to replace `libc` with [`winapi`](https://docs.rs/winapi) or even omit both libraries, but ran into similar problems there (all my failed attempts are in the git commit history of this PR :stuck_out_tongue:). The gist of it was that I just couldn't get it to properly link the C runtime library (referred to as "crt" in other places, or a variation of that), even when passing the [`-C target-feature=+crt-static`](https://rust-lang.github.io/rfcs/1721-crt-static.html) compiler flag in order to statically link the C runtime library into the binary. For now it should properly work on Windows, as long as your binary depends on the Rust standard library (the proof for this claim should be delivered once the CI finishes and none of the other Windows test cases are failing, because all of the current Windows tests depend on the Rust standard library).

prusti-tests/README.md
```diff
@@ -2,3 +2,13 @@ Prusti-tests
...
 
 Crate containing the end-to-end tests (input program and expected result) of Prusti.
+
+
+The tests are organized into several categories under the `tests/` folder:
+
+* `cargo_verify` is meant to test `cargo-prusti` and the `prusti-contracts` crate. It checks on a set of crates that (1) cargo build succeeds and (2) `cargo-prusti` produces the expected output.
+* `parse` tests the parsing of specifications. This doesn't run the verifier.
+* `typecheck` tests the type-checking of specifications. This doesn't run the verifier.
+* `verify` runs the verifier.
```
fpoli:
```suggestion
* `verify` runs the verifier with overflow checks disabled.
```

fpoli (approved):
Thanks! It looks good to me. Is it ready to be merged? Could you also open an issue for `no_std` not supported on Windows?

Pointerbender:
Thanks for the review! I committed the suggested changes and it's now ready to merge (when CI finishes).

> Could you also open an issue for no_std not supported on Windows?

This is more of an issue with Rust than with Prusti. Documentation on building `#![no_std]` binaries on Windows is scarce and I didn't have access to a Windows machine which makes it hard to get a proof-of-concept working. I do expect that if anyone can get a Rust binary to work with `#![no_std]` on Windows then the `prusti-conracts` library will also work okay, I just wasn't able to back this up with a test case, due that I couldn't get anything `#![no_std]` to compile on Windows :joy: 

There is a smoke test in this PR though that also passes on Windows: `prusti-tests/tests/parse/ui/nostd.rs`. This test expects a compile error due to the missing `std` library (confirming that `std` is not accidentally linked). However, this doesn't cover the case for compiling actual Windows `#![no_std]` binaries.

fpoli:
> This is more of an issue with Rust than with Prusti.

We can label that issue [`pending-on-external`](https://github.com/viperproject/prusti-dev/issues?q=is%3Aopen+is%3Aissue+label%3Apending-on-external).

Pointerbender:
Created #775

PR #767, opened 2021-11-19
Encode move of copy types as copy

### LINKED ISSUE #389: Incorrect loop invariant permissions due to structure initialized before the invariant
The following program causes an internal error ("generating fold-unfold Viper statements failed") caused by a limitation in the encoding of initialization of local variables.
```rust
fn test() {
    let mut k: usize = 0;
    while k - 123 > 0 { }
}

fn main() {}
```

The reason is that the `k - 123` in the loop guard is desugared to an overflow check, which stores the result in a tuple `_6` of type `(usize, bool)`. The second element of the tuple is moved out to check if the overflow happened before reaching the loop invariant. As a result, `_6` is a local variable that:
1. is initialized to (`_6 = CheckedSub(...)`) during the execution of the loop (i.e. loop body + loop guard evaluation), and
2. is partially uninitialized when reaching the loop body invariant (placed at the beginning of the loop, after the loop guard evaluation).
Currently, the algorithm that computes which permissions should be in the loop invariant uses just (1) to decide that there should be a full permission for `_6`, while only its first field is actually initialized.

CFG of the `test` method (click to open the image with a better resolution):
cfg

The loop head is block 0, the loop guard is checked in block 3 and the loop invariant is placed at the beginning of block 4, which is the beginning of the loop body. Cleanup and imaginary edges are ignored by Prusti.

fpoli:
This PR encodes the *moves* of Copy types as *copies*. As a consequence, copy types are still considered initialized even after a MIR move. This avoids an issue with the computation of the permissions of the loop invariant.

Fixes #389

### DIFF
--- a/analysis/src/domains/definitely_initialized/state.rs
+++ b/analysis/src/domains/definitely_initialized/state.rs
@@ -182,20 +182,13 @@ impl<'mir, 'tcx: 'mir> DefinitelyInitializedState<'mir, 'tcx> {
 
     /// If the operand is move, make the place uninitialised
     /// If the place is a Copy type, uninitialise the place iif `move_out_copy_types` is true.
-    fn apply_operand_effect(
-        &mut self,
-        operand: &mir::Operand<'tcx>,
-        location: mir::Location,
-        move_out_copy_types: bool,
-    ) {
+    fn apply_operand_effect(&mut self, operand: &mir::Operand<'tcx>, move_out_copy_types: bool) {
         if let mir::Operand::Move(place) = operand {
             let mut uninitialise = true;
             if !move_out_copy_types {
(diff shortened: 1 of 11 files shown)


fpoli:
`ty.is_copy_modulo_regions` is failing for mysterious reasons. I wrote a question at https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Using.20is_copy_modulo_regions

fpoli:
Encoding the move of `Copy` type as copy (i.e. leaving the source initialized) requires to remove copy types from the precondition of method calls, which would be a small but significant change. Fixing the computation of loop invariants seems a better way of fixing #389.

PR #774, opened 2021-11-22
add `no_verify_deps` configuration flag to disable verification for local dependencies

### LINKED ISSUE #713: Prusti attempts to verify local dependencies
I am attempting to use Prusti to verify a crate that has other local crates as dependencies. Currently when I run the verifier, it is attempting to also validate the dependencies as they compile. Is it possible to treat the local crates like external libraries, such that they are not validated, and I can write external specifications for them instead.

Pointerbender:
Adds a new `no_verify_deps` configuration flag to disable verification for local dependencies.
Includes a `cargo_verify/no_deps` regression test case to verify that the flag is correctly recognized.

Closes #713.

### DIFF
--- a/prusti/src/driver.rs
+++ b/prusti/src/driver.rs
@@ -133,13 +133,13 @@ fn main() {
     let prusti_be_rustc = config::be_rustc();
     // This environment variable will not be set when building dependencies.
     let is_primary_package = env::var("CARGO_PRIMARY_PACKAGE").is_ok();
-    let are_lints_disabled = arg_value(&original_rustc_args, "--cap-lints", |val| val == "allow")
-        .is_some()
-        || (!is_primary_package && config::no_verify_deps());
+    let is_no_verify_crate = !is_primary_package && config::no_verify_deps();
+    let are_lints_disabled =
+        arg_value(&original_rustc_args, "--cap-lints", |val| val == "allow").is_some();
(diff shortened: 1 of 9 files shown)


prusti/src/driver.rs
```diff
@@ -131,8 +131,11 @@ fn main() {
     // If the environment asks us to actually be rustc, or if lints have been disabled (which
     // indicates that an upstream dependency is being compiled), then run `rustc` instead of Prusti.
     let prusti_be_rustc = config::be_rustc();
-    let are_lints_disabled =
-        arg_value(&original_rustc_args, "--cap-lints", |val| val == "allow").is_some();
+    // This environment variable will not be set when building dependencies.
+    let is_primary_package = env::var("CARGO_PRIMARY_PACKAGE").is_ok();
+    let are_lints_disabled = arg_value(&original_rustc_args, "--cap-lints", |val| val == "allow")
+        .is_some()
+        || (!is_primary_package && config::no_verify_deps());
```
fpoli: Could you rewrite to `let is_no_verify_crate = !is_primary_package && config::no_verify_deps();` (or a local variable with a similar name) adding a disjunction to the `if ... { rustc_driver::main() }`? It should be easier to read.

Pointerbender:
Thanks for the review! The commit with the suggested changes has been pushed, ready to review/merge :+1:

fpoli:
Thanks!

Issue #770, opened 2021-11-21
ICE: `assertion failed: env.has_errors()`


Pointerbender:
The following rust program:

```rust
use prusti_contracts::*;

pub struct A {
    b: isize,
}

impl A {
    #[pure]
    #[ensures(result == true)]
    pub const fn invariant(&self) -> bool {
        self.is_negative() != self.is_non_negative()
    }

    #[pure]
    // #[ensures(result == (self.b >= 0))]
    pub const fn is_non_negative(&self) -> bool {
        self.b >= 0
    }

    #[pure]
    // #[ensures(result == (self.b < 0))]
    pub const fn is_negative(&self) -> bool {
        self.b < 0
    }

    #[requires(self.invariant())]
    pub const fn test(&self) {}
}

fn main() {}
```

fails with an internal compiler error and does not output any Prusti error messages. The compiler error message is `thread 'rustc' panicked at 'assertion failed: env.has_errors()', prusti/src/verifier.rs:73:17`. When uncommenting both `#[ensures(...)]` attributes, the program verifies okay and without errors.

When running Prusti with debug logging enabled, the verification errors are visible:

```
[2021-11-21T16:37:48Z DEBUG viper::verifier] Starting: Viper consistency checks
[2021-11-21T16:37:48Z DEBUG viper::verifier] Completed: Viper consistency checks (0.1 seconds)
[2021-11-21T16:37:48Z DEBUG viper::verifier] Starting: Viper verification
[2021-11-21T16:37:50Z DEBUG viper::verifier] Completed: Viper verification (1.81 seconds)
[2021-11-21T16:37:50Z DEBUG viper::verifier] Viper verification result: Verification failed with 1 errors:
      [postcondition.violated:assertion.false] Postcondition of m_invariant__$TY$__Snap$struct$m_A$$bool$ might not hold. Assertion result might not hold. (@16.5)
[2021-11-21T16:37:50Z INFO  prusti_common::stopwatch::log_level] [prusti-server] Completed: verification (1.884 seconds)
[2021-11-21T16:37:50Z DEBUG jni::wrapper::java_vm::vm] Detached thread rustc (ThreadId(2)). 0 threads remain attached
[2021-11-21T16:37:50Z INFO  prusti_common::stopwatch::log_level] [prusti-viper] Completed: verifying Viper program (17.786 seconds)
[...]

Issue #738, opened 2021-10-31
value might not fit into the target type (due to check_overflows)


Pointerbender:
```rust
use prusti_contracts::*;

#[derive(Clone, Copy)]
pub struct A {
    count: isize,
}

impl A {
    #[pure]
    #[ensures(result <= isize::MAX as usize)]
    pub const fn len(&self) -> usize {
        if self.count < 0 {
            ((1_isize + self.count) + isize::MAX) as usize
        } else {
            self.count as usize
            //~^ ERROR value might not fit into the target type.
        }
    }
}

#[pure]
#[requires(slice.len() > 0)]
#[requires(slice[0].len() == 0)]
pub fn test(slice: &[A]) -> bool {
    true
}

fn main() {}
```

The above program yields the error `[Prusti: verification error] value might not fit into the target type.` for the Viper program of `fn test` (the Viper program for `len` verifies okay as expected). There seems to be a problem in the pure encoding of the struct fields for `A` in the precondition `#[requires(slice[0].len() == 0)]` of `fn test`. It looks like `acc(isize(_1.f$count))` is not properly applied during the pure lookup of `slice[0]` in Viper `function m_len__$TY$__Snap$struct$m_A$$int$(_1: Snap$struct$m_A): Int`. Overflow checks should to be enabled to trigger this error.

Edit: the relevant generated Viper statements for the Viper program for `fn test` are:

```
domain MirrorDomain {
  
  function mirror_simple$m_len__$TY$__Snap$struct$m_A$$int$(_1: Snap$struct$m_A): Int
}

domain Snap$struct$m_A {
  
  function Snap$struct$m_A$0$field$f$count__$TY$__Snap$struct$m_A$$int$(self: Snap$struct$m_A): Int
}

function m_len__$TY$__Snap$struct$m_A$$int$(_1: Snap$struct$m_A): Int
  requires true
  requires true
  ensures result <= builtin$cast$isize$usize__$TY$__$int$$$int$(9223372036854775807)
  ensures 0 <= result
  ensures result <= 18446744073709551615
  ensures [result == mirror_simple$m_len__$TY$__Snap$struct$m_A$$int$(_1), true]
{
[...]

PR #782, opened 2021-11-28
Issue 738

Pointerbender:
While familiarizing myself a bit with the various VIR encodings in the Prusti processing stage, I started playing to see if I could make a temporary patch for #738 in await of the refactoring of `vir_high` landing on `master`. During this investigation I unearthed two problems:

1. The required type bounds for Rust primitives are deleted from the Viper program here on line 242:

https://github.com/viperproject/prusti-dev/blob/1afad4718f44f4f451a4113d7323d70bb97bbfd6/prusti-viper/src/encoder/definition_collector.rs#L227-L251

This seems to be too rigorous, because Prusti does generate `axiom snap$valid {}` Viper statements which axiomatize the type bounds for Rust primitives. However, because the axioms are cleared, these never make it into the Viper program.

2. The type bounds for arrays and slices of Rust primitives are currently not generated. There is an exception in the above linked code that allows all axioms for arrays and slices to remain in the Viper program, but because these are absent from the domain axioms, these are also absent in the resulting Viper program.

In order to address both issues, I made code changes that add `snap$valid` axioms for arrays and slices of Rust primitives in the Viper program. Also, I updated to logic to selectively retain the `snap$valid` axioms for non-arrays/slices, but only if any actually used functions depend on the axiom. It turned out that always keeping any `snap$valid` axiom is too liberal, because this adds unnecessary statements, but can also make the Viper program invalid if an axiom refers to one of the unused domain functions.

I added a passing regression test for #738 and another for one of the two cases in #770.


vakaras (approved):
The changes look great, thanks!

vakaras:
By the way, this part of the code is going to change significantly during the refactoring, so having tests is very useful.

Pointerbender:
Thank you! I'll see if I can come up with some additional tests in a separate PR to aid the refactoring :)

PR #784, opened 2021-11-30
Fix for definition collector

### LINKED ISSUE #783: Inconsistency when generating `function builtin$cast$isize$usize__$TY$__$int$$$int$(number: Int): Int`
I've got another fun bug which left me scratching my head for a while :) This one is related to casts from `isize` to `usize`. Typically, these casts appear in a Viper program (when overflow checks are enabled) as:

```
function builtin$cast$isize$usize__$TY$__$int$$$int$(number: Int): Int
  requires -9223372036854775808 <= number
  requires number <= 9223372036854775807
  requires 0 <= number
  requires number <= 18446744073709551615
  ensures 0 <= result
  ensures result <= 18446744073709551615
{
  number
}
```

I've got a Rust project which is relatively large in comparison to the tests in `prusti-tests`, and when I run `cargo-prusti` on it, it generates a Viper program which shows the cast as an abstract function without a function body:

```
function builtin$cast$isize$usize__$TY$__$int$$$int$(number: Int): Int
  requires -9223372036854775808 <= number
  requires number <= 9223372036854775807
  requires 0 <= number
  requires number <= 18446744073709551615
  ensures 0 <= result
  ensures result <= 18446744073709551615
```

So far I have not been able to reproduce this with a minimal example, I can only reproduce it with the larger code base. In the latter case, this causes the verification to fail with error `Postcondition of m_len__$TY$__Snap$struct$m_PageIndex$$int$ might not hold. Assertion Snap$struct$m_PageIndex$0$field$f$entry_count__$TY$__Snap$struct$m_PageIndex$$int$(_1) == -9223372036854775808 ==> result == 0 might not hold.` In this case the only discernible difference is the presence of the function body of `builtin$cast$isize$usize__$TY$__$int$$$int$`. This Rust program creates a successfully verifying Viper program for the `test` method:

```rust
use prusti_contracts::*;

pub struct PageIndex {
    entry_count: isize,
}

impl PageIndex {
    #[pure]
    #[ensures(self.entry_count == isize::MIN ==> result == 0)]
    #[ensures(self.entry_count == 0 ==> result == 0)]
    #[ensures(result <= isize::MAX as usize)]
    pub const fn len(&self) -> usize {
        if self.entry_count < 0 {
            ((1_isize + self.entry_count) + isize::MAX) as usize
        } else {
            self.entry_count as usize
            
        }
    }

    #[pure]
    #[ensures(result == (self.len() == 0))]
    pub const fn is_empty(&self) -> bool {
        self.entry_count == isize::MIN || self.entry_count == 0
    }

    #[requires(self.is_empty())]
    pub fn test(&self) {
        assert!(self.is_empty());
    }
}

pub fn main() {}
```

If you manually edit the resulting Viper program to remove the `{ number }` function body of `builtin$cast$isize$usize__$TY$__$int$$$int$`, then you arrive at the same error as mentioned above (there is a second cascading error `Postcondition of m_is_empty__$TY$__Snap$struct$m_PageIndex$$bool$ might not hold. Assertion result == (m_len__$TY$__Snap$struct$m_PageIndex$$int$(_1) == 0) might not hold.`, too).

Looking at the code that encodes the number casts, it does set a `Some(_)` in its `body` on line 563:

https://github.com/viperproject/prusti-dev/blob/7e3cb69fe1348beeb652101ca6b80eb9548b8c0c/prusti-viper/src/encoder/encoder.rs#L546-L569

so it must be converted into an abstract function somewhere else. One suspect is here on line 209:

https://github.com/viperproject/prusti-dev/blob/7e3cb69fe1348beeb652101ca6b80eb9548b8c0c/prusti-viper/src/encoder/definition_collector.rs#L196-L211

If I update that code to skip any function for which the name starts with `builtin$`, then the errors go away in my code base.

I'm planning to create a PR for this fix, but have some questions that could help me prepare the PR:

* I'm not able to create a minimal reproduction. Might anyone see a way to reproduce this minimally, so that I can add that as a regression test to the PR?
* Is it sound to skip all `builtin$*` functions in the solution? When it arrives there, the `builtin$` function may or may not have a function body (e.g. the [`builtin$unreach*` family](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/builtin_encoder.rs#L127-L241) are all abstract functions), but if it has one, I believe it should not be purged. Because, somehow this prevents the back-end from following the passed-in variable and it gets stuck during the automated proof.

Thanks!

Pointerbender:
The definition collector should revisit the pre- and post-conditions in case the function was not directly called yet.

Added a regression test.

Closes #783.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-783-1.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-783-1.rs
@@ -1,0 +1,38 @@
+use prusti_contracts::*;
+
+pub struct PageIndex {
+    entry_count: isize,
+}
+
+impl PageIndex {
+    #[pure]
+    #[ensures(self.entry_count == isize::MIN ==> result == 0)]
+    #[ensures(self.entry_count == 0 ==> result == 0)]
+    #[ensures(result <= isize::MAX as usize)]
+    pub const fn len(&self) -> usize {
+        if self.entry_count < 0 {
+            ((1_isize + self.entry_count) + isize::MAX) as usize
+        } else {
+            self.entry_count as usize
+            
+        }
+    }
+
+    #[pure]
+    #[ensures(result == (self.len() == 0))]
+    pub const fn is_empty(&self) -> bool {
+        self.entry_count == isize::MIN || self.entry_count == 0
+    }
+
+    #[pure]
+    pub const fn foo(&self) -> bool {
+        self.len() < 4
+    }
+
+    #[requires(self.foo())]
+    pub const fn bar(&self) -> bool {
+        !self.is_empty()
+    }
+}
+
+fn main() {}
--- a/prusti-viper/src/encoder/definition_collector.rs
+++ b/prusti-viper/src/encoder/definition_collector.rs
@@ -378,13 +378,15 @@ impl<'p, 'v: 'p, 'tcx: 'v> FallibleExprWalker for Collector<'p, 'v, 'tcx> {
     ) -> SpannedEncodingResult<()> {
         let identifier: vir::FunctionIdentifier =
             compute_identifier(function_name, formal_arguments, return_type).into();
-        let have_visited = !self.used_functions.contains(&identifier);
-        let have_visited_in_directly_calling_state =
+        let should_visit = !self.used_functions.contains(&identifier);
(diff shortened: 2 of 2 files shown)


vakaras (approved):
Thanks!

PR #795, opened 2021-12-10
Improved specification refinement lookup

vl0w:
This PR fixes two issues regarding specification refinement:

- The macro `refine_trait_spec` must be used on trait implementations and generates the _specification functions_ for specs inside that trait impl block as a sibbling impl block in order to make the expanded Rust code compilable. The existing implementation of `refine_trait_spec`, however, changed the original trait impl block in such a way that associated items were not included in the new trait impl block after expansion. As an example, 
```rust
#[refine_trait_spec]
impl Iterator for Foo {
    type Item = ...
    // ...
}
```
did not compile, because after macro expansion, `type Item = ...` was not included in the new impl block:
```rust
impl Foo {
    // specs for type checking
}
impl Iterator for Foo {
   type Item = ... // <-- missing!
    // ...
}
```

- When a trait method is called, Prusti performs a spec refinement: Use the spec of the impl if it is present or use the spec of the trait otherwise. The lookup of the impl spec can fail because the existing logic simply returns any implementation of a trait and not the actual implementation that will be called. This is problematic when we refine specs which have blanket implementations (as an example, refer to `refinement-involving-blanket-impls.rs` in this commit). The altered logic uses Rust compiler internals to find the implementation to-be called to get the correct specs.


prusti-interface/src/environment/mod.rs
```diff
@@ -289,16 +290,38 @@ impl<'tcx> Environment<'tcx> {
...
+                Note: In order to run ty::Instance::resolve, we disable diagnostics. In some cases
+                if method lookup fails, this method attaches delayed span bugs to the compiler session,
+                which will eventually be printed to stderr. With disabled diagnostics, these errors are ignored.
+                There is a change request pending for the Rust compiler to change that behaviour [1] which is not yet implemented.
+                [1]  https://github.com/rust-lang/compiler-team/issues/449
+             */
+            let diagnostic: &rustc_errors::Handler = self.tcx().sess.diagnostic();
+            let resolved_instance: Result<Option<ty::Instance<'tcx>>, rustc_errors::ErrorReported> = diagnostic.with_disabled_diagnostic(||  {
+                let param_env = ty::ParamEnv::reveal_all();
+                return ty::Instance::resolve(self.tcx(), param_env, proc_def_id, substs);
```
```suggestion
                ty::Instance::resolve(self.tcx(), param_env, proc_def_id, substs)
```

Issue #754, opened 2021-11-08
Interrupting the test suite doesn't kill the server


fpoli:
I currently have 2 `prusti-server` and 5 `prusti-server-driver` unused programs running in the background:
```
$ ps ax | grep [p]rusti
 5638 pts/18   Sl     0:00 ../target/debug/prusti-server --port=0
 5640 pts/18   Sl     0:40 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port=0
 7408 pts/18   Sl     0:00 ../target/debug/prusti-server --port=0
 7410 pts/18   Sl     0:39 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port=0
14870 pts/18   Sl     0:39 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port=0
18838 pts/18   Sl     0:39 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port=0
21675 pts/18   Sl     1:15 /home/fpoli/src/prusti/prusti-dev/target/debug/prusti-server-driver --port=0
```

I suspect that they are leaked when I `Ctrl + C` a test suite run.

gavinleroy:
Hi @fpoli 👋   I think I have an idea as to what is happening here.

I did a few experiments and the zombie processes are also leaked in a normal test suite run (without a CTRL+C event)!

For example (on MacOS):
```
% ./x.pyt test test_prusti_rustc_with_server &> /dev/null
% ps ax | grep [p]rusti                                   
35549 s003  S      0:00.15 /Users/gavinleroy/dev/prj/prusti-dev/prusti-launch/../target/debug/prusti-server-driver --port=0
```

So what is happening? If you recall from MR #344 when handling a SIGINT `prusti-server` will kill its entire process tree. Examining the above test [code ](https://github.com/viperproject/prusti-dev/blob/95e8fa6e2d89f8b927c9b413eb8d285873651146/prusti-launch/tests/test_binaries.rs#L122-L154) we see that a server process is spawned and wrapped in a `ChildGuard` such that when the guard goes out of scope the process is killed.

[...]

PR #799, opened 2021-12-13
Cache verification of unchanged methods

JonasAlaif:
This caching will be implemented at the `VerificationRequest` level


JonasAlaif:
With the above changes the `Hash` of repeated `VerificationRequest`s should always be the same... except for the fact that the fold-unfold "algorithm" is not deterministic, thus some of those statements may be reordered depending on whether it's in a good mood or not.

fpoli:
> With the above changes the `Hash` of repeated `VerificationRequest`s should always be the same... except for the fact that the fold-unfold "algorithm" is not deterministic, thus some of those statements may be reordered depending on whether it's in a good mood or not.

Bummer. You could try sorting the `reqs` at the beginning of the `fn obtain_all(&mut self, reqs: Vec<Perm>) -> ...` function to make that more stable.

prusti-viper/src/encoder/mir/pure/specifications/encoder.rs
```diff
@@ -255,7 +255,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> SpecificationEncoder<'p, 'v, 'tcx> {
                                         self.encode_quantifier_arg(
                                             *arg,
                                             arg_ty,
-                                            &format!("{}_{}", vars.spec_id, vars.pre_id),
```
vakaras, line 258: Have you checked whether this does not lead to strange name conflicts in encoded programs that are insanely hard to understand on the source level?

vakaras: It would be great to have a bunch of tests that check this.

JonasAlaif: I think that might be possible; I'll take a look at creating some test cases for this. However, in general I think this is very hard to avoid with deterministic names - one could just look at the generated Viper code and name another local variable in a way which clashes. The two solutions to avoid this that I see: (A) use hash of the HIR/MIR of the function as a sort of 'spec_id' which could be used in these cases (then if another variable is defined after the fact, this hash will change) or (B) look at all defined variables in scope and pick a name that doesn't clash.

vakaras: > (B) look at all defined variables in scope and pick a name that doesn't clash.

We could do this when lowering to `vir_legacy` – at that point, we have all variable names present.

fpoli: > one could just look at the generated Viper code and name another local variable in a way which clashes

Would that still be possible if we add a prefix `quantified_` (or just `q_`) to all quantified variables?

Even with that, as Vytautas asked I wonder what happens to the encoding of `forall x: u32 :: forall x: i32 :: ...`. Will the Viper names clash?

JonasAlaif: Aha I forgot that the user doesn't have control over MIR names. In that case we probably don't need to resort to the drastic measures I mention above? I'll think about where this might not hold

Aurel300: FYI in the spec refactor I base this ID on the `DefId` of the closure it comes from.

JonasAlaif: @Aurel300 do you mean the `spec_id`? The `pre_id` is just the level of nesting and shouldn't change in the refactor, right?

Aurel300: In my WIP branch, quantifier variables are named with `format!("_{}_quant_{}", arg_idx, body_def_id.index.index())`, where `arg_idx` is the position in the quantifier's arguments and `body_def_id` is the `DefId` of the closure containing the quantifier body. It is a local `DefId`, so `index` is the only number we use here.

Pointerbender:
I have one small additional feature request for the caching of the verification, if time allows :) Would it be possible to introduce a new Prusti configuration flag that allows to disable the caching for a certain run? The use case for me is that I'm currently trying to tweak the performance of some parts in Prusti, but if the benchmark I'm trying to run is in the cache, then I won't be able to see the actual performance difference :smile: I could also add the configuration flag myself in a separate PR, if that is more convenient :) Thanks!

JonasAlaif:
@Pointerbender good suggestion, I'll add such a flag.

prusti-viper/src/encoder/mir/pure/specifications/encoder.rs
JonasAlaif: @Aurel300 Hmm, but using the `DefId` is problematic in this case since it won't be stable if other stuff in the file/crate is added/deleted. I'm not exactly sure what position in quantifiers args (`arg_idx`) means (do you mean that for a `forall(|a: i32, b: i32| forall(|c: i32| ...))`, then `a` is 1, `b` is 2, `c` is 1?), but could we not use only this, with also a `quantifier_nesting_count` in the name as well?

fpoli:
Can't wait to use the verification cache on CI :)

prusti-viper/src/encoder/mir/pure/specifications/encoder.rs
Aurel300: Yes, it wasn't chosen with caching in mind, and it is still better than the original approach of random UUIDs. A nesting count would work, but I'm not sure it would be easy to introduce? When encoding specifications into Viper expressions we would generally work inside out, so it's hard for the inner `forall` to know its nesting depth.

JonasAlaif: The current `vars.pre_id` seems to do this now though (at least from looking at the generated Viper), I'm not sure how it's implemented but maybe if we just copied the ideas from that?

Aurel300: I thought `vars.pre_id` corresponded to another specification UUID? Maybe it's being interned somewhere?

JonasAlaif:
Caching should work quite nicely now.

Saving to disk is done by implementing a destructor here: https://github.com/viperproject/prusti-dev/blob/0eea62c7af012f6740b4a22845859a43ad71722b/viper/src/verification_result.rs#L129-L139
I'm not sure how well that interacts with the Prusti server and IDE plugin?
The test I added still fails in some cases I think.

Error reporting seems to work out of the box with cached results.

I'm going on holiday for two weeks from tomorrow so won't be able to finish merging before I'm back, but I'd be happy to let someone else take over - most of the work should be done (and this branch can be used if needed just fine) and I'll be online if needed.

prusti-server/src/process_verification.rs
```diff
@@ -8,12 +8,45 @@ use crate::{VerificationRequest, ViperBackendConfig};
...
+    request: VerificationRequest,
+    cache: impl Cache,
+) -> viper::VerificationResult {
+    if config::disable_cache() || config::print_hash() {
+        process_verification_request(verification_context, request)
+    } else {
+        let hash = request.get_hash();
+        // Try to load from cache and return `result`
+        if let Some(result) = cache.get(hash) {
+            return result;
```
fpoli, line 26:
```suggestion
            info!("Using verification result from the cache");
            return result;
```

Just to make debugging easier in the IDE, when users include their log when reporting issues.

`info!("Verification request hash: {}", hash);` two lines above might be useful too.

prusti-server/src/process_verification.rs
```diff
@@ -8,12 +8,45 @@ use crate::{VerificationRequest, ViperBackendConfig};
...
+}
 
 pub fn process_verification_request<'v, 't: 'v>(
     verification_context: &'v VerificationContext<'t>,
     request: VerificationRequest,
 ) -> viper::VerificationResult {
+    // Print hash of the `VerificationRequest`
+    if config::print_hash() {
+        println!(
+            "Recieved verification request for: {}",
```
fpoli:
```suggestion
            "Received verification request for: {}",
```

prusti-server/src/server.rs
```diff
@@ -54,35 +54,36 @@ where
     let viper = Arc::new(Viper::new_with_args(config::extra_jvm_args()));
     stopwatch.finish();
 
-    let build_verification_request_handler = |viper_arc: Arc<Viper>| {
+    let cache_loc = config::cache_dir() + "data.json";
```
fpoli: `let cache_loc = PathBuf::from(config::cache_dir()).join("data.json");` is more robust, e.g. if `cache_dir` doesn't end with a `/`.

prusti-server/src/server.rs
```diff
@@ -54,35 +54,36 @@ where
...
+    let cache = Arc::new(Mutex::new(CacheData::load_cache(&cache_loc)));
+    let build_verification_request_handler = |viper_arc: Arc<Viper>, cache| {
         move |request: VerificationRequest| {
             // Here it would be easy to cache the result of verification requests, as soon as
             // `VerificationRequest` implements `Hash`.
```
fpoli:
```suggestion
```

prusti-viper/src/verifier.rs
```diff
@@ -422,8 +422,10 @@ fn verify_programs(env: &Environment, programs: Vec<vir::Program>)
         stopwatch.start_next("attach current thread to the JVM");
         let viper_thread = viper.attach_current_thread();
         stopwatch.finish();
+        let cache_loc = config::cache_dir() + "data.json";
```
fpoli: `PathBuf::from(...).join(...)` is more robust.

Also, what about renaming `cache_dir` to `cache_path`, making the default `"cache/data.json"`? That way, the user (or the IDE) can use e.g. `/tmp/prusti_cache.json`.

prusti-viper/tests/test_caching.rs
```diff
@@ -0,0 +1,105 @@
...
+            .env("RUST_BACKTRACE", "1")
+            .env("PRUSTI_DUMP_VIPER_PROGRAM", "true")
+            .env("PRUSTI_PRINT_HASH", "true")
+            .env("PATH", env!("PATH"))
+            .output()
+            .expect("failed to execute prusti-rustc");
+        assert!(out.status.success(), "Failed to compile: {:?}\n{}", program, String::from_utf8(out.stderr).unwrap());
+        let stdout = String::from_utf8(out.stdout).unwrap();
+        let mut hash_lines = stdout.lines()
+            .skip_while(|line| !line.starts_with("Recieved verification request for: "));
```
fpoli:
```suggestion
            .skip_while(|line| !line.starts_with("Received verification request for: "));
```

prusti-viper/tests/test_caching.rs
```diff
@@ -0,0 +1,105 @@
...
+        while let Some(l1) = hash_lines.next() {
+            let full_name = l1.strip_prefix("Recieved verification request for: ").unwrap();
```
fpoli:
```suggestion
            let full_name = l1.strip_prefix("Received verification request for: ").unwrap();
```

viper/src/verification_result.rs
```diff
@@ -70,3 +78,82 @@ pub struct JavaExceptionWithOrigin {
     /// The actual exception.
     pub exception: JavaException,
 }
+
+pub trait Cache {
+    fn get(&self, request: u64) -> Option<VerificationResult>;
+    fn insert(self, request: u64, result: VerificationResult) -> Option<VerificationResult>;
+}
+#[derive(Debug, Clone)]
+pub struct CacheData {
```
fpoli: What about `PersistentCache` instead of `CacheData`?

Also, I'd move all the cache-related code to a new `src/cache.rs` file.

viper/src/verification_result.rs
```diff
@@ -70,3 +78,82 @@ pub struct JavaExceptionWithOrigin {
...
+    updated: bool,
+    load_loc: String,
```
fpoli: `PathBuf` instead of `String`?

fpoli:
> I'm not sure how well that interacts with the Prusti server and IDE plugin?

There is some stuff to fix (e.g. the IDE should terminate with `SIGINT` instead of `SIGKILL`) and to add (e.g. a "clear cache" command); after holidays. It'd be much easier to start with in-memory cache enabled but on-disk cache disabled, because the IDE already has a "restart server" command that can be used to clear the in-memory cache.

prusti-common/src/config.rs
```diff
@@ -101,6 +102,8 @@ lazy_static! {
         settings.set_default("print_collected_verification_items", false).unwrap();
         settings.set_default("hide_uuids", false).unwrap();
         settings.set_default("counterexample", false).unwrap();
+        settings.set_default("print_hash", false).unwrap();
+        settings.set_default("disable_cache", false).unwrap();
```
fpoli: `enable_cache` might be more immediate (one less negation = one less neuron required)

prusti-viper/tests/hash/layout_1.rs
```diff
@@ -0,0 +1,157 @@
+use prusti_contracts::*;
+
+// Borrows are not stable since rustc::Loan indicies aren't
+// i.e. the `N` used in `DeadBorrowToken$(N)` for wands
+// will differ between compilation
+// This includes fns which return borrows, or e.g. slices
+// where the fold-unfold algorithm doesn't handle the exhale well
```
Pointerbender, line 7: Are the loan indices unstable because of how Polonius returns the indices, or are these renumbered in a non-deterministic fashion somewhere in Prusti? (e.g. due to the use of a randomly seeded HashSet/HashMap somewhere).

Pointerbender: Indeed this turned out to be caused by the use of randomly seeded `HashSet`/`HashMap` in various places. PR #817 to the `master` branch should solve this!

JonasAlaif:
The CI is now failing on #827

viper/src/cache.rs
```diff
@@ -0,0 +1,126 @@
...
+        let mut data_res: Option<ResultCache> = None;
+        if !cache_loc.as_os_str().is_empty() {
+            if let Ok(f) = fs::File::open(&cache_loc) {
+                match bincode::deserialize_from(&mut io::BufReader::new(f)) {
+                    Ok(data) => {
+                        info!("Loaded cache from \"{}\"", cache_loc.display());
+                        data_res = Some(data);
+                    }
+                    Err(e) =>
+                        error!("Failed to read cache from \"{}\": {e}", cache_loc.display()),
```
fpoli: Small nit: if the file doesn't exist we can omit this error.

fpoli: Oh, but then `open` fails. Ignore my comment.

fpoli (approved):

viper/src/cache.rs
JonasAlaif: Yeah, I had a more detailed structure where the first two `if`s would produce `info!` output in the `else` cases but then removed that in order to simplify things and instead just use the generic `info!("Cache file doesn't exist or is invalid. Using fresh cache.")` thing at the end.
But yes, this error here is if we try to load the cache from some file that has some random data which we fail to parse.

fpoli:
The reason the cache is not saved when sending a `SIGINT` to the server is probably that `prusti-server` abruptly kills the child server process. Sending `Signal::SIGINT` should be better.
https://github.com/viperproject/prusti-dev/blob/be28ed5137384504c43e9e84944df0b1b4eb2553/prusti-launch/src/lib.rs#L207

If we do so, we should also remove the `"/F"` argument on Windows:
https://github.com/viperproject/prusti-dev/blob/be28ed5137384504c43e9e84944df0b1b4eb2553/prusti-launch/src/lib.rs#L218

(A related issue is https://github.com/viperproject/prusti-dev/issues/754)

JonasAlaif:
I tried replacing the `SIGKILL` with a `SIGINT` but this doesn't help. I'm pretty sure that killing the process like this doesn't call Rust destructors at all - the only reason caching was working for me is that I was running Prusti without the server (as is the default) and then things are destructed properly. It seems that there are two options:
- Catch SIGINT/SIGTERM in prusti-server and save the cache at that point
- Add the `POST /save` to prusti-server which should be called before quitting

I've implemented the latter for now as it was easier, but we may want to consider switching in the future (the advantage would be when running prusti-server from the command line rather than IDE)

tillarnold:
The caching appears to also affect the [automatic benchmarks](https://github.com/viperproject/prusti-dev/blob/gh-pages/benchmark-output/benchmark1642783441.8613358.json) (especially the speedup for Knights_tour.rs is very impressive). Would it maybe be a good idea to set the `disable_cache` flag for the benchmarks to get more representative results?

vakaras:
> The caching appears to also affect the [automatic benchmarks](https://github.com/viperproject/prusti-dev/blob/gh-pages/benchmark-output/benchmark1642783441.8613358.json) (especially the speedup for Knights_tour.rs is very impressive). Would it maybe be a good idea to set the `disable_cache` flag for the benchmarks to get more representative results?

Thank you, @tillarnold, for noticing this. @JonasAlaif Could you please fix this?

PR #800, opened 2021-12-14
Fix the script that dumps lifetime information

vakaras:
Replace the old `dump_borrowck_info.rs` with new code.


Aurel300:
Does this fix any existing issues?

vakaras:
> Does this fix any existing issues?

No. This is the code for visualising compiler's internal information.

fpoli:
What's the way to run the script?

vakaras:
> What's the way to run the script?

Same as before: `PRUSTI_DUMP_BORROWCK_INFO=true`.

PR #801, opened 2021-12-14
Fix glob patterns

fpoli:
There was more than one copy-paste error in the glob patterns.

### DIFF
--- a/prusti-launch/tests/pass/assert_true.rs
+++ b/prusti-launch/tests/pass/assert_true.rs
@@ -1,5 +1,5 @@
 fn test() {
-    assert!(false);
+    assert!(true);
 }
 
 fn main() {}
--- a/prusti-launch/tests/test_binaries.rs
+++ b/prusti-launch/tests/test_binaries.rs
@@ -5,13 +5,10 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use glob::glob;
-use prusti_launch::find_java_home;
 use std::{
-    collections::HashMap,
-    env,
     io::{BufRead, BufReader},
     path::PathBuf,
-    process::{Child, Command, ExitStatus, Stdio},
+    process::{Child, Command, Stdio},
 };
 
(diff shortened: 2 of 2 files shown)


JonasAlaif:
Ah sorry I forgot to mention the issue with `clear_env`. The only thing you need to add is `.env("PATH", env!("PATH"))` I'd say, no need for the extra
```
.envs(&filtered_env)
.stdout(Stdio::null())
.stderr(Stdio::null())
```
(The stdout/err should only be shown if the cargo test fails anyway)

I would also say that the current glob you have is fine.

fpoli:
I'd still like to keep `clear_env`, because it should make the test more robust in case the environment is not clean.

JonasAlaif:
I meant using `... .env_clear().env("PATH", env!("PATH")) ...` - it's probably nicer than the `filter` and `.envs(&filtered_env)` solution. I guess the `unwrap_or_default` version also works, though it'll just fail later on anyway if PATH is not set.

JonasAlaif:
There are also two
```
        println!(
            "Running {:?} on {:?}...",
             prusti_rustc.display(),
             program.display()
         );
```
You can replace that with ` println!("Running {:?} on {:?}...", prusti_rustc, program);` imo
Edit: I wonder if this is used elsewhere

JonasAlaif:
I would also maybe use
```
for entry in fs::read_dir("tests/pass/").unwrap_or_else(|_| panic!("Did not find dir: {}", dir)) {
        let path = entry.unwrap().path();
```
rather than
```
let pass_entries = glob("tests/pass/**/*.rs").expect("failed to read glob pattern");
for entry in pass_entries {
```
Such that we don't need to import glob

fpoli:
Ideally it would be nice to pipe both stderr and stdout of the child processes to stdout, but I don't see an easy way to do that. Removing the glob dependency feels a small improvement to me... I like the idea that we can easily add folders if needed.

fpoli:
bors r+

For some reason, Windows doesn't like `env_clean()`.

PR #803, opened 2021-12-16
Add ignore-regions configuration flag that ignores regions in dumps.

vakaras:
bors r+

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -35,6 +35,7 @@ jobs:
           RUSTFLAGS: -Zinstrument-coverage
           PRUSTI_DUMP_DEBUG_INFO: true
           PRUSTI_DUMP_VIPER_PROGRAM: true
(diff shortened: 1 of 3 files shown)


.github/workflows/coverage.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           RUSTFLAGS: -Zinstrument-coverage
           PRUSTI_DUMP_DEBUG_INFO: true
           PRUSTI_DUMP_VIPER_PROGRAM: true
+          PRUSTI_IGNORE_REGIONS: true
```
fpoli, line 38: Oh, now I see, CI fails on `master` because this `master`-only run enables `PRUSTI_DUMP_DEBUG_INFO` when measuring coverage of the test suite. Thanks for the fix! Removing `PRUSTI_DUMP_DEBUG_INFO: true` from the build would also be a perfectly fine alternative solution IMO.

PR #804, opened 2021-12-16
Fix exists being translated to forall

JonasAlaif:
Hmmm, well [this](https://github.com/viperproject/prusti-dev/commit/8777f800133f56b27f59ad8cf79c9a1a2d6fb23b) is embarrassing... 😄
Will add test before merging


JonasAlaif:
I was quite confused why the hash of a precondition with `exists` is the same as that of a `forall` haha

PR #809, opened 2021-12-22
Fixes for issue #721

### LINKED ISSUE #721: ASSERT_TIMEOUT=0 not properly passed to Viper?
I have an example which I believe should verify:

```rust
use prusti_contracts::*;

pub const MAX_DEPTH: usize = usize::BITS as usize;

pub struct Path {
    current: usize,
    depth: usize,
    entries: [usize; MAX_DEPTH],
    max_counts: [usize; MAX_DEPTH],
}

impl Path {
    #[pure]
    pub fn invariant(&self) -> bool {
        self.entries.len() == MAX_DEPTH
            && self.max_counts.len() == MAX_DEPTH
            && self.entries.len() == self.max_counts.len()
            && self.current < MAX_DEPTH
            && self.depth < MAX_DEPTH
            && self.current <= self.depth
    }

    #[requires(self.invariant())]
    #[ensures(self.current == 0)]
    #[ensures(self.depth == 0)]
    #[ensures(self.entries[0] == 0)]
    #[ensures(self.invariant())]
    pub fn reset(&mut self) {
        self.current = 0;
        self.depth = 0;
        self.entries[0] = 0;
        assert!(self.entries.len() == self.max_counts.len());
        assert!(self.current < MAX_DEPTH);
        assert!(self.depth < MAX_DEPTH);
        assert!(self.current <= self.depth);
        assert!(self.entries[0] == 0);
    }
}

pub fn main() {}
```

Prusti fails the verification with the error message (with setting [`ASSERT_TIMEOUT=0`](https://viperproject.github.io/prusti-dev/dev-guide/config/flags.html#assert_timeout) "Set to 0 to disable timeout.") after about 1 minute and 25 seconds:

```
error: [Prusti: verification error] the asserted expression might not hold
  --> src/lib.rs:37:9
   |
37 |         assert!(self.entries[0] == 0);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `prusti-example` due to previous error
Command exited with non-zero status 101
```

The weird part is, if I dump the resulting program with [`DUMP_VIPER_PROGRAM=true`](dump_viper_program) and try to verify the program for `target/verify/log/viper_program/lib.rs-Path::reset.vpr` using Viper IDE in VS Code, it tells me `Viper: Verification of lib.rs-Path::reset.vpr timed out after 100000ms` the first time I run the example (i.e. during a "cold run"), but it verifies okay and without errors during a subsequent "warm run" in 98 seconds (and then okay in further subsequent runs in under 2 seconds).

I have the line `assert_timeout = 0` in my `Prusti.toml` and I'm running the Cargo project as `PRUSTI_DUMP_VIPER_PROGRAM=true PRUSTI_ASSERT_TIMEOUT=0 ASSERT_TIMEOUT=0 cargo-prusti`. If I remove the last `assert!()` from the example, it verifies okay from the command line using `cargo-prusti`. Note that the identical post-condition `#[ensures(self.entries[0] == 0)]` is verified okay, despite the last assertion being identical. Unfortunately I was not able to make an even more minimal example. Removing any line makes it verify okay, further hinting at a timeout issue. 

This gives me the suspicion that the `ASSERT_TIMEOUT=0` argument is not passed to the Prusti server properly (or to the back-end). I took a quick look at the code, but did not see anything extremely suspicious there (assuming it uses the Silicon back-end by default). The only thing I could not verify, is if it is passed down properly in file `prusti-common/src/verification_service.rs`, as I couldn't find the manual for the CLI arguments passed to Silicon (in this case `vec!["--assertTimeout".to_string(), config::assert_timeout().to_string()]`) nor the code that would pass the extended arguments to the backend. There is some commented out logic in `prusti-viper/src/verifier.rs`, that seems like it used to handle it, but couldn't find where it is handled now in the latest version from the `master` branch.

Did I hit some kind of bug/out-of-date documentation/unsupported feature here by any chance? :) Thanks!

p.s. is there any way to speed up the verification of the example program? 100+ seconds seems like a lot for such a relatively simple program. Other programs of similar size often verify in matter of mere seconds (sometimes even in under a second), which is why the verification time seems a little surprising in this case. I guess it's related to the array size (64 elements on my architecture), which seems to generate a lot of statements in the Viper program.

Pointerbender:
This PR got a little bigger than I anticipated, so here is a little explanation of what was changed and tested and why :)

I set out to implement [this idea](https://github.com/viperproject/prusti-dev/issues/721#issuecomment-997446669) from #721 in Prusti, but had to slightly deviate from the initial plan regarding the `Snap$Array` domain:

```
domain Snap$Array$64$usize {

  function cast$Snap$Array$64$usize$__$TY$__Snap$Slice$usize(array: Snap$Array$64$usize): Snap$Slice$usize
  
  // ...

  axiom Snap$Array$64$usize$unsize {
    (forall data: Seq[Int] ::
      { cast$Snap$Array$64$usize$__$TY$__Snap$Slice$usize(cons$Snap$Array$64$usize$__$TY$__Seq$$int$$Snap$Array$64$usize(data)) }
      cast$Snap$Array$64$usize$__$TY$__Snap$Slice$usize(cons$Snap$Array$64$usize$__$TY$__Seq$$int$$Snap$Array$64$usize(data)) == cons$Snap$Slice$usize$__$TY$__Seq$$int$$Snap$Slice$usize(data)
    )
  }
}
```

Shortly after working on it I realized that this would create a dependency on `Snap$Slice`, but there is no guarantee a program contains the slice type, so in some corner cases this could lead to a consistency error. Another problem I ran into is that when we're building the `Snap$Array` domain, we don't have access to a slice type (i.e. a `rustc_middle::ty::Ty`). There exists [`mk_slice`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.mk_slice) that could help out here, but instead I chose to use an abstract function instead which is automatically generated when unsizing an array (where both the array and slice types are readily available):

```
function builtin$unsize$Array$64$usize$Slice$usize__$TY$__Snap$Array$64$usize$Snap$Slice$usize(array: Snap$Array$64$usize): Snap$Slice$usize
  ensures (exists data: Seq[Int] ::
		array == cons$Snap$Array$64$usize$__$TY$__Seq$$int$$Snap$Array$64$usize(data)
		&& result == cons$Snap$Slice$usize$__$TY$__Seq$$int$$Snap$Slice$usize(data)
  )
```

Using this abstract function instead of an axiom had a another benefit, namely that I was able to use one quantifier less in snapshot encoding of an array (note that the `ensures [exists data: Seq[Int] :: result == cons$Snap$Array$64$usize$__$TY$__Seq$$int$$Snap$Array$64$usize(data) && |data| == 64, true]` post-condition is intentionally left out on `function snap$__$TY$__Array$64$usize$Snap$Array$64$usize()`, this works okay):

```
function seq_collect$Array$64$usize$__$TY$__Array$64$usize$$int$$Seq$$int$(self: Ref, start: Int): Seq[Int]
  requires acc(Array$64$usize(self), read$())
  requires 0 <= start
  ensures start >= 64 ==> result == Seq[Int]()
  ensures start < 64 ==> |result| == 64 - start
  ensures start < 64 ==> result[0] == lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, start)
  ensures start < 64 ==> (forall i: Int, j: Int :: { lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, i),result[j] } start <= i && i < 64 && (0 <= j && j < 64 - start && i == j + start) ==> lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, i) == result[j])
{
  (start >= 64 ? Seq[Int]() : Seq(lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, start)) ++ seq_collect$Array$64$usize$__$TY$__Array$64$usize$$int$$Seq$$int$(self, start + 1))
}

function snap$__$TY$__Array$64$usize$Snap$Array$64$usize(self: Ref): Snap$Array$64$usize
  requires acc(Array$64$usize(self), read$())
  ensures (forall i: Int :: { read$Snap$Array$64$usize$__$TY$__Snap$Array$64$usize$$int$$$int$(result, i) } { lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, i) } 0 <= i && i < 64 ==> read$Snap$Array$64$usize$__$TY$__Snap$Array$64$usize$$int$$$int$(result, i) == lookup_pure__$TY$__Array$64$usize$$int$$$int$(self, i))
{
  cons$Snap$Array$64$usize$__$TY$__Seq$$int$$Snap$Array$64$usize(seq_collect$Array$64$usize$__$TY$__Array$64$usize$$int$$Seq$$int$(self, 0))
}
```

The `seq_collect` function is based on Johannes' `slice_collect` function, but slightly renamed in order to benefit from the code re-use between arrays and slices (through `SnapshotEncoder::encode_seq_collect_func()`).

After using a single quantifier for all array operations (instantiate, construct, cast) the example program from #721 was passing verification, but there were 4 regressions in the test suite. The most interesting case was `prusti-tests/tests/verify/pass/arrays/simple-equality.rs`:

```rust
fn main() {
    let a = [0; 3];
    let b = [0, 0, 0];

    assert!(a == b);
}
```

Now that Prusti no longer triggered all the quantifiers for arrays upon its first use, this test case was failing. The reason for this was as surprising as interesting, to understand it better I'll have to back-up my explanation with footnote 6 on page 15 of Johannes' thesis: "*In practice the injectivity axiom seems sufficient, so an axiom for surjectivity, encoding that all values of type Snap$T must correspond to an application of the cons function, is currently not generated.*". My theory is that the surjectivity axiom was not required at the time, because all array quantifiers were fully instantiated for all elements everywhere. Z3 was then able to infer the surjectivity of array constructors somehow. But now that the cascade of quantifier instantiations was stopped, this no longer worked. So, I implemented the surjectivity axiom for both arrays and slices and added a `simple-equality` test case for slices, too (this one was missing still, I had to make one additional fix in `prusti-viper/src/encoder/encoder.rs` to make it pass for slices as well).

The 3 other test cases were all failing for the same reason: these no longer could benefit from the information made available by the many inhale statements (now replaced by a single quantifier) and the program was not complete enough to trigger to right quantifiers. I added extra [dummy statements](https://github.com/viperproject/prusti-dev/issues/721#issuecomment-992472952) to help the test cases pass, without changing the intended functionality:

* `prusti-tests/tests/verify/pass/arrays/merge.rs`
* `prusti-tests/tests/verify/pass/arrays/selection_sort.rs`
* `prusti-tests/tests/verify/pass/slices/slice-split.rs`

Last but not least, I also added the example programs that came up during the in discussions in #721 as additional regression tests.

I also gave some thought to the question whether we should create multiple inhale statements instead of a single quantifier for small array lengths ([relevant comment](https://github.com/viperproject/prusti-dev/issues/721#issuecomment-948495663)). I think we should always use quantifiers instead of multiple statements for the following reasons:

* Encoding individual inhales for every element will trigger a lot of quantifiers, which may not be needed for every program, leading to performance loss instead of performance gain.
* If we would not use quantifiers for small array lengths, this could lead to different verification results for different array sizes, as witnessed in the `prusti-tests/tests/verify/pass/arrays/simple-equality.rs` test case. There may exist other corner cases as well. This would not be not a good user experience.

Finally, I also removed one FIXME here:

https://github.com/viperproject/prusti-dev/blob/69325d35ec51a45118ec93ed4e46957e1d08f903/prusti-viper/src/encoder/snapshot/encoder.rs#L846-L857

Now that the performance problem with large arrays is adressed, this FIXME was resolved.

That about summarizes all the changes and tests. Thanks for all the tips, examples and pointers so far, it has been a great help in attempting this! All feedback is very much welcome :-)

Closes #721.

### DIFF
--- a/prusti-tests/tests/verify/pass/arrays/merge.rs
+++ b/prusti-tests/tests/verify/pass/arrays/merge.rs
@@ -50,10 +50,12 @@ fn merge(a: [i32; 3], b: [i32; 3]) -> [i32; 6] {
         body_invariant!(res_pos > 0 && b_pos < 3 ==> res[res_pos - 1] <= b[b_pos]);
 
         if b_pos == 3 || a_pos < 3 && a[a_pos] <= b[b_pos] {
-            res[res_pos] = a[a_pos];
+            // res[res_pos] = a[a_pos];
+            set_res(&mut res, res_pos, a[a_pos]);
             a_pos += 1;
         } else {
-            res[res_pos] = b[b_pos];
+            // res[res_pos] = b[b_pos];
+            set_res(&mut res, res_pos, b[b_pos]);
             b_pos += 1;
         }
         res_pos += 1;
@@ -62,4 +64,14 @@ fn merge(a: [i32; 3], b: [i32; 3]) -> [i32; 6] {
     assert!(res_pos == res.len());
 
     res
+}
+
+#[requires(res_pos < 6)]
+#[requires(forall(|i: usize, j: usize| (0 <= i && i < j && j < res_pos) ==> res[i] <= res[j]))]
+#[requires(forall(|i: usize| (0 <= i && i < res_pos) ==> res[i] <= value))]
+#[ensures(forall(|i: usize| (0 <= i && i < res_pos) ==> res[i] == old(res[i])))]
+#[ensures(res[res_pos] == value)]
+#[ensures(forall(|i: usize, j: usize| (0 <= i && i < j && j <= res_pos) ==> res[i] <= res[j]))]
+fn set_res(res: &mut [i32; 6], res_pos: usize, value: i32) {
+    res[res_pos] = value
 }
--- a/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
+++ b/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
@@ -44,11 +44,29 @@ fn selection_sort(mut a: [i32; 10]) {
             j += 1;
         }
 
-        let a_i = a[i];
-        let a_min = a[min];
-        a[i] = a_min;
-        a[min] = a_i;
+        // let a_i = a[i];
+        // let a_min = a[min];
+        // a[i] = a_min;
+        // a[min] = a_i;
+        set_a(&mut a, i, min);
 
         i += 1;
     }
+}
+
+#[requires(i < 10)]
+#[requires(min < 10)]
+#[requires(i <= min)]
+#[requires(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < k2 && k2 < i) ==> a[k1] <= a[k2]))]
+#[requires(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < i && i <= k2 && k2 < 10) ==> a[k1] <= a[k2]))]
+#[ensures(forall(|k1: usize| (0 <= k1 && k1 < 10 && k1 != i && k1 != min) ==> (a[k1] == old(a[k1]))))]
+#[ensures(a[i] == old(a[min]))]
+#[ensures(a[min] == old(a[i]))]
+#[ensures(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < k2 && k2 <= i) ==> a[k1] <= a[k2]))]
+#[ensures(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < i && i <= k2 && k2 < 10) ==> a[k1] <= a[k2]))]
+fn set_a(a: &mut [i32; 10], i: usize, min: usize) {
+    let a_i = a[i];
+    let a_min = a[min];
+    a[i] = a_min;
+    a[min] = a_i;
 }
--- a/prusti-tests/tests/verify/pass/issues/issue-721-1.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-721-1.rs
@@ -1,0 +1,41 @@
+//! See https://github.com/viperproject/prusti-dev/issues/721
+use prusti_contracts::*;
+
+pub const MAX_DEPTH: usize = usize::BITS as usize;
+
+pub struct Path {
+    current: usize,
+    depth: usize,
+    entries: [usize; MAX_DEPTH],
+    max_counts: [usize; MAX_DEPTH],
+}
+
+impl Path {
+    #[pure]
+    pub fn invariant(&self) -> bool {
+        self.entries.len() == MAX_DEPTH
+            && self.max_counts.len() == MAX_DEPTH
+            && self.entries.len() == self.max_counts.len()
+            && self.current < MAX_DEPTH
+            && self.depth < MAX_DEPTH
+            && self.current <= self.depth
+    }
+
+    #[requires(self.invariant())]
+    #[ensures(self.current == 0)]
+    #[ensures(self.depth == 0)]
+    #[ensures(self.entries[0] == 0)]
+    #[ensures(self.invariant())]
+    pub fn reset(&mut self) {
+        self.current = 0;
+        self.depth = 0;
+        self.entries[0] = 0;
+        assert!(self.entries.len() == self.max_counts.len());
+        assert!(self.current < MAX_DEPTH);
+        assert!(self.depth < MAX_DEPTH);
+        assert!(self.current <= self.depth);
+        assert!(self.entries[0] == 0);
+    }
+}
+
+fn main() {}
--- a/prusti-tests/tests/verify/pass/issues/issue-721-2.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-721-2.rs
@@ -1,0 +1,20 @@
+use prusti_contracts::*;
+use std::convert::TryFrom;
+
+#[trusted]
+#[requires(s.len() == 16)]
+#[ensures(forall(|i: usize| (0 <= i && i < 16) ==> (old(s[i]) == result[i] )))]
+#[ensures(result[0] > 100)]
+pub fn slice_to_array16(s: &[i32]) -> [i32; 16] {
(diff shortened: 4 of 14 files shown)


vakaras:
Thanks! Most of our team are currently on vacation, so it will take time to review this PR.

A quick question: do you have a small enough test that can be added to our test suite which with current Prusti takes significantly longer to verify than with our improvements? Since this code is something that is going to be touched by my refactorings I would like to make sure that any regressions are not unnoticed.

Pointerbender:
The test in `prusti-tests/tests/verify/pass/issues/issue-721-1.rs` tests for this, it will timeout and fail verification on the current version of Prusti. I might be able to come up with an even more minimal test case with bigger array sizes (close to the stack size limit). I will play with this a bit and add the extra test case :) Thanks for the suggestion!

Pointerbender:
I added `prusti-tests/tests/verify/pass/issues/issue-721-4.rs`. On the current version of Prusti it will take a long time during the encoding step (it's very noticeable, I'm not sure if it will finish in reasonable time, I didn't see it finish yet), but on the version of this PR the whole compilation and verification finishes in 6 seconds flat :)

vakaras (commented):
LGTM

Pointerbender:
Thanks!

I just pushed one extra commit, I found a corner case where the array length was not known to Z3 when unsizing an array into a slice inside nested pure functions. I added an extra `&& |data| == 64` clause to the post-condition of `function builtin$unsize` and added an extra regression test for this, too.

vakaras:
Closing because it was merged as part of #813

Pointerbender:
Thanks!

PR #820, opened 2022-01-07
Improve name interner

JonasAlaif:
The current name interner will pick between two names: a short version if there is no clash with another name it has already given out, or the full unique long version if there is a clash. This sucks for the following example:
```
trait T {
  fn foo() {}
}
fn foo() {}
```
since the trait function is given the name `m_foo` and the other fn gets `m_filename::foo` - looking at the names alone, one can hardly tell which is which. The naming here is also based on the order functions are defined within a file; not great for caching (#799).
Instead, the interner should return a "name-token" for a given `def_id` and only once all names are interned can these tokens be exchanged for string names. Internally, it will probably save a mapping from name-token to a path of uniquely identifying names (as returned by `tcx.def_path(def_id)`; the names between the `::` in e.g. `crate_name::mod_name::mod_name::trait_name::item_name`). Each path segment would also have an associated "active" boolean stored alongside it, with all but the last (`item_name`) initially set to `false`. Imagine the type of the map as:
```
FxHashMap<Uuid, Vec<(bool, String)>>
```
When adding to the map, if there is already an element with the same `item_name` (an element that clashes, since on the new item only this segment is "active"), we traverse the vector and find the **most generic** segment which differs between the two and set it active on **both**. The `Vec<(bool, String)>` to `String` translation is then easy, once tokens are being exchanged for names.


JonasAlaif:
This PR is very much not yet finished; the name_interner does not yet use the `tcx.def_path(def_id)` vector as described above, but the main thing to figure out are the "name-tokens" and what functionality they might need to implement (maybe the `WithIdentifier` trait should return these tokens rather than a String?).

prusti-viper/src/encoder/encoder.rs
```diff
@@ -789,14 +789,15 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
 
-    pub fn encode_item_name(&self, def_id: DefId) -> String {
-        let full_name = format!("m_{}", encode_identifier(self.env.get_item_def_path(def_id)));
-        let short_name = format!("m_{}", encode_identifier(
-            self.env.tcx().opt_item_name(def_id)
-                .map(|s| s.name.to_ident_string())
-                .unwrap_or_else(|| self.env.get_item_name(def_id))
-        ));
-        self.intern_viper_identifier(full_name, short_name)
+    pub fn encode_item_name(&self, def_id: DefId, prefix: &str) -> String {
```
vakaras, line 792: I think it would be better to have two methods `encode_item_name` that takes only `def_id` and `encode_item_name_with_prefix` that corresponds to this one.

prusti-viper/src/encoder/mir/procedures/encoder.rs
```diff
@@ -54,7 +54,7 @@ struct ProcedureEncoder<'p, 'v: 'p, 'tcx: 'v> {
 
 impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
     fn encode(&mut self) -> SpannedEncodingResult<cfg::ProcedureDecl> {
-        let name = self.encoder.encode_item_name(self.def_id);
+        let name = self.encoder.encode_item_name(self.def_id, "");
```
vakaras, line 57:
```suggestion
        let name = self.encoder.encode_item_name(self.def_id);
```

prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
```diff
@@ -427,7 +427,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
     }
 
     pub fn encode_function_name(&self) -> String {
-        self.encoder.encode_item_name(self.proc_def_id)
+        self.encoder.encode_item_name(self.proc_def_id, "")
```
vakaras, line 430:
```suggestion
        self.encoder.encode_item_name(self.proc_def_id)
```

prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -61,19 +61,19 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
     }
 
     fn encode_enum_name(&self, did: DefId) -> String {
-        format!("enum${}", self.encoder.encode_item_name(did))
+        self.encoder.encode_item_name(did, "enum$")
```
vakaras, line 64:
```suggestion
        self.encoder.encode_item_name_with_prefix(did, "enum$")
```

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -66,6 +136,7 @@ impl Default for NameInterner {
     }
 }
 
+/*
```
vakaras, line 139: Please do not forget to reenable tests.

vir/src/common/identifier.rs
```diff
@@ -1,3 +1,18 @@
...
 }
+
+// TODO: this will need to be some extensible name, which supports all of the
+// name operations that we might want to do, but is not a string.
+
+/// Temporary standing for a name, returned by `NameInterner::intern`.
+/// Can be exchanged for a real `String` name with `NameInterner::get_interned`,
+/// but only once all names have been interned.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, PartialOrd, Ord)]
+pub struct NameHash(u64, u64);
```
vakaras, line 12: Why is this not just a wrapper around `u64`? The interner could have a counter that is incremented each time it needs to intern a new identifier.

vakaras (changes requested):

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -4,60 +4,130 @@
...
+    pub fn intern(&mut self, prefix: &str, def_id: DefId, tcx: TyCtxt) -> String {
+        let def_path = tcx.def_path(def_id);
+        // The `to_string` gives each `DisambiguatedDefPathData` separated by "::"
+        let crate_name = tcx.crate_name(def_path.krate).to_string();
+        let item_name = def_path.to_string_no_crate_verbose();
+        // Assume that the friendly `tcx.opt_item_name` is the last one, use second-to-last as the mod unique, and others as crate unique.
+        let mut name_split = item_name.rsplitn(3, "::");
+        let short_name = name_split.next().unwrap_or_default().to_string();
+        let mod_unique = name_split.next().unwrap_or_default().to_string();
+        // TODO: How does this work when we have a crate with multiple files
```
fpoli, line 83: What do you mean?

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -4,60 +4,130 @@
...
+            let full_name = format!("{}{}::{}::{}::{}", prefix, crate_name, crate_unique, mod_unique, short_name);
+            let short_name = format!("{}{}", prefix, short_name);
+            panic!(
+                "Attempting to add {} to the `NameInterner` after names have been read out, \
+                this could invalidate these names if they clash with {}!",
+                full_name,
+                short_name
+            );
+        }
+        let can_be_unqiue = config::disable_name_mangling() || config::intern_names();
```
fpoli, line 107:
```suggestion
        let can_be_unique = config::disable_name_mangling() || config::intern_names();
```

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -4,60 +4,130 @@
...
+        // Assume that the friendly `tcx.opt_item_name` is the last one, use second-to-last as the mod unique, and others as crate unique.
+        let mut name_split = item_name.rsplitn(3, "::");
+        let short_name = name_split.next().unwrap_or_default().to_string();
+        let mod_unique = name_split.next().unwrap_or_default().to_string();
+        // TODO: How does this work when we have a crate with multiple files
+        let crate_unique = name_split.next().unwrap_or_default().to_string();
+
+        let path_hash = tcx.def_path_hash(def_id);
+        //let key = NameHash::new(path_hash.stable_crate_id().to_u64(), path_hash.local_hash());
+        let key = path_hash.stable_crate_id().to_u64().to_string() + &path_hash.local_hash().to_string();
```
fpoli, line 88: Why not just using `item_name` as `key`?

fpoli:
> Instead, the interner should return a "name-token" for a given `def_id`

~~Could the "name-token" be just `ItemName(DefId)`?~~ I moved the question to the diff.

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -4,60 +4,130 @@
...
-        let symbol = readable_names.iter()
-            .find(|r| !self.used_symbols.contains(r.as_ref()))
-            .map(|r| r.as_ref())
-            .unwrap_or(full_unique_name);
-        debug!("Interning of {:?} is {:?}", full_unique_name, symbol);
-        self.used_symbols.insert(symbol.to_string());
-        self.name_to_symbol.insert(full_unique_name.to_string(), symbol.to_string());
-
-        symbol.to_string()
+    fn intern_name(&mut self, prefix: String, crate_name: String, crate_unique: String, mod_unique: String, short_name: String) -> ShortName {
```
fpoli, line 96: Is this signature going to change before the PR is ready? I think it's confusing that the method returns a `ShortName` that might be invalidated by future `intern_name` calls.

vir/src/common/identifier.rs
fpoli: Also, why not wrapping around `DefId` or, alternatively,  the full unique `item_name` string? The latter would be more stable than a counter when the order of the interned elements change.

prusti-viper/src/encoder/name_interner.rs
```diff
@@ -4,60 +4,130 @@
...
 use log::debug;
 
+// Tracks if crate-unqiue and module-unique names are needed to ensure uniqueness
+#[derive(Debug)]
+struct ShortName {
+    prefix: String,
+    project_unique: (bool, String),
+    crate_unique: (bool, String),
+    mod_unique: (bool, String),
+    short_name: String,
```
fpoli, line 22: `ShortName { path: Vec<String>, suffix_len: usize }` with `1 <= suffix_len <= path.len()` would be more generic and might even require less code.

prusti-viper/src/encoder/name_interner.rs
fpoli: (I think yes)

JonasAlaif:
Closing due to age, I'll try this again after the refactoring lands

PR #821, opened 2022-01-10
Extern trait specs

### LINKED ISSUE #779: Support trait's spec in `extern_spec`
Currently, the `extern_spec` attribute reject the specifications attached to a trait's function declaration:
```rust
// This will panics
#[extern_spec]
pub trait Max {
    #[ensures(result >= a && result >= b)]
    #[ensures(result == a || result == b)]
    fn max(a: i32, b: i32) -> i32;
}
```
This is because, in [rewrite](https://github.com/viperproject/prusti-dev/blob/e9a23ebf4d6b364cd9944e197ea906f57e3aaaea/prusti-specs/src/lib.rs#L442), it's only accepting `mod`, `fn` and `impl`. 

Since we might not need to rewrite the trait itself when it's external, simply returning the trait can be one option?

vl0w:
Provides extern spec support for traits as demonstrated in the following code snippet:

```rust
trait Op {
    fn op(&self, x: i32) -> i32;
}

#[extern_spec]
trait Op {
    
    #[requires(...)]
    #[ensures(...)]
    fn op(&self, x: i32) -> i32;
}
```

## Why is this needed?
In Prusti, we can attach specifications to trait implementations via the `#[refine_trait_spec]` macro.

Prusti obeys behavioral subtyping rules. That is:
- The impls precondition must be as least as weak as the traits precondition
- The impls postcondition must be as least as strong as the traits postcondition

If a trait has no explicit user-typed pre- and postcondition, this brings up a challenge: One cannot provide a precondition to the implementation of the trait.

This change allows one to specifically provide specifications for traits which are not user-typed and therefore cannot be specified directly.

## Implementation
The implementation of this feature is almost identical as extern specs on impl blocks.

Given an external trait spec for some trait `Foo`
```rust
#[extern_spec]
trait Foo {
    type AssocType;
    
    #[requires(...)]
    #[ensures(...)]
    fn foo(&self, ...) -> ...;
}
```
we generate an auxiliary struct `Aux` which typechecks the specification and serves as *lookup* definition for future method calls.

```rust
struct Aux<T_AssocType, T_Self> 
    where T_Self: Foo<AssocType = T_AssocType>
{
    // phantom data for generics
}
impl<...> Aux<...> where ... {
    
    #[requires(...)]
    #[ensures(...)]
    #[prusti::extern_spec]
    fn foo(_self: &T_Self, ...) -> ... {
        Foo::foo(_self); // <-- used when searching for specs of function calls
        unimplemented();
    }
}
```

For each associated type definition, a generic parameter on the auxiliary structure is generated.

Additionally, a generic parameter `T_Self` is added to `Aux`. `T_Self` represents some type which implements the to-be specified trait. We cannot directly implement the trait on `Aux`, because one might only add specifications for a subset of `Foo` methods.

## Remarks
- `where` clauses are explicitly not supported
One could add a `where Self: Bar` clause to the extern spec. This imposes challenges during the lookup of the specs for a method call and is out of scope of this PR.
- Default methods in extern trait specs are invalid
- Defaults for associated types in extern trait specs are invalid
- Generics are partially supported. Given a trait `MyTrait<T>`, we can either add an extern spec for a generic `T` or exactly _one_ concrete type (but not multiple)
- Maybe closes #779


prusti-specs/src/lib.rs
```diff
@@ -459,20 +456,23 @@ pub fn extern_spec(_attr: TokenStream, tokens:TokenStream) -> TokenStream {
...
+                extern_spec_rewriter::impls::rewrite_impl(&mut item_impl, Box::from(struct_ty))
             );
 
             quote_spanned! {item_span=>
                 #new_struct
                 #rewritten_item
             }
         }
+        syn::Item::Trait(item_trait) => {
+            handle_result!(extern_spec_rewriter::traits::handle_extern_spec_trait(&item_trait))
```
Aurel300: I wonder if the other `syn::Item::*` bits can have a corresponding `handle_*` method so this method becomes more readable.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
-            syn::GenericParam::Type(tp) => {
-                let ident = tp.ident.clone();
-                syn::parse_quote!{ ::core::marker::PhantomData<#ident> }
+/// Process external specifications in Rust modules marked with the
+/// #[extern_spec] attribute. Nested modules are processed recursively.
+/// Specifications are collected from functions and function stubs.
+///
+/// Modules are rewritten so that their name does not clash with the module
+/// they are specifying.
+pub mod mods {
```
Aurel300: Maybe `extern_specs` should be a folder with multiple files in it? It's a bit more readable than having inlined modules in one file.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -169,20 +29,95 @@ fn rewrite_self(tokens: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
...
+fn add_phantom_data_for_generic_params(item_struct: &mut syn::ItemStruct) {
+    item_struct.generics = item_struct.generics.clone();
+    let generics = &item_struct.generics;
+
+    let mut fields: Vec<syn::Field> = Vec::new();
+    for param in generics.params.iter() {
+        let field: ParsableUnnamedField = match param {
+            syn::GenericParam::Type(tp) => {
+                let ident = tp.ident.clone();
+                syn::parse_quote! { ::core::marker::PhantomData<#ident> }
```
Aurel300: Just wondering—do you really need to get a `syn` type here? Why not just `quote!` to get a `proc_macro2::TokenTree`, since in the end you are only using this to do `#(#fields),*`, i.e. tokenising. Then the `ParsableUnnamedField` type would not be necessary.

(Using `quote_spanned!` would also allow you to preserve the span of the original generics, although I don't think these fields should really appear in error messages.)

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -169,20 +29,95 @@ fn rewrite_self(tokens: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
...
+}
+
+/// We take the Generics (parameters) defined with the `#[extern_spec] impl<...>` (the `<...>`)
+/// but then need to pass those as arguments: `SomeStruct<...>`. This function translates from
+/// the syntax of one to the other; e.g. `<T: Bound, 'l: Bound, const C: usize>` -> `<T, 'l, C>`
+pub fn rewrite_generics(gens: &syn::Generics) -> syn::AngleBracketedGenericArguments {
+    let args: Vec<syn::GenericArgument> = gens
+        .params
+        .clone()
+        .iter()
```
Aurel300: If you are cloning here, you can use `.into_iter` and maybe avoid the `clone` in the `Lifetime` case?

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+                                new_tokens.extend(
+                                    Group::new(proc_macro2::Delimiter::Brace, TokenStream::new())
+                                        .to_token_stream(),
+                                );
+                                continue;
+                            }
+                        }
+                        new_tokens.extend(token.to_token_stream());
+                    }
+                    let res: Result<syn::Item, _> = syn::parse2(new_tokens);
```
Aurel300: I wonder if we could instead try to parse as `syn::TraitItemMethod`.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
-struct ParsableUnnamedField(syn::Field);
-impl syn::parse::Parse for ParsableUnnamedField {
-    fn parse(input: syn::parse::ParseStream<'_>) -> syn::parse::Result<Self> {
-        Ok(ParsableUnnamedField( syn::Field::parse_unnamed(input)? ))
+pub mod impls {
+    use super::*;
+
+    pub fn generate_new_struct(item_struct: &syn::ItemImpl) -> syn::Result<syn::ItemStruct> {
+        let name_generator = NameGenerator::new();
+        let struct_name = match name_generator.generate_struct_name(item_struct) {
```
Aurel300: Maybe not in scope for this PR, but `NameGenerator::generate_struct_name` should not be returning a `Result`. The [implementation](https://github.com/viperproject/prusti-dev/blob/f7f3e1b27e678c28c078c1be0796ab974dfc7c62/prusti-specs/src/specifications/common.rs#L142-L144) should handle any kind of `syn::Type` and stringify it (somehow) instead of errorring out. This might be one of the reasons why extern specs on slices don't work @JonasAlaif ?

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+    /// The result of this rewriting is then parsed in `ExternSpecResolver`.
+    pub fn rewrite_impl(
+        impl_item: &mut syn::ItemImpl,
+        new_ty: Box<syn::Type>,
+    ) -> syn::Result<TokenStream> {
+        let item_ty = &mut impl_item.self_ty;
+        if let syn::Type::Path(type_path) = item_ty.as_mut() {
+            for seg in type_path.path.segments.iter_mut() {
+                if let syn::PathArguments::AngleBracketed(genargs) = &mut seg.arguments {
+                    genargs.colon2_token = Some(syn::token::Colon2::default());
```
Aurel300:
```suggestion
                    genargs.colon2_token = Some(syn::Token![::].default());
```

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+
+                    method
+                        .attrs
+                        .push(parse_quote_spanned!(item_span=> #[prusti::extern_spec]));
+                    method
+                        .attrs
+                        .push(parse_quote_spanned!(item_span=> #[trusted]));
+
+                    let mut method_path: syn::ExprPath = parse_quote_spanned! {ident.span()=>
+                        #item_ty :: #ident
```
Aurel300: If we want to support slices here I think we need to wrap this in brackets:

```suggestion
                        < #item_ty > :: #ident
```

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+
+                    method.block = parse_quote_spanned! {item_span=>
+                        {
+                            #method_path (#args);
+                            unimplemented!()
+                        }
+                    };
+                }
+                _ => {
+                    return Err(syn::Error::new(
```
Aurel300: I assume this would be an error of our implementation, never the user, so we can panic instead of having a `Result` return type.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+    /// ```rust
+    /// #[extern_spec]
+    /// trait SomeTrait where Self: OtherTrait {
+    ///     type ArgTy;
+    ///     type RetTy;
+    ///
+    ///     fn foo(&self, arg: Self::ArgTy) -> Self::RetTy;
+    /// }
+    /// ```
+    /// it produces a s struct
```
Aurel300:
```suggestion
    /// it produces a struct
```

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+    /// Returns an error when the macro was used in a wrong way on the trait
+    fn validate_macro_usage(item_trait: &syn::ItemTrait) -> syn::Result<()> {
+        // Default methods not allowed
+        for item in item_trait.items.iter() {
+            if let syn::TraitItem::Method(item_method) = item {
+                if item_method.default.is_some() {
+                    return Err(syn::Error::new(
+                        item_method.default.as_ref().unwrap().span(),
+                        "Default methods in external trait specs are disallowed \
+                        because external specs are always trusted",
```
Aurel300: I would leave out this second line of the justification.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+                    ));
+                }
+            }
+        }
+
+        // Generics not supported
+        if item_trait.generics.params.len() > 0 {
+            return Err(syn::Error::new(
+                item_trait.span(),
+                "Generics in external trait specs are not supported",
```
Aurel300: Why? Is it not just a matter of parsing the list on the declaration, stripping out the bounds(?), and prepending it `where TSelf: SomeTrait<(here), AssocT1 = ...>`?

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+        let trait_impl = generated_struct.generate_impl();
+        let new_struct = generated_struct.generated_struct;
+        Ok(quote_spanned! {item_trait.span()=>
+                #new_struct
+                #trait_impl
+        })
+    }
+
+    /// Returns an error when the macro was used in a wrong way on the trait
+    fn validate_macro_usage(item_trait: &syn::ItemTrait) -> syn::Result<()> {
```
Aurel300: Might be nicer to do this check in the place where methods are actually dealt with, especially if the generics limitation is lifted. This way the logic for what constitutes valid usages is not split into multiple places.

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+            ));
+        }
+
+        return Ok(());
+    }
+
+    /// Given a trait declaration, returns the bounds to `Self` in the where clause
+    /// # Example
+    /// ```
+    /// trait Foo where Self: Bar<Baz> + Qux, Self: Quux
```
Aurel300: Does Rust accept things like `where Foo<Self>: Bar`? If so, I guess the bounds should more generally be copied over with `Self` replaced with `TSelf` for the wrapper struct?

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
-        syn::parse2::<syn::GenericArgument>(ts).unwrap()
-    }).collect();
-    syn::parse_quote!{ < #(#args),* > }
+        // Find associated types in trait
+        let mut assoc_types_to_generics_map = HashMap::new();
+        let associated_type_decls = get_associated_types(item_trait);
+        for associated_type_decl in associated_type_decls {
+            let associated_type_ident = &associated_type_decl.ident;
+            let generic_ident = syn::Ident::new(
+                format!("T_AssocType_{}", associated_type_ident).as_str(),
```
Aurel300: Unlikely, but this (and `T_Self` more so) might still have a name clash with user-declared types. `PrustiT_...`?

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+
+        add_phantom_data_for_generic_params(&mut new_struct);
+
+        let gen = GeneratedStruct {
+            generated_struct: new_struct,
+            item_trait,
+            assoc_types_to_generics_map,
+            self_type_param_ident,
+        };
+        Ok(gen)
```
Aurel300:
```suggestion
        Ok(GeneratedStruct {
            generated_struct: new_struct,
            item_trait,
            assoc_types_to_generics_map,
            self_type_param_ident,
        })
```

prusti-specs/src/extern_spec_rewriter.rs
```diff
@@ -201,76 +136,470 @@ fn rewrite_method_inputs(item_ty: &syn::Type, method: &mut ImplItemMethod) ->
...
+                }
+            };
+        }
+    }
+
+    fn get_associated_types(item_trait: &syn::ItemTrait) -> Vec<&syn::TraitItemType> {
+        let mut result = Vec::new();
+        for trait_item in item_trait.items.iter() {
+            if let syn::TraitItem::Type(assoc_type) = trait_item {
+                // TODO: How to handle associated types with defaults?
```
Aurel300: I think specifying a default for an associated type in an `extern_spec` block is probably a user error.

prusti-specs/src/extern_spec_rewriter.rs
vl0w: where clauses are disabled entirely now

PR #826, opened 2022-01-18
Update to the latest Viper pre-release

fpoli:
Make the bot update to the latest Viper version, *including pre-releases*.


fpoli:
Updating to the latest Viper version currently fails because `VerificationError` no longer has a direct `counterexample` field. It has been moved into `failureContexts`.

I'll make the script update the viper version anyway. As soon as someone has time and knowledge to fix that, we can actually commit the upgrade to master.

PR #831, opened 2022-01-21
Specification encoding refactor

Aurel300:
## Summary of changes

- Syntax in specifications is re-entrant: Prusti and Rust syntax can be mixed.
- Pre-typing encoding of specifications is simplified, no JSON or Prusti AST, instead specifications are embedded directly into Rust AST.
- Small improvements to quantifier-related error messages.
- `after_expiry_if` removed (untested, undocumented).

Thanks @JustinHuPrime (and @alexanderjsummers) for prototyping the re-entrant parser.


vir/src/legacy/ast/expr.rs
```diff
@@ -1140,11 +1140,11 @@ impl Expr {
...
             | Expr::Variant(_, Field { ref typ, .. }, _)
             | Expr::Field(_, Field { ref typ, .. }, _)
-            | Expr::AddrOf(_, ref typ, _)
-            | Expr::LetExpr(LocalVar { ref typ, .. }, _, _, _) => typ,
+            | Expr::AddrOf(_, ref typ, _) => typ,
             Expr::LabelledOld(_, box ref base, _)
             | Expr::Unfolding(_, _, box ref base, _, _, _)
-            | Expr::UnaryOp(_, box ref base, _) => base.get_type(),
+            | Expr::UnaryOp(_, box ref base, _)
+            | Expr::LetExpr(_, _, box ref base, _) => base.get_type(),
```
Aurel300, line 1147: This one was annoying to find. @vakaras are `let`s assigned the correct type (i.e. the type of the `base` expression, *not* the type of the bound variable) in the new layers?

vakaras: Shouldn't the types match exactly? We probably should add an assert that ensures that.

prusti-specs/src/rewriter.rs
```diff
@@ -105,90 +99,127 @@ impl AstRewriter {
...
+        self.generate_spec_item_fn(
+            spec_type,
+            spec_id,
+            parse_prusti(tokens)?,
+            item
+        )
+    }
+/*
+    /// Parse a pledge without lhs into a Rust expression
+    pub fn process_pledge_rhs_only(
```
vakaras: Why is this commented out? I would prefer to not have commented out code.

vir/src/legacy/ast/expr.rs
Aurel300: `let foo == 1 in (foo + 1 == 2)`: the let-expression has type `Bool`, but the bound variable is `Int`.

prusti-tests/tests/verify/fail/quantifiers/triggers.rs
```diff
@@ -28,7 +28,7 @@ fn dummy(_n: isize, _res: isize) -> bool { true }
 
 #[requires(
     // This is an unsound quantifier to check whether triggerring works.
-    forall(|n: isize, res: isize| fib(n) == res, triggers=[(dummy(n, res),)])
+    forall(|n: isize, res: isize| fib(n) == res, triggers=[(dummy(n, res))])
```
vakaras: This change looks wrong: triggers are vectors of tuples. How it is recognized that `(dummy(n, res))` is a tuple?

prusti-viper/src/encoder/mir/pure/specifications/encoder.rs
```diff
@@ -1,824 +1,373 @@
...
-                        upper_bound,
-                    ));
-                }
-            } else if config::encode_unsigned_num_constraint() {
-                if let ty::TyKind::Uint(_) = ty.kind() {
-                    let expr =
-                        vir_high::Expression::less_equals(0u8.into(), encoded_arg.clone().into());
-                    bounds.push(expr);
-                }
+// ================== high ==================
```
vakaras: Could you please split this into separate files because the comments are hard to notice?

prusti-viper/src/encoder/mir_encoder/mod.rs
```diff
@@ -50,6 +48,15 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
         format!("{:?}", local)
     }
 
+    fn encode_local_high(&self, local: mir::Local) -> SpannedEncodingResult<vir_crate::high::VariableDecl> {
+        let var_name = self.encode_local_var_name(local);
```
vakaras, line 53: Why not using:

```suggestion
self.encoder().encode_local_high(self.mir, local)
```

vakaras (approved):
LGTM

vir/src/legacy/ast/expr.rs
vakaras: Oh, I see.

prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
```diff
@@ -210,7 +180,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
         trace!("apply_terminator {:?}, states: {:?}", term, states);
         use rustc_middle::mir::TerminatorKind;
         let span = term.source_info.span;
-        let location = self.mir.terminator_loc(bb);
+        let _location = self.mir.terminator_loc(bb);
```
fpoli: Remove?

prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
```diff
@@ -746,7 +744,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
     ) -> Result<(), Self::Error> {
         trace!("apply_statement {:?}, state: {}", stmt, state);
         let span = stmt.source_info.span;
-        let location = mir::Location {
+        let _location = mir::Location {
```
fpoli: Remove?

prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
```diff
@@ -830,7 +828,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                                     let (encoded_rhs, is_value) = self.encode_operand(operand).with_span(span)?;
                                     if is_value {
```
fpoli, line 864: It is now true that in this file that all pure function arguments are passed as values (i.e. primitive values or snapshot values)?

prusti-tests/tests/verify/fail/quantifiers/triggers.rs
Aurel300: I'll undo the change and allow trailing-comma syntax. However, `triggers=[(foo)]` will also be recognised as the singular trigger set with `foo` as its one expression. This is not parsed by Rust or `syn`, so the comma is optional even for one-expression trigger sets.

vakaras: Sounds good.

prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
Aurel300: I think yes, so the second element of the `encode_operand` tuple should be removable.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -1061,23 +986,6 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
-            let variant_field = if let ty::TyKind::Adt(adt_def, _) = place_ty.kind() {
-                let tcx = self.encoder.env().tcx();
-                let variant_name = adt_def.variants[variant_idx].ident(tcx).to_string();
-                self.encoder.encode_enum_variant_field(&variant_name)
-            } else {
-                unreachable!()
-            };
-            downcast_stmts.push(vir::Stmt::Downcast( vir::Downcast {base: encoded_place, field: variant_field} ));
-        }
-
```
fpoli, line 1080: This looks suspicious to me. Without this, are fold-unfold statements for variants of enums generated correctly?

prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
fpoli: Then, I think that the `state.substitute_value(&encoded_lhs, snapshot);` at (new) line 846 should be enough to encode this case. That is, the `state.substitute_value(&field_place, encoded_rhs.clone());` should be redundant because `encoded_lhs` is always going to be a proper prefix of `field_place`. The same should hold for many other cases (`AggregateKind::Adt` etc).

PR #835, opened 2022-01-27
Extern trait implementation specs

vl0w:
Adds support for extern trait impl specifications:
```rust
#[extern_spec]
    type AssocType = i32;
    //...
}
```

## Supported
* Traits with associated types

## Not supported
* Generics


vl0w:
We need to make sure that we can handle code like the following:
```rust
trait A {
    fn a(&self);
}

struct Foo;
impl Foo {
    fn a(&self) {println!("Foo::a()")}
}
impl A for Foo {
    fn a(&self) {println!("<Foo as A>::a()")}
}

fn main() {
    let foo = Foo{};
    foo.a();
    <Foo as A>::a(&foo);
}
```
i.e. shadowing of trait methods.

vl0w:
Shadowing works as expected (tests are enclosed in last commit).

Please note that there is a change how we specify extern specs for trait implementations.
Before, it was possible to attach extern specifications like this:
```rust
trait A { fn a(); }
struct S;
impl A for S { 
    fn a() { ... }
}
#[extern_spec]
impl S {
    // the specs
    fn a();
```

The extern spec above provided a specification for the method `A::a` implemented on `S`. This causes an error in Prusti now, the extern spec needs to be declared explicitly for the trait implementation:
```rust
#[extern_spec]
impl A for S {
    // the specs
    fn a();
}
```

prusti-specs/src/specifications/common.rs
```diff
@@ -127,38 +127,3 @@ fn generate_name_for_type(ty: &syn::Type) -> Option<String> {
         }
     }
 }
-
-
-#[cfg(test)]
-mod tests {
```
Aurel300: @vl0w FYI I deleted this: I didn't want regex as a dependency (even as `dev-dependency` it somehow shows up in `Cargo.lock` of the cargo test…), and this test isn't particularly useful. If we really want to test stability of the mangled names, it can be a UI test.

vl0w: That's ok. Afaik the other tests (also in the other PR) don't use Regex.

Issue #529, opened 2021-06-01
Publish prusti-contracts on crates.io


fpoli:
As [discussed on Zulip](https://prusti.zulipchat.com/#narrow/stream/236614-general/topic/using.20cargo-prusti.20for.20generating.20executable.20code.20from.20an.2E.2E.2E), publishing `prusti-contracts` on crates.io would make it much easier to add contracts to existing projects.

Some requirements:
* Since for the moment we don't want to provide backward-compatibility guarantees for `prusti-contracts`, we should publish the crate using just `0.x.y` versions.
* At verification time, Prusti should check that the version of `prusti-contracts` used by the crate is compatible with the given build of Prusti. For example, when compiling Prusti we can store the version of `prusti-contracts` in a static variable and then check at runtime that it's equal to the one used by the crate to be verified.
* It would be nice to automatically publish `prusti-contracts` versions on crates.io, at the end of the Prusti release workflow (CD).

Are there more requirements to be aware of?

PR #836, opened 2022-01-28
Type models

vl0w:
# Summary
Support for models of arbitrary (local, non-local) types. A model represents an abstraction over a type that can be used in specifications.

# Motivation
Some structs, mostly non-local structs from other crates, contain non-public code which are essential for verification.
An example is the `std::slice::Iter` struct from the standard library. We would like to provide a specification for the `Iterator` implementation of `Iter`:
```rust
impl Iterator<'a, T> for std::slice::Iter<'a, T>{
    // spec???
    fn next(&mut self) -> Option<&'a T>;
}
```
There is not really a way to provide a concise specification for this iterator implementation, since `Iter`'s fields [do not allow a straightforward](https://doc.rust-lang.org/src/core/slice/iter.rs.html#65-71) specification.

We can instead provide a *model* `Iter` in the following way: The contained data and the current position in the iteration:
```rust
#[model]
struct std::slice::Iter<'a, T> where T: 'a {
    position: usize,
    data: &'a [T],
}
```
This model then can be used in the specification (just an outline, actual specs would look different):
```rust
impl Iterator<'a, T> for std::slice::Iter<'a, T>{
    #[ensures( self.model().data.len == old(self.model().data.len)) )]
    #[ensures( old(self.model().pos) < self.model().data.len ==> (result.is_some() && result.unwrap() == self.model().data[old(self.model().pos)]))]
    // ... more ...
    fn next(&mut self) -> Option<&'a T>;
}
```

# Implementation
This PR adds the following things:
- A procedural macro `#[model]` -> Please refer to the [code docs](https://github.com/vl0w/prusti-dev/blob/type-models/prusti-specs/src/type_model/mod.rs) and the [tests](https://github.com/vl0w/prusti-dev/blob/type-models/prusti-specs/src/type_model/mod.rs#L199) to see how it works
- An extension to `SpecChecker` to ensure that the `.model()` method gets only called in `#[prusti::spec_only]` annotated code.


prusti-contracts-internal/src/lib.rs
```diff
@@ -52,3 +52,40 @@ pub fn extern_spec(attr: TokenStream, tokens: TokenStream) -> TokenStream {
...
+/// ```
+/// ## Caution
+/// The model is only to be used in specifications (pre- and postconditions of functions or methods)
+/// and never in code which will be executed. Using `.model()` will cause a panic during runtime.
+///
+/// ## Remarks
+/// * Specifications involving models can only be used in trusted functions or methods, i.e. on
+///   explicitly `#[trusted]` methods or on `#[extern_spec]`s.
+/// * Models can not be generic, i.e. creating a model for `Vec<T>` is not possible. It is however
+///   possible to model concrete generic types, i.e. `Vec<i32>` and `Vec<u32>`
```
Aurel300: Good documentation, but this would probably fit better in the user guide. I'm not sure docs show up correctly on Prusti attributes in the IDE. If yes, then it should be a short explanation + a link to the user guide.

prusti-contracts/src/lib.rs
```diff
@@ -78,8 +81,10 @@ mod private {
     /// A macro for defining a predicate using prusti expression syntax instead
     /// of just Rust expressions.
     pub use prusti_contracts_internal::predicate;
-}
 
+    /// Macro for creating type [models](prusti_contracts_internal::model).
```
Aurel300:
```suggestion
    /// Macro for creating type models.
```
(Do re-export docs do anything in the IDE?)

prusti-interface/src/specs/checker.rs
```diff
@@ -126,12 +131,76 @@ impl<'v, 'tcx> Visitor<'tcx> for CheckPredicatesVisitor<'v, 'tcx> {
...
+            if let Some(local_def_id) = maybe_local_def_id {
+                let decl_hir_id = self.tcx.hir().local_def_id_to_hir_id(local_def_id);
+                return Some(decl_hir_id);
+            }
+        }
+        None
+    }
+
+    fn is_embedded_in_spec(&self, hir_id: hir::HirId) -> bool {
+        for (parent_local_def_id, _) in self.tcx.hir().parent_owner_iter(hir_id) {
```
Aurel300: Why is the approach here different from `CheckPredicatesVisitor`? That is, why do we iterate over parents for calls to model methods rather than simply not recursing into spec functions in the first place?

prusti-interface/src/utils.rs
```diff
@@ -307,6 +307,10 @@ pub fn has_extern_spec_attr(attrs: &[ast::Attribute]) -> bool {
     has_prusti_attr(attrs, "extern_spec")
 }
 
+pub fn has_model_attr(attrs: &[ast::Attribute]) -> bool {
+    has_prusti_attr(attrs, "model_generator")
```
Aurel300: I don't like this name, we might be using Rust generators to model some things soonish :)

prusti-specs/src/type_model/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        for param in item_struct.generics.params.iter() {
+            if let syn::GenericParam::Type(ty_param) = param {
+                name.push_str(ty_param.ident.to_string().as_str());
+            }
+        }
+
+        GeneratedIdents {
+            model_struct_ident: Ident::new(
+                format!("Prusti{}Model", name).as_str(),
```
Aurel300: These should probably get some UUIDs to avoid conflicts.

prusti-specs/src/type_model/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+    let idents = GeneratedIdents::generate(&item_struct);
+    let model_struct = create_model_struct(&item_struct, &idents)?;
+    let to_model_trait = create_to_model_trait(&item_struct, &model_struct, &idents);
+    let model_impl = create_model_impl(&item_struct, &model_struct, &to_model_trait)?;
+
+    Ok(TypeModel {
+        model_struct,
+        model_impl,
+        to_model_trait,
+    })
```
Aurel300:
```suggestion
    Ok(TypeModel {
        model_struct: create_model_struct(&item_struct, &idents)?,
        model_impl: create_model_impl(&item_struct, &model_struct, &to_model_trait)?,
        to_model_trait: create_to_model_trait(&item_struct, &model_struct, &idents),
    })
```

prusti-specs/src/type_model/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+) -> TypeModelGenerationResult<syn::ItemStruct> {
+    if item_struct.fields.is_empty() {
+        return Err(TypeModelGenerationError::MissingStructFields(
+            item_struct.span(),
+        ));
+    }
+
+    let model_struct_ident = &idents.model_struct_ident;
+    let mut model_struct: syn::ItemStruct = parse_quote_spanned! {item_struct.span()=>
+        #[derive(Copy, Clone)]
```
Aurel300: `PartialEq` and `Eq` as well?

prusti-specs/src/type_model/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    /// Thrown when using a const type param in the model
+    ConstParamDisallowed(proc_macro2::Span),
+}
+
+impl std::convert::From<TypeModelGenerationError> for syn::Error {
+    fn from(err: TypeModelGenerationError) -> Self {
+        match err {
+            TypeModelGenerationError::MissingStructFields(span) => {
+                syn::Error::new(span, "Missing fields in model")
```
Aurel300: This error message is not great: it could be misunderstood as saying there are some *particular* fields which should be in the model. How about "Type model must have at least one field"?

prusti-tests/tests/typecheck/fail/type-model/generic-param.rs
```diff
@@ -0,0 +1,12 @@
+extern crate prusti_contracts;
+use prusti_contracts::*;
+
+use std::vec::Vec;
+
+#[model]
+struct Vec<T> { //~ ERROR: cannot find type `T` in this scope [E0412]
```
Aurel300: So the generics here are the instantiations? Is it not possible to define a type model for a generic type?

prusti-contracts-internal/src/lib.rs
vl0w: I can at confirm that they don't show up in my IDE (clion). 
Maybe this becomes more relevant when `prusti_contracts` will be distributed crate on crates.io (#529)
For now, I will simply add this documentation to the user guide :+1:

prusti-tests/tests/typecheck/fail/type-model/generic-param.rs
vl0w: Currently it is not possible to define a model for a generic type.
It is, however, possible to define models for different concrete types (see [this test](https://github.com/vl0w/prusti-dev/blob/650ac711ca2e1af82144cdf560086e00a27954b1/prusti-tests/tests/verify/pass/type-model/models-for-different-monomorphizations.rs)).

To support generic type parameters as well, we would need annotations whether we refer to a _concrete_ or _generic_ type as we did in extern trait specs (e.g. [here](https://github.com/vl0w/prusti-dev/blob/extern-trait-specs/prusti-tests/tests/verify/pass/extern-spec/traits/generics-1.rs) or [here](https://github.com/vl0w/prusti-dev/blob/extern-trait-specs/prusti-tests/tests/verify/pass/extern-spec/traits/generics-2.rs))

prusti-specs/src/type_model/mod.rs
vl0w: Your suggestion is good :+1: 

If I recall correctly, there is no technical reason for this error. Imo, it just makes no sense to have a model without fields.

prusti-interface/src/utils.rs
vl0w: This is a simple marked used to check that `.model()` is called from spec code only. Suggestion for new name: `to_model_fn`?

prusti-tests/tests/typecheck/fail/type-model/generic-param.rs
vl0w: It might be necessary to add this feature, because we want to have a generic model for `std::iter::Iter`

prusti-contracts-internal/src/lib.rs
vl0w: PR #865 created

prusti-interface/src/utils.rs
Aurel300: `to_model_fn` or even just `to_model` should be good.

prusti-interface/src/specs/checker.rs
vl0w: Good idea, I guess this will simplify the code

vl0w: I did a somewhat bigger refactoring of `SpecChecker`. The checker for type models now behaves like `CheckPredicatesVisitor`.

prusti-specs/src/type_model/mod.rs
vl0w: Simply deriving those two traits will have no immediate effect since we cannot assume a specification for `PartialEq::eq`.
We could generate the specification during the macro expansion, but it might be overkill?

Aurel300: IIRC we treat types with derived `PartialEq` and `Eq` specially, because we use [`has_structural_eq_impls`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.has_structural_eq_impls)?

prusti-tests/tests/verify/pass/type-model/models-for-same-type-unsound.rs
```diff
@@ -0,0 +1,35 @@
+// This test instantiates two instances of `A` and checks that their models are independent.
+// Notice that `A` has no fields, which introduces a soundness issues.
+// The reason for this behavior is that the encoding of `A` (w/o fields) in Viper
+// assumes that all instances are equal, i.e. there is no notion of inequality for two `A`s.
+// Hence, we cannot refer to different models for different instances of `A`.
+// If `A` had fields, the code verifies as expected.
+//
+// This code will print a warning to a user.
```
Aurel300: If this test is meant to check that the warning is printed, we can put it into `verify(_overflow)/ui`.

vl0w: There is a separate test which checks the warning in `verify/ui`.
This test is meant as a "documentation" of the behavior. It is maybe not needed, but I found it somewhat useful. Mentioning it in the user docs might be good enough though.

Aurel300: Ok, then I would delete this and only keep the `ui` test. As it is, it is just testing the same thing again, without checking the warning. Oh also, by default tests should be in `verify_overflow`, as that is the default in Prusti.

vl0w: Test deleted, other tests moved

prusti-contracts-internal/src/lib.rs
vl0w: I incorporated your feedback in the original PR

prusti-tests/tests/typecheck/fail/type-model/generic-param.rs
vl0w: This feature should be implemented by [e76b3ac](https://github.com/viperproject/prusti-dev/pull/836/commits/e76b3ac6f4a0791f2601cb0a1fe340a7f78eda7e)

Aurel300:
Let's merge this now. Overall I think there are some things to simplify, e.g. `model` as a predicate function to avoid the duplicate logic that checks if models/predicates are used in non-spec code, but let's do that in a future PR.

bors r+

Issue #815, opened 2021-12-31
[Prusti internal error] `Consistency error: Local variable _5 not found` inside a pure function


Pointerbender:
The following Rust program fails verification with error `[Prusti internal error] consistency error in A::test: Consistency error: Local variable _5 not found. (@0.0)`.

```rust
use prusti_contracts::*;

#[derive(Clone, Copy)]
pub struct A {
    inner: isize,
}

impl A {
    #[pure]
    #[requires(num < isize::MAX as usize)]
    pub fn new(num: usize) -> Self {
        Self {
            inner: num as isize + isize::MIN,
        }
    }

    #[pure]
    pub fn test(&self) {
        let _ = Self::new(2);
    }
}

fn main() {}
```

This error is found in the resulting Viper program for `A::test` in the pure function for `A::new`:

```
function m_new__$TY$__(_1: Int): Snap$struct$m_A
  requires true
  requires _1 < builtin$cast$isize$usize__$TY$__(9223372036854775807)
  requires 0 <= _1
  requires _1 <= 18446744073709551615
  ensures true
  ensures [result == mirror_simple$m_new__$TY$__(_1), true]
{
  (
    !(
      builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 < -9223372036854775808
      || builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 > 9223372036854775807
    ) ? cons$0$__$TY$__Snap$struct$m_A$(_5.tuple_0.val_int) : builtin$unreach__$TY$__Snap$struct$m_A$()
  )
}
```

where it tries to access `_5.tuple_0.val_int` (the result of the checked addition `num as isize + isize::MIN`), but `_5` is not in the scope of the function. In fact, `_5` is not declared anywhere in the Viper program for `A::test` (it seems to have spilled over from the Viper program for `A::new` somehow, where it is declared inside `method m_new()`).

fpoli:
Thanks for the report! This seems a bug in the replacements done to encode pure code. Probably

https://github.com/viperproject/prusti-dev/blob/e9edf18d1c6fa99cb4eaac1180d67e3837a661c1/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L224-L226

or 

https://github.com/viperproject/prusti-dev/blob/e9edf18d1c6fa99cb4eaac1180d67e3837a661c1/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L992-L994

[...]

PR #837, opened 2022-01-29
Fix for first example of issue #815, plus regression test case

Pointerbender:
The `PureFunctionBackwardInterpreter` contained a bug during the `CheckedBinaryOp` case:

https://github.com/viperproject/prusti-dev/blob/e9edf18d1c6fa99cb4eaac1180d67e3837a661c1/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L983-L994

Here accessing the `value_field_value`/`check_field_value` of the respective fields is not necessary and this is causing the substitution to fail. The "new" encoder already seems to handle this correctly:

https://github.com/viperproject/prusti-dev/blob/9a30632bec36724fbef5e0d82526acfce15991dd/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L213-L222

And now `PureFunctionBackwardInterpreter` does, too :)

I also added a regression test for good measure. This PR does not close the related issue just yet, but already fixes one of the two known cases.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-815-1.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-815-1.rs
@@ -1,0 +1,23 @@
+use prusti_contracts::*;
+
+#[derive(Clone, Copy)]
+pub struct A {
+    inner: isize,
+}
+
+impl A {
+    #[pure]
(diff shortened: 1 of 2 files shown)


vakaras (commented):
LGTM. Thank you!

PR #841, opened 2022-01-31
Accept -> as an implication.

vakaras:
Accept `->` as an implication because it is used by many other Rust tools.


Aurel300 (approved):

Aurel300:
@vakaras We should document this also.

PR #844, opened 2022-02-01
Fold and walk triggers in quantifiers

### LINKED ISSUE #815: [Prusti internal error] `Consistency error: Local variable _5 not found` inside a pure function

Aurel300:
Closes #815.

### DIFF
--- a/prusti-tests/tests/verify/pass/issues/issue-815-1.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-815-1.rs
@@ -1,0 +1,23 @@
+use prusti_contracts::*;
+
+#[derive(Clone, Copy)]
+pub struct A {
+    inner: isize,
+}
+
+impl A {
+    #[pure]
+    #[requires(num < isize::MAX as usize)]
+    pub fn new(num: usize) -> Self {
+        Self {
+            inner: num as isize + isize::MIN,
+        }
+    }
+
+    #[pure]
+    pub fn test(&self) {
+        let _ = Self::new(2);
+    }
+}
+
+fn main() {}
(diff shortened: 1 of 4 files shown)


Pointerbender:
Ahhh, it was due to the trigger folding (+walking) after all :-) Many thanks for fixing this one!!

PR #845, opened 2022-02-02
Dump the Viper program even on a successful cache hit

### LINKED ISSUE #843: Cached programs do not dump their Viper encoding
When using the Prusti Server with `PRUSTI_DUMP_VIPER_PROGRAMS=true`, cached programs do not write their Viper encoding to disk. We might want that to happen anyway.

JonasAlaif:
Fixes #843 
The fix is ugly so if there is an obviously better way to do things let me know

### DIFF
--- a/prusti-server/src/process_verification.rs
+++ b/prusti-server/src/process_verification.rs
@@ -16,23 +16,26 @@ pub fn process_verification_request_cache<'v, 't: 'v>(
     cache: impl Cache,
 ) -> viper::VerificationResult {
     if !config::enable_cache() || config::print_hash() {
-        process_verification_request(verification_context, request)
+        process_verification_request::<true>(verification_context, request)
     } else {
         let hash = request.get_hash();
         info!("Verification request hash: {}", hash);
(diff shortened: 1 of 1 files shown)


prusti-server/src/process_verification.rs
```diff
@@ -16,23 +16,26 @@ pub fn process_verification_request_cache<'v, 't: 'v>(
...
-        let result = process_verification_request(verification_context, request);
+        let result = process_verification_request::<true>(verification_context, request);
         // Save `result` to cache
         cache.insert(hash, result.clone());
         result
     }
 }
 
-pub fn process_verification_request<'v, 't: 'v>(
+pub fn process_verification_request<'v, 't: 'v, const ACTUALLY_VERIFY: bool>(
```
Aurel300, line 38: Why not a normal argument? It doesn't look like the compilation should be affected much…

fpoli:
I think merging the two `process_verification_request` and `process_verification_request_cache` might be cleaner. I can give it a try myself.

PR #846, opened 2022-02-03
Dump viper on cache hit (version 2)

### LINKED ISSUE #843: Cached programs do not dump their Viper encoding

fpoli:
Fixes https://github.com/viperproject/prusti-dev/issues/843.

An alternative to https://github.com/viperproject/prusti-dev/pull/845.

### DIFF
--- a/prusti-server/src/process_verification.rs
+++ b/prusti-server/src/process_verification.rs
@@ -16,45 +16,52 @@ pub fn process_verification_request<'v, 't: 'v>(
     cache: impl Cache,
 ) -> viper::VerificationResult {
     let ast_utils = verification_context.new_ast_utils();
-    ast_utils.with_local_frame(16, || {
-        let hash = request.get_hash();
-        info!("Verification request hash: {}", hash);
 
-        let build_or_dump_viper_program = || {
-            let mut stopwatch = Stopwatch::start("prusti-server", "construction of JVM objects");
-            let ast_factory = verification_context.new_ast_factory();
-            let viper_program = request.program.to_viper(&ast_factory);
+    let hash = request.get_hash();
+    info!("Verification request hash: {}", hash);
 
-            if config::dump_viper_program() {
-                stopwatch.start_next("dumping viper program");
-                dump_viper_program(&ast_utils, viper_program, request.program.get_name());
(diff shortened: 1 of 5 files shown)


prusti-server/src/process_verification.rs
```diff
@@ -10,67 +10,70 @@ use prusti_common::{config, report::log::report, vir::ToViper, Stopwatch};
...
+
+        // Print the hash and skip verification. Used for testing.
+        if config::print_hash() {
+            println!(
+                "Received verification request for: {}",
+                request.program.get_name()
+            );
+            println!("Hash of the request is: {}", hash);
+            // Some tests need the dump to report a diff of the Viper programs.
+            let _ = build_or_dump_viper_program();
```
JonasAlaif: There is no need to run this if `!config::dump_viper_program()`

prusti-server/src/process_verification.rs
```diff
@@ -10,67 +10,70 @@ use prusti_common::{config, report::log::report, vir::ToViper, Stopwatch};
...
+pub fn process_verification_request<'v, 't: 'v>(
     verification_context: &'v VerificationContext<'t>,
     request: VerificationRequest,
     cache: impl Cache,
 ) -> viper::VerificationResult {
-    if !config::enable_cache() || config::print_hash() {
-        process_verification_request(verification_context, request)
-    } else {
+    let ast_utils = verification_context.new_ast_utils();
+    ast_utils.with_local_frame(16, || {
```
JonasAlaif: I'm not sure how expensive `with_local_frame` is? If it's cheap then this is fine, but if not then it's problematic for the common case where `!config::dump_viper_program()` but we get a cache hit

fpoli: I've not benchmarked it, but it should be cheaper than doing a method call in the JVM.

fpoli: (avoided)

JonasAlaif (approved):

PR #847, opened 2022-02-03
Persist Prusti cache across CI runs

fpoli:
This PR persists the Prusti cache across CI runs. Updating the dependencies, which should happen twice per month, will clear the cache. The tests in the `deploy.yml` workflow intentionally do not persist the cache, just to be sure that the cache-miss path is still periodically tested.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ env:
   PRUSTI_ASSERT_TIMEOUT: 60000
 
 jobs:
-  # Run a subset of the tests that used to fail often.
+  # Run Prusti reusing the cache from previous CI runs.
   # The goal here is to fail fast and give quick feedback to the developers.
   quick-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


vakaras:
I am not sure that this is a good idea because it will hide intermittent failures. Also, bugs in caching will make the CI completely useless and when discovered, they will be insanely hard to debug because you will need to figure out which commit introduced the problem. I think it would be better to duplicate the test run:
1. have a cached version that allows to much faster see if there is some problem in the PR;
2. a non-cached version that is ultimately used for determining whether to merge or not.

fpoli:
That's why the `deploy.yml` doesn't persist the cache; it's a trade-off. Duplicating all tests won't reduce the longest running job, which is what I was aiming for. I'd prefer to not persist the cache rather than duplicating, to postpone the moment that we'll hit the maximum number of concurrent CI runs in the organization.

fpoli:
Or, we can just duplicate one OS run rather than all three.

vakaras:
> Duplicating all tests won't reduce the longest running job, which is what I was aiming for.

Why do you care about this? With bors you just `r+` the PR and forget it.

> Or, we can just duplicate one OS run rather than all three.

Since intermittent bugs can be OS-specific, in this case, we should have three non-cached runs and a single cached run. I would be fine with this.

fpoli:
Oh, but the the test suite doesn't yet write the cache to disk when it terminates... That's a blocking issue.

Issue #819, opened 2022-01-07
`verify/pass/arrays/{merge,selection_sort}.rs` timeout issue


Pointerbender:
On commit 7e259b8 from PR #817 the test case `prusti-tests/tests/verify/pass/arrays/selection_sort.rs` is reliably failing due to a timeout error during the CI run on Ubuntu. This timeout issue is not reproducible on the Windows or Mac CI runs, nor on my local Ubuntu machine, where the test case is verifying okay. @vakaras [pointed out](https://github.com/viperproject/prusti-dev/pull/817#issuecomment-1006942560) that such a performance issue could very well be due to matching loop.

There are a couple of things we could try to investigate this, per the instructions from @fpoli:

> > Typically such large performance variations are caused by matching loops. @fpoli If I am not mistaken you had used AxiomProfiler recently, would you mind adding some brief instructions on how to use it?
> 
> I didn't; my logs were too big (my timeout was 30 minutes) and the tool kept crashing on Linux. However, this failing test in Prusti seems much smaller than that. The [readme](https://github.com/viperproject/axiom-profiler) has instructions on how to run the tool and there is a working Docker image that can be used even from Mac. To use the profiler, see [here](https://github.com/dafny-lang/dafny/wiki/Investigating-slow-verification-performance#4-interpreting-axiomprofiler-output); they link to the paper that describes the tool. The "Help" menu of the tool also contains useful instructions.
> 
> Alternatively, Z3's `smt.qi.profile=true` flag is very promising. Its output is explained [here](https://github.com/Z3Prover/z3/issues/4522) and with a bit of processing I used it to see how many times each quantifier is instantiated. You need to obtain first from Silicon the smt2 file with the commands sent to Z3, making sure that you can reproduce the performance issue just by running Z3 on it.

After tracking down and fixing the performance issue, we should re-enable the `prusti-tests/tests/verify/pass/arrays/selection_sort.rs` test case, as it was ignored for now until this is resolved.

Pointerbender:
I also observed `selection_sort.rs` failing spuriously in the wild on Mac OS X when fetching the latest changes on `master` from upstream to my fork of `prusti-dev`. [This particular commit](https://github.com/Pointerbender/prusti-dev/commit/14a436f9efa93918d82453f0fa1b060912109770) doesn't have the `FxHash` changes yet, so it's probably not reproducible there: https://github.com/Pointerbender/prusti-dev/runs/4737295553?check_suite_focus=true#step:8:707

[...]

Issue #812, opened 2021-12-25
Slice and array indexing should be supported in triggers


vakaras:
Currently, using `s[i]` in a trigger where `s` is an array or a slice causes Prusti to crash.

Pointerbender:
@vakaras In case the refactoring is not a blocker for adding support for triggers on domain functions, and if it's not a super non-trivial fix, maybe I could take a stab at implementing this one? :)

vakaras:
Thanks! This code is likely to be rewritten during the refactoring, but I think it would be still worth trying to figure out how to actually do it.

fpoli:
Implemented in #858 by @Pointerbender. Thanks! 

@vakaras Do you prefer to keep this issue open because of the following todo, or can I close?

> The test prusti-tests/tests/verify/pass/issues/issue-812-4.rs is currently ignored because it triggers a todo!() in the new vir high encoder here: https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/specifications/encoder_high.rs#L26-L35

vakaras:
> @vakaras Do you prefer to keep this issue open because of the following todo, or can I close?
> 
> > The test prusti-tests/tests/verify/pass/issues/issue-812-4.rs is currently ignored because it triggers a todo!() in the new vir high encoder here: https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/specifications/encoder_high.rs#L26-L35

Let's keep it open.

PR #848, opened 2022-02-03
partial fix for #819

Pointerbender:
In issue #819 the `prusti-tests/tests/verify/pass/arrays/selection_sort.rs` test was entirely disabled in anticipation of some future fixes. Those future fixes have not landed yet, but PRs #831 and #844 opened a window for a temporary work-around. I was curious if I could get it to work and was able to achieve a stable test run for `selection_sort` (consistently finishes in under 60 seconds on my PC). 

The changes I made were to selectively disable the `fix_quantifiers` and `optimize_folding` optimization passes and use wrapper functions for array accesses instead of using array indices directly (those are blocked by #812 for now).

I don't consider this a proper fix for #819, but I think it could still be a nice regression test and now that its execution time has stabilized I think it's safe to turn back on. Let's keep #819 open to track implementing a proper fix in the future.

### DIFF
--- a/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
+++ b/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
@@ -1,12 +1,16 @@
-// ignore-test: FIXME performance issue, see https://github.com/viperproject/prusti-dev/issues/819
-// compile-flags: -Pverification_deadline=180
+// FIXME performance issue, see https://github.com/viperproject/prusti-dev/issues/819
+// disabled `fix_quantifiers` and `optimize_folding` optimizations for now:
+// compile-flags: -Poptimizations=inline_constant_functions,delete_unused_predicates,remove_empty_if,purify_vars,remove_unused_vars,remove_trivial_assertions,clean_cfg -Pverification_deadline=180
 
 use prusti_contracts::*;
 
 fn main() {}
 
 
-fn selection_sort(mut a: [i32; 10]) {
+#[ensures(forall(|k1: usize, k2: usize|(0 <= k1 && k1 < k2 && k2 < 10)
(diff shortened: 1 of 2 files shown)


prusti-tests/tests/verify/pass/arrays/selection_sort.rs
```diff
@@ -15,61 +19,66 @@ fn selection_sort(mut a: [i32; 10]) {
...
-            body_invariant!(forall(|k: usize| (0 <= k && k < i) ==> a[k] <= a[min]));
+            body_invariant!(forall(|k: usize| (0 <= k && k < i)
+                                    ==> get(a, k) <= get(a, min),
+                                    triggers=[(get(a, k))]));
 
             // all not-yet-sorted that we checked yet are bigger than the current min
             body_invariant!(forall(|k: usize| (i <= k && k < j && k < 10)
-                                   ==> a[min] <= a[k]));
+                                    ==> get(a, min) <= get(a, k),
+                                    triggers=[(get(a, k))]));
```
Pointerbender, line 53: Note that these 2 quantifiers trigger themselves, but at most 1 time, so there is no matching loop. If we add `get(a, min)` to the trigger sets here, the program will not verify, unfortunately.

Pointerbender:
Hmm, looks like a different test (`tests/verify/pass/slices/slice-split.rs`) [failed](https://github.com/viperproject/prusti-dev/runs/5051346317?check_suite_focus=true#step:8:1018) on Mac OS with the error `Prusti failed to finish within 20 seconds. It finished in 20.241862046s.`. Coincidentally, that test is also quantifier related. Slightly increasing the verification deadline should fix that, I'll update this PR with that, too, shortly.

vakaras:
Thank you!

bors r+

PR #855, opened 2022-02-09
Add option to override default settings with env variables

### LINKED ISSUE #853: Prusti cache path from Prusti.toml overriden by cargo-prusti
I'm not sure if this is intended behavior or not, but currently it is impossible to provide a CACHE_PATH within a Prusti.toml file to cargo-prusti, because it is over-written with an environment variable in `cargo-prusti.rs` which takes priority that sets it to the default path of ./target/verify/CACHE.bin.

JonasAlaif:
Fixes #853
Will also need to update the documentation

### DIFF
--- a/prusti-common/src/config.rs
+++ b/prusti-common/src/config.rs
@@ -128,7 +128,7 @@ lazy_static! {
         check_keys(&settings, &allowed_keys, "default environment variables");
 
         // 3. Override with an optional TOML file specified by the `PRUSTI_CONFIG` env variable
-        let file = env::var("PRUSTI_CONFIG").unwrap_or("./Prusti.toml".to_string());
+        let file = env::var("PRUSTI_CONFIG").unwrap_or_else(|_| "./Prusti.toml".to_string());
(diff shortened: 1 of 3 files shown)


prusti-common/src/config.rs
```diff
@@ -121,18 +121,24 @@ lazy_static! {
...
-        check_keys(&settings, &allowed_keys, "Prusti.toml file");
+        check_keys(&settings, &allowed_keys, "default environment variables");
 
         // 3. Override with an optional TOML file specified by the `PRUSTI_CONFIG` env variable
         if let Ok(file) = env::var("PRUSTI_CONFIG") {
             // Since this file is explicitly specified by the user, it would be
             // nice to tell them if we cannot open it.
             settings.merge(File::with_name(&file)).unwrap();
             check_keys(&settings, &allowed_keys, &format!("{} file", file));
+        } else {
```
fpoli: Or even `let file = env::var("PRUSTI_CONFIG").unwrap_or("Prusti.toml".to_string());`

PR #856, opened 2022-02-09
Cache in CI

JonasAlaif:
Wan't sure how else to trigger a CI run


.github/workflows/test.yml
```diff
@@ -161,6 +169,14 @@ jobs:
...
       - name: Build with cargo
         run: python x.py build --all --verbose
+      - name: Set Prusti cache path and key
+        run: |
+          echo "YEAR=$(date +%Y)" >> $GITHUB_ENV
+      - name: Persist Prusti cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.PRUSTI_CACHE_PATH }}
+          key: prusti-cache-${{ matrix.os }}-${{ env.YEAR }}
```
fpoli: In https://github.com/viperproject/prusti-dev/pull/847 we concluded that only the `quick-tests` job should persist the cache, in order to avoid hiding intermittent failures.

fpoli:
@JonasAlaif @vakaras This brings the CI time down from 2h30min to about 1h30min. Could you give a look and tell me if it's fine to merge in your opinion?

.github/workflows/test.yml
vakaras: Do I understand correctly that this PR uses Prusti cache for all tests and would, therefore, potentially hide intermittent CI failures?

fpoli: Yes, but I wonder if we can find a better tradeoff than what at the end of #847. Faster CI would encourage smaller PRs. All tests are rerun on master without cache anyway, so we'll able to notice hidden CI failures and in the worst case revert this PR.

Perhaps I can
- add the Viper version to the cache key, so that upgrading Viper invalidates the cache.
- run the quick tests with Silicon and Carbon, to keep testing the code that launches Viper.

JonasAlaif: I'm not sure I understand the point about "intermittent CI failures" - caching is done at the viper program level, as soon as a bug would introduce an encoding error this would necessarily be a different viper program to the passing one which was cached. That is, caching would only hide timing/benchmarking errors and "sending the viper program to Scala" errors

vakaras: > All tests are rerun on master without cache anyway, so we'll able to notice hidden CI failures and in the worst case revert this PR.

Do you mean once it is fully merged? That is probably sufficient.

vakaras: > That is, caching would only hide timing/benchmarking errors

These are the errors I am talking about. For example, non-linear arithmetic or problematic quantifiers can cause a test to fail non-deterministically.

fpoli: > Do you mean once it is fully merged? That is probably sufficient.

Yes, the `deploy.yml` workflow compiles in release mode (to prepare the artifact) and run all tests (in release mode) without using the cargo nor the verification cache. I added the Viper version to the key of the verification cache and I disabled the verification cache for the quick tests as it was just saving seconds.

> These are the errors I am talking about. For example, non-linear arithmetic or problematic quantifiers can cause a test to fail non-deterministically.

That's a good point, but I think that we can address flaky tests it in other ways. For example, we could have a nightly run that reruns all tests on master and that automatically creates and assigns an issue or PR to one of us, asking to disable or fix the flaky tests.

vakaras: > > Do you mean once it is fully merged? That is probably sufficient.
> 
> Yes, the `deploy.yml` workflow compiles in release mode (to prepare the artifact) and run all tests (in release mode) without using the cargo nor the verification cache. I added the Viper version to the key of the verification cache and I disabled the verification cache for the quick tests as it was just saving seconds.

Sounds good.

> > These are the errors I am talking about. For example, non-linear arithmetic or problematic quantifiers can cause a test to fail non-deterministically.
> 
> That's a good point, but I think that we can address flaky tests it in other ways. For example, we could have a nightly run that reruns all tests on master and that automatically creates and assigns an issue or PR to one of us, asking to disable or fix the flaky tests.

I like this idea.

fpoli:
I opened https://github.com/viperproject/prusti-dev/issues/1176 to address the problem of the flaky tests.

Issue #796, opened 2021-12-12
Use snapshot encoding for Rust assertions


Pointerbender:
The following Rust program:

```rust
use prusti_contracts::*;

#[requires(forall(|i: usize| (0 <= i && i < 16) ==> (array[i] > 100) ))]
pub fn assert_example(array: &[usize; 16]) {
    assert!(array[0] > 100);
}

fn main() {}
```

fails verification with the error:

```
error: [Prusti: verification error] the asserted expression might not hold
 --> src/lib.rs:9:5
  |
9 |     assert!(array[0] > 100);
  |     ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `prusti-example` due to previous error
```

Digging a bit deeper, I noticed this is because the precondition is inhaled using snapshot encoding:

```
inhale (forall _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101: Int ::
  0 <= _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101
  && _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101 <= 18446744073709551615
    ==> 0 <= _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101
        && _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101 < 16
             ==> _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101 < 16
                  && read$Snap$Array$16$usize$__$TY$__Snap$Array$16$usize$$int$$$int$(
                    snap$__$TY$__Array$16$usize$Snap$Array$16$usize(_1.val_ref),
                    _2_quant_2c47419d7b5c4fa681d9a6f889ecb294_101
                  ) > 100
)
```

Note that because this is a user specified precondition, there are no explicit triggers defined and Viper is tasked with inferring the correct triggers. Then the `assert!()` is checked in the body as:

```
  __t5 := builtin$havoc_int()
  inhale __t5 >= 0 && 18446744073709551615 >= __t5
  assert acc(Array$16$usize(_1.val_ref), read$())
  inhale __t5 >= 0 && 18446744073709551615 >= __t5
  inhale lookup_pure__$TY$__Array$16$usize$$int$$$int$(_1.val_ref, _6) == __t5
  _5 := builtin$havoc_int()
  _5 := __t5
  label l0
[...]

PR #858, opened 2022-02-12
add support for array/slice accesses in quantifier triggers

Pointerbender:
This PR adds support for using slice/array access in quantifier triggers (issue #812). These are encoded including bound checks, but the bound checks are not needed (nor allowed) inside the triggers. The fix/hack applied here removes those bound checks from the triggers.

This seems to fix the encoding problem, but it doesn't appear to be a full solution for `prusti-tests/tests/verify/pass/arrays/selection_sort.rs` (see #819). If I replace all the triggers there with direct array accesses (but not the lookup and the assignment statements in the loop body wrapped in `get`/`set`), then the verification time for this test jumps from 1 minute to 4 minutes on my PC. If I also replace the lookup and assign statements in the loop bodies with direct array accesses, then the verification times out. I suspect this is a matter of adding a few dummy assert statements here and there, but as long as #796 hasn't landed yet this does not seem to improve the situation (it still times out, even with dummy assertions atm). I tried adding `body_invariant`s instead of assert statements, but this also still results in timeouts. For these reasons, I decided not to include the updated version of `selection_sort.rs`. Using a custom pure lookup function such as `get` still yields superior performance over interactions with `lookup_pure` array/slice accesses, even with the fix from this PR, unfortunately :)

The test `prusti-tests/tests/verify/pass/issues/issue-812-4.rs` is currently ignored because it triggers a `todo!()` in the new vir high encoder here:

https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/specifications/encoder_high.rs#L26-L35

I included the test anyway so that it can be enabled later when that part is implemented during the refactoring.

I'd like to investigate deeper what is the best way to trigger the quantifiers (which are snap encoded) from a Viper method body (where Rust statements are not snap encoded, unless routed through a pure Rust function), but I thought I'd first do a check-in first to see if the proposed fix is heading in the right direction and inquire about if there are any other alternatives than #796 or custom pure lookup functions to work-around the performance issues that seem to be caused by the mixed snap/non-snap encoding :) 

Thanks!

### DIFF
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
@@ -184,7 +184,7 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                true,
+                false, // quantifier triggers might not evaluate to boolean
                 parent_def_id,
                 substs,
             );
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
@@ -22,6 +22,26 @@ type FunctionConstructor<'v, 'tcx> = Box<
     ) -> SpannedEncodingResult<vir_high::FunctionDecl>,
 >;
 
+/// Depending on the context of the pure encoding,
+/// panics will be encoded slightly differently.
+#[derive(PartialEq, Eq)]
+pub(crate) enum PureEncodingContext {
+    /// Panics will be encoded as calls to unreachable functions
+    /// (which have a `requires false` pre-condition).
+    Code,
+    /// Indicates that a panic should be encoded to a `false` boolean value.
+    /// This flag is used to distinguish whether an assert terminators generated e.g. for an
+    /// integer overflow should be translated into `false` and when to an "unreachable" function
+    /// call with a `false` precondition.
+    Assertion,
+    /// Whether the current pure expression that's being encoded sits inside a trigger closure.
+    /// Viper limits the type of expressions that are allowed in quantifier triggers and
+    /// this requires special care when encoding array/slice accesses which may come with
+    /// bound checks included in the MIR. For this purpose, paths that might panic will
+    /// be stripped away during the trigger encoding.
+    Trigger,
+}
+
 #[derive(Default)]
 pub(crate) struct PureFunctionEncoderState<'v, 'tcx: 'v> {
     bodies_high: RefCell<FxHashMap<Key, vir_high::Expression>>,
@@ -184,8 +204,12 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                // quantifier triggers might not evaluate to boolean
-                !self.is_encoding_trigger.get(),
+                if self.is_encoding_trigger.get() {
+                    // quantifier triggers might not evaluate to boolean
+                    PureEncodingContext::Trigger
+                } else {
+                    PureEncodingContext::Assertion
+                },
                 parent_def_id,
                 substs,
             );
@@ -242,7 +266,7 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                false,
+                PureEncodingContext::Code,
                 proc_def_id,
                 substs,
             );
@@ -367,7 +391,7 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                false,
+                PureEncodingContext::Code,
                 parent_def_id,
                 substs,
             );
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
@@ -184,7 +184,8 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                false, // quantifier triggers might not evaluate to boolean
+                // quantifier triggers might not evaluate to boolean
+                !self.is_encoding_trigger.get(),
                 parent_def_id,
                 substs,
             );
--- a/prusti-tests/tests/verify/fail/simple-specs/panic-in-post.rs
+++ b/prusti-tests/tests/verify/fail/simple-specs/panic-in-post.rs
@@ -1,0 +1,16 @@
+use prusti_contracts::*;
+
+#[requires(x != 0)]
+#[ensures({
+    assert!(x != 0); // OK
+    true
+})]
+fn good(x: u32) {}
+
+#[ensures({
+    assert!(x != 0); //~ ERROR postcondition might not hold
+    true
+})]
+fn bad(x: u32) {}
+
+fn main() {}
--- a/prusti-tests/tests/verify/fail/simple-specs/panic-in-pre.rs
+++ b/prusti-tests/tests/verify/fail/simple-specs/panic-in-pre.rs
@@ -1,0 +1,18 @@
+use prusti_contracts::*;
+
+#[requires({
+    assert!(x != 0);
+    true
+})]
+fn foo(x: u32) {
+    assert!(x != 0); // OK
+    assert!(false);  //~ ERROR the asserted expression might not hold
+}
+
+fn bad(x: u32) {
+    foo(x); //~ ERROR precondition might not hold
+}
+
+fn main() {
+    foo(1); // OK
+}
--- a/prusti-tests/tests/verify/pass/issues/issue-812-1.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-812-1.rs
@@ -1,0 +1,6 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[requires(forall(|i: usize| (0 <= i && i < slice.len()) ==> (slice[i] > 10), triggers=[(slice[i],)]))]
+fn test(slice: &[i32]) {}
--- a/prusti-tests/tests/verify/pass/issues/issue-812-2.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-812-2.rs
@@ -1,0 +1,6 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[requires(forall(|i: usize| (0 <= i && i < array.len()) ==> (array[i] > 10), triggers=[(array[i],)]))]
+fn test(array: &[i32; 10]) {}
--- a/prusti-tests/tests/verify/pass/issues/issue-812-3.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-812-3.rs
@@ -1,0 +1,17 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[requires(forall(|i: usize| (0 <= i && i < slice.len()) ==> (bar(slice[i])), triggers=[(bar(slice[i]),(slice.len()),)]))]
+fn foo(slice: &[i32]) {}
+
+#[pure]
+#[trusted]
+fn bar(a: i32) -> bool { true }
+
+#[requires(slice.len() == 1)]
+fn baz(slice: &[i32]) {
+    if bar(slice[0]) {
+        foo(slice);
+    }
+}
--- a/prusti-tests/tests/verify/pass/issues/issue-812-4.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-812-4.rs
@@ -1,0 +1,18 @@
+// ignore-test
+// FIXME: this test passes for the original encoder, but it fails on a todo!() in the new vir high encoder
+// FIXME: see https://github.com/viperproject/prusti-dev/issues/812
+
+use prusti_contracts::*;
+
+fn main() {}
+
+#[pure]
+#[requires(forall(|i: usize| (0 <= i && i < slice.len()) ==> (slice[i] > 10), triggers=[(slice[i],slice.len(),baz(slice))]))]
+fn foo(slice: &[i32]) {}
+
+#[requires(forall(|i: usize| (0 <= i && i < array.len()) ==> (array[i] > 10), triggers=[(array[i],array.len(),baz(array),)]))]
+fn bar(array: &[i32; 10]) {}
+
+#[pure]
+#[trusted]
+fn baz(slice: &[i32]) {}
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -27,7 +27,7 @@ use vir_crate::common::identifier::WithIdentifier;
 use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_middle::mir;
 use rustc_middle::ty;
-use std::cell::{RefCell, RefMut, Ref};
+use std::cell::{Cell, RefCell, RefMut, Ref};
 use rustc_hash::{FxHashMap};
 use std::io::Write;
 use std::rc::Rc;
@@ -90,6 +90,11 @@ pub struct Encoder<'v, 'tcx: 'v> {
     name_interner: RefCell<NameInterner>,
     /// Maps locals to the local of their discriminant.
     discriminants_info: RefCell<FxHashMap<(ProcedureDefId, String), Vec<String>>>,
+    /// Whether the current pure expression that's being encoded sits inside a trigger closure.
+    /// Viper limits the type of expressions that are allowed in quantifier triggers and
+    /// this requires special care when encoding array/slice accesses which may come with
+    /// bound checks included in the MIR.
+    pub(super) is_encoding_trigger: Cell<bool>,
 }
 
 pub type EncodingTask<'tcx> = (ProcedureDefId, Vec<(ty::Ty<'tcx>, ty::Ty<'tcx>)>);
@@ -157,6 +162,7 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             encoding_errors_counter: RefCell::new(0),
             name_interner: RefCell::new(NameInterner::new()),
             discriminants_info: RefCell::new(FxHashMap::default()),
+            is_encoding_trigger: Cell::new(false),
         }
     }
 
--- a/prusti-viper/src/encoder/mir/pure/mod.rs
+++ b/prusti-viper/src/encoder/mir/pure/mod.rs
@@ -4,7 +4,8 @@ mod specifications;
 
 pub(crate) use self::{
     pure_functions::{
-        PureFunctionBackwardInterpreter, PureFunctionEncoderInterface, PureFunctionEncoderState,
+        PureEncodingContext, PureFunctionBackwardInterpreter, PureFunctionEncoderInterface,
+        PureFunctionEncoderState,
     },
     specifications::SpecificationEncoderInterface,
 };
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
@@ -10,7 +10,7 @@ use crate::encoder::{
     encoder::SubstMap,
     errors::{ErrorCtxt, SpannedEncodingError, SpannedEncodingResult, WithSpan},
     high::{generics::HighGenericsEncoderInterface, types::HighTypeEncoderInterface},
-    mir::pure::SpecificationEncoderInterface,
+    mir::pure::{PureEncodingContext, SpecificationEncoderInterface},
     mir_encoder::PlaceEncoder,
     mir_interpreter::run_backward_interpretation,
     snapshot::interface::SnapshotEncoderInterface,
@@ -45,7 +45,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
         encoder: &'p Encoder<'v, 'tcx>,
         proc_def_id: DefId,
         mir: &'p mir::Body<'tcx>,
-        is_encoding_assertion: bool,
+        pure_encoding_context: PureEncodingContext,
         parent_def_id: DefId,
         tymap: &'p SubstMap<'tcx>,
     ) -> Self {
@@ -54,7 +54,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
             encoder,
             mir,
             proc_def_id,
-            is_encoding_assertion,
+            pure_encoding_context,
             parent_def_id,
             tymap.clone(),
         );
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
@@ -77,7 +77,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
             .unwrap_or_else(|| panic!("Procedure {:?} contains a loop", self.proc_def_id));
         let body_expr = state.into_expr().unwrap();
         debug!(
-            "Pure function {} has been encoded with expr: {}",
+            "Pure function body {} has been encoded with expr: {}",
             function_name, body_expr
         );
         let substs = &self
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
@@ -663,31 +663,47 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                             }
                         }
                     } else {
-                        // FIXME: Refactor the common code with the procedure encoder.
-
                         // Encoding of a non-terminating function call
-                        let error_ctxt = match full_func_proc_name {
-                            "std::rt::begin_panic"
-                            | "core::panicking::panic"
-                            | "core::panicking::panic_fmt" => {
-                                // This is called when a Rust assertion fails
-                                // args[0]: message
-                                // args[1]: position of failing assertions
-
-                                let panic_cause = self.mir_encoder.encode_panic_cause(span);
-                                ErrorCtxt::PanicInPureFunction(panic_cause)
+                        // FIXME: Refactor the common code with the procedure encoder.
+                        match self.pure_encoding_context {
+                            PureEncodingContext::Trigger => {
+                                // We are encoding a trigger, so all panic branches must be stripped.
+                                ExprBackwardInterpreterState::new(None)
                             }
+                            PureEncodingContext::Assertion => {
+                                // We are encoding an assertion, so all failures should be equivalent to false.
+                                debug_assert!(matches!(
+                                    self.mir.return_ty().kind(),
+                                    ty::TyKind::Bool
+                                ));
+                                ExprBackwardInterpreterState::new(Some(false.into()))
+                            }
+                            PureEncodingContext::Code => {
+                                // We are encoding a pure function, so all failures should be unreachable.
+                                let error_ctxt = match full_func_proc_name {
+                                    "std::rt::begin_panic"
+                                    | "core::panicking::panic"
+                                    | "core::panicking::panic_fmt" => {
+                                        // This is called when a Rust assertion fails
+                                        // args[0]: message
+                                        // args[1]: position of failing assertions
 
-                            _ => ErrorCtxt::DivergingCallInPureFunction,
-                        };
-                        let pos = self.encoder.error_manager().register(
-                            term.source_info.span,
-                            error_ctxt,
-                            self.caller_def_id,
-                        );
-                        ExprBackwardInterpreterState::new_defined(
-                            unreachable_expr(pos).with_span(term.source_info.span)?,
-                        )
+                                        let panic_cause = self.mir_encoder.encode_panic_cause(span);
+                                        ErrorCtxt::PanicInPureFunction(panic_cause)
+                                    }
+
+                                    _ => ErrorCtxt::DivergingCallInPureFunction,
+                                };
+                                let pos = self.encoder.error_manager().register(
+                                    term.source_info.span,
+                                    error_ctxt,
+                                    self.caller_def_id,
+                                );
+                                ExprBackwardInterpreterState::new_defined(
+                                    unreachable_expr(pos).with_span(term.source_info.span)?,
+                                )
+                            }
+                        }
                     };
 
                     state
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
@@ -732,7 +732,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                 match self.pure_encoding_context {
                     PureEncodingContext::Trigger => {
                         // We are encoding a trigger, so all panic branches must be stripped.
-                        ExprBackwardInterpreterState::new(None)
+                        states[target].clone()
                     }
                     PureEncodingContext::Assertion => {
                         // We are encoding an assertion, so all failures should be equivalent to false.
@@ -762,20 +762,6 @@ impl<'p, 'v: 'p, 'tcx: 'v> BackwardMirInterpreter<'tcx>
                         ))
                     }
                 }
-
-                //     let failure_encoding = if self.encode_panic_to_false == PureEncodingContext::Assertion {
-                //         // TODO PB
-                //         // We are encoding an assertion, so all failures should be equivalent to false.
-                //         debug_assert!(matches!(self.mir.return_ty().kind(), ty::TyKind::Bool));
-                //         false.into()
-                //     } else {
-                //         // We are encoding a pure function, so all failures should be unreachable.
-                //         unreachable_expr(pos).with_span(term.source_info.span)?
-                //     };
-
-                //     ExprBackwardInterpreterState::new(states[target].expr().map(|target_expr| {
-                //         vir::Expr::ite(viper_guard.clone(), target_expr.clone(), failure_encoding)
-                //     }))
             }
 
             TerminatorKind::Yield { .. }
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs
@@ -11,7 +11,8 @@ use crate::encoder::{
         generics::HighGenericsEncoderInterface, types::HighTypeEncoderInterface,
     },
     mir::{
-        generics::MirGenericsEncoderInterface, pure::specifications::SpecificationEncoderInterface,
+        generics::MirGenericsEncoderInterface,
+        pure::{specifications::SpecificationEncoderInterface, PureEncodingContext},
         types::MirTypeEncoderInterface,
     },
     mir_encoder::{MirEncoder, PlaceEncoder, PlaceEncoding, PRECONDITION_LABEL, WAND_LHS_LABEL},
@@ -36,11 +37,8 @@ pub(crate) struct PureFunctionBackwardInterpreter<'p, 'v: 'p, 'tcx: 'v> {
     mir: &'p mir::Body<'tcx>,
     /// MirEncoder of the pure function being encoded.
     mir_encoder: MirEncoder<'p, 'v, 'tcx>,
-    /// True if a panic should be encoded to a `false` boolean value.
-    /// This flag is used to distinguish whether an assert terminators generated e.g. for an
-    /// integer overflow should be translated into `false` and when to an "unreachable" function
-    /// call with a `false` precondition.
-    encode_panic_to_false: bool,
+    /// How panics are handled depending on the encoding context.
+    pure_encoding_context: PureEncodingContext,
     /// DefId of the caller. Used for error reporting.
     caller_def_id: DefId,
     tymap: SubstMap<'tcx>,
@@ -54,7 +52,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionBackwardInterpreter<'p, 'v, 'tcx> {
         encoder: &'p Encoder<'v, 'tcx>,
         mir: &'p mir::Body<'tcx>,
         def_id: DefId,
-        encode_panic_to_false: bool,
+        pure_encoding_context: PureEncodingContext,
         caller_def_id: DefId,
         tymap: SubstMap<'tcx>,
     ) -> Self {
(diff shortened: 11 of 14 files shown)


Pointerbender:
~(Looks like the Ubuntu tests failed to download the Viper tools and that's why the test run failed, would someone be able to restart the Ubuntu tests one more time? :) Thanks!)~

Edit: Nvm, it looks like more tests failed :o) I will investigate and push again!

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
```diff
@@ -184,7 +184,7 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
                 self,
                 proc_def_id,
                 procedure.get_mir(),
-                true,
+                false, // quantifier triggers might not evaluate to boolean
```
vakaras: This change looks wrong. Does it panic with `true` when encoding triggers?

Aurel300:
This relies on the fact that we don't currently allow much in trigger expressions, so we assume that any conditional in a trigger should be stripped? If so, this is quite hacky…

vakaras:
> This relies on the fact that we don't currently allow much in trigger expressions, so we assume that any conditional in a trigger should be stripped? If so, this is quite hacky…

Triggers are allowed to contain only function calls and variables, so as a temporary fix this sounds fine to me (before I reach this code with my refactorings).

Pointerbender:
> > `false, // quantifier triggers might not evaluate to boolean`
>
> This change looks wrong. Does it panic with true when encoding triggers?

This is a fun one :) There is a debug assertion here:

https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L733-L740

which panics when encoding an array/slice index. It complains about the return type being `I32` while it expects it to be a boolean (when slicing into a `[i32]`). This is due to the added bound checks on the access. As far as I can tell, it's not Prusti that adds these bound checks, but these in fact come from the Rust core library:

https://github.com/rust-lang/rust/blob/5d8767cb229b097fedb1dd4bd9420d463c37774f/library/core/src/slice/index.rs#L156-L163

and

https://github.com/rust-lang/rust/blob/5d8767cb229b097fedb1dd4bd9420d463c37774f/library/core/src/slice/index.rs#L186-L190

which generates MIR for the trigger that looks like:

20220211-issue_812_1 prusti_pre_item_test_8e3c41a77fbb42d58cba0a8b5961dfce-{closure#0} ------- renumber 0 dot

and it's the `assert(move _5, ..)` that then triggers the panic in Prusti.

> This relies on the fact that we don't currently allow much in trigger expressions, so we assume that any conditional in a trigger should be stripped? If so, this is quite hacky…

I share the feeling that it's a bit hacky. I tried to mitigate it (for the purpose of a temporary work-around) by pattern matching on just the cases where we can encounter bound checks and not in other scenarios, although it is not an air-tight solution nor is it future-proof when triggers can handle more kinds of expressions :sweat_smile: 

Another way to do it, and solve the `encode_panic_to_false = false` quirk at the same time, could be to intercept the call to `core::ops::Index::index` (sort-of like is happening [in a few places](https://github.com/viperproject/prusti-dev/search?q=%22core%3A%3Aops%3A%3AIndex%3A%3Aindex%22) already [#1](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L508) [#2](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L476-L486) [#3](https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/procedure_encoder.rs#L2243-L2253)) and remove the bound checks here already. However, this would probably not scale well when end-users start adding custom `Index` trait implementations for their own types (this PR has that same problem, too) and in some cases we still want to keep the bound checks, too. A solution that could work-around those problems would then maybe have to wrap the whole `core::ops::Index::index` method into a wrapper Viper function, which can then be used as a trigger assuming that same function is used everywhere else, too. This would probably be a pretty big change, and there might be better solutions that we haven't thought of yet.

vakaras:
Thank you for the investigation!

> Another way to do it, and solve the `encode_panic_to_false = false` quirk at the same time, could be to intercept the call to `core::ops::Index::index` (sort-of like is happening [in a few places](https://github.com/viperproject/prusti-dev/search?q=%22core%3A%3Aops%3A%3AIndex%3A%3Aindex%22) already [#1](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L508) [#2](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L476-L486) [#3](https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/procedure_encoder.rs#L2243-L2253)) and remove the bound checks here already.

This sounds like the best solution for now. We probably should have a flag `encoding_trigger` to make sure that interception is done only when encoding triggers.

> However, this would probably not scale well when end-users start adding custom `Index` trait implementations for their own types

That should not be a problem because in that case we should have just a regular function call.

Pointerbender:
> This sounds like the best solution for now. We probably should have a flag encoding_trigger to make sure that interception is done only when encoding triggers.

Thanks! I'll give this a shot to see if anything else we can learn comes out of it :)

> That should not be a problem because in that case we should have just a regular function call.

Ah right, and that function is then perfectly fine as a trigger, nice! I guess this only affects triggers that try to encode domain functions using special logic in the Prusti internals. Do there currently exist any other use cases for triggers that would end up encoding domain function apps, other than pure lookups?

vakaras:
> > This sounds like the best solution for now. We probably should have a flag encoding_trigger to make sure that interception is done only when encoding triggers.
> 
> Thanks! I'll give this a shot to see if anything else we can learn comes out of it :)

Great! Thanks!

> > That should not be a problem because in that case we should have just a regular function call.
> 
> Ah right, and that function is then perfectly fine as a trigger, nice! I guess this only affects triggers that try to encode domain functions using special logic in the Prusti internals. Do there currently exist any other use cases for triggers that would end up encoding domain function apps, other than pure lookups?

Nothing comes to my mind. @Aurel300 ?

Aurel300:
In the closures work, when encoding spec entailments and call descriptions, we turn them into quantifiars triggered on specification functions (which are domain functions) of the corresponding closure type. We haven't had cases when we needed to trigger a quantifier itself with some closure call, but maybe this should be allowed and maybe such a trigger should then also be the specification function call.

Pointerbender:
> Another way to do it, and solve the encode_panic_to_false = false quirk at the same time, could be to intercept the call to core::ops::Index::index (sort-of like is happening [in a few places](https://github.com/viperproject/prusti-dev/search?q=%22core%3A%3Aops%3A%3AIndex%3A%3Aindex%22) already [#1](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L508) [#2](https://github.com/viperproject/prusti-dev/blob/66c90d5dae9040eb68b2f0d2689faa480b10e8d6/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L476-L486) [#3](https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/procedure_encoder.rs#L2243-L2253)) and remove the bound checks here already.

I hit a small setback with this one: it looks like the knowledge about `core::ops::Index::index` for `usize` indexing is already lost at the MIR level and instead it's [translated to snap encoding](https://github.com/viperproject/prusti-dev/blob/master/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L157-L192) by looking at `PlaceEncoding::ArrayAccess`/`PlaceEncoding::SliceAccess` in the MIR in `postprocess_place_encoding` in file `prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs`. Here is the MIR outputted for this program to get a feel for the information we can work with:

```rust
pub fn main() {
    let _cl = |a: &[i32], i: usize| -> i32 { a[i] };
}

pub fn test(a: &[i32], i: usize) -> i32 {
    a[i]
}
```

The MIR is (from the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9680f89569bee686ac225ffaa7ce4e2d), but the `renumber` MIR version shows the same thing, too):

```
fn main::{closure#0}(_1: &[closure@src/main.rs:2:15: 2:52], _2: &[i32], _3: usize) -> i32 {
    debug a => _2;                       // in scope 0 at src/main.rs:2:16: 2:17
    debug i => _3;                       // in scope 0 at src/main.rs:2:27: 2:28
    let mut _0: i32;                     // return place in scope 0 at src/main.rs:2:40: 2:43
    let _4: usize;                       // in scope 0 at src/main.rs:2:48: 2:49
    let mut _5: usize;                   // in scope 0 at src/main.rs:2:46: 2:50
    let mut _6: bool;                    // in scope 0 at src/main.rs:2:46: 2:50

    bb0: {
        _4 = _3;                         // scope 0 at src/main.rs:2:48: 2:49
        _5 = Len((*_2));                 // scope 0 at src/main.rs:2:46: 2:50
        _6 = Lt(_4, _5);                 // scope 0 at src/main.rs:2:46: 2:50
        assert(move _6, "index out of bounds: the length is {} but the index is {}", move _5, _4) -> bb1; // scope 0 at src/main.rs:2:46: 2:50
    }

    bb1: {
        _0 = (*_2)[_4];                  // scope 0 at src/main.rs:2:46: 2:50
        return;                          // scope 0 at src/main.rs:2:52: 2:52
    }
}

fn test(_1: &[i32], _2: usize) -> i32 {
    debug a => _1;                       // in scope 0 at src/main.rs:5:13: 5:14
    debug i => _2;                       // in scope 0 at src/main.rs:5:24: 5:25
    let mut _0: i32;                     // return place in scope 0 at src/main.rs:5:37: 5:40
    let _3: usize;                       // in scope 0 at src/main.rs:6:7: 6:8
    let mut _4: usize;                   // in scope 0 at src/main.rs:6:5: 6:9
    let mut _5: bool;                    // in scope 0 at src/main.rs:6:5: 6:9

    bb0: {
        _3 = _2;                         // scope 0 at src/main.rs:6:7: 6:8
        _4 = Len((*_1));                 // scope 0 at src/main.rs:6:5: 6:9
        _5 = Lt(_3, _4);                 // scope 0 at src/main.rs:6:5: 6:9
        assert(move _5, "index out of bounds: the length is {} but the index is {}", move _4, _3) -> bb1; // scope 0 at src/main.rs:6:5: 6:9
    }

    bb1: {
        _0 = (*_1)[_3];                  // scope 0 at src/main.rs:6:5: 6:9
        return;                          // scope 0 at src/main.rs:7:2: 7:2
    }
}

fn main() -> () {
    let mut _0: ();                      // return place in scope 0 at src/main.rs:1:15: 1:15
    let _1: [closure@src/main.rs:2:15: 2:52]; // in scope 0 at src/main.rs:2:9: 2:12
    scope 1 {
        debug _cl => _1;                 // in scope 1 at src/main.rs:2:9: 2:12
    }

    bb0: {
        return;                          // scope 0 at src/main.rs:3:2: 3:2
    }
}
```

As far as I can tell, the only 2 places where Prusti is aware of being in a trigger are:

1. [`encode_quantifier()`](https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs#L105) where it recombines the specs generated by the proc macros into a pure expression.
2. During the pre-parsing of the procedural macros in the `prusti-specs` crate.

In order to correctly set a new `is_encoding_trigger` flag, we would have to add the required information to do so in one of these 2 places, I think. For the first option, a proof-of-concept is made in this PR. The second option would require some kind of markers outputted by the proc macro which the encoder can then pick up on in order to toggle the `is_encoding_trigger` flag.

I slightly improved the pattern matching on the latest commit from just now in order to yield less false-positives for when triggers support more syntax in the future.

I'm not entirely happy about adding a `pub(super) is_encoding_trigger: Cell<bool>` field to the `Encoder` from a code design perspective, but it seemed to be the only way without having to update the trait definitions as well (which seemed like such a flag did not belong in the parameters of its methods). A proper long-term fix would probably include some modifications to the proc macro pre-parser to make this part nicer. In theory the `is_encoding_trigger` flag and bound-check-stripping logic can then both be moved to the `PureFunctionEncoder` where it would feel more at home. But even then we still need some way to detect the bound checks from the MIR which are different for arrays and slices, e.g. in the case of an array of length 4 it is:

```
_5 = const 4_usize;
_6 = Lt(_4, _5); 
assert(move _6, "index out of bounds: the length is {} but the index is {}", move _5, _4) -> bb1;
```

For now, hopefully this PR holds an acceptable temporary fix that uses an `is_encoding_trigger` flag to intercept the bound checks encoding :)

vakaras:
> I hit a small setback with this one: it looks like the knowledge about `core::ops::Index::index` for `usize` indexing is already lost at the MIR level

This is interesting! I guess the reason is that indexing into slices is a compiler built-in operation.

> As far as I can tell, the only 2 places where Prusti is aware of being in a trigger are:
> 
>     1. [`encode_quantifier()`](https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs#L105) where it recombines the specs generated by the proc macros into a pure expression.
> 
>     2. During the pre-parsing of the procedural macros in the `prusti-specs` crate.

I think adding changes to `encode_quantifier` is clearly the better option.

vakaras:
> I'm not entirely happy about adding a `pub(super) is_encoding_trigger: Cell<bool>` field to the Encoder from a code design perspective

This is clearly not ideal, but I would say that it is good enough for now.

Pointerbender:
Thanks for the review! I just pushed one more commit which addresses some failing corner cases where `is_encoding_trigger` was not restored properly. I think we're there now :)

fpoli:
> > This relies on the fact that we don't currently allow much in trigger expressions, so we assume that any conditional in a trigger should be stripped? If so, this is quite hacky…
> 
> I share the feeling that it's a bit hacky. I tried to mitigate it (for the purpose of a temporary work-around) by pattern matching on just the cases where we can encounter bound checks and not in other scenarios, although it is not an air-tight solution nor is it future-proof when triggers can handle more kinds of expressions sweat_smile

If we add a new `is_encoding_trigger: bool` field to `ExpressionBackwardInterpreter`, then the following two changes should be enough to robustly remove all panic checks from the encoding of the pure expression:
1. Add a third case that, if encoding a trigger, exits from the match with `ExprBackwardInterpreterState::new(None)`: https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L733-L740 (Even better, `encode_panic_to_false` and `is_encoding_trigger` could be merged in a single field of type e.g. `enum PureEncodingContext { Code, Assertion, Trigger }`, but we can do that after the refactoring.)
2. When encoding a switch, all targets associated to a `None` state should be skipped instead of assuming that they are `Some`. In particular, if only one target state is defined, then there is no need to generate the `Expr::ite(guard, ...)` check. The code to be rewritten is this: https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L381-L395

fpoli:
> 2\. When encoding a switch, all targets associated to a `None` state should be skipped instead of assuming that they are `Some`. In particular, if only one target state is defined, then there is no need to generate the `Expr::ite(guard, ...)` check. The code to be rewritten is this: https://github.com/viperproject/prusti-dev/blob/2a71e807e637751d62728a2627c71913acfeed85/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L381-L395

I implemented this change in #861.

Pointerbender:
> I implemented this change in https://github.com/viperproject/prusti-dev/pull/861.

Thank you! I have a bit of downtime tomorrow during which I can try to implement step 1 :) I will also try to implement the single enum field already, that seems the cleanest way to do it.

Pointerbender:
I added the `PureEncodingContext` enum in the last commit. It's now panicking on the `.unwrap()` on line 78 here:

https://github.com/viperproject/prusti-dev/blob/5b09b527fff90ca767abe35a1fb08a624713e422/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs#L71-L89

I was not able to investigate what the next step is, but hopefully I'll find another hour or so tonight to dig a little deeper :) (converted the PR to draft for now pending the investigation of the panic)

fpoli:
The new changes are exactly what I meant :+1: I would expect that unwrap to fail if e.g. the trigger is `panic!()`. Mmm, I also need more time to investigate.

Pointerbender:
The trace logs show some interesting things near the end:

```
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: Some(Field(FieldExpr { base: DomainFuncApp(DomainFuncApp { domain_function: DomainFunc { name: "read$Snap$Slice$i32$", type_arguments: [], formal_args: [self: Snapshot(Slice$i32), idx: Int], return_type: Int, unique: false, domain_name: "Snap$Slice$i32" }, arguments: [Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Local(Local { variable: _1: Ref(ref$closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: closure_0: Ref(ref$ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), Field(FieldExpr { base: Local(Local { variable: _3: Ref(usize), position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 90, id: 29 } })], position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 98, id: 27 } })), substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] States before: {bb1: ExprBackwardInterpreterState { expr: Some(Field(FieldExpr { base: DomainFuncApp(DomainFuncApp { domain_function: DomainFunc { name: "read$Snap$Slice$i32$", type_arguments: [], formal_args: [self: Snapshot(Slice$i32), idx: Int], return_type: Int, unique: false, domain_name: "Snap$Slice$i32" }, arguments: [Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Local(Local { variable: _1: Ref(ref$closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: closure_0: Ref(ref$ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), Field(FieldExpr { base: Local(Local { variable: _3: Ref(usize), position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 90, id: 29 } })], position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 98, id: 27 } })), substs: {} }, bb2: ExprBackwardInterpreterState { expr: Some(FuncApp(FuncApp { function_name: "builtin$unreach", type_arguments: [Int], arguments: [], formal_arguments: [], return_type: Int, position: Position { line: 6, column: 90, id: 25 } })), substs: {} }}
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply terminator Terminator { source_info: SourceInfo { span: prusti-tests/tests/verify/pass/issues/issue-812-1.rs:6:90: 6:98 (#0), scope: scope[0] }, kind: assert(move _5, "index out of bounds: the length is {} but the index is {}", move _4, _3) -> [success: bb1, unwind: bb2] }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_terminator Terminator { source_info: SourceInfo { span: prusti-tests/tests/verify/pass/issues/issue-812-1.rs:6:90: 6:98 (#0), scope: scope[0] }, kind: assert(move _5, "index out of bounds: the length is {} but the index is {}", move _4, _3) -> [success: bb1, unwind: bb2] }, states: {bb1: ExprBackwardInterpreterState { expr: Some(Field(FieldExpr { base: DomainFuncApp(DomainFuncApp { domain_function: DomainFunc { name: "read$Snap$Slice$i32$", type_arguments: [], formal_args: [self: Snapshot(Slice$i32), idx: Int], return_type: Int, unique: false, domain_name: "Snap$Slice$i32" }, arguments: [Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Field(FieldExpr { base: Local(Local { variable: _1: Ref(ref$closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(closure$m_prusti_pre_item_test_ac0f455d5c7546118ecf57dc97e38028$$$opencur$closure$sharp$0$closecur$), position: Position { line: 6, column: 90, id: 29 } }), field: closure_0: Ref(ref$ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(ref$Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), field: val_ref: Ref(Slice$i32), position: Position { line: 6, column: 90, id: 29 } }), Field(FieldExpr { base: Local(Local { variable: _3: Ref(usize), position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 90, id: 29 } })], position: Position { line: 6, column: 90, id: 29 } }), field: val_int: Int, position: Position { line: 6, column: 98, id: 27 } })), substs: {} }, bb2: ExprBackwardInterpreterState { expr: Some(FuncApp(FuncApp { function_name: "builtin$unreach", type_arguments: [Int], arguments: [], formal_arguments: [], return_type: Int, position: Position { line: 6, column: 90, id: 25 } })), substs: {} }}
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode operand expr move _5
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode place _5
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode projection _5, []
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::high::types::fields] Encode value field for type 'Bool'
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::errors::error_manager] Register span MultiSpan { primary_spans: [prusti-tests/tests/verify/pass/issues/issue-812-1.rs:6:90: 6:98 (#0)], span_labels: [] } at position id 30
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::errors::error_manager] Register error BoundsCheckAssert of DefId(0:9 ~ issue_812_1[4c80]::test) at position id 30
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State before: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply statement _5 = Lt(_3, _4)
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_statement _5 = Lt(_3, _4), state: expr=undefined
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode place _5
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode projection _5, []
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] Encoding assignment to LHS Local(Local { variable: _5: Ref(bool), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] use_place Local(Local { variable: _5: Ref(bool), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] The state does not mention Local(Local { variable: _5: Ref(bool), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State before: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply statement _4 = Len((*(*((*_1).0: &&[i32]))))
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_statement _4 = Len((*(*((*_1).0: &&[i32])))), state: expr=undefined
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode place _4
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode projection _4, []
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] Encoding assignment to LHS Local(Local { variable: _4: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] use_place Local(Local { variable: _4: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] The state does not mention Local(Local { variable: _4: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State before: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply statement _3 = _2
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_statement _3 = _2, state: expr=undefined
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode place _3
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_encoder] Encode projection _3, []
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] Encoding assignment to LHS Local(Local { variable: _3: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] use_place Local(Local { variable: _3: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] The state does not mention Local(Local { variable: _3: Ref(usize), position: Position { line: 0, column: 0, id: 0 } })
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State before: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply statement StorageLive(_3)
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_statement StorageLive(_3), state: expr=undefined
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State after: ExprBackwardInterpreterState { expr: None, substs: {} }
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs:78:43
stack backtrace:
   0: rust_begin_unwind
             at /rustc/498eeb72f590e518e19746b346be53713689e207/library/std/src/panicking.rs:577:5
   1: core::panicking::panic_fmt
             at /rustc/498eeb72f590e518e19746b346be53713689e207/library/core/src/panicking.rs:135:14
   2: core::panicking::panic
             at /rustc/498eeb72f590e518e19746b346be53713689e207/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/498eeb72f590e518e19746b346be53713689e207/library/core/src/option.rs:752:21
   4: prusti_viper::encoder::mir::pure::pure_functions::encoder::PureFunctionEncoder::encode_body
             at ./prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs:78:25
   5: <prusti_viper::encoder::encoder::Encoder as prusti_viper::encoder::mir::pure::pure_functions::interface::PureFunctionEncoderInterface>::encode_pure_expression
             at ./prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs:216:24
   6: prusti_viper::encoder::mir::pure::specifications::encoder_poly::inline_closure
             at ./prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs:55:8
   7: prusti_viper::encoder::mir::pure::specifications::encoder_poly::encode_quantifier
             at ./prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs:167:42
   8: <prusti_viper::encoder::encoder::Encoder as prusti_viper::encoder::mir::pure::specifications::interface::SpecificationEncoderInterface>::encode_prusti_operation
             at ./prusti-viper/src/encoder/mir/pure/specifications/interface.rs:142:72
   9: <prusti_viper::encoder::mir::pure::pure_functions::interpreter::PureFunctionBackwardInterpreter as prusti_viper::encoder::mir_interpreter::BackwardMirInterpreter>::apply_terminator
             at ./prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs:597:44
  10: prusti_viper::encoder::mir_interpreter::run_backward_interpretation
             at ./prusti-viper/src/encoder/mir_interpreter.rs:69:30
  11: prusti_viper::encoder::mir::pure::pure_functions::encoder::PureFunctionEncoder::encode_body
             at ./prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs:76:21
  12: <prusti_viper::encoder::encoder::Encoder as prusti_viper::encoder::mir::pure::pure_functions::interface::PureFunctionEncoderInterface>::encode_pure_expression
             at ./prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs:216:24
  13: prusti_viper::encoder::mir::pure::specifications::encoder_poly::inline_spec_item
             at ./prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs:100:8
  14: <prusti_viper::encoder::encoder::Encoder as prusti_viper::encoder::mir::pure::specifications::interface::SpecificationEncoderInterface>::encode_assertion
             at ./prusti-viper/src/encoder/mir/pure/specifications/interface.rs:166:37
  15: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode_precondition_expr
             at ./prusti-viper/src/encoder/procedure_encoder.rs:3464:25
  16: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode_preconditions
             at ./prusti-viper/src/encoder/procedure_encoder.rs:3695:13
  17: prusti_viper::encoder::procedure_encoder::ProcedureEncoder::encode
             at ./prusti-viper/src/encoder/procedure_encoder.rs:352:9
  18: prusti_viper::encoder::encoder::Encoder::encode_procedure
             at ./prusti-viper/src/encoder/encoder.rs:608:36
  19: prusti_viper::encoder::encoder::Encoder::process_encoding_queue
             at ./prusti-viper/src/encoder/encoder.rs:851:40
  20: prusti_viper::verifier::Verifier::verify
             at ./prusti-viper/src/verifier.rs:247:9
  21: prusti_driver::verifier::verify
             at ./prusti/src/verifier.rs:51:39
  22: <prusti_driver::callbacks::PrustiCompilerCalls as rustc_driver::Callbacks>::after_analysis::{{closure}}
             at ./prusti/src/callbacks.rs:109:17
  23: rustc_interface::passes::QueryContext::enter::{{closure}}
             at /rustc/498eeb72f590e518e19746b346be53713689e207/compiler/rustc_interface/src/passes.rs:834:42
  24: rustc_middle::ty::context::tls::enter_context::{{closure}}
             at /rustc/498eeb72f590e518e19746b346be53713689e207/compiler/rustc_middle/src/ty/context.rs:1771:50
  25: rustc_middle::ty::context::tls::set_tlv
             at /rustc/498eeb72f590e518e19746b346be53713689e207/compiler/rustc_middle/src/ty/context.rs:1755:9
  26: rustc_middle::ty::context::tls::enter_context
             at /rustc/498eeb72f590e518e19746b346be53713689e207/compiler/rustc_middle/src/ty/context.rs:1771:9
  27: rustc_interface::passes::QueryContext::enter
             at /rustc/498eeb72f590e518e19746b346be53713689e207/compiler/rustc_interface/src/passes.rs:834:9
  28: <prusti_driver::callbacks::PrustiCompilerCalls as rustc_driver::Callbacks>::after_analysis
             at ./prusti/src/callbacks.rs:71:9
  29: <rustc_interface::interface::Compiler>::enter::<rustc_driver::run_compiler::{closure#1}::{closure#2}, core::result::Result<core::option::Option<rustc_interface::queries::Linker>, rustc_errors::ErrorReported>>
  30: rustc_span::with_source_map::<core::result::Result<(), rustc_errors::ErrorReported>, rustc_interface::interface::create_compiler_and_run<core::result::Result<(), rustc_errors::ErrorReported>, rustc_driver::run_compiler::{closure#1}>::{closure#1}>
  31: rustc_interface::interface::create_compiler_and_run::<core::result::Result<(), rustc_errors::ErrorReported>, rustc_driver::run_compiler::{closure#1}>
```

These statements in particular caught my eye:

```
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply terminator Terminator { source_info: SourceInfo { span: prusti-tests/tests/verify/pass/issues/issue-812-1.rs:6:90: 6:98 (#0), scope: scope[0] }, kind: assert(move _5, "index out of bounds: the length is {} but the index is {}", move _4, _3) -> [success: bb1, unwind: bb2] }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_terminator Terminator { source_info: SourceInfo { span: prusti-tests/tests/verify/pass/issues/issue-812-1.rs:6:90: 6:98 (#0), scope: scope[0] }, kind: assert(move _5, "index out of bounds: the length is {} but the index is {}", move _4, _3) -> [success: bb1, unwind: bb2] }, states: ...
....
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] State before: ExprBackwardInterpreterState { expr: None, substs: {} }
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir_interpreter] Apply statement _5 = Lt(_3, _4)
[2022-02-16T19:13:53Z TRACE prusti_viper::encoder::mir::pure::pure_functions::interpreter] apply_statement _5 = Lt(_3, _4), state: expr=undefined
```

The terminator [from block bb0](https://user-images.githubusercontent.com/81013316/153728974-2c14c048-03b3-45c8-8cec-ad694e5624c4.svg) results in a fully empty state, and this `None`/`undefined` state then carries over to the previous statement during the backwards MIR interpretation (and so on). Instead, I would expect to see something resembling preserving at least some of the state and then interpreting the Assert terminator more like a Goto terminator (when encoding triggers).

I'm still poking round the codebase to test this theory and find out how this might be implemented, but wanted to share the insight here already in case it helps :)

Pointerbender:
A couple of new insights to go with the new push from just now:

* In step 1, returning `states[target].clone()` instead of `ExprBackwardInterpreterState::new(None)` seems to do the trick of emulating the Goto terminator behavior when encoding quantifier triggers.
* There is currently still a hack in `encode_quantifier` to strip the `val_XXX` field from a trigger expression such as `slice[i]`. My gut feeling says there is probably a better way to solve this :) (should the field expression even be generated? I don't see it in any of the other non-trigger snapshot expressions)
* The test suite passed locally (~although I did not run it all the way to the end~ it finished okay after posting this), the CI didn't start when I moved the PR back to "ready for review" for some reason, a force push also didn't trigger it.
* We don't have a solution yet for when end-users put `panic!()` as a trigger, this could also use an extra test case which is currently not there.

fpoli:
That seems a bug, indeed. A `return` terminator should generate a state where the expression is `Some(_0)`.

fpoli:
> In step 1, returning `states[target].clone()` instead of `ExprBackwardInterpreterState::new(None)` seems to do the trick of emulating the Goto terminator behavior when encoding quantifier triggers.

Thanks for fixing this! Yes, I completely agree. I didn't remember that `assert` terminators have a target to be executed when the assertion succeeds.

fpoli:
> the CI didn't start when I moved the PR back to "ready for review" for some reason, a force push also didn't trigger it.

I rebased on master. We had a bug in the workflows.

fpoli:
> We don't have a solution yet for when end-users put `panic!()` as a trigger, this could also use an extra test case which is currently not there.

We should eventually report an error instead of crashing on the `unwrap()`, but it doesn't need to be in this PR.

fpoli:
> There is currently still a hack in `encode_quantifier` to strip the `val_XXX` field from a trigger expression such as `slice[i]`. My gut feeling says there is probably a better way to solve this :) (should the field expression even be generated? I don't see it in any of the other non-trigger snapshot expressions)

@vakaras @Aurel300 Any ideas? If not, I'd merge the PR.

Aurel300:
I guess the `.val_...` is there because we are encoding the trigger to match an access in an impure context, right? In which case it's not pretty but probably works for now.

fpoli:
bors cancel

(all tests were passing)

Pointerbender:
Thank you! I appreciate all the help, this was a fun one to fix and I learned a bit more about the MIR to VIR conversion :smile:

PR #865, opened 2022-02-16
Type models: Documentation

vl0w:
Documentation PR #836

### DIFF
--- a/user-guide/src/SUMMARY.md
+++ b/user-guide/src/SUMMARY.md
@@ -28,4 +28,5 @@
   - [Trait contract refinement](verify/traits.md)
   - [Closures](verify/closure.md)
   - [Specification entailments](verify/spec_ent.md)
+  - [Type Models](verify/type-models.md)
 - [Specification Syntax](syntax.md)
--- a/user-guide/src/SUMMARY.md
+++ b/user-guide/src/SUMMARY.md
@@ -28,5 +28,5 @@
   - [Trait contract refinement](verify/traits.md)
   - [Closures](verify/closure.md)
   - [Specification entailments](verify/spec_ent.md)
-  - [Type Models](verify/type-models.md)
+  - [Type models](verify/type-models.md)
 - [Specification Syntax](syntax.md)
--- a/user-guide/src/verify/type-models.md
+++ b/user-guide/src/verify/type-models.md
@@ -1,0 +1,60 @@
+# Type Models
+
+Some structs, mostly non-local structs from other crates, contain non-public code which are essential for verification.
+An example is the `std::slice::Iter` struct from the standard library. 
+We would like to provide a specification for the `Iterator` implementation of `Iter<'_, i32>`:
+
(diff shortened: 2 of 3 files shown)


user-guide/src/SUMMARY.md
```diff
@@ -28,4 +28,5 @@
   - [Trait contract refinement](verify/traits.md)
   - [Closures](verify/closure.md)
   - [Specification entailments](verify/spec_ent.md)
+  - [Type Models](verify/type-models.md)
```
Aurel300: Please also add the link in the file`verify/summary.md`.

vl0w: Done

user-guide/src/verify/type-models.md
```diff
@@ -0,0 +1,60 @@
...
+    // ... more ...
+    fn next(&mut self) -> Option<&'a T>;
+}
+```
+
+## Remarks
+* A model needs to be copyable, i.e. all fields need to be `Copy`.
+* The model is only to be used in specifications (pre- and postconditions of functions or methods)
+and never in code which will be executed. Using `.model()` in non-spec code will cause a Prusti error.
+* Models can not be generic, i.e. creating a model for `Vec<T>` is not possible. It is however
```
Aurel300, line 54: But this is a temporary limitation?

user-guide/src/verify/type-models.md
```diff
@@ -0,0 +1,60 @@
...
+  possible to model concrete generic types, i.e. `Vec<i32>` and `Vec<u32>`.
+
+> **Caution**
+> 
+> As of yet, iterators are not fully supported in Prusti. 
+> The example above is a draft and is meant as a possible real-world usage of `#[model]` in the near future.
```
Aurel300, line 60: I agree, and maybe it would be wiser to explain the feature with something simpler, even if the code is very artificial. Especially e.g. the spec for `next` can be quite confusing to look at in this section.

user-guide/src/verify/type-models.md
vl0w: I guess you are referring to the last bullet point?
=> Redirecting to the [feature PR discussion](https://github.com/viperproject/prusti-dev/pull/836#discussion_r802032594)

user-guide/src/verify/type-models.md
vl0w: You might be right, especially about the `next` spec. I dislike artificial examples, however :sweat:

fpoli:
Could you move these changes to the `docs/` folder of the `master` branch? We are going to disable/delete `gh-pages-src`.

user-guide/src/verify/type-models.md
Aurel300: Yes, it's that discussion but the point here is that I think we should say that it is a temporary limitation.

vl0w:
Docs are now updated as part of PR #836

PR #869, opened 2022-02-17
Fix the path filter of worflows

fpoli:
This way, tests should run as expected.


fpoli:
Yep

Aurel300:
Is it not `paths-ignore` @fpoli ? :D

fpoli:
Oh, you are right. It's really bad that it doesn't error on unknown keys.

PR #876, opened 2022-02-19
use array indices as quantifier triggers for `selection_sort` test

Pointerbender:
Now that there is support for array/slice indices in quantifier triggers (#812), I updated the `selection_sort.rs` test case to make use of this. This results in a ~15 second faster verification time :tada: increasing performance further per #819.

Prusti can't verify this test yet when using array assignments instead of `set()`, so there are still 2 calls to `set()` in the test with a FIXME to address these later. Will post a separate comment with some ideas about what that would take.


vakaras (approved):
Thank you!

PR #880, opened 2022-02-21
Spec refinement refactoring

vl0w:
This is a refactoring of the refinement of specifications. 
Refinement of specifications happens lazily ("on the fly") whenever they are accessed in a separate module. Previously, refinements were conducted in `ProcedureEncoder::encode_spec_refinement`.

In particular:
- There is a new module `mir/specifications` which provides access to everything specification related (e.g. `is_pure`, `is_trusted`).
- The module initially contains user-typed specifications (`DefSpecificationMap`). When a query for a `DefId` is submitted, the module checks whether there is an applicable user-typed specification and then refines that specification if needed.
    - Refinement happens when the `DefId` belongs to an implementation of a trait method with a user-typed specification
- This makes the actual refinement in `ProcedureEncoder::encode_spec_refinement` obsolete (we just need to encode the already refined specs)
- Additionally, `ProcedureSpecification`s contain information about the refinement: 
  Everything that is _refinable_ can be wrapped in a `SpecificationItem<T>`, which has variants `Empty`, `Inherent`, `Inherited` and `Refined`. For example, given preconditions `PreTrait` for a trait method `trait_def_id` and preconditions `PreImpl` for a corresponding impl method `impl_def_id`, a query to the preconditions of `impl_def_id` will return a `SpecificationItem::Refined(PreTrait, PreImpl)`

Apart from the refactoring, there is a small change how Prusti interprets `#[pure]` annotations:
When an implementation of a trait method is not marked pure but the trait method is marked pure, Prusti treats the implementation as pure anyways. In other words: `#[pure]` annotations will be inherited by trait method implementations.


prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -0,0 +1,120 @@
...
+                .specs
+                .borrow_mut()
+                .compute_is_trusted(self.env(), def_id)
+        });
+        trace!("is_trusted {:?} = {}", def_id, result);
+        *result
+    }
+
+    fn get_predicate_body(&self, def_id: DefId) -> Option<LocalDefId> {
+        let mut specs = self.specifications_state.specs.borrow_mut();
```
Aurel300, line 72: Does this need to be a local variable? Why not `.borrow_mut` in a chain like the other methods?

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn get_loop_specs(&self, def_id: DefId) -> Option<typed::LoopSpecification> {
+        self.specifications_state
+            .specs
+            .borrow()
+            .get_loop_spec(def_id)
+            .cloned()
+    }
+
+    fn get_procedure_specs(&self, def_id: DefId) -> Option<typed::ProcedureSpecification> {
+        let mut specs = self.specifications_state.specs.borrow_mut();
```
Aurel300, line 105: (same comment: does this need to be a local var?)

prusti-viper/src/encoder/mir/specifications/specs.rs
```diff
@@ -0,0 +1,139 @@
...
+use prusti_interface::{
+    environment::Environment,
+    specs::typed::{DefSpecificationMap, LoopSpecification, ProcedureSpecification, Refinable},
+};
+use rustc_hash::FxHashMap;
+use rustc_hir::def_id::{DefId, LocalDefId};
+use std::collections::HashMap;
+
+/// Provides access to specifications, handling refinement if needed
+pub(super) struct Specifications {
```
Aurel300, line 11: Can this state (actually the fields of `SpecMap`) not be fields of the `SpecificationsState` struct directly?

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -0,0 +1,120 @@
...
+use log::trace;
+use prusti_interface::specs::{typed, typed::DefSpecificationMap};
+use rustc_hash::FxHashMap;
+use rustc_hir::def_id::{DefId, LocalDefId};
+use std::cell::RefCell;
+
+pub(crate) struct SpecificationsState {
+    specs: RefCell<Specifications>,
+    pure_cache: RefCell<FxHashMap<DefId, bool>>,
+    trusted_cache: RefCell<FxHashMap<DefId, bool>>,
```
Aurel300: Why is there a separate "cache" for these? IIUC to know if a method is pure or trusted, we need to first figure out its specification set anyway. That is, the `SpecMap` should always contain an entry for the `DefId`, so we might as well look up pureness or trustedness in that map, right?

prusti-interface/src/specs/typed.rs
```diff
@@ -32,79 +45,232 @@ impl SpecificationSet {
...
-            trusted: false,
-        }
-    }
-    /// Trait implementation method refinement
-    /// Choosing alternative C as discussed in
-    /// https://ethz.ch/content/dam/ethz/special-interest/infk/chair-program-method/pm/documents/Education/Theses/Matthias_Erdin_MA_report.pdf
-    /// pp 19-23
-    ///
-    /// In other words, any pre-/post-condition provided by `other` will overwrite any provided by
-    /// `self`.
```
Aurel300: Can we keep this comment somewhere nearby?

vl0w:
@Aurel300 Are you aware of any other methods that I should put into `mir/specifications`?

Aurel300:
`Encoder::is_spec_closure` would also fit.

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: Rust complains: `temporary value dropped while borrowed E0716`

prusti-interface/src/specs/typed.rs
vl0w: I think the refactoring obfuscated the "refinement strategy" completely. The strategy is deeply hidden in `SpecItemIter`, which I am not super happy about in retrospective.
I think the we could add methods to `ProcedureSpecification` to get the `LocalDefId`s of pre-/postconditions. There it will be clearer what what actual specification will be encoded.

vl0w: For example, to get the preconditions, we would have the following function:
```rust
    pub fn get_preconditions(&self) -> Vec<LocalDefId> {
        match self.pres.clone() {
            SpecificationItem::Empty => vec![],
            SpecificationItem::Inherent(items) => items,
            SpecificationItem::Inherited(items) => items,
            SpecificationItem::Refined(_, items) => items,
        }
    }
```
Which makes it pretty explicit that we use the (C) strategy. 
The `clone()` is a bit ugly though and was the reason of the weird `SpecItemIter`. In an ideal world, I would want `get_preconditions` to return a slice, which is (I think) not possible due to the `SpecificationItem::Empty` variant. Technically, we could drop this variant, but I like it (especially when debugging -> one can observe that the user defined did not specify anything).

Aurel300: Regardless of how the implementation achieves the goal, we should document that this is still our refinement strategy. Reworded a little bit it can be the module-level documentation for `mir/specifications` for example.

vl0w: I agree that the documentation needs to be somewhere and ideally, it is also immediately obvious when viewing the code.

I think `mir/specifications` is not the correct place for this doc. My refactoring basically just adds more information to specs (see changes in `prusti-interface/specs/typed.rs`) and `mir/specifications` just provides (lazy) access to these enriched specifications (including computing refinements if needed).
The actual to-be-encoded assertions are accessed in `ProcedureContractGeneric::{functional_precondition, functional_postcondition, pledges}`, i.e. this is the place where we can choose how to handle this additional information.

vl0w: I made the usage of the strategy more explicit by rearranging stuff in `prusti-interface/specs/typed.rs`

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: Caches removed

prusti-viper/src/encoder/mir/specifications/specs.rs
vl0w: I removed one layer of indirection (the `SpecMap`).
I would still like to keep `Specifications`, which is responsible for invoking the refinement.

vl0w:
> Encoder::is_spec_closure

Done

prusti-interface/src/specs/typed.rs
```diff
@@ -96,96 +95,74 @@ pub enum SpecificationItem<T> {
...
+            SpecificationItem::Refined(from, to) => Some((Some(from), to)),
         }
     }
 }
 
-pub struct SpecItemIter<'a, T> {
-    items: Option<&'a Vec<T>>,
-    pos: usize,
+/// A type to access possibly refined [SpecificationItem] values with different strategies
+pub struct ExtractedRefinement<'a, T> {
```
Aurel300: Why this extra wrapper? Can the "refinement strategies" not be defined directly on `SpecificationItem`?

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -44,36 +39,43 @@ pub(crate) trait SpecificationsInterface {
...
-        });
+        let result = self
+            .specifications_state
+            .specs
+            .borrow_mut()
+            .get_and_refine_proc_spec(self.env(), def_id)
+            .and_then(|spec| {
+                spec.trusted
+                    .extract_refinements()
+                    .with_selective_replacement()
```
Aurel300: `is_pure` looks different that `is_trusted`: I'm guessing the intent is that `pure` is inherited but `trusted` is not. However, there is no way to declare a function as *not* trusted, correct? So if the base method is trusted, `with_selective_replacement` will always inherit the trustedness?

prusti-interface/src/specs/typed.rs
vl0w: They could be directly on `SpecificationItem`. The wrapper makes it more readable imo.

Aurel300: Not sure I agree :) I would especially like to avoid the call to `.extract_refinements()` at the call sites.

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: Yes, I'm not aware of any way the user can mark a function as "not trusted".
The default, when there is no specification, is that a function is not trusted though (same for pure).

`with_selective_replacement` does not inherit anything, it ignores the possibly refined-from value. This is why there is the separate `inherit_refined` which can be used for the pure flag.

I am not sure whether we should handle `#[pure]` and `#[trusted]` equally in this case. When we attach an extern spec to a trait, the trait's specification will automatically be trusted, even if the user does not annotate it directly. The refinement would then inherit this trustedness on any local impl.

Aurel300: Ok, thanks. I agree that `pure` and `trusted` should be treated differently. But I wrongly assumed `with_selective_replacement` would inherit. All is good.

prusti-interface/src/specs/typed.rs
vl0w: I personally like the idea to make explicit that we are extracting information from a `SpecificationItem`, i.e. preparing it to be used for encoding. 

But this is of course just cosmetics, these methods can all life on `SpecificationItem` directly, I guess this is what you suggest.

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: Hmm, the name maybe seems a bit "weird", but this is basically how "strategy (C) in the paper" is called (the code docs refer to that).

Apart from that, with the new structure of the `DefSpecificationMap`, we could also say that `ProcedureSpecification::{pure, trusted}` are `SpecificationItem::Empty` by default to account for non-user-typed specs and resolve it later.

prusti-interface/src/specs/typed.rs
vl0w: I've moved the methods to `SpecificationItem`

Issue #56, opened 2020-06-12
Add support for quantifying over copy types


vakaras:
Currently, Prusti allows quantification only over primitive types. However, once we have snapshots as part of #53, we could support quantification over `Copy` types by quantifying over snapshots.

FabianWolff:
Speaking of quantification, how would you write a specification for this function?
```Rust
fn foo (v: &mut Vec<&mut i32>) {
    for num in v {
        **num += 1;
    }
}
```
In Rust (from my understanding), there isn't even a clear meaning of "`v` contains the same references", because references are treated "by-value", meaning that `v` actually _doesn't_ contain the same references.

So perhaps something like this should work:
```
ensures: forall ref: &mut i32 :: old(v).contains(&ref)
             ==> (exists nref: &mut i32 :: v.contains(&nref) && *nref == *ref + 1)
```
But this needs both existential and universal quantification over a non-copy type (namely, `&mut i32` in this example).

And it gets worse:
```Rust
fn bar (v: &mut Vec<&mut i32>) {
    for num in v {
        **num = 0;
    }
}

let mut x = 1;
let mut v = vec! [&mut x];
bar (&mut v);
assert_eq! (x, 0);
```
Here, in the postcondition of `bar()`, it is _not_ enough to say that all references in `v` dereference to zero, we _also_ need to know that `&mut x` is still in `v` in order to prove the assertion...

What are your thoughts on this?

cmatheja:
I'm not entirely sure about these examples because we do not properly support Vec yet.

Assuming support for Vec, I guess we can deal at least with the first example without quantifying over references. It should suffice to quantify over all positions in the vector, i.e., we could try something along the lines of

```
old(v).len() == v.len() && forall i: i32 :: old(*v[i]) == *v[i] + 1
```

The second example seems to be more of an issue.

fpoli:
To annotate your `foo` and `bar` functions I would rather quantify over the index and use something like `old(*v[i]) == *v[i] + 1` in the postcondition, because there are many tricky things going on:
* `&mut i32` is not a copy type
* the semantics of quantifying over references is undefined (e.g. should it quantify over existing references? over all memory regions? over all memory regions for which holding a mutable reference is not UB? ...)
* types like `&mut Vec<&mut i32>` that contain references are not yet supported, mostly because of examples involving reborrowing (btw, @frececroka you might be interested in Fabian's examples)

cmatheja:
[...]

PR #882, opened 2022-02-22
Small fixes

### LINKED ISSUE #878: Incorrect encoding of enum discriminant type
When encoding discriminants Prusti assumes that their type is `isize` (encoding [here](https://github.com/viperproject/prusti-dev/blob/7977d30a028edfe0e300bb2931992d4029385b41/prusti-viper/src/encoder/high/pure_functions/interface.rs#L59)), but this isn't always the case and so Prusti may panic due to the call on this [line](https://github.com/viperproject/prusti-dev/blob/7977d30a028edfe0e300bb2931992d4029385b41/prusti-viper/src/encoder/mir/pure/interpreter/mod.rs#L232) since the `encoded_lhs` has the correct type. An example which triggers this error is:
```rust
use prusti_contracts::*;
use std::cmp::{Ord, Ordering};

fn main() {}

#[extern_spec]
trait Ord {
    #[pure]
    fn cmp(&self, other: &Self) -> Ordering;
}

#[pure]
pub fn eq<T: Ord>(a: &T, b: &T) -> bool {
    match a.cmp(b) {
        Ordering::Equal => true,
        Ordering::Less => false,
        Ordering::Greater => false,
    }
}
```
Where:
```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
  left: `Int(I8)`,
 right: `Int(Isize)`: _6 → discriminant(_3)', vir/src/../gen/vir_gen.rs:12610:25
```

JonasAlaif:
Kinda fixes #878 but would prefer if @vakaras looked at it/made it less hacky


prusti-viper/src/encoder/foldunfold/footprint.rs
```diff
@@ -115,9 +115,6 @@ impl ExprFootprintGetter for vir::Expr {
                 box body,
                 ..
             }) => {
-                assert!(variables
```
Aurel300, line 118: Is this meant to be part of this PR? If we do this we should look into actually using the `is_quantifiable` function (there is a TODO somewhere in the code mentioning that).

vakaras (approved):
I am not sure about the removed `assert`s, but other changes look good to me.

prusti-viper/src/encoder/foldunfold/footprint.rs
JonasAlaif: When I removed the same assert that you mentioned in https://github.com/viperproject/prusti-dev/issues/56#issuecomment-1046966018 I ended up running into this one causing a panic. Since it was the same one I had already removed (in [foldunfold/requirements.rs](https://github.com/viperproject/prusti-dev/blob/7977d30a028edfe0e300bb2931992d4029385b41/prusti-viper/src/encoder/foldunfold/requirements.rs#L328-L330)), I tried removing it here as well. @vakaras I agree that maybe we don't want to entirely remove the asserts, but just modify `!var.typ.is_typed_ref_or_type_var()` to `!var.typ.is_typed_ref()` or something?

vakaras: > @vakaras I agree that maybe we don't want to entirely remove the asserts, but just modify `!var.typ.is_typed_ref_or_type_var()` to `!var.typ.is_typed_ref()` or something?

Yes.

JonasAlaif:
@vakaras if I convert this line:
https://github.com/viperproject/prusti-dev/blob/d53bdd7af1dda31d763e5fce83b2829aa1164a79/prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs#L137
to:
```rust
let qvar_ty = encoder.encode_type(arg_ty).unwrap().convert_to_snapshot();
```
Quantification over generic types works... but I probably shouldn't use `convert_to_snapshot` here, but instead `SnapshotEncoder::encode_type`. The issue is that `fn encode_quantifier()` only gets a `encoder: &Encoder` and not `SnapshotEncoder` - is there any way to get the latter to this function?

vakaras:
> @vakaras if I convert this line:
> 
> https://github.com/viperproject/prusti-dev/blob/d53bdd7af1dda31d763e5fce83b2829aa1164a79/prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs#L137
> 
> to:
> 
> ```rust
> let qvar_ty = encoder.encode_type(arg_ty).unwrap().convert_to_snapshot();
> ```
> 
> Quantification over generic types works... but I probably shouldn't use `convert_to_snapshot` here, but instead `SnapshotEncoder::encode_type`. The issue is that `fn encode_quantifier()` only gets a `encoder: &Encoder` and not `SnapshotEncoder` - is there any way to get the latter to this function?

Isn't the snapshot encoder, a field of the `encoder`? Anyway, the code was written by @Aurel300, so he could probably tell more.

Aurel300:
Yeah the snapshot encoder works with the usual interface trait + state mechanism now. So, if you import `crate::encoder::snapshot::SnapshotEncoderInterface`, you can call the [`encode_snapshot_type`](https://github.com/viperproject/prusti-dev/blob/d53bdd7af1dda31d763e5fce83b2829aa1164a79/prusti-viper/src/encoder/snapshot/interface.rs#L39) method on `Encoder`. It also takes a `SubstMap` (which should eventually be `SubstRef` maybe) so generics should get properly monomorphised at the callsite.

prusti-viper/src/encoder/foldunfold/footprint.rs
JonasAlaif: Removed by [this recent commit](https://github.com/viperproject/prusti-dev/commit/07f7d6ce5fb96faadbbb58122ae5052778f44b3c#diff-464a6cfdb5cefcf2b9e5ae3d8b06fadd22f506153b31b3791230611df9671b8aL328-L330) so this is no longer an issue

JonasAlaif:
Requires #895 now

JonasAlaif:
Will merge this today or tomorrow if there are no objections? Though it's a small change it should actually improve generics support quite a bit

Aurel300:
@JonasAlaif Can I try to get #899 in first?

JonasAlaif:
@Aurel300 sounds good, looks like that might also fix some of the issues I might be running into with the test I just added. Let's see

Aurel300:
Status: I found some problems with generics and extern specs still. WIP but I think the test you added @JonasAlaif should be supported soon.

(Some notes:)

The wrappers (from extern spec desugaring) are not inherited the same way that `is_pure` is. This can result in an annoying de-sync where an implementation (e.g. of `Ord`) is known to be pure but cannot be encoded because it would attempt to encode a non-local `DefId`. Extern spec wrappers should really disappear completely, and the `DefSpecificationMap` should have keys of `DefId` instead of only `LocalDefId`.

The reason the wrappers exist in the first place is because `PureFunctionEncoder` (and by extension the backwards interpreter and MIR encoder) are mis-used to encode call infos (`encode_pure_function_use`) and Prusti expressions (`encode_body`). I extracted the latter to a separate method and made fewer things in the `PureFunctionEncoder` depend on an MIR body to fix the former.

Issue #884, opened 2022-02-23
Wrong encoding of snapshots in postcondition


JonasAlaif:
The following snippet causes a fold-unfold error in `main`. However, this is misleading and not due to the fold-unfold algorithm.
```rust
use prusti_contracts::*;

#[pure]
fn foo<T>(y: &T) -> bool { true }
#[ensures(foo(&x))]
fn bar<T>(x: T) {}

struct S(i32);
fn main() {
    let s = S(0);
    bar(s);
}
```
The real culprit is that the `foo(&x)` postcondition is encoded as `foo(snap(x))` rather than the correct `foo(old(snap(x)))`.
This means that at the call of `bar(s)` we generate:
```vim
label l1
exhale acc(struct_S(s), write)
inhale foo(snap(s))
```
rather than the correct:
```vim
inhale foo(old[l1](snap(s)))
```

JonasAlaif:
There may be other related issues to this, but I found them impossible to find due to the super ambiguous error of `cannot generate fold-unfold Viper statements`

fpoli:
> this is misleading and not due to the fold-unfold algorithm

In general, a fold-unfold error means that the encoding has permission errors that cannot be fixed just by folding or unfolding predicates. For example, trying to exhale a predicate twice.

The current error message is
```
cannot generate fold-unfold Viper statements. The required permission Pred(_3, read) cannot be obtained.
```

Would it be better to expand it to the following?
```
Details: cannot generate fold-unfold Viper statements. The required permission Pred(_3, read) cannot be obtained by folding or unfolding other predicates.
```

fpoli:
In case of fold-unfold error, it might also be useful to to generate a special dump, mentioning in the error message the specific CFG block and statement where the error occurred.

JonasAlaif:
[...]

PR #885, opened 2022-02-25
Move PositionManager out of ErrorManager

fpoli:
This PR is a refactoring that extracts a `PositionManager` from the `ErrorManager`. In order to translate VIR back to MIR and to the source code (see the discussion at https://github.com/viperproject/prusti-dev/issues/884), each VIR expression and statement should have their own unique position id. This could be done by a future VIR factory that generates positions via a reference to the `PositionManager`.

Notes:
* `ErrorManager::register` becomes `ErrorManager::register_error`
* `Encoder::encode_assertion` loses its `ErrorCtxt` argument. The caller should use `ErrorManager::set_error` on the returned expression to set the error context.
* Ideally, each expression should have its own `Position` obtained via `ErrorManager::register_span`. This is not mandatory, but it would make error reporting and the back-translation from VIR to source code more precise.

### DIFF
--- a/prusti-viper/src/encoder/errors/error_manager.rs
+++ b/prusti-viper/src/encoder/errors/error_manager.rs
@@ -5,18 +5,18 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use vir_crate::polymorphic::Position;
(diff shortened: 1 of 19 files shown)


fpoli:
bors merge

PR #887, opened 2022-02-28
Automatically add loop guards without preconditions as invariants

### LINKED ISSUE #448: Derive the body invariant from the loop condition
Currently, the following fails to check:

```rust
fn index_all<T>(slice: &[T]) {
    let mut i = 0;
    while i < slice_len(slice) {
        let el = index_slice(slice, i);
        i += 1;
    }
}
```

Though adding `body_invariant!(i < slice_len(slice))` into the loop body fixes the issue, Prusti could derive the invariant from the loop condition.

### LINKED ISSUE #995: While let fails with internal error due to discriminant being havocked by loop invariant
The following example:
```rust
#![feature(box_patterns)]

use prusti_contracts::*;

enum List<T> {
    Nil,
    Cons(Node<T>),
}
use List::*;
type Node<T> = Box<(T, List<T>)>;

fn destruct<T>(l: List<T>) {
    let mut head = l;
    while let Cons(mut curr) = head {
        head = curr.1;
        //body_invariant!(true);
    }
}

fn main() {}
```
fails to verify with the following error:
```
error: [Prusti internal error] Prusti encountered an unexpected internal error
  |
  = help: This could be caused by too small assertion timeout. Try increasing it by setting the configuration parameter ASSERT_TIMEOUT to a larger value.
  = note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
  = note: Details: unregistered verification error: [unfold.failed:insufficient.permission] Unfolding m_List$_beg_$__TYPARAM__$_T$0$__$_end_Cons(_2.enum_Cons) might fail. There might be insufficient permission to access m_List$_beg_$__TYPARAM__$_T$0$__$_end_Cons(_2.enum_Cons) (<no position>)

Verification failed
```
The failure occurs because the loop invariant is injected before `head = curr.1;` and it havocs the knowledge that `head` is `Cons`. A workaround would be manually specifying `body_invariant!` at a different point in the loop.

@JonasAlaif Shouldn't your change prevent this? 

The credit for finding this bug goes to Xavier Denis.

JonasAlaif:
Currently, when encoding loops:
```rust
while { g = G; g } { B1; invariant!(I); B2 }
```
Prusti will translate this into the Viper:
```boogie
// Encoding 1
g = G
if (g) {
  B1
  exhale I
  // ... havoc local variables modified in G, B1, or B2
  inhale I
  B2
  g = G
  if (g) {
    B1
    exhale I
    assume false
  }
  // We know !g at this point
}
```
We can try to assume the loop guard by adding the following:
```boogie
// Encoding 2
g = G
if (g) {
  B1
  exhale I
  // ... havoc local variables modified in G, B1, or B2

  g = G
  assume g
  B1

  inhale I
  ...
}
```
This works great, except for when `G` or `B1` include `exhale`/`assert` statements (including function calls with preconditions). In such a case, these statements would very likely fail (everything has been havoced, and we have-not/cannot yet assume the invariant) resulting in an unavoidable verification failure. The solution I take is to use encoding `2` if we find no such preconditions, and otherwise fall back to encoding `1` with a warning to the user.
Fixes #448, fixes #995


JonasAlaif:
The check for `fun_preconds` is quite surely wrong - I'll need help figuring out where to find the preconditions of relevant functions (I assume that non-pure method calls will already have been inlined as `requires`/`ensures`)

prusti-interface/src/prusti_error.rs
```diff
@@ -100,6 +100,17 @@ impl PrustiError {
         )
     }
 
+    /// Issue with loop invariants (e.g. cannot automatically include loop guard)
+    pub fn loop_invariant<S: ToString>(message: S, span: MultiSpan) -> Self {
```
fpoli: We could rename this method to `warning` to make it more reusable.

JonasAlaif: Sure, would the `[Prusti: warning_category] ...` (the `warning_category` part) be passed as an argument?

fpoli: Oh, I was thinking of leaving it as `[Prusti: warning]`. If we keep the number of kinds of messages low, then we can (in the future) easily generate a summary like "verified with xx unsupported features, yy invalid specifications, zz verification errors and qq warnings". If we choose more fine-grained categories (loop invariants and each language feature) the summary would be harder to generate.

fpoli: Since we don't currently generate that summary, in the end I'm fine either way.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -510,6 +512,115 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+            fn walk_assert(&mut self, statement: &vir::Assert) {
+                self.0.push(statement.expr.clone());
+            }
+            fn walk_method_call(&mut self, statement: &vir::MethodCall) {
+                // Note: We know that in Prusti method's preconditions and postconditions are empty
+                for arg in &statement.arguments {
+                    self.walk_expr(arg);
+                }
+            }
+            fn walk_expr(&mut self, expr: &vir::Expr) {
```
Aurel300: Why not use an `ExprWalker` here?

Aurel300:
I think this approach is too hacky. We generate stuff (the invariant) then figure out if what we generated fits into a particular shape (a "pure" invariant). It might also be a bit unpredictable as far as the user is concerned.

prusti-viper/src/encoder/procedure_encoder.rs
JonasAlaif: Hmm I missed that since it's in a separate file compared to the definition of `Expr`, whereas the Statement one isn't

prusti-interface/src/prusti_error.rs
JonasAlaif: No, your suggestion sounds good

JonasAlaif:
@Aurel300 maybe you saw the version from the first commit, but the second one 1b9ed1a fixes the issue with unnecessarily generating stuff. The way it now works is that we generate (line by line):
```rust
g = G
if (g) {
  B1
  exhale I
  // ... havoc local variables modified in G, B1, or B2

  // Check if the above generated G and B1 have preconditions
  // Only generate if yes:
  g = G
  assume g
  B1
 
  // Continue as normal
  inhale I
  ...
```
There is no looking back at what we generated anymore.
Regarding the second point, I think that making use of the `ExprFolder` has made things more durable, to the point that it's very predictable to a user which encoding is used. Most of the time if we decide to not use the new Encoding 2, then one couldn't simply just copy and paste the loop guard as a `body_invariant!`. Moreover, Prusti will emit a warning in such a case, telling you to manually try copy-pasting it, if that's what you need.
I do agree though that this isn't ideal. A better solution would be to always use Encoding 2 and then go through and `assume`/`inhale` the required "preconditions" at the correct places.

prusti-interface/src/prusti_error.rs
```diff
@@ -100,6 +100,17 @@ impl PrustiError {
         )
     }
 
+    /// Issue with loop invariants (e.g. cannot automatically include loop guard)
```
Aurel300: (also change the comment here)

Aurel300:
> There is no looking back at what we generated anymore.

What I'm referring to is that you walk the generated VIR at all, and figure out which functions were called based on matching `FuncApp` and which methods based on `Exhale`s. I think more ideally we would have a way to remember that information as we were encoding the loop guard in the first place.

fpoli:
> What I'm referring to is that you walk the generated VIR at all, and figure out which functions were called based on matching `FuncApp` and which methods based on `Exhale`s.

I agree with this; it's the same issue that we have with the fold-unfold pass. A consequence is that the "it requires the viper preconditions: ..." messages in this PR are going to be as unreadable as the current fold-unfold errors are. The difference is that fold-unfold errors are always bugs of some kind, so the user is not supposed to see them in the long term, while these loop invariant warnings only exist to help the user, so they should be fully understandable.

> I think more ideally we would have a way to remember that information as we were encoding the loop guard in the first place.

Either that, or `block_preconditions` could be redefined on MIR blocks, collecting readable reasons why the block cannot be added to a loop invariant. The existing `block_preconditions` *on VIR* could still be used in a `debug_assert!(..)` to keep the two definitions in sync.

JonasAlaif:
It seems that with `Change asserts to assumes after unrolling` this actually succeeds in unrolling every time except for when it runs into pure functions. The impure ones are handily inlined with the functional spec encoded with `assert`s rather than `exhale`s. Therefore I think we could, in most cases, just report the set of pure functions which prevented the unrolling - this would be readable, without much extra work. Otherwise have just some message that the user should manually add the loop guard as a body_invariant, without further explanation (ideally with some way to disable the warning)

Issue #877, opened 2022-02-20
Snap encoding of array assignments


Pointerbender:
This issue is a spin-off from #819, which revealed that currently Prusti is unable to verify programs like:

```rust
#[ensures(result <= 3)]
fn foo() -> usize {
    let mut a = [1, 2, 3];
    let mut i = 0;
    while i < 3 {
        body_invariant!(i < 3);
        body_invariant!(forall(|j: usize| (0 <= j && j < i) ==> (a[j] <= j + 2),triggers=[(a[j],)]));
        body_invariant!(forall(|j: usize| (i <= j && j < 3) ==> (a[j] <= j + 1),triggers=[(a[j],)]));
        body_invariant!(a[i] <= i + 1);
        let tmp = a[i] + 1;
        a[i] = tmp;
        i += 1;
    }
    a[1]
}
```

but Prusti can verify this program okay:

```rust
#[ensures(result <= 3)]
fn bar() -> usize {
    let mut a = [1, 2, 3];
    let mut i= 0;
    while i < 3 {
        body_invariant!(i < 3);
        body_invariant!(forall(|j: usize| (0 <= j && j < i) ==> (a[j] <= j + 2),triggers=[(a[j],)]));
        body_invariant!(forall(|j: usize| (i <= j && j < 3) ==> (a[j] <= j + 1),triggers=[(a[j],)]));
        body_invariant!(a[i] <= i + 1);
        let tmp = a[i] + 1;
        set(&mut a, i, tmp);
        i += 1;
    }
    a[1]
}

#[requires(0 <= i && i < 3)]
#[ensures(forall(|j: usize| (0 <= j && j < 3 && j != old(i)) ==> (a[j] == old(a[j])), triggers=[(a[j],)]))]
#[ensures(a[old(i)] == old(v))]
fn set(a: &mut [usize; 3], i: usize, v: usize) {
    a[i] = v;
}
```

The underlying issue seems to be related to the snapshot encoding of `a`.

Some ideas that were suggested in #819 are:

> 1. Wait for #796 which addresses snap encoding of Rust assertions. Then we can add `(debug_)assert!` statements at the end of the loop which will invoke `snap$__$TY$__Snap$Array$10$i32$`. This one is currently blocked by the refactoring.
[...]

PR #888, opened 2022-02-28
use `old` in trigger for direct array assignments

Pointerbender:
This PR experiments with addressing 2 things:

1. It implements the suggested work-around for #877 together with a regression test (the test suite indicated no regressions locally).
2. The `selection_sort.rs` test case was updated to use this fix in order to remove the auxiliary `set` function (#819).

Some notes apply to the approach chosen to make `selection_sort.rs` verify:

* A "dummy" `body_invariant!(a[0] <= a[i]);` was added to the outer and inner loops in order to correctly trigger the quantifiers.
* A change was made in `prusti-viper/src/encoder/foldunfold/mod.rs` to prevent Prusti from running into an encoder error when the `old` expresion was only present in the trigger, but not in the quantifier body. Without this change, `selection_sort.rs` will complain about a fold-unfold error. Since access predicates are not needed when the `old` expression is only present in the trigger, this should be sound (but please do double check this during the peer review).
* Without the auxiliary `set` function, the verification time roughly doubled, but it is still well within the `-Pverification_deadline=180` compiler parameter. Execution time now hovers around ~2 minutes on my PC.

If the chosen approach sounds good, then this PR can be merged. I'm still a bit suspicious about why the `set` function yields better performance, I could look into this deeper with some guidance (possibly in a separate PR if that makes more sense). For now I came up empty-handed for what could be causing this performance regression (if this happens to be related to how Viper gives hints to Z3 based on the Viper/Silver syntax, then my current experience may not be of much help, but am willing to learn more!).

Thanks!

p.s. Issue #877 can be left open for now in case we want to file a bug report with the Viper back-end developers to investigate the `old` trigger issue. I posted a ~30 lines example there that hopefully is small enough to investigate.

### DIFF
--- a/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
+++ b/prusti-tests/tests/verify/pass/arrays/selection_sort.rs
@@ -16,6 +16,7 @@ fn selection_sort(a: &mut [i32; 10]) {
 
     while i < a.len() {
         body_invariant!(0 <= i && i < 10);
+        body_invariant!(a[0] <= a[i]);
 
         // sorted below i
         body_invariant!(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < k2 && k2 < i)
@@ -32,6 +33,7 @@ fn selection_sort(a: &mut [i32; 10]) {
         while j < a.len() {
             // these three are the same as the outer loop
             body_invariant!(0 <= i && i < 10);
+            body_invariant!(a[0] <= a[i]);
             body_invariant!(forall(|k1: usize, k2: usize| (0 <= k1 && k1 < k2 && k2 < i)
                                     ==> a[k1] <= a[k2],
                                     triggers=[(a[k1],a[k2])]));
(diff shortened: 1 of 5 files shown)


vakaras:
I have opened a Silicon issue https://github.com/viperproject/silicon/issues/603. Before we merge this PR, let's see what the Silicon developers think.

prusti-tests/tests/verify/pass/arrays/selection_sort.rs
```diff
@@ -16,6 +16,7 @@ fn selection_sort(a: &mut [i32; 10]) {
 
     while i < a.len() {
         body_invariant!(0 <= i && i < 10);
+        body_invariant!(a[0] <= a[i]);
```
vakaras, line 21: Could you add a comment saying that this is needed for triggering?

vakaras: Also, could you make a copy of this example, so that we have both versions?

vakaras (approved):
Seems like this is a Viper backend bug and we could merge this workaround until it is fixed.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5206,10 +5206,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         let idx_conditions = vir_expr!{ [zero_le_i] && ([i_lt_len] && [i_ne_idx]) };
         let tymap = FxHashMap::default();
         let lookup_ret_ty = self.encoder.encode_snapshot_type(array_types.elem_ty_rs, &tymap).with_span(span)?;
-        let lookup_array_i = array_types.encode_lookup_pure_call(self.encoder, encoded_array.clone(), i_var, lookup_ret_ty.clone());
-        let lookup_same_as_old = vir_expr!{ [lookup_array_i] == [old(lookup_array_i.clone())] };
+        let lookup_array_i = array_types.encode_lookup_pure_call(self.encoder, encoded_array.clone(), i_var.clone(), lookup_ret_ty.clone());
+        // FIXME: using `old` here is a work-around for https://github.com/viperproject/prusti-dev/issues/877
```
vakaras, line 5210: Could you also link this issue here: https://github.com/viperproject/silicon/issues/603 ?

Pointerbender:
Thanks for the feedback @vakaras! I will address the requested changes some time tomorrow :)

Pointerbender:
(The requested changes are ready and have been pushed just now, fyi.)

vakaras (approved):
Great, thank you!

bors r+

PR #891, opened 2022-03-01
Fix calls to extern spec functions with `Self`

Aurel300:
@JonasAlaif This should help.

### DIFF
--- a/prusti-tests/tests/verify/pass/extern-spec/traits/generics-3.rs
+++ b/prusti-tests/tests/verify/pass/extern-spec/traits/generics-3.rs
@@ -1,0 +1,15 @@
+use prusti_contracts::*;
+use std::cmp::{Ord, Ordering};
+
+#[extern_spec]
+trait Ord {
+    #[pure]
+    fn cmp(&self, other: &Self) -> Ordering;
+}
+
+#[pure]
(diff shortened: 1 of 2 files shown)


prusti-viper/src/encoder/encoder.rs
```diff
@@ -838,9 +838,13 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
             fn tcx(&self) -> ty::TyCtxt<'tcx> {
                 self.tcx
             }
-            fn fold_ty(&mut self, ty: ty::Ty<'tcx>) -> ty::Ty<'tcx> {
-                let rep = self.tymap.get(&ty).unwrap_or(&ty);
-                rep.super_fold_with(self)
+            fn fold_ty(&mut self, mut ty: ty::Ty<'tcx>) -> ty::Ty<'tcx> {
+                // strange SubstMaps from extern specs can require multiple
+                // replacements before the final type is reached
+                while let Some(rep) = self.tymap.get(&ty) {
```
vl0w: Can we guarantee that there are no cyclic dependencies in tymap?

Aurel300: Maybe this is why the test suite timed out :)

PR #893, opened 2022-03-02
fix issue with nested unfolding expressions when `optimize_folding` is disabled

### LINKED ISSUE #892: Nested `unfolding (...) in ...` statements trying to unfold the same predicate
The following Rust program:

```rust
fn main() {}

#[derive(Clone, Copy)]
pub enum Foo {
    Bar,
}

#[trusted]
#[ensures(match result {
    Ok(Some(r)) => r < 100,
    _ => true,
})]
pub fn foo() -> Result<Option<usize>, Foo> {
    unimplemented!()
}

pub fn test() -> Result<usize, Foo> {
    match foo() {
        Err(error) => Err(error),
        Ok(opt) => match opt {
            Some(res) => Ok(res),
            None => Err(Foo::Bar),
        },
    }
}
```

Generates the following error when the `optimize_folding` optimization is disabled:

> error: [Prusti internal error] Prusti encountered an unexpected internal error
>  |
>  = help: This could be caused by too small assertion timeout. Try increasing it by setting the configuration parameter ASSERT_TIMEOUT to a larger value.
>  = note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
>  = note: Details: unregistered verification error: [inhale.failed:insufficient.permission; 0] Inhale might fail. There might be insufficient permission to access `m_Result$_beg_$m_Option$_beg_$usize$_end_$_sep_$struct$m_Foo$_end_Ok(_1.enum_Ok)` (@0.0)

The error goes away and the program verifies okay when enabling the `optimize_folding` optimization. However, this optimization is causing some of my other programs to fail, so I'd like to keep `optimize_folding` disabled. Note that `optimize_folding` will go away when the refactoring lands.

The relevant Viper code looks like this:

```
inhale /* ... */
      unfolding acc(m_Result$_beg_$m_Option$_beg_$usize$_end_$_sep_$struct$m_Foo$_end_Ok(_1.enum_Ok), write) in (
        unfolding acc(m_Option$_beg_$usize$_end_(_1.enum_Ok.f$0), write) in (
          // unfolding acc(m_Result$_beg_$m_Option$_beg_$usize$_end_$_sep_$struct$m_Foo$_end_Ok(_1.enum_Ok), write) in (
            // unfolding acc(m_Option$_beg_$usize$_end_(_1.enum_Ok.f$0), write) in (
              unfolding acc(m_Option$_beg_$usize$_end_Some(_1.enum_Ok.f$0.enum_Some), write) in (
                unfolding acc(usize(_1.enum_Ok.f$0.enum_Some.f$0), write) in _1.enum_Ok.f$0.enum_Some.f$0.val_int
              )
            // )
          // )
        ) < 100
      )
      /* ... */
```

With the 2 `unfolding` expressions commented out it verifies, but when those are present the verification fails (= with `optimize_folding` disabled).

I have a local patch that can solve this when `optimize_folding` is disabled. I'm just creating this issue to create the regression test and PR against :) I'm currently still running some tests, the PR will follow when all tests are passing.

Pointerbender:
This PR adds a new `fix_unfoldings` optimization pass. This is useful for when the `optimize_folding` optimization can't be enabled, which can trigger problems when matching on enums (see #892).

This is probably a temporary work-around, because `optimize_folding` is going away when the refactoring arrives and then Prusti should also no longer generate invalid `unfolding` expressions. This PR can address some problematic cases until then and the accompanied regression test should come in handy when testing the refactoring :)

Closes #892.


vakaras (approved):
LGTM. This PR clearly shows that I should hurry up with my refactorings. Thanks!

@fpoli If you do not have any concerns regarding this PR, I will merge it.

PR #895, opened 2022-03-04
Treat slices more like arrays to improve support

### LINKED ISSUE #683: Fold-unfold error when a pure function returns a non-pure value

### LINKED ISSUE #718: Unhandled verification error when using slices
The following regression test fails to verify with `prusti-rustc`, but when running `x.py test` the compilation succeeds.

```rust
#![no_std]
use prusti_contracts::*;

#[derive(Clone, Copy)]
pub struct A {
    _inner: usize
}
pub struct B {
    inner: [A]
}

impl B {
    /// Obtain the length of a slice.
    #[pure]
    pub const fn len(&self) -> usize { //~ ERROR unhandled verification error
        self.inner.len()
    }
}

pub fn test(b: &mut B) -> usize {
    b.len()
}

pub fn main() {}
```

The output of `prusti-rustc` is (slightly reformatted to avoid sideways scrolling):

```
$ ./target/debug/prusti-rustc prusti-tests/tests/verify/fail/issues/TBD.rs 
  __          __        __  ___             
 |__)  _\/_  |__) |  | /__`  |   ____\/_  | 
 |      /\   |  \ \__/ .__/  |       /\   | 


Verification of 4 items...
error: [Prusti internal error] unhandled verification error: PureFunctionPostconditionValueRangeOfResult 
[postcondition.violated:assertion.false] Postcondition of m_len__$TY$__Snap$struct$m_B$$int$ might not hold.
Assertion result <= 18446744073709551615 might not hold. (@16.5)
  --> prusti-tests/tests/verify/fail/issues/TBD.rs:16:5
   |
16 | /     pub const fn len(&self) -> usize { //~ ERROR unhandled verification error
17 | |         self.inner.len()
18 | |     }
   | |_____^
   |
   = help: This could be caused by too small assertion timeout.
      Try increasing it by setting the configuration parameter ASSERT_TIMEOUT to a larger value.
note: the failing assertion is here
  --> prusti-tests/tests/verify/fail/issues/TBD.rs:16:5
   |
16 | /     pub const fn len(&self) -> usize { //~ ERROR unhandled verification error
17 | |         self.inner.len()
18 | |     }
   | |_____^

Verification failed
error: aborting due to previous error
```

Additionally, the imposed post-condition `result <= 18446744073709551615` in the error message is too liberal (unsound?). `18446744073709551615` is `usize::MAX`, which can never be valid for the length of slice reference. The length of a slice reference should be at most `isize::MAX as usize` bytes. There exists no valid way in Rust to construct slice references with more elements than can fit inside `isize::MAX` bytes (i.e. `len * sizeof(T) <= isize::MAX`), see <https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety>. Note that raw pointers and `NonNull` do allow pointing to slices with more than `isize::MAX` bytes, this is only disallowed for references to slices, because internally the Rust core library performs "inbounds GEP" LLVM instructions on pointers derived from the slice reference, such as the `<*const [T]>::offset`, `<&mut [T]>::add` methods, which are used for slicing/indexing.)

I have 2 regression tests that replicate both these issues. But since `x.py test` is not recognizing the first one, I will open a draft PR for these regression tests (separately from issue #709 where I encountered these two problems when making more regression tests).

<sub>Edit: clarified explanation around maximum slice reference length</sub>

### LINKED ISSUE #733: Update post-condition of `<[T]>::len`
The post-condition of `<[T]>::len` should be updated to:

```rust
impl<T> [T] { // &[T; N] is coerced to &[T] when calling .len() on it
  #[ensures(result * core::mem::size_of::<T>() <= isize::MAX as usize)]
  pub fn len(&self)  -> usize;
}
```

This way, the post-condition also holds for `[(); usize::MAX]`, which can safely be constructed. A instance of `&[u8; N]` (where `N > isize::MAX`) is unsound (see [`[T]::get`](https://doc.rust-lang.org/src/core/slice/index.rs.html#195-206) and [`[T]::get_unchecked`](https://doc.rust-lang.org/src/core/slice/index.rs.html#218-225), this would violate the safety contracts for `ptr::slice_from_raw_parts` and `<*const T>::add`, which both require that the computed offset fits inside an `isize::MAX`.

In order to encode `core::mem::size_of::<T>()` in the contract of `len` we need to obtain the size of a type at compilation time using some [compiler calls](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html).

### LINKED ISSUE #886: Slices don't work. Prusti doesn't check "a[0] = 1; assert!(a[0] == 1)"
I use current master (d53bdd7af1dda31d763e5fce83b2829aa1164a79) and provided Dockerfile. Here is my code:
```rust
fn f(a: &mut [i32]) {
    if a.len() == 0 {
        return;
    }
    a[0] = 1;
    assert!(a[0] == 1);
}

fn main() {
}
```
Here is prusti output:
```text
# prusti-rustc /tmp/o.rs 
  __          __        __  ___             
 |__)  _\/_  |__) |  | /__`  |   ____\/_  | 
 |      /\   |  \ \__/ .__/  |       /\   | 

Prusti version: commit  <unknown>, built on 2022-02-27 18:02:18 UTC
warning: function is never used: `f`
 --> /tmp/o.rs:1:4
  |
1 | fn f(a: &mut [i32]) {
  |    ^
  |
  = note: `#[warn(dead_code)]` on by default

Verification of 2 items...
error: [Prusti: verification error] the asserted expression might not hold
 --> /tmp/o.rs:6:5
  |
6 |     assert!(a[0] == 1);
  |     ^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

Verification failed
error: aborting due to previous error; 1 warning emitted
```

JonasAlaif:
Test cases from issues moved to passing dir
Fixes #886, fixes #718, fixes #733, fixes #683


prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2361,7 +2365,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         Ok(result)
     }
 
-    fn encode_slice_len_call(
+    fn len(
```
Aurel300: This name is not descriptive enough.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -5121,36 +5112,40 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
 
         let old = |e| { vir::Expr::labelled_old(&label, e) };
 
+        // inhale that len unchanged
+        let len_eq = vir_expr!{ [ sequence_len ] == [ old(sequence_len.clone()) ] };
```
Aurel300: Can we only emit this for slices?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -2451,37 +2455,29 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         let base_seq = self.mir_encoder.encode_operand_place(&args[0])?.unwrap();
         let base_seq_ty = self.mir_encoder.get_operand_ty(&args[0]);
 
+        match base_seq_ty {
+            a if a.peel_refs().is_array() => (),
+            s if s.is_slice() => (),
+            _ => return Err(EncodingError::unsupported(
+                format!("Slicing is only supported for arrays/slices currently, not '{:?}'", base_seq_ty)
```
Aurel300, line 2459: Do you know if this is reachable for any type? Custom implementations will go through an impure method call, right?

prusti-viper/src/encoder/procedure_encoder.rs
JonasAlaif: I think this was an accidental change due to a replace-all

prusti-viper/src/encoder/procedure_encoder.rs
JonasAlaif: I'm think it is, this is called for any `std::ops::Index::index` call, so anything that implements [Index](https://doc.rust-lang.org/nightly/std/ops/trait.Index.html). Better to keep this error in.

JonasAlaif:
The current array encoding is absolutely horrible, with just random Viper statements everywhere. We should shift to `extern_spec`ing things as soon as possible

JonasAlaif:
Fixes #718

JonasAlaif:
The error now is that the statements from `encode_sequence_lookup_shared` (e.g. the havoc of the temporary) gets pulled into the packaging of the magic wand when encoding `issue-709-5.rs` and `issue-709-11.rs`. I don't really know how this can be fixed. Any idea @fpoli? The panic is [here](https://github.com/viperproject/prusti-dev/blame/master/prusti-viper/src/encoder/foldunfold/process_expire_borrows.rs#L511) since the havoc is a `vir::Stmt::MethodCall`.
If I change that line to return `stmt.clone()` just like for the other cases above, then things work up to Viper where `Assert might fail. There might be insufficient permission to access _1.val_ref` for the assert after `l5`:
```boogie
// obtain ((acc(struct$m_A(_0.val_ref), read)) && (true)) && (true)
label l4
package acc(DeadBorrowToken$(-1), write) && acc(struct$m_A(old[l4](_0.val_ref)), read$()) --* acc(struct$m_B(old[pre](_1.val_ref)), read$()) {
  // expire_borrows ReborrowingDAG(L4,L3,)

  if (__t1) {
    // expire loan L4
    exhale acc(struct$m_A(old[l4](_0.val_ref)), read$())
  }
  if (__t1 && __t1) {
    // expire loan L3
    __t4 := builtin$havoc_ref()
    inhale acc(struct$m_A(__t4), write)
    label l5
    assert acc(Array$16$m_A(_1.val_ref.f$0), read$())
    inhale lookup_pure_Array$16_m_A$(old[l4](_1.val_ref).f$0, old[l4](_4)) == snap$m_A(__t4)
    exhale acc(struct$m_A(old[l4](_3.val_ref)), read$())
    inhale acc(struct$m_A(__t3), write - read$())
  }
  // Fold predicates for &mut args
  // transfer perm _1.val_ref --> old[pre](_1.val_ref) // unchecked: false
  fold acc(struct$m_B(old[pre](_1.val_ref)), read$())
  // obtain acc(struct$m_B(old[pre](_1.val_ref)), write)
}
```
Is this just an issue with shared borrows that we don't support? In that case, how could I change things to avoid the panic?

fpoli:
When a method with no argument containing a mutable reference returns a shared reference, Prusti should *not* generate a magic wand. See [this](https://prusti.zulipchat.com/#narrow/stream/288456-prusti-developers/topic/Shared.20references/near/256379281) comment in an internal discussion. Since the code currently tries (and fails) to generate a magic wand, methods that return shared references are unsupported (in which case Prusti should have raised a "xyz is unsupported" error message, which is missing).

To avoid generating magic wands for the case above (shared reference return type with no argument containing a mutable reference) the fix should be done in `compute_procedure_contract` or the code around it, I think. 

(The `patch_place` originates from a1f987c3c4a48dd4dbf689e97cc9aa01fbaa030d, which is a Viper technicality. Just cloning `__t4 := builtin$havoc_ref()` seems correct, since there are no field accesses. If you really want to keep that magic wand, I'll give a look next week.)

JonasAlaif:
Thanks! I was able to update the `compute_procedure_contract` function to implement this change. For mutable reborrows with slices we still fail (`709-4` and `709-10`) but there it isn't a crash so will just leave it as is.

fpoli:
> I was able to update the `compute_procedure_contract` function to implement this change.

Awesome!

> For mutable reborrows with slices we still fail (`709-4` and `709-10`)

That's something that I can give a look at, as soon as I have some time.

JonasAlaif:
@fpoli So the issue after was that Prusti now complains for some other cases that there are 0 magic wands (e.g. `equality.rs`). I've updated the code that returns this error in the latest PR, but I'm not sure if this is the correct fix? Should we even be calling `construct_vir_reborrowing_node_for_call` when there are no magic wands?

fpoli:
> @fpoli So the issue after was that Prusti now complains for some other cases that there are 0 magic wands (e.g. `equality.rs`). I've updated the code that returns this error in the latest PR, but I'm not sure if this is the correct fix?

The fix looks good to me. Methods like `fn get(&self, index: usize) -> &A` create a reference at the call site but no magic wands. 

> Should we even be calling `construct_vir_reborrowing_node_for_call` when there are no magic wands?

I suspect the code needs to build an edge in the reborrowing DAG anyway, but I'm not 100% sure about it.

JonasAlaif:
I'll merge this today if there are no objections?

JonasAlaif:
@vakaras is there some way to make Bors squash the merge into one commit? Or at least to squash the commits in a merge which do not pass the CI. Otherwise we still get commits in master with failing CI

vakaras:
> @vakaras is there some way to make Bors squash the merge into one commit? Or at least to squash the commits in a merge which do not pass the CI. Otherwise we still get commits in master with failing CI

There is [`use_squash_merge` option](https://bors.tech/documentation/). However, if we use it, we should encourage everyone to make only a single logical change per PR.

JonasAlaif:
Hmm hard to say, I guess that enabling this is reasonable as a default idk

Pointerbender:
> The current array encoding is absolutely horrible, with just random Viper statements everywhere. We should shift to `extern_spec`ing things as soon as possible

@JonasAlaif  Just a heads up that not all array/slice operations can be `extern_spec`'ed. The compiler special cases direct accesses by lowering these straight into MIR without going through the `Index` trait. Check out the example MIR output here: https://rust.godbolt.org/z/hYcz7W4dz

PR #896, opened 2022-03-04
Better support for generics (in traits / impls)

vl0w:
This PR aims to provide better support for generics with trait implementations.
Currently, when Prusti encodes a trait method and the corresponding implementation has generics like in this example:

```rust
trait Trait {
    #[pure]
    fn foo(&self) -> i32;
}

struct Foo<T>(std::marker::PhantomData<T>);
impl<T> Trait for Foo<T> {
    #[pure]
    fn foo(&self) -> i32 {
        42
    }
}

fn main() {
    let f = Foo::<u32>(std::marker::PhantomData);
    assert!(f.foo() == 42);
    let f = Foo::<i32>(std::marker::PhantomData);
    assert!(f.foo() == 42);
}
```
we get a wrong encoding because the generic parameter `T` is not instantiated with the callsite arguments `u32` and `i32`, which leads to a type error in Viper.
In this PR, I adjusted tymap to also contain information from the trait implementation block.

### DIFF
--- a/prusti-interface/src/environment/tymap.rs
+++ b/prusti-interface/src/environment/tymap.rs
@@ -1,0 +1,127 @@
+use super::{traits, Environment};
+use crate::data::ProcedureDefId;
+use log::{trace, warn};
+use rustc_hash::{FxHashMap, FxHashSet};
+use rustc_middle::{ty, ty::subst::SubstsRef};
+
+type Key<'tcx> = ty::Ty<'tcx>;
+type Value<'tcx> = ty::Ty<'tcx>;
+
+#[derive(Default, Clone, Debug)]
+pub struct SubstMap<'tcx> {
+    map: FxHashMap<Key<'tcx>, Value<'tcx>>,
+}
+
+impl<'tcx> SubstMap<'tcx> {
+    /// Builds a SubstMap from a call to a function.
+    /// The built tymap considers whether the called function resolves to a trait implementation
+    /// and adds the necessary substitutions.
+    pub fn build<'a>(
+        env: &'a Environment<'tcx>,
+        def_id: ProcedureDefId,
+        call_substs: SubstsRef<'tcx>,
+    ) -> Self {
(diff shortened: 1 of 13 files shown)


prusti-interface/src/environment/tymap.rs
```diff
@@ -0,0 +1,130 @@
...
+            None
+        } else {
+            env.find_impl_of_trait_method_call(def_id, call_substs)
+        };
+
+        if let Some(impl_did) = maybe_impl_did {
+            let impl_own_substs = ty::List::identity_for_item(env.tcx(), impl_did);
+            let trait_did = env.tcx().trait_of_item(def_id).unwrap();
+            let trait_ref = ty::TraitRef::from_method(env.tcx(), trait_did, call_substs);
+            let trait_binder = ty::Binder::bind_with_vars(trait_ref, ty::List::empty());
```
Aurel300: Is this the same as `ty::Binder::dummy(trait_ref)`?

vl0w: This seems to be the proper choice when I read the docs :+1:

prusti-interface/src/environment/tymap.rs
```diff
@@ -0,0 +1,122 @@
...
+        // In case this is a trait method implementation, we also extend
+        // tymap with information from the implementation itself
+        let maybe_impl_did = if call_substs.is_empty() {
+            None
+        } else {
+            env.find_impl_of_trait_method_call(def_id, call_substs)
+        };
+
+        if let Some(impl_did) = maybe_impl_did {
+            let impl_own_substs = ty::List::identity_for_item(env.tcx(), impl_did);
```
Aurel300, line 65: Maybe add a comment for these lines explaining what is happening here?

vl0w: Done, I hope it's understandable :)

PR #899, opened 2022-03-08
(WIP) Remove `tymap`

Aurel300:
(generics improvements)

TODO:
- [x] remove dead code (definition of `SubstMap`)
- [x] remove specs closures collector
- [x] improve extern specs (make sure usage is valid wrt. generics, forbid specialisations, deal with associated types better)
- [x] figure out what to do in cases where we used to create an empty tymap or substs
- [ ] resolve all comments marked `TODO(tymap)`
- [ ] (for a future PR) make sure we can serialise unoptimised MIR
- [ ] (for a future PR) remove generics from VIR


prusti-interface/src/environment/mod.rs
```diff
@@ -320,21 +337,65 @@ impl<'tcx> Environment<'tcx> {
...
+        let impl_def_id = self.tcx.impl_of_method(impl_method_def_id)?;
+        let trait_def_id = self.tcx.trait_id_of_impl(impl_def_id)?;
+
+        // the input method was in an impl for a trait: finding a `None` in
+        // further calls would be unexpected, so we unwrap instead of ?
+
+        let trait_method_def_id = self.get_assoc_item(
+            trait_def_id,
+            self.tcx().item_name(impl_method_def_id),
+        ).unwrap().def_id;
```
vakaras, line 374: I think it would be better to change the return type to `Result` and return `Result::Err` here to be able to show a user-friendly internal error message to the user instead of crashing.

prusti-interface/src/environment/traits/mod.rs
```diff
@@ -1,5 +1,4 @@
 mod rustc_codegen;
 mod rustc_instance;
 
-pub(super) use rustc_instance::resolve_instance;
-pub(super) use rustc_codegen::codegen_fulfill_obligation;
```
vakaras, line 5: Did you forget to delete `codegen_fulfill_obligation`?

prusti-tests/tests/verify/fail/unsupported/type_param_array.rs
```diff
@@ -1,3 +1,5 @@
+// ignore-test slices of generics
+
```
vakaras, line 2: Does it crash now?

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
```diff
@@ -191,33 +191,42 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
...
             .bodies_poly
             .borrow()
             .contains_key(&key)
         {
+            // TODO(tymap): why the procedure?
+            let mir = self.env().local_mir(proc_def_id.expect_local(), substs);
+
+            /*
             let procedure = self.env().get_procedure(proc_def_id);
+            let mir = procedure.get_mir();
```
vakaras: Delete?

vakaras (approved):
LGTM

prusti-interface/src/environment/mod.rs
Aurel300: I don't want to clutter the interface. If we get an unwrap error now (after the two successful `?` unwraps just above), there is probably something wrong inside the compiler. The two calls establish that:

- the given method belongs to an `impl` block
- that `impl` block is implementing a trait

So we should definitely be able to get the associated item (or else what is the method implementing?) and we should definitely be able to get the trait ref (or else how could the `impl` block be implementing a trait?).

prusti-interface/src/environment/traits/mod.rs
Aurel300: No, it is still used from withing `resolve_instance`… I might delete the (one) usage of that method eventually, but not in this PR.

prusti-tests/tests/verify/fail/unsupported/type_param_array.rs
Aurel300: Yes. As far as I can tell It's because the inhale generated because of the statement refers to `.val_ref` of the generic value, but that is not a thing in our generics. I think it would make things easier for us if we treated typarams as `Ref`s, just like ADTs.

vakaras: Does it panic or does it report an unsupported error?

prusti-interface/src/environment/mod.rs
vakaras: In that case, I would clarify the comment to explicitly say that we expect a compiler bug in this case.

prusti-tests/tests/verify/fail/unsupported/type_param_array.rs
Aurel300: Right now it is a panic.

vakaras: I see.

PR #906, opened 2022-03-17
Ghost constraints

vl0w:
Adds ghost constraints to Prusti, which we plan to use for `Iterator` support in Prusti.

A ghost constraint defines a constraint for a specification which must hold for a function call.
That means, the specification is only "active" for a function call when the constraint is fulfilled on callsite.
At this point, a ghost constraint is simply a trait bound for a type (usually a generic / `Self`). More technically, a ghost constraint is conceptually a Rust `where` clause attached to a spec.

```rust
#[ghost_constraint(T: SomeTrait, [
    requires(R), ensures(E)
])]
fn foo<T>(...) {...}

fn client() {
    // Assume T1 implements SomeTrait
    // The precondition R and the postcondition E of foo are active
    foo::<T1>(...); 

    // Assume T2 does not implement SomeTrait
    // The precondition R and the postcondition E of foo are _NOT_ active
    foo::<T2>(...); 
}

```

Conceptually, a contract of a function is split into a specification without constraints and several specifications which have constraints (constrained specs). A constrained spec is **active** for a function call when it's bounds are satisfied. If no constrained spec is active, the base spec is active. Only one spec may be active at a time. This PR should be restrictive enough such that multiple active specs are not possible.
The constrained specs gain knowledge about their bounds, so the traits in the bound can be used in the specification.

## Goal
The goal of this PR is to enable modular specifications on iterators and iterator adapters.
A draft specification outline is the following:

```rust
// Defines two abstract predicates
trait IteratorSpecExt {
    fn enumerated(&self) -> bool;
    fn completed(&self) -> bool;
}


#[extern_spec]
trait Iterator {
    // Abstract predicates used in the spec
    // of Iterator
    #[requires(false)] // Self without IteratorSpecExt is uncallable
    #[ghost_constraint(Self: IteratorSpecExt, [
        requires(self.enumerated()),
        ensures(self.enumerated())
    ])]
    fn next(&mut self) -> Option<Self::Item>;
}

impl<...> IteratorSpecExt for std::iter::Map<...> {
    // Give a meaningful implementation to the abstract predicates
}

#[extern_spec]
impl<...> Iterator for std::iter::Map<...> {

    // More specs
    fn next(&mut self) -> Option<Self::Item>;
}

fn client(iter: std::iter::Map<...>) {
    iter.next();
    // iter.enumerated() holds
}

```

## Challenge
Checking correct behavioral subtyping rules involving ghost constraints is rather involved because a contract may contain arbitrarily constrained specs with overlapping constraints. 
Consider a a user-typed contract with a base spec `BS` and a set of constrained specs `{CS_1, CS_2, ..., CS_n}`, i.e.
```rust
#[requires(pre(BS))]
#[ensures(post(BS))]
#[ghost_constraint(bounds(CS_1), [requires(pre(CS_1)), ensures(post(CS_1))])]
// ...
#[ghost_constraint(bounds(CS_n), [requires(pre(CS_n)), ensures(post(CS_n))])]
fn foo<T>(...) {...}
```
For example:
```rust
#[requires(requires(...))]
#[ensures(ensures(...))]
#[ghost_constraint(T: A, [...])]
#[ghost_constraint(T: B + C, [...])]
#[ghost_constraint(T: A + C, [...])]
fn foo<T>(...) {...}
```
Generally, we face the following challenges:
- What is the notion of subtyping amongst `BS` and `CS_1, ..., CS_n`
- Which spec should be active on a callsite?
- How should `foo<T>` be modularly verified?

Since this PR is intended to enable modular verification support for iterators only,
this PR does not fully resolve these challenges.

### Refinement is not automatically verified
- There is no check that every `CS_i` is a valid refinement of `BS`
- There is no check that `CS_i` is a valid refinement of `CS_j` when the bounds defined on `CS_i` induce a subtyping check w.r.t. the bounds of `CS_j`

The Prusti user needs to make sure that no soundness issues are introduced when using constrained specs. This is one of the main reasons why we add a feature flag for ghost constraints.

### Only allowed in `#[trusted]` functions (including extern specs)
For a modular verification, it is not immediately obvious which spec of a contract is active for any call to that function. This PR is thus very restrictive and allows constrained specs only on trusted functions (and external specs).

### It is not possible to declare multiple ghost constraints with different bounds
This is ensured during constraint resolving of a function contract. It basically checks that `bounds(CS_i) == bounds(CS_j)` for every `i != j`.
The following spec will be rejected by Prusti:
```rust
#[ghost_constraint(T: A, [...])]
#[ghost_constraint(T: B, [...])]
fn foo<T>() { }
```
The following is possible however: 
```rust
#[ghost_constraint(T: A, [...])]
#[ghost_constraint(T: A, [...])]
fn foo<T>() { }
```
Which will unify both ghost constraints.

### Inheriting from the base spec to the constrained spec
A constrained spec "inherits" the postconditions, pure flag, trusted flag, pledges from the base spec. This is mainly a convenience feature to not need to repeat the specs.

Preconditions are not copied, since this would not lead to a valid refinement.

- [ ] Maybe pledges should not be copied @Aurel300 ?

### Trait contract refinement
Trait contract refinement is aware of ghost constraints to account for ghost constraints on the trait method with a `Self` bound.

Given a contract `T_BS, T_CS_1` on a trait and `S_BS, S_CS_1`, contract refinement for a function call is performed like this:
- Get the contract of the impl contract
- Resolve the obligations of the impl 
    - -> Selected spec `X`
- Get the contract of the trait
- Resolve the obligations of the trait contract
    - -> Selected spec `Y`
- Refine `X` with `Y` -> Refined `Z`
- `Z` is the spec used for encoding a function call
- `X` and `Y` are used for contract refinement checks

Note 1: There cannot be multiple ghost constraints on either the trait or the impl.

Note 2: Obligation resolving does not happens not pure/trusted checks (we still refine though)

Note 3: This refinement process happens once for the encoding and modular verification of the function _and_ once for any encoding of a call. Specifically, this does refinement does not happen for any outcome of obligation resolving. We might need to adjust this in a future PR.

### Parsing
- [x] Support generic methods in trait bounds
- [x] Macro can not be parsed in certain cases (see corresponding TODO in code)
- [x] Trait/Impls -> Disallow usage in refined trait specs
- [x] Further parsing unit tests
- [x] Extern specs
- [x] Fully qualified paths in trait bounds

### Architectural
- [x] Reiterate structure in `typed.rs`
- [x] Naming: Ghost constraint vs obligation (in struct names etc)
- [x] Associated types in trait bounds (lead to an error)

### General
- [x] Resolve all  `TODO hansenj: ...`
- [x] Disallow multiple ghost constraints
- [x] ~User Docs~
- [x] `Self` in ghost constraints (and extern specs)
- [x] Add feature flag
- [x] Explain/Document ghost constraints in the context of refinements


prusti-interface/src/environment/mod.rs
```diff
@@ -389,6 +389,7 @@ impl<'tcx> Environment<'tcx> {
     /// Checks whether the given type implements the trait with the given DefId.
     /// Accounts for generic params on the trait by the given substs.
     /// TODO hansenj: name, params, docs
+    /// TODO hansenj: Do we need this?
```
Aurel300: I assume yes? A ghost constraint `T: Foo<i32>` is not the same as `T: Foo<bool>`.

prusti-common/src/config.rs
```diff
@@ -533,3 +534,6 @@ pub fn full_compilation() -> bool {
 pub fn intern_names() -> bool {
     read_setting("intern_names")
 }
+
+/// Enables ghost constraint parsing and resolving.
+pub fn enable_ghost_constraints() -> bool { read_setting("enable_ghost_constraints") }
```
Aurel300:
```suggestion
pub fn enable_ghost_constraints() -> bool {
    read_setting("enable_ghost_constraints")
}
```

docs/dev-guide/src/config/flags.md
```diff
@@ -105,6 +106,10 @@ When enabled, reborrowing DAGs will be output in debug files.
 
 When enabled, the encoded Viper program will be output.
 
+## `ENABLE_GHOST_CONSTRAINTS`
+
+Enables ghost constraints in Prusti specifications.
```
Aurel300, line 111: Point out that this is an experimental feature.

prusti-interface/src/environment/mod.rs
```diff
@@ -495,10 +495,16 @@ impl<'tcx> Environment<'tcx> {
 
     /// Checks whether the given type implements the trait with the given DefId.
     pub fn type_implements_trait(&self, ty: ty::Ty<'tcx>, trait_def_id: DefId, param_env: ty::ParamEnv<'tcx>) -> bool {
```
Aurel300, line 523: This method should probably not exist: I assume the empty `ty::List` will cause a crash when checking against a generic trait.

prusti-interface/src/specs/mod.rs
```diff
@@ -71,65 +72,50 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
...
-                        pres.push(*self.spec_functions.get(spec_id).unwrap());
+                        spec.add_precondition(*self.spec_functions.get(spec_id).unwrap(), self.env);
                     }
                     SpecIdRef::Postcondition(spec_id) => {
-                        posts.push(*self.spec_functions.get(spec_id).unwrap());
+                        spec.add_postcondition(*self.spec_functions.get(spec_id).unwrap(), self.env);
                     }
                     SpecIdRef::Pledge { lhs, rhs } => {
-                        pledges.push(typed::Pledge {
+                        let pledge = typed::Pledge {
```
Aurel300: Any reason for this local variable? Seems inconsistent with the other cases.

prusti-interface/src/specs/typed.rs
```diff
@@ -1,68 +1,157 @@
...
-    pub fn into_procedure(self) -> ProcedureSpecification {
-        if let SpecificationSet::Procedure(spec) = self {
-            return spec;
+impl SpecGraph<ProcedureSpecification> {
+    pub fn add_precondition<'tcx>(&mut self, pre: LocalDefId, env: &Environment<'tcx>) {
+        match self.get_constraint(pre, env) {
+            None => {
+                self.base_spec.pres.push(pre);
+                // Preconditions are explicitly not copied (as opposed to postconditions)
+                // This would always violate behavioral subtyping rules
```
Aurel300, line 188: Maybe explain this in more detail (with an example) in a comment above `impl SpecGraph<ProcedureSpecification>`, and mention briefly that this does not happen here and *does* happen in postconditions.

prusti-interface/src/specs/typed.rs
```diff
@@ -1,68 +1,157 @@
...
+    fn get_partitioned_mut(
+        &mut self,
+        obligation: SpecConstraintKind,
+    ) -> &mut ProcedureSpecification {
+        self.specs_with_constraints
+            .entry(obligation)
+            .or_insert_with(|| self.base_spec.clone())
+    }
+
+    /// Note: First wins, we do currently support not multiple constraints
```
Aurel300:
```suggestion
    /// Note: First wins, we currently do not support multiple constraints
```

prusti-interface/src/specs/typed.rs
```diff
@@ -126,11 +215,55 @@ impl<T> SpecificationItem<T> {
     pub fn extract_with_selective_replacement(&self) -> Option<&T> {
         self.extract_with_strategy(|(_, refined)| refined)
     }
+
+    pub fn expect_empty_or_inherent(&self) -> Option<&T> {
```
Aurel300, line 326: These types of methods (all the `expect_*`) are nicer with `#[track_caller]`.

prusti-specs/src/extern_spec_rewriter/common.rs
```diff
@@ -141,11 +142,40 @@ impl<'a> AssociatedTypeRewriter<'a> {
...
+    }
+
+    pub fn rewrite_attribute(&mut self, attr: &mut syn::Attribute) {
+        self.visit_attribute_mut(attr);
+    }
+
+    fn rewrite_tokens(&self, self_replacement_as_str: &str, tokens: TokenStream) -> TokenStream {
+        let it = tokens.into_iter();
+        it.map(|tt| {
+            match tt{
```
Aurel300:
```suggestion
            match tt {
```

prusti-specs/src/extern_spec_rewriter/common.rs
```diff
@@ -141,11 +142,40 @@ impl<'a> AssociatedTypeRewriter<'a> {
...
+        }).collect()
     }
 }
 
 impl<'a> syn::visit_mut::VisitMut for AssociatedTypeRewriter<'a> {
+    fn visit_attribute_mut(&mut self, attr: &mut syn::Attribute) {
+        let self_replacement = self.self_type.to_token_stream().to_string();
+
+        // Note: Attribute tokens are not visited by a syn visitor,
+        // so we use visit the attribute tokens "manually"
```
Aurel300:
```suggestion
        // so we visit the attribute tokens "manually"
```

prusti-specs/src/lib.rs
```diff
@@ -501,4 +515,4 @@ pub fn predicate(tokens: TokenStream) -> TokenStream {
         #[prusti::pred_spec_id_ref = #spec_id_str]
         #cleaned_fn
     }
-}
+}
```
```suggestion
}

```

(and in general, please try to keep the line ends)

prusti-specs/src/specifications/preparser.rs
```diff
@@ -131,9 +139,12 @@ impl PrustiTokenStream {
                     // to avoid later mis-identifying its suffix
                     tokens.push_back(PrustiToken::Token(token.clone()));
                     while let Some(token @ TokenTree::Punct(p)) = source.get(pos) {
+                        if p.as_char() == ',' { // Treat comma as a delimiter for macros
+                            break;
+                        }
                         pos += 1;
                         tokens.push_back(PrustiToken::Token(token.clone()));
-                        if p.spacing() != Joint {
+                        if p.spacing() != Alone {
```
Aurel300: I'm not sure I understand these changes. Why stop parsing when a `Joint`-spaced token is encountered?

prusti-specs/src/specifications/preparser.rs
```diff
@@ -421,14 +452,15 @@ impl PrustiTokenStream {
...
-            .map(|group| Self { tokens: group.iter().cloned().collect() })
+            .map(|group| Self { tokens: group.iter().cloned().collect(), source_span: self.source_span })
             .collect::<Vec<_>>();
         if allow_trailing && res.len() > 1 && res[res.len() - 1].tokens.is_empty() {
             res.pop();
         }
         res
     }
 
+
```
Aurel300:
```suggestion
```

prusti-tests/tests/compiletest.rs
```diff
@@ -126,6 +126,7 @@ fn run_verification_base(group_name: &str, filter: &Option<String>) {
         TemporaryEnvVar::set("PRUSTI_FULL_COMPILATION", "true"),
         TemporaryEnvVar::set("PRUSTI_ENCODE_UNSIGNED_NUM_CONSTRAINT", "true"),
         TemporaryEnvVar::set("PRUSTI_QUIET", "true"),
+        TemporaryEnvVar::set("PRUSTI_ENABLE_GHOST_CONSTRAINTS", "true"),
```
Aurel300: Let's only enable it per-test for now?

prusti-tests/tests/parse/fail/ghost-constraints/trait-bounds-illegal-4.rs
```diff
@@ -0,0 +1,14 @@
+use prusti_contracts::*;
+
+trait A { }
+trait B { }
+
+#[ghost_constraint(T: Fn(A) -> B, [ //~ ERROR: unexpected Prusti syntax
```
Aurel300: I'm not sure I understand why this should fail.

prusti-tests/tests/verify/fail/ghost-constraints/invalid-trait-refinement-1.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn pre(&self) -> bool;
+    #[pure]
+    fn post(&self) -> bool;
+}
+
+trait MyTrait {
+    fn foo(&mut self);
+}
+
+#[extern_spec] //~ ERROR: the method's precondition may not be a valid weakening of the trait's precondition
```
Aurel300: Why is it reported here?

prusti-tests/tests/verify_overflow/fail/ghost-constraints/different-bounds.rs
```diff
@@ -0,0 +1,23 @@
+// Note: For now, it is completely fine to _declare_ two ghost constraints with different bounds
+// since resolving happens for specific callsites. That is, without the call in main, this
+// file verifies.
```
Aurel300, line 5: This is not great but we're going to have multiple ghost constraints later anyway so that's fine.

prusti-tests/tests/verify_overflow/pass/ghost-constraints/ghost-constraints-extend-base-attributes.rs
```diff
@@ -0,0 +1,58 @@
...
+// Refinement is currently not checked, so we can introduce unsound stuff
+fn constrained_contract_inherits_posts_unsound<T>(_x: T) -> i32 {
+    42
+}
+
+fn verify_constrained_contract_inherits_posts_unsound() {
+    let res = constrained_contract_inherits_posts_unsound(32 as i32);
+    assert!(res == 42);
+    assert!(res == 24);
+    assert!(false);
```
Aurel300: This is unfortunate but let's not actually test for asserting false (:

prusti-tests/tests/verify_overflow/pass/ghost-constraints/splitted-constraint.rs
```diff
@@ -0,0 +1,22 @@
+use prusti_contracts::*;
```
Aurel300: (name: `split-constraint.rs`?)

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -4770,7 +4770,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                     _,
                     mir::Rvalue::Aggregate(box mir::AggregateKind::Closure(cl_def_id, cl_substs), _),
                 )) = stmt.kind {
-                    if let Some(spec) = self.encoder.get_loop_specs(cl_def_id) {
+                    if let Some(spec) = self.encoder.get_loop_specs(cl_def_id,  cl_substs) {
```
```suggestion
                    if let Some(spec) = self.encoder.get_loop_specs(cl_def_id, cl_substs) {
```

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -1,39 +1,98 @@
...
+        let mut copy = *self;
+        copy.called_def_id = called_def_id;
+        copy.call_substs = call_substs;
+        copy
+    }
+}
 
-    fn is_trusted(&self, def_id: DefId) -> bool;
+pub(crate) trait SpecificationsInterface<'tcx> {
+    fn is_pure(&self, def_id: DefId, substs: Option<SubstsRef<'tcx>>) -> bool;
```
Aurel300, line 73: Unsure how I feel about `substs` being passed for `is_pure` and `is_trusted`. These are flags which we cannot change with ghost constraints.

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -1,39 +1,98 @@
...
         Self {
             specs: RefCell::new(Specifications::new(user_typed_specs)),
         }
     }
 }
 
-pub(crate) trait SpecificationsInterface {
-    fn is_pure(&self, def_id: DefId) -> bool;
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct SpecQuery<'tcx> {
```
Aurel300: Should this really be `pub`? Is it used outside of `specifications`?

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -1,39 +1,98 @@
...
+    pub(crate) fn adapt_to_call(&self, called_def_id: DefId, call_substs: SubstsRef<'tcx>) -> Self {
+        let mut copy = *self;
+        copy.called_def_id = called_def_id;
+        copy.call_substs = call_substs;
+        copy
+    }
+}
 
-    fn is_trusted(&self, def_id: DefId) -> bool;
+pub(crate) trait SpecificationsInterface<'tcx> {
```
Aurel300, line 72: General TODO (add in a comment, not for this PR): since the calls to `get_loop_specs` etc will receive `substs`, we could already monomorphise the MIR bodies.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -543,7 +541,7 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
     pub fn encode_procedure(&self, def_id: ProcedureDefId) -> SpannedEncodingResult<()> {
         debug!("encode_procedure({:?})", def_id);
         assert!(
-            !self.is_trusted(def_id),
+            !self.is_trusted( def_id, None),
```
```suggestion
            !self.is_trusted(def_id, None),
```

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -0,0 +1,267 @@
...
+            return Ok(&self.base_spec);
+        }
+
+        match query.cause {
+            // For simple pure or trusted checks, we do not need to consider obligations
+            // since they can a pure/trusted flag can not change in a constrained spec
+            SpecQueryCause::PureOrTrustedCheck => {
+                trace!(
+                "No need to resolve obligations for checking whether function is pure or trusted"
+            );
```
Aurel300:
```suggestion
                    "No need to resolve obligations for checking whether function is pure or trusted"
                );
```

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -0,0 +1,267 @@
...
+            debug!("Trait bound constraint resolving for {:?}", query);
+
+            let param_env_constraint = extract_param_env(env, proc_spec);
+            let param_env_constraint =
+                perform_param_env_substitutions(env, query, param_env_constraint);
+
+            // There is no caller when encoding a function.
+            // We still resolve obligations to account for constrained specs on a trait
+            // for which we encode it's implementation. The corresponding encoding will
```
Aurel300:
```suggestion
            // for which we encode its implementation. The corresponding encoding will
```

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -0,0 +1,267 @@
...
+        SpecConstraintKind::ResolveGenericParamTraitBounds => {
+            resolvers::trait_bounds::resolve(env, query, proc_spec)
+        }
+    }
+}
+
+mod resolvers {
+    use super::*;
+
+    pub mod trait_bounds {
```
Aurel300: Why these strange nested modules?

prusti-tests/tests/verify_overflow/pass/ghost-constraints/ghost-constraints-extend-base-attributes.rs
vl0w: I believe it might be valuable to illustrate how one can introduce unsound verification results.
But yes, this is rather documentation than a test.

prusti-viper/src/encoder/mir/specifications/constraints.rs
vl0w: The idea was that we have one mod per `SpecConstraintKind`. I am not sure though whether we we will ever have multiple variants in there.
I will keep `SpecConstraintKind` but remove the outermost mod `resolvers` if thats ok for you. Keeping trait bound resolving logic in a separate module is fine to give `constraints.rs` some structure.

prusti-tests/tests/parse/fail/ghost-constraints/trait-bounds-illegal-4.rs
vl0w: Ah good point. It should not fail but it currently does because the parser can't handle it.
I will remove the test.

prusti-tests/tests/verify_overflow/fail/ghost-constraints/different-bounds.rs
vl0w: Yeah, that's indeed a bit special with the current implementation. I guess when we'll implement behavioral subtyping checks of ghost constraints, we need to preprocess all of them somehow and catch these errors there.

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: You are right, we cannot override `is_pure` and `is_trusted` with ghost constraints. We technically need `substs` for refinement though. Currently this is not handled properly as you mentioned in [this comment](https://github.com/viperproject/prusti-dev/pull/906#discussion_r844195348).

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: Nope. I adjusted its visibility to `pub(in crate::encoder::mir::specifications)`

prusti-tests/tests/compiletest.rs
vl0w: With the feature generally activated, we can check for regressions against code without ghost constraints.
Concretely, I would like to check that the specs module (w/ the changes in this PR) is still is correct.
I would therefore suggest that we keep the flag enabled. Eventually, ghost constraints won't be feature-gated anymore.

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: I indeed already use the correct method to resolve to the trait spec [here](https://github.com/viperproject/prusti-dev/blob/d5f1ac726df268dc71e4cd7235555985092ce0c3/prusti-viper/src/encoder/mir/specifications/specs.rs#L66)

prusti-interface/src/specs/typed.rs
vl0w: I added a big comment to `SpecGraph`. Hope it clarifies things.

prusti-interface/src/environment/mod.rs
vl0w: Afaik this is not used in the `specifications` module or is used anyhow involved in constraint resolving (`constraints.rs`)?

prusti-viper/src/encoder/mir/specifications/interface.rs
Aurel300: That's just `pub(super)`, no?

vl0w: Hmm, yes 😅

prusti-tests/tests/verify/fail/ghost-constraints/invalid-trait-refinement-1.rs
vl0w: The full error message is:
```
error: [Prusti: verification error] the method's precondition may not be a valid weakening of the trait's precondition.
  --> verify-me.rs:15:1
   |
15 | #[extern_spec] //~ ERROR: the method's precondition may not be a valid weakening of the trait's precondition
   | ^^^^^^^^^^^^^^
...
29 |     #[requires(self.x >= 15)]
   |                ^^^^^^^^^^^^
   |
   = help: The trait's precondition should imply the implemented method's precondition.
   = note: this error originates in the attribute macro `extern_spec` (in Nightly builds, run with -Z macro-backtrace for more info)
```
which is fine imo.
The `#[extern_spec]` is part of the error span because the precondition function is generated during the expansion of the extern spec.
I will create UI tests for `invalid-trait-refinement-1.rs` and `invalid-trait-refinement-2.rs` to resolve this confusion (afaik it's not possible to test a multiline span nicely with these error patterns).

prusti-tests/tests/compiletest.rs
Aurel300: In the meanwhile though we are not testing any part of Prusti with the feature disabled, and it being disabled will be the default for any user. We can check for regressions when we make the feature enabled by default, not now. (And of course you can see locally and in the already-executed CI run if it caused any issues yet.)

prusti-specs/src/specifications/preparser.rs
vl0w: I removed this and adjusted the tests (as discussed)

prusti-interface/src/specs/typed.rs
```diff
@@ -1,70 +1,263 @@
...
 
-#[derive(Debug, Clone)]
-pub enum SpecificationSet {
-    Procedure(ProcedureSpecification),
-    Loop(LoopSpecification),
+/// A map of specifications keyed by crate-local DefIds.
+#[derive(Default, Debug, Clone)]
+pub struct DefSpecificationMap {
+    pub proc_specs: HashMap<DefId, SpecGraph<ProcedureSpecification>>,
+    pub loop_specs: HashMap<DefId, SpecGraph<LoopSpecification>>,
```
Aurel300: Does it make sense for this to be a `SpecGraph` as well? I guess for now there is no way to declare a ghost constrained loop invariant (if that even makes sense).

prusti-interface/src/specs/typed.rs
```diff
@@ -1,70 +1,263 @@
...
+/// which means:
+/// - The postconditions of any constrained spec always contain the postconditions of the base spec
+/// - The [ProcedureSpecificationKind] for the base spec and constrained specs is always the same
+/// - The `trusted` flag for the base spec and constrained specs is always the same
+/// - The [Pledge]s for the base spec and constrained specs are always the same
+///
+/// Note: Unlike postconditions, the preconditions are not copied amongst the base spec
+/// and constrained specs.
+///
+/// # A note about behavior subtyping
```
Aurel300: `behavioural subtyping`

(or `behavioral` for AmEng 🤮)

prusti-viper/src/encoder/mir/specifications/specs.rs
```diff
@@ -10,100 +16,169 @@ use prusti_interface::{
...
+                call_substs: substs,
+                ..
+            })
+            | SpecQuery::FunctionDefEncoding(def_id, substs)
+            | SpecQuery::PureOrTrustedCheck(def_id, substs) => {
+                let (trait_def_id, trait_substs) = env.find_trait_method_substs(*def_id, substs)?;
+
+                // We need to subst the trait substs because they might be generic
+                // (with substitutions in the call substs)
+                let trait_substs = if trait_substs.needs_subst() {
```
vl0w: Why do we need to do this? Should this not be handled by `find_trait_method_substs`? @Aurel300

prusti-specs/src/extern_spec_rewriter/common.rs
```diff
@@ -142,11 +143,40 @@ impl<'a> AssociatedTypeRewriter<'a> {
...
     }
 }
 
 impl<'a> syn::visit_mut::VisitMut for AssociatedTypeRewriter<'a> {
+    fn visit_attribute_mut(&mut self, attr: &mut syn::Attribute) {
+        let self_replacement = self.self_type.to_token_stream().to_string();
+
+        // Note: Attribute tokens are not visited by a syn visitor,
+        // so we visit the attribute tokens "manually"
+        attr.tokens = self.rewrite_tokens(self_replacement.as_str(), attr.tokens.clone());
```
Aurel300, line 175: Does this mean we process here the user-written attribute for `Self` translation? This seems weird, I would expect us to do it after preparsing…

prusti-specs/src/extern_spec_rewriter/traits.rs
```diff
@@ -184,6 +184,12 @@ impl<'a> GeneratedStruct<'a> {
         let trait_method_inputs =
             rewrite_method_inputs(&self.self_type_ident, &mut trait_method_sig.inputs);
 
+        // We need to rewrite `Self` in the attributes tokens as well
+        // to account for ghost constraints with a bound on `Self`
+        for attr in trait_method_attrs.iter_mut() {
+            rewriter.rewrite_attribute(attr);
```
Aurel300, line 190: Same comment as above.

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+            // Add bounds as a where clause
+            item_fn.sig.generics.where_clause = Some(generate_where_clause_for_spec(
+                &trait_bounds,
+                item_fn.sig.generics.where_clause.as_ref(),
+            ));
+
+            // Add attribute to mark this as a "specification with constraint" (used for processing the contract in `SpecCollector`)
+            item_fn.attrs.push(parse_quote_spanned!(tokens_span=>
+                #[prusti::ghost_constraint_trait_bounds_in_where_clause]
+            ));
```
Aurel300: I think it's cleaner/more consistent to use `{...}` for the `(parse_)quote(_spanned)!` macros.

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+        ));
+        where_clause
+    } else {
+        parse_quote_spanned!(span=>
+            where #trait_bounds
+        )
+    }
+}
+
+fn verify_provided_trait_bounds(trait_bounds: &ProvidedTraitBounds) -> syn::Result<()> {
```
Aurel300: In the context of a verifier, `verify` is a big word :D `validate`?

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+            syn::TypeParamBound::Trait(trait_bound) => {
+                verify_provided_trait_bound(trait_bound)?;
+            }
+        }
+    }
+
+    Ok(())
+}
+
+fn verify_provided_trait_bound(trait_bound: &syn::TraitBound) -> syn::Result<()> {
```
Aurel300: Does this really need to be a method?

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+        new_items.extend(generated_items);
+        new_attrs.extend(generated_attrs);
+    }
+
+    Ok((new_items, new_attrs))
+}
+
+fn generate_where_clause_for_spec(
+    trait_bounds: &ProvidedTraitBounds,
+    preliminary_where_clause: Option<&syn::WhereClause>,
```
Aurel300: `preliminary` might be misleading, if this is referring to the `where` clause written outside a ghost constraint. `existing`?

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+                lifetimes: None,
+                bounded_ty: input.parse()?,
+                colon_token: input.parse()?,
+                bounds: {
+                    let mut bounds = syn::punctuated::Punctuated::new();
+
+                    loop {
+                        if input.is_empty() {
+                            break;
+                        }
```
Aurel300:
```suggestion
                    while !input.is_empty() {
```

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    Ok(())
+}
+
+/// A `syn::Parse`able representaiton of `syn::PredicateType` (ignoring bound lifetimes)
+struct ProvidedTraitBounds {
+    predicate_type: syn::PredicateType,
+}
+
+impl syn::parse::Parse for ProvidedTraitBounds {
```
Aurel300, line 100: This parsing logic and the fact that we parse the bounds *at all* is only so that we can raise a parsing error for invalid bounds (e.g. lifetimes)?

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+                        }
+
+                        let value = input.parse()?;
+                        bounds.push_value(value);
+
+                        if !input.peek(syn::Token![+]) {
+                            break;
+                        }
+
+                        let punct = input.parse()?;
```
Aurel300: Why the `value` and `punct` temporaries?

prusti-specs/src/ghost_constraints/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl syn::parse::Parse for ProvidedTraitBounds {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        Ok(ProvidedTraitBounds {
+            predicate_type: syn::PredicateType {
+                lifetimes: None,
+                bounded_ty: input.parse()?,
+                colon_token: input.parse()?,
+                bounds: {
+                    let mut bounds = syn::punctuated::Punctuated::new();
```
Aurel300: I'm surprised that this is implemented with a loop: does [`Punctuated::parse_separated_nonempty`](https://docs.rs/syn/latest/syn/punctuated/struct.Punctuated.html#method.parse_separated_nonempty) not do the same thing?

prusti-specs/src/specifications/preparser.rs
```diff
@@ -50,6 +50,12 @@ pub fn parse_prusti_assert_pledge(tokens: TokenStream) -> syn::Result<(TokenStre
     Ok((lhs, rhs))
 }
 
+pub fn parse_ghost_constraint(tokens: TokenStream) -> syn::Result<(TokenStream, Vec<NestedSpec<TokenStream>>)> {
+    let pts = PrustiTokenStream::new(tokens);
+    let (trait_bounds_ts, nested_specs_ts) = pts.parse_ghost_constraint()?;
+    Ok((trait_bounds_ts, nested_specs_ts))
```
Aurel300:
```suggestion
    pts.parse_ghost_constraint()
```

prusti-specs/src/specifications/preparser.rs
```diff
@@ -147,7 +155,20 @@ impl PrustiTokenStream {
...
+            tokens: VecDeque::default(),
+            source_span,
+        }
+    }
+
+    fn extend(&mut self, other: PrustiTokenStream) {
+        for token in other.tokens.into_iter() {
+            self.tokens.push_back(token)
+        }
     }
```
Aurel300: I don't think these should need to exist…

prusti-specs/src/specifications/preparser.rs
```diff
@@ -214,6 +235,44 @@ impl PrustiTokenStream {
...
+        // This is due to the fact that the prusti token stream also splits on commas
+        // inside the ghost constraint.
+        let mut arg_iter = arguments.into_iter().peekable();
+        let mut trait_bounds_ts = PrustiTokenStream::empty(span);
+        let mut nested_specs: Option<Vec<NestedSpec<TokenStream>>> = None;
+        while let Some(mut arg) = arg_iter.next() {
+            if arg_iter.peek().is_some() {
+                // arg is part of ghost constraint
+                trait_bounds_ts.extend(arg);
+                trait_bounds_ts.tokens.push_back(PrustiToken::Token(TokenTree::Punct(Punct::new(',', Spacing::Alone))))
```
Aurel300: Why are you extending a `PrustiTokenStream`? Why not just a `TokenStream`?

prusti-tests/tests/verify/ui/ghost-constraints/invalid-trait-refinement-2.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn pre(&self) -> bool;
+    #[pure]
+    fn post(&self) -> bool;
+}
+
+trait MyTrait {
+    fn foo(&mut self);
+}
+
+#[extern_spec] //~ ERROR: the method's postcondition may not be a valid strengthening of the trait's postcondition.
```
Aurel300: Does the `//~ ERROR` annotation do anything in UI tests? (Also in the `invalid-trait-refinement-1.rs` file.)

prusti-tests/tests/verify_overflow/pass/ghost-constraints/split-constraint.rs
```diff
@@ -0,0 +1,24 @@
...
+
+trait A {}
+
+impl A for i32 {}
+
+#[trusted]
+#[ghost_constraint(T: A, [
+    ensures(result % 2 == 0)
+])]
+#[ghost_constraint(T: A, [
```
Aurel300, line 13: Is this meant to be allowed? Is it a syntactic check for whether the bound is the same?

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -0,0 +1,273 @@
...
+                }
+            });
+
+        if all_bounds_satisfied {
+            trace!("Constraint fulfilled");
+            true
+        } else {
+            trace!("Constraint not fulfilled");
+            false
+        }
```
Aurel300:
```suggestion
        trace!("Constraint fulfilled: {all_bounds_satisfied}");
        all_bounds_satisfied
```

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        // There is no caller when encoding a function.
+        // We still resolve obligations to account for constrained specs on a trait
+        // for which we encode its implementation. The corresponding encoding will
+        // contain a behavioral subtyping check which will be performed on the
+        // resolved spec.
+        let param_env_lookup = if let Some(caller_def_id) = context.caller_proc_def_id {
+            env.tcx().param_env(caller_def_id)
+        } else {
+            ty::ParamEnv::reveal_all()
```
Aurel300: Should this not be the `param_env` of the method itself?

prusti-interface/src/environment/mod.rs
```diff
@@ -534,4 +540,45 @@ impl<'tcx> Environment<'tcx> {
...
+        debug!("Evaluating predicate {:?}", predicate);
+        use rustc_trait_selection::traits;
+        use crate::rustc_trait_selection::traits::query::evaluate_obligation::InferCtxtExt;
+
+        let obligation = traits::Obligation {
+            cause: traits::ObligationCause::dummy(),
+            param_env,
+            recursion_depth: 0, // Arbitrarily chosen magic number
+            predicate,
+        };
```
```suggestion
        let obligation = traits::Obligation::new(
            traits::ObligationCause::dummy(),
            param_env,
            predicate,
        );
```

(also gets rid of the not-quite-magic magic)

prusti-interface/src/environment/mod.rs
```diff
@@ -534,4 +540,45 @@ impl<'tcx> Environment<'tcx> {
...
+            let mut obligations = vec![];
+
+            traits::normalize_to(
+                &mut selcx,
+                ty::ParamEnv::reveal_all(),
+                traits::ObligationCause::dummy(),
+                normalizable,
+                &mut obligations,
+            )
+        })
```
Aurel300, line 581: Do our tests pass if we `assert_eq!(obligations.len(), 0);` here?

prusti-specs/src/specifications/preparser.rs
vl0w: Because the ghost constraints are `PrustiTokenStream`s (PTS) and I first want to collect all ghost constraint into one single `PrustiTokenStream` before I can finally parse it with `parse_rust_only`.
When having something like `#[ghost_constraint(T: A<Foo, Baz = Bar>, ...)]`, there will be 3 PTS:
- (1): `T: A<Foo`
- (2): ` Baz = Bar>`
- (3): `...`
I think need to combine (1) and (2) into a new PTS first such that I do not get into unnecessary parsing errors.
Do you think I should immediately call `parse_rust_only` on (1) and (2) and combine the corresponding `TokenStream`s afterwards?

prusti-tests/tests/verify_overflow/pass/ghost-constraints/split-constraint.rs
vl0w: This is just a check that you indeed can have two constraints IF the bounds are equal.
In this case, two spec functions with a where clause `where T: A` will be generated.
When I resolve the constraints, I first collect all of the spec function's `ParamEnv`s with their included where clauses and then compare them. If not equal => throw error. But they will be equal in that case, so no need to be restrictive there.

prusti-interface/src/specs/typed.rs
vl0w: My own variety of English is by far the most superior of all of them 😎

prusti-interface/src/specs/typed.rs
vl0w: Nope, I think it's fine to omit the `SpecGraph` there

prusti-specs/src/extern_spec_rewriter/common.rs
vl0w: I think in order to do that we need to process the nested specs in an extern spec with `generate_spec_and_assertions` like we do for trait spec refinements.
I think this is generally a good idea.
Can we do this after we merged abstract predicates #938? I moved this logic to a new place and would like to avoid nasty merges. So my suggestion:
- Leave this as it is rn
- Finish abstract predicates
- New PR where we eagerly process nested specs in extern specs

prusti-specs/src/ghost_constraints/mod.rs
vl0w: I think it also comes in handy when generate the where clause for the spec function. Ghost constraint trait bounds are actually `syn::PredicateType`s, but they are not parsable by `syn` (see comment on the type). So I made this custom struct.

Do you think this could be done easier?

prusti-specs/src/ghost_constraints/mod.rs
vl0w: I think it does! :)

prusti-tests/tests/verify/ui/ghost-constraints/invalid-trait-refinement-2.rs
vl0w: No, I think this is a leftover, they once were not UI tests

prusti-interface/src/environment/mod.rs
vl0w: No, sadly not. And it might be a problem.
Take the test `associated_types_3.rs`.
When we encode the function `Foo::foo`, we resolve the ghost constraints, but the associated type projection can not be fully normalized. This means that we do no really consider the constrained spec in that case (=> behavioral subtyping will also not be checked).

prusti-specs/src/ghost_constraints/mod.rs
Aurel300: Well, if we simply passed through the predicate as typed by the user, without parsing it with `syn`, we wouldn't need this. But then some validation (e.g. that lifetime bounds aren't allowed) would need to happen post-typing and maybe that's too much hassle.

prusti-specs/src/specifications/preparser.rs
Aurel300: Yes, I think parsing as "Rust only" and pushing into a `TokenStream` is nicer than extending the `PrustiTokenStream` API to support this.

vl0w: Done

Aurel300:
bors retry

Aurel300:
bors retry

PR #907, opened 2022-03-18
Properly derive Hash for Type

### LINKED ISSUE #894: Caching prevents resolution of consistency errors
The following test case, which would cause a `consistency error` was recently fixed
```bash
echo "use prusti_contracts::*;
use std::cmp::{Ord, Ordering};
#[extern_spec]
trait Ord {
    #[pure]
    fn cmp(&self, other: &Self) -> Ordering;
}
#[pure]
fn compare<T: Ord>(a: &T, other: &T) -> Ordering {
    a.cmp(other)
}
fn main() {}" >> test_file.rs
```
If one creates a cache file (running into the `consistency error`) before the fix
```bash
git checkout 6c0f3dedd8ce141f9df4792750f3d90d8da8b7b4
./x.py build --all
PRUSTI_CACHE_PATH=./cache.bin ./target/debug/prusti-rustc --edition=2018 ./test_file.rs
```
And then uses that cache file with a version of Prusti where the `consistency error` shouldn't happen, caching will still erroneously report an error:
```bash
git checkout dde61004e91da0ceafe1addb26fddf3894b513ea
./x.py build --all
PRUSTI_CACHE_PATH=./cache.bin ./target/debug/prusti-rustc --edition=2018 ./test_file.rs
./target/debug/prusti-rustc --edition=2018 ./test_file.rs
```
(The command with caching fails, but the one without caching does not)

JonasAlaif:
Fixes #894

### DIFF
--- a/vir/src/legacy/ast/common.rs
+++ b/vir/src/legacy/ast/common.rs
@@ -129,7 +129,7 @@ impl Ord for PermAmount {
     }
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum Float {
     F32,
     F64,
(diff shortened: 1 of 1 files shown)


vakaras (approved):
LGTM

PR #912, opened 2022-03-22
Generate debugging launch.json for VS Code.

vakaras:
When running `./x.py verify-test` generate `launch.json` file that can be used for debugging that file in VS Code with CodeLLDB extension.


vakaras:
bors ping

PR #915, opened 2022-03-23
Squash commits when merging.

vakaras:
Squash all commits when merging the PR.

@Aurel300 @fpoli @JonasAlaif Let me know if you have any concerns.


fpoli:
Does this means that from now on all commits should be compiling, passing all tests, clippy and fmt? Even when manually merging?

vakaras:
> Does this means that from now on all commits should be compiling, passing all tests, clippy and fmt? Even when manually merging?

No, it simply means that bors will squash the commits into one when merging the PR.

JonasAlaif:
I would say so yeah (that this _should_ be what happens), one would have to select "Squash and merge" from the dropdown when merging manually

Aurel300:
This makes it more difficult to bisect though…

fpoli:
Should we move the decision to the next technical meeting?

vakaras:
> This makes it more difficult to bisect though…

@JonasAlaif pointed out that this potentially would be the opposite because all commits would be the compiling ones.

PR #918, opened 2022-03-23
Add correct lifetime identifiers

pascal-huber:
# Changes
 - [x] Add constant lifetime for references in pure context
 - [x] Add real lifetimes for references created in mixed contexts (pure and impure)
 - [x] Replace real lifetimes with constant lifetimes (again) for pure contexts and make sure all cases are covered 
 - [x] Add function to create viper compatible variable names for Regions

### DIFF
--- a/prusti-viper/src/encoder/encoder.rs
+++ b/prusti-viper/src/encoder/encoder.rs
@@ -713,6 +713,29 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
         self.intern_viper_identifier(full_name, short_name)
     }
 
+    pub fn encode_lifetime_name(&self, region: &ty::Region) -> String {
+        match region.kind() {
+            ty::ReEarlyBound(_) => {
+                unimplemented!("ReEarlyBound: {}", format!("{}", region));
+            },
+            ty::ReLateBound(_, _) => {
+                unimplemented!("ReLateBound: {}", format!("{}", region));
+            },
+            ty::ReFree(_) => {
+                unimplemented!("ReFree: {}", format!("{}", region));
+            },
+            ty::ReStatic=> String::from("lft_static"),
+            ty::ReVar(region_vid) => format!("lft_{}", region_vid.index()),
+            ty::RePlaceholder(_) => {
+                unimplemented!("RePlaceholder: {}", format!("{}", region));
+            },
+            ty::ReEmpty(_) => {
+                unimplemented!("ReEmpty: {}", format!("{}", region));
+            },
+            ty::ReErased => String::from("lft_erased"),
+        }
+    }
+
     pub fn encode_invariant_func_app(
         &self,
         ty: ty::Ty<'tcx>,
--- a/prusti-viper/src/encoder/mir/types/encoder.rs
+++ b/prusti-viper/src/encoder/mir/types/encoder.rs
@@ -61,6 +61,10 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
         format!("fndef${}", self.encoder.encode_item_name(did))
     }
 
+    fn encode_lifetime_name(&self, region: &ty::Region) -> String {
+        self.encoder.encode_lifetime_name(region)
+    }
+
     fn compute_array_len(&self, size: ty::Const<'tcx>) -> u64 {
         self.encoder
             .const_eval_intlike(size.val())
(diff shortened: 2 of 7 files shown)


prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -98,12 +98,10 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
             }
 
-            ty::TyKind::Ref(_, ty, _) => {
-                // TODO: add real lifetime here?
-                let fake_lft = vir::ty::Lifetime {
-                    name: String::from("lft_fake"),
-                };
-                vir::Type::reference(self.encoder.encode_type_high(*ty)?, fake_lft)
+            ty::TyKind::Ref(region, ty, _) => {
+                let lft_name = format!("{}", region);
```
vakaras: The lifetime name probably needs to be a valid Viper identifier.

prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -301,13 +299,11 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
...
             }
-            ty::TyKind::Ref(_, ty, _) => {
+            ty::TyKind::Ref(region, ty, _) => {
                 let target_type = self.encoder.encode_type_high(*ty)?;
-                // TODO: add real lifetime here?
-                let fake_lft = vir_crate::high::type_decl::Lifetime {
-                    name: String::from("lft_fake"),
-                };
-                vir::TypeDecl::reference(target_type, fake_lft)
+                let lft_name = format!("{}", region);
```
vakaras: Same comment about the lifetime name.

vir/defs/high/operations_internal/expression.rs
```diff
@@ -72,6 +103,16 @@ impl Expression {
             _ => false,
         }
     }
+    pub fn erase_lifetime(&mut self) {
```
vakaras: A canonical way of doing this would be to have a folding visitor that replaces all types. Similarly to this: https://github.com/viperproject/prusti-dev/blob/6a9291b92f1d742e36715a3fde15786abd950ea8/vir/defs/high/operations_internal/expression.rs#L245-L260

prusti-viper/src/encoder/mir/types/encoder.rs
pascal-huber: > The lifetime name probably needs to be a valid Viper identifier.

Valid as in not containing any special characters?

vir/defs/high/operations_internal/expression.rs
pascal-huber: agree, will do

prusti-viper/src/encoder/mir/types/encoder.rs
vakaras: Yes. And starting with a letter.

prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -27,6 +27,12 @@ pub struct TypeEncoder<'p, 'v: 'p, 'tcx: 'v> {
     ty: ty::Ty<'tcx>,
 }
 
+// TODO: move this?
+fn encode_lifetime_name(lft: String) -> String {
+    // TODO: check if all lifetime names have the parttern '_#7r
```
Aurel300: That number can be more that one digit long...

prusti-viper/src/encoder/mir/types/encoder.rs
```diff
@@ -99,7 +105,7 @@ impl<'p, 'v, 'r: 'v, 'tcx: 'v> TypeEncoder<'p, 'v, 'tcx> {
             }
 
             ty::TyKind::Ref(region, ty, _) => {
-                let lft_name = format!("{}", region);
+                let lft_name = encode_lifetime_name(format!("{}", region));
```
Aurel300: Maybe there should be a function from `RegionKind` to `String`? I expect many of the `RegionKind` variants will not be visible to us anyway.

prusti-viper/src/encoder/mir/types/encoder.rs
pascal-huber: Yes, currently the first 3 and the last digits are removed and "lft" is prefixed

prusti-viper/src/encoder/encoder.rs
```diff
@@ -713,6 +713,29 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
+            ty::ReVar(region_vid) => format!("lft_{}", region_vid.index()),
+            ty::RePlaceholder(_) => {
+                unimplemented!("RePlaceholder: {}", format!("{}", region));
+            },
+            ty::ReEmpty(_) => {
+                unimplemented!("ReEmpty: {}", format!("{}", region));
+            },
+            ty::ReErased => String::from("lft_erased"),
+        }
+    }
```
pascal-huber, line 737: @vakaras I did this as @Aurel300 suggested (in an other comment). Using the default formatter of Region with slicing or regular expressions did not work very well.

vakaras: Makes sense. Actually, you may also need to use a custom `RegionKind` instead of `String` because we may need to distinguish the static lifetime from others. Also, for erased lifetimes it would be cleaner to have an enum variant instead of a special name. However, let's leave this change for later once we hit the actual tests.

vakaras (approved):
bors r+

PR #919, opened 2022-03-24
Reenable datatype invariants [Attempt 2]

JonasAlaif:
When run under PR #686 these test cases mostly work - I've listed some considerations/issues. @Aurel300 you can take this PR over when you have time.


JonasAlaif:
Ah sorry I missed the update on the original PR. I guess you'll try to get that one working in the end? I can move the examples over there if that's the case

Issue #925, opened 2022-03-26
[Prusti internal error] Consistency error: No matching local variable _2 found with type Ref, instead found Int.


Pointerbender:
The following Rust program, which is a slight variation on #815, triggers a consistency error in Prusti/Viper:

```rust
use prusti_contracts::*;

fn main() {
    let a = A::new(2, 3);
    A::test(a, 4);
}

#[derive(Clone, Copy)]
pub struct A {
    foo: isize,
    bar: usize,
}

impl A {
    #[pure]
    #[requires(num < isize::MAX as usize)]
    pub fn new(num: usize, bar: usize) -> Self {
        Self {
            foo: num as isize + isize::MIN,
            bar,
        }
    }

    #[pure]
    pub fn bar(&self) -> usize {
        self.bar
    }

    #[pure]
    #[requires(num < isize::MAX as usize)]
    pub fn test(a: A, num: usize) -> A {
        A::new(num, a.bar)
    }
}
```

The error message is:

```
error: [Prusti internal error] Prusti encountered an unexpected internal error
  |
  = note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
  = note: Details: consistency error in prusti_example::A::test: Consistency error: No matching local variable _2 found with type Ref, instead found Int. (@36.13)
```

This happens for both Viper programs for `fn main` and `fn test`. In both cases, the relevant Viper code is:

```
function m_new__$TY$__(_1: Int, _2: Int): Snap$struct$m_A
  requires true
  requires _1 < builtin$cast$isize$usize__$TY$__(9223372036854775807)
  requires 0 <= _1
  requires _1 <= 18446744073709551615
  requires 0 <= _2
  requires _2 <= 18446744073709551615
  ensures true
  ensures [result == mirror_simple$m_new__$TY$__(_1, _2), true]
{
  (
    !(
      builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 < -9223372036854775808
      || builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 > 9223372036854775807
    )
      ? cons$0$__$TY$__Snap$struct$m_A$(
          builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808,
           _2.val_int // <-- the error comes from here
        )
      : builtin$unreach__$TY$__Snap$struct$m_A$()
  )
}
```

[...]

Issue #854, opened 2022-02-09
Memory crash during Prusti encoding step


Pointerbender:
Somewhere between commits bb3cb3fe25 and f8e70f9 a regression was introduced that significantly increased memory usage, leading to a crash during the Prusti encoding step in both debug and release mode. On my computer with 16GB of physical memory and 32GB swap file (Ubuntu), the `prusti-driver` process ends up consuming ~100% CPU and 10-11GB of physical memory and another ~25GB of virtual memory, utilizing all available memory (and bringing the whole system to a near-freeze because of excessive swapping in the rest of the OS) before exiting the process during the encoding of a predicate.

I did not have a chance yet to deep dive into this during working weekdays, but I'm posting my preliminary findings here already in the hopes that someone might be able to spot where the problem could lie. I'll have some more time on the weekend to look deeper :)

The problem appears in a relatively large (WIP and unpublished) project, which makes debugging harder than usual. My plan of approach to narrow it down is:

1. Post initial findings (this post).
2. Try to build a minimal reproduction (I think this might be hard, commenting out random parts of the predicate seem to bring down the resource consumption to acceptable levels and it's not possible to pin-point a certain clause in the predicate that might be responsible).
3. Try to narrow down the exact commit that introduced the regression, in the hopes that a smaller diff may yield new insights.
4. If all else fails, publish a draft version of the project for closer inspection (although I hope I can prevent burdening anyone with having to debug the whole thing :) )

Currently I'm on commit f8e70f9037 and the `cargo-prusti` command I use to verify the project is:

```shell
# Release mode
RUST_BACKTRACE=1 PRUSTI_LOG=info PRUSTI_DUMP_VIPER_PROGRAM=true time prusti-dev/target/release/cargo-prusti
# Debug mode
RUST_BACKTRACE=1 PRUSTI_LOG=debug PRUSTI_DUMP_VIPER_PROGRAM=true time prusti-dev/target/debug/cargo-prusti
```

Its `Prusti.toml` contents are:

```toml
check_overflows = true
check_panics = true
skip_unsupported_features = true
assert_timeout = 60000
optimizations = "inline_constant_functions,delete_unused_predicates,optimize_folding,remove_empty_if,purify_vars,remove_unused_vars,remove_trivial_assertions,clean_cfg" # all minus fix_quantifiers
extra_verifier_args = ["--z3ConfigArgs=\"smt.random_seed=42\""]
```

In the moment right before crashing, the debug logs show the following:

```
[...]

PR #928, opened 2022-03-27
fix for #925

Pointerbender:
This PR addresses issue #925 where a problem was uncovered with encoding the arguments of a pure function.

There was already an earlier attempt to fix this which already lives here:

https://github.com/viperproject/prusti-dev/blob/6a9291b92f1d742e36715a3fde15786abd950ea8/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs#L142-L167

However, it does not seem to catch all corner cases. I pretty much copied and pasted this code to `PureFunctionEncoder::encode_function_given_body` after the snapshot patching here:

https://github.com/viperproject/prusti-dev/blob/6a9291b92f1d742e36715a3fde15786abd950ea8/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs#L343-L350

Originally, I wanted to move the existing logic there instead of copying it, but it turns out that in order to patch the snapshots the arguments need to be "fixed", and in order to fix the arguments (again) the snapshots need to be patched :smile: 

The need for code duplication in this case and the presence of this circular code dependency, suggests there might exist a better way to address this. There is one other avenue I explored, although I couldn't piece it together, I'm including an explanation of it here, just in case:

To recap, the offending Viper code for #925 is:

```
function m_new__$TY$__(_1: Int, _2: Int): Snap$struct$m_A
  requires true
  requires _1 < builtin$cast$isize$usize__$TY$__(9223372036854775807)
  requires 0 <= _1
  requires _1 <= 18446744073709551615
  requires 0 <= _2
  requires _2 <= 18446744073709551615
  ensures true
  ensures [result == mirror_simple$m_new__$TY$__(_1, _2), true]
{
  (
    !(
      builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 < -9223372036854775808
      || builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808 > 9223372036854775807
    )
      ? cons$0$__$TY$__Snap$struct$m_A$(
          builtin$cast$usize$isize__$TY$__(_1) + -9223372036854775808,
           _2.val_int // <-- the error comes from here
        )
      : builtin$unreach__$TY$__Snap$struct$m_A$()
  )
}
```

Here the second argument to the constructor of `A` is not properly encoded, it passes `_2.val_int` as a value instead of `_2`. The way the `PureFunctionBackwardInterpreter` walks the MIR to find this parameter, goes like this:

mir

It starts at the end since it goes backwards, so the first time it encounters the second parameter is at MIR statement `_0 = A { foo: move _3, bar: move _7 }`. Next, it sees the assignment `_7 = _2;`. In theory, here the backwards interpreter can already recognize that `_2` is a (pure) function argument and that there will be no further statements encountered that would replace `_2`. At this point, it could already patch this assignment by wrapping `_2` in a field expression. This would happen here:

https://github.com/viperproject/prusti-dev/blob/e9edf18d1c6fa99cb4eaac1180d67e3837a661c1/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L798-L808

As long as `_2` is not wrapped in a field expression, the existing code in `PureFunctionEncoder::encode_function` is not able to fix this argument. The existing code replaces `target_place` which is assembled by calling `Encoder::encode_value_expr` (which in this case indirectly determines that the wrapping field is of type `vir::Type::Int` and has field name `val_int` and wraps it as such).

I was not able to properly puzzle this together in the code itself, but if it turns out that this is the preferred way to fix it, I could take another shot at it later this week :) Alternatively, an even better fix could also be deferred until a forward MIR interpreter is implemented (there was some talk of this not too long ago in #854, in order to bring down the memory consumption for encoding pure expressions with many branches).

I also included the example from #925 as a regression test.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-925.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-925.rs
@@ -1,0 +1,34 @@
+use prusti_contracts::*;
+
+fn main() {
+    let a = A::new(2, 3);
+    A::test(a, 4);
+}
+
+#[derive(Clone, Copy)]
+pub struct A {
+    foo: isize,
+    bar: usize,
+}
+
+impl A {
+    #[pure]
+    #[requires(num < isize::MAX as usize)]
+    pub fn new(num: usize, bar: usize) -> Self {
+        Self {
+            foo: num as isize + isize::MIN,
+            bar,
+        }
+    }
+
+    #[pure]
+    pub fn bar(&self) -> usize {
+        self.bar
+    }
+
+    #[pure]
+    #[requires(num < isize::MAX as usize)]
+    pub fn test(a: A, num: usize) -> A {
+        A::new(num, a.bar)
+    }
+}
--- a/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
+++ b/prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs
@@ -346,6 +346,30 @@ impl<'p, 'v: 'p, 'tcx: 'v> PureFunctionEncoder<'p, 'v, 'tcx> {
             .patch_snapshots_function(function, &self.tymap)
             .with_span(self.mir.span)?;
 
+        // Fix arguments
+        if let Some(mut body) = function.body.take() {
+            for arg in self.mir.args_iter() {
+                let arg_ty = self.interpreter.mir_encoder().get_local_ty(arg);
+                let span = self.get_local_span(arg);
+                let target_place = self
+                    .encoder
+                    .encode_value_expr(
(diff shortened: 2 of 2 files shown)


fpoli:
Thanks for explaining the current behavior! The fix looks good to me, although the current "is value"/"is not value" distinction is a leftover from the before-snapshots era. I believe a lot of the interpreter can be simplified by treating everything as value (but without the final `.val_int` fields in VIR) inside the pure function encoder, applying some conversion only at the boundary with impure code, but I had not enough time so far to look into it.
https://github.com/viperproject/prusti-dev/blob/6a9291b92f1d742e36715a3fde15786abd950ea8/prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs#L866-L867

PR #929, opened 2022-03-28
Test that Prusti doesn't crash on popular crates

fpoli:
This PR adds a "test kind" column to the list of public crates tested by `test-crates` [1]. The supported values are:
* `NoErrors`: Test that Prusti doesn't crash nor report "internal/invalid" errors.
* `NoCrash`: Test that Prusti doesn't crash nor report "invalid" errors.
* `Skip`: Skip the crate, because Prusti crashes or the crate doesn't compile with the standard compiler.

A new `INTERNAL_ERRORS_AS_WARNINGS` configuration flag makes Prusti report internal errors as warnings.

After this PR is merged, `test-crates` will only be triggered for push events on `master` and on the branches managed by bors. So, every PR should be merged via bors only, not via the merge button.

[1]: With `check_panics=false`, `check_overflows=false` and `skip_unsupported_features=true`


fpoli:
I estimate that after this PR `test-crates` will take 3 hours and a half, in every CI run. (I suspect that the bottleneck is Polonius, but I haven't checked carefully.)

fpoli:
> I estimate that after this PR `test-crates` will take 3 hours and a half, in every CI run. (I suspect that the bottleneck is Polonius, but I haven't checked carefully.)

Checking a random crate (`hex`), most of the time is actually spent verifying Viper programs:
* `cargo +nightly-2022-03-15 build` runs in 2s
* `RUSTFLAGS="-Zpolonius" cargo +nightly-2022-03-15 build` (and more of the flags added by Prusti) runs in 5s
* `cargo prusti` runs in 3m25s:
    * Initially, the compiler runs for 17s
        * Prusti probably slows down the compiler by e.g. cloning the `BodyWithBorrowckFacts` types in the `mir_storage` component.
    * Prusti encodes the crate to VIR in 22s
    * Prusti-server verifies the programs, one by one, in 2m45s
        * Each program takes from 0.6s to 7s. Most of them seem to take ~0.7s.

PR #832, opened 2022-01-21
Add Caching Documentation

JonasAlaif:
To be merged once "Prusti Assistant" performs caching

PR #931, opened 2022-03-29
Update flags in dev guide

Aurel300:
Updated the list in the dev guide to reflect the new config flags. Also made all the docs for the config methods match those of the dev guide, so there aren't two slightly different versions.

@JonasAlaif I took the documentation for the cache flags from #832, because the flags are already there and work. However, #832 should still add the actual section on caching when appropriate.


JonasAlaif:
The caching doc PR is now #973 
When are you planning on merging this @Aurel300?

PR #938, opened 2022-04-01
Abstract predicates

vl0w:
This PR introduces *abstract predicates*, which are predicates which are declared on a trait:
```rust
trait SomeTrait {
    // Define an abstract predicate
    predicate! {
        fn pred(&self) -> bool;
    }
}
struct SomeStruct { ... }
#[refine_trait_spec]
    // Implement the abstract predicate
    predicate! {
        fn pred(&self) -> bool {
            ...
        }
    }
}
```

These abstract predicates can then be used in spec code:
```rust
#[requires(s.pred())]
#[ensures(s.pred())]
fn foo(s: SomeStruct) { ... }
```
And even in a generic setting:
```rust
#[requires(s.pred())]
#[ensures(s.pred())]
fn bar<T: SomeTrait>(s: T) { ... }
```

An abstract predicate (like in `bar` above) is always encoded as a pure bodyless Viper function. Thus, abstract predicates can never have a body (i.e. a default implementation on the trait method). This restriction also allows us to not need to worry about refinement.

## Other notable changes
- `#[refine_trait_spec]` now supports implementing a local trait on a non-local type. This allows for implementing abstract predicates on non-local types.
    Implementation detail: Spec functions in `#[refine_trait_spec] impl A for B` are not anymore generated in an `impl B` block, but are free-standing functions now.

# Work to do
- [x] Parsing
- [x] Ensure that abstract predicates are not used in non-spec code
- [x] Ensure abstract predicates don't have bodies
- [x] Ensure that we don't declare predicates in extern specs
- [x] Support extern types in `#[refine_trait_spec]`
- [x] Resolve all `TODO hansenj` and `TODO abstract-predicates`
- [x] Adapt to new `ProcedureSpecificationKind` and handle abstract predicates
- [x] Testing (Parsing, Validation, Usage)


prusti-interface/src/specs/checker/predicate_checks.rs
```diff
@@ -114,16 +144,33 @@ impl<'v, 'tcx: 'v> NonSpecExprVisitor<'tcx> for CheckPredicatesVisitor<'tcx> {
...
     fn visit_expr(&mut self, ex: &'tcx hir::Expr<'tcx>) {
+        let owner_def_id = ex.hir_id.owner;
+
+        // General check: The "path" of a predicate doesn't appear anywhere
+        // (e.g. as in a function call or an argument when we pass the predicate to another function)
         if let hir::ExprKind::Path(ref path) = ex.kind {
-            let def_id = ex.hir_id.owner;
-            if self.tcx.is_mir_available(def_id) && !self.tcx.is_constructor(def_id.to_def_id()) {
-                let res = self.tcx.typeck(def_id).qpath_res(path, ex.hir_id);
+            if self.tcx.is_mir_available(owner_def_id) && !self.tcx.is_constructor(owner_def_id.to_def_id()) {
```
Aurel300, line 152: Why are we checking for `is_constructor` (and `is_mir_available`) here and in line 165?

prusti-specs/src/common.rs
```diff
@@ -1,29 +1,338 @@
...
+    /// }
+    /// ```
+    /// will be rewritten to
+    /// ```ignore
+    /// fn foo(_self: &mut T, arg1: Bar) -> Baz {
+    ///     _self.qux()
+    /// }
+    /// ```
+    pub(crate) trait RewritableReceiver {
+        fn rewrite_receiver<T: ToTokens>(&mut self, new_ty: &T);
```
Aurel300: Can the `T` here be something specific, e.g. `TypePath`? Is there a use for providing different tokenisable types?

prusti-specs/src/common.rs
```diff
@@ -1,29 +1,338 @@
...
+                    TokenTree::Group(new_group)
+                }
+                TokenTree::Ident(ident) if ident == "self" => {
+                    TokenTree::Ident(proc_macro2::Ident::new("_self", ident.span()))
+                },
+                other => other,
+            }));
+            parse_quote_spanned! {tokens_span=>
+            #rewritten
+        }
```
Aurel300: Indent mishap.

prusti-specs/src/common.rs
```diff
@@ -1,29 +1,338 @@
...
-impl HasGenerics for syn::ItemImpl {
-    fn get_generics(&self) -> &syn::Generics {
-        &self.generics
+/// Copies the [syn::Generics] of `source` to the generics of `target`
+pub(crate) fn merge_generics<T: HasGenerics>(target: &mut T, source: &T) {
+    let generics_target = target.generics_mut();
+    let generics_source = source.generics();
+
+    generics_target.params.extend(generics_source.params.clone());
+    if let (Some(target_where), Some(source_where)) =
```
Aurel300: What if `generics_target.where_clause` is `None` but `generics_source.where_clause` is `Some(_)`? Then `generics_target` should become `Some(_)`, right?

prusti-specs/src/extern_spec_rewriter/common.rs
```diff
@@ -1,27 +1,138 @@
...
+    // - Rewrite "self" to "_self"
+    let self_type_path = parse_quote_spanned! {self_type.span()=> #self_type };
+
+    let mut stub_method = stub_method.expect_impl_item();
+    stub_method.attrs.extend(generated_attributes);
+    stub_method.rewrite_self_type(&self_type_path, self_type_trait);
+    stub_method.rewrite_receiver(&self_type_path);
+
+    // Set span of generated method to externally specified method for better error reporting
+    syn::visit_mut::visit_impl_item_method_mut(&mut SpanOverrider::new(method_sig_span), &mut stub_method);
```
Aurel300, line 83: Which spans will this override at this point? Is it still necessary?

prusti-specs/src/extern_spec_rewriter/common.rs
```diff
@@ -1,27 +1,138 @@
...
+                FnArg::Receiver(_) => {
+                    args.push(parse_quote_spanned! {span=>self})
                 }
-                ident.into_token_stream()
+                _ => unimplemented!(),
             }
-            _ => token.into_token_stream(),
-        }))
+        }
+        args
```
Aurel300: `Punctuated` has [`FromIterator`](https://docs.rs/syn/latest/syn/punctuated/struct.Punctuated.html#impl-FromIterator%3CT%3E), so all this should be expressible with an iterator chain + `map` + `collect`. The `push_punct` in particular seems unneccessary if it is not correctly spanned anyway.

prusti-specs/src/lib.rs
```diff
@@ -394,43 +409,52 @@ pub fn refine_trait_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream
...
                 };
                 new_items.push(new_item);
             },
+            syn::ImplItem::Macro(makro) if is_predicate_macro(&makro) => {
+                let parsed_predicate = handle_result!(predicate::parse_predicate_in_impl(makro.mac.tokens.clone()));
+
+                let predicate = match parsed_predicate {
+                    ParsedPredicate::Impl(predicate) => predicate,
+                    _ => unreachable!(),
+                };
```
Aurel300: Can we use our `force_match!` macro for these?

prusti-tests/tests/typecheck/fail/abstract-predicates/invalid-sig-1.rs
```diff
@@ -0,0 +1,22 @@
...
+    }
+}
+
+struct SomeStruct;
+
+#[refine_trait_spec]
+impl SomeTrait for SomeStruct {
+    predicate! {
+        fn foo(&self) -> i32 {
+            42 //~ ERROR: mismatched types
```
Aurel300: Why this error?

prusti-tests/tests/verify_overflow/pass/trait-contracts-refinement/refine-non-local-type-with-generics.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[extern_spec]
+impl<T: PartialEq> std::option::Option<T> {
+    #[pure]
+    #[ensures(matches ! (* self, Some(_)) == result)]
+    pub fn is_some(&self) -> bool;
+
+    #[pure]
+    #[ensures(self.is_some() == ! result)]
+    #[ensures(matches ! (* self, None) == result)]
```
Aurel300: Is the funky spacing here intentional?

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
```diff
@@ -275,15 +276,19 @@ impl<'v, 'tcx: 'v> PureFunctionEncoderInterface<'v, 'tcx>
...
-                        (
-                            pure_function_encoder.encode_predicate_function(&predicate_body)?,
-                            false,
-                        )
-                    } else if self.is_trusted(proc_def_id, Some(substs)) {
+                let (mut function, needs_patching) = match (
+                    self.is_trusted(proc_def_id, Some(substs)),
+                    self.get_proc_kind(proc_def_id, Some(substs)),
+                ) {
+                    // Predicate bodies are encoded even though the function is trusted
```
Aurel300: This seems strange: do we need to support this? Can we not error out in parsing and/or not mark a predicate as trusted?

prusti-viper/src/encoder/mir/specifications/interface.rs
```diff
@@ -63,26 +65,29 @@ impl<'tcx> SpecQuery<'tcx> {
...
             }
-            PureOrTrustedCheck(_, _) => PureOrTrustedCheck(new_def_id, new_substs),
+            GetProcKind(_, _) => GetProcKind(new_def_id, new_substs),
             FetchSpan(_) => FetchSpan(new_def_id),
         }
     }
 }
 
 pub(crate) trait SpecificationsInterface<'tcx> {
+    // TODO abstract-predicates: Maybe this should be deleted (and ProcedureSpecificationKind::is_pure)
```
Aurel300, line 75: I agree, it might be confusing that predicates also return `true` here. How many usages are there?

prusti-specs/src/common.rs
vl0w: I think the original idea of this was because the implementation is purely token based, i.e. `TypePath` would be tokenized anyway. But I think this is a good idea, the interface should be designed according to client's needs, not the implementation.

vl0w: And actually, I think this was not considered during my refactoring (of the refactoring). `T` makes no sense anymore.

prusti-specs/src/common.rs
vl0w: Indeed! Also, type params could be duplicated. I fixed this and added some tests for it.

prusti-specs/src/extern_spec_rewriter/common.rs
vl0w: I re-added this line because some tests were failing.

prusti-tests/tests/typecheck/fail/abstract-predicates/invalid-sig-1.rs
vl0w: I created a UI test, the error message is better there.

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
vl0w: My change there resulted from existing code, so I did not try to change existing behavior.
I double-checked: We can not mark a predicate as trusted.

vl0w: My question then is how should the encode work if we had trusted predicates? Should they be encoded as bodyless functions (like trusted pure functions) or should the body still be encoded? Before my change, a predicate would not have been encoded by a bodyless function.

prusti-tests/tests/verify_overflow/pass/trait-contracts-refinement/refine-non-local-type-with-generics.rs
vl0w: No :)

prusti-viper/src/encoder/mir/specifications/interface.rs
vl0w: I found 8 in total.
We maybe could just rename the method to `must_be_encoded_as_pure` or something similar, afaik this is how we interact with this method.

prusti-interface/src/specs/checker/predicate_checks.rs
vl0w: Idk to be honest. These checks were already there and I did not question them. Let's see if the build fails.

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
Aurel300: I guess they would be abstract/bodyless, yes… The motivation for predicates was to support Prusti syntax inside methods, but with a trusted method the body should be ignored anyway. So in the end, the only thing a trusted predicate does vs a trusted pure method is that it cannot be used in non-spec code. I guess that would be the case if you could declare these in an extern spec but maybe that's a stretch.

So, if we cannot mark a predicate as trusted anyway (not even via extern specs), can this code be simplified any (or at least the comment changed)?

prusti-specs/src/common.rs
```diff
@@ -1,29 +1,358 @@
...
+    // small amount of type parameters in general (n < 10)
+    for param_source in generics_source.params.iter() {
+        // Lifetimes and consts are currently not handled
+        if let GenericParam::Type(type_param_source) = param_source {
+            let in_target = generics_target.params.iter_mut()
+                .find(|gp| match gp {
+                    GenericParam::Type(tp) => tp.ident.eq(&type_param_source.ident),
+                    _ => false
+                }
+                ).map(|gp| force_matches!(gp, GenericParam::Type(tp) => tp));
```
Aurel300: This sounds like you want a `HashSet` of the identifiers already inserted...

prusti-specs/src/lib.rs
```diff
@@ -394,43 +409,49 @@ pub fn refine_trait_spec(_attr: TokenStream, tokens: TokenStream) -> TokenStream
...
+            syn::ImplItem::Macro(makro) if is_predicate_macro(&makro) => {
+                let parsed_predicate = handle_result!(predicate::parse_predicate_in_impl(makro.mac.tokens.clone()));
+
+                let predicate = force_matches!(parsed_predicate, ParsedPredicate::Impl(p) => p);
+
+                // Patch spec function: Rewrite self with _self: <SpecStruct>
+                let mut spec_function = match predicate.spec_function {
+                    syn::Item::Fn(item_fn) => item_fn,
+                    _ => unreachable!(),
+                };
```
Aurel300: Is this also not a `force_matches!`?

prusti-tests/tests/typecheck/ui/abstract-predicates/invalid-sig-1.stderr
```diff
@@ -0,0 +1,9 @@
+error[E0308]: mismatched types
+  --> $DIR/invalid-sig-1.rs:15:13
+   |
+15 |             42 //~ ERROR: mismatched types
```
Aurel300: This still doesn't explain what this error is. Are predicates supposed to only return `bool`? I don't think this should be the case. (If it were, the error here should highlight the return type, not the expression.) I'm guessing this error is raised because of a `(... : bool)` type check expression, which should not exist for predicate spec items (but should for pre/postconditions).

vl0w: I believed that predicates are meant to return `bool`. However, the user docs don't mention that and I think my assumption was wrong after all.
However, this PR doesn't change the encoding/behavior of predicates at all.
I think if we want to fix this we should do it in a separate PR then and I will delete this test.

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
vl0w: Yes, we also cannot add a `trusted` flag via extern spec, even though the error is not really helpful ("cannot use predicate in non-spec code" because of the expansion of the extern spec).

If the general behavior is to always encode a trusted function as bodyless I will adjust it.

prusti-tests/tests/typecheck/ui/abstract-predicates/invalid-sig-1.stderr
vl0w: I deleted the test

prusti-specs/src/common.rs
vl0w: It screams for `HashABC`s ^^
I refactored it again. Maybe it could be done more elegantly?

prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs
vl0w: My naive understanding and refactoring caused `ui/predicate.rs` to fail.

I (maybe) understand this better now: Predicate functions *are* trusted. In contrast to pure functions, we cannot add a functional spec to the predicate, the predicate *is* the spec. Therefore, in the pure function encoder, we have to check first whether a function is a predicate to encode it, because they will otw. not do anything at all.

Does my explanation make sense? I think I can adjust the comment in the code a bit to clarify.

Aurel300: This indirection of a spec item in a separate method is an implementation detail. I don't want to conflate this with `trusted` methods in the code or comments too much.

Aurel300:
Did somebody cancel the run?

bors retry

PR #944, opened 2022-04-05
Disallow spacing between Prusti operators

JonasAlaif:
Works with the changes here


Aurel300:
What is this meant to fix? The test failure looks relevant.

There are some things wrong with the spacing parsing, in particular something like `>,` (the end of a generic argument list before a comma) might misparse, but I'm not sure this helps.

vakaras:
@JonasAlaif This is the wrong approach. You need to actually get the span and check that the spans are next to each other.

JonasAlaif:
Hmm it's strange that this fails though (when used with `quote!`), running it on actual code it seems to work.
Passing:
pass

Failing:
fail

JonasAlaif:
@vakaras, @Aurel300 CI passes now, what do you think?

vakaras (commented):
Looks good to me, but I am not an expert so would be good to get @Aurel300 feedback.

Aurel300:
LGTM

bors r+

PR #945, opened 2022-04-05
some fixes and regression tests for #738 related to retaining domain axioms

Pointerbender:
I found a few more cases where the definition collector is a little too strict when removing domain axioms (#738). I included some fixes and regression tests in this PR. A short summary of the changes:

I extended the logic that determines whether to retain the domain axioms to also include cases such as:

* Retain the injectivity axiom when the constructor is "used".
* Also retain the field axiom when the field access snap function is used.

On top of that, I had to change a bit of logic in `prusti-viper/src/encoder/mir/pure/pure_functions/encoder.rs` so that Prusti encoded the type bound preconditions before the other preconditions. Apparently the order in which these are defined matters for Viper (tested by regression test `issue-738-4.rs`).

(This PR does not fully fix the related issue #738)

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-738-2.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-738-2.rs
@@ -1,0 +1,11 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[pure]
+pub fn get(a: &usize) -> usize {
+    *a
+}
+fn foo(a: &usize) {
+    let v = get(a);
+}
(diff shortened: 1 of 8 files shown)


Pointerbender:
p.s. I also found a fix for [example 2](https://github.com/viperproject/prusti-dev/issues/738#issuecomment-1046935825) of #738 and just pushed this as a second commit with some extra regression tests. This PR now addresses all known examples from #738, but that issue is probably still best left open until the refactoring is completed.

vakaras (approved):
Looks great! Thanks!

bors r+

PR #947, opened 2022-04-06
Specifications: Add new enum which indicates whether a procedure is impure/pure/predicate

vl0w:
Adds a new enum `ProcedureSpecificationKind` which indicates whether a `ProcedureSpecification` is impure, pure or a predicate.

This PR is a predecessor of #938, in which we introduce the notion of abstract predicates (i.e. mark trait methods as predicates).

**Notable (breaking) change:**
When refining a trait, the `ProcedureSpecificationKind` is not automatically inherited. For example, the following code causes a Prusti error:
```rust
trait MyTrait {
    #[pure]
    fn foo();
}

struct MyStruct;
impl MyTrait for MyStruct {
    // not marked pure!
    fn foo() { }
}
```
We require the user to explicitly mark `MyStruct::foo` as pure.

Generally, functions are either impure, pure or predicates with the following relation: predicate functions ⊂ pure functions ⊂ impure functions. This error is emitted whenever this relation is violated.


prusti-interface/src/specs/typed.rs
```diff
@@ -134,7 +136,67 @@ impl SpecificationItem<bool> {
...
+    /// Ensures that refined [ProcedureSpecificationKind]s are valid.
+    /// See [ProcedureSpecificationKindError] for detailed error descriptions.
+    pub fn validate(&self) -> Result<(), ProcedureSpecificationKindError> {
+        if let SpecificationItem::Refined(base, refined) = self {
+            match (base, refined) {
+                (ProcedureSpecificationKind::Predicate(_), ProcedureSpecificationKind::Pure) |
+                (ProcedureSpecificationKind::Predicate(_), ProcedureSpecificationKind::Impure) |
+                (ProcedureSpecificationKind::Pure, ProcedureSpecificationKind::Impure) =>
+                    Err(ProcedureSpecificationKindError::InvalidSpecKindRefinement(*base, *refined)),
+                _ => Ok(())
```
Aurel300: Let's flip this match around: match on kind refinements that are valid, and make the `_` case raise the error. This should be a little bit more robust against future extensions.

vl0w: Done

prusti-interface/src/specs/typed.rs
```diff
@@ -134,7 +136,71 @@ impl SpecificationItem<bool> {
...
+    /// See [ProcedureSpecificationKindError] for detailed error descriptions.
+    pub fn validate(&self) -> Result<(), ProcedureSpecificationKindError> {
+        use ProcedureSpecificationKind::*;
+        if let SpecificationItem::Refined(base, refined) = self {
+            match (base, refined) {
+                // An impure method can be refined to anything
+                (Impure, _) |
+                // An pure method can be refined to pure method or a predicate
+                (Pure, Pure) |
+                (Pure, Predicate(_)) |
```
Aurel300: Thinking about this some more: I don't think pure -> predicate or impure -> predicate should be allowed.

```rust
trait Foo        {             fn blah(&self) -> bool;                 }
impl Bar for Foo { predicate!{ fn blah(&self) -> bool { forall ... } } }
```

Here `Foo::blah` is impure and callable from non-spec code but a predicate should only be callable from specs. So a method like this might result in a call to the predicate from non-spec code:

```rust
fn client<T: Foo>(x: &T) { x.blah(); }
```

vl0w: Yeah that's right, this validation also needs to consider that no calls can happen to predicates. 
For your `client`, we would need to validate all callsites, which is not really practical and modular.
So we should only allow the following refinements:
- Impure -> {Impure, Pure}
- Pure -> Pure
- Predicate -> Predicate

PR #958, opened 2022-04-14
Add lifetime encoding at end of block

pascal-huber:
# Work

 - [x] Count number of lifetimes for read permission amount
 - [x] Use NewLft statements for lifetimes needed by the target block of the goto-terminator
 - [x] Implement and use ShortenLifetime statement to shorten derived lifetimes as needed by the target block of the goto-terminator

# Example

Here the if-branch creates lifetimes `lft6 ⊆ bw0`, the else-branch creates `lft6 ⊆ bw1`. After the branch we require the lifetimes to be the same. This means that the if-branch also needs to create `bw1` and the else-branch also needs to create `bw0` and `lft6` must be shortened to  `lft6 ⊆ bw0 ∩ bw1`.

```
fn main(c: bool){
    let mut a: i32 = 4;
    let mut b: i32 = 5;
    let x;
    if c {
        x = &mut a;
    } else {
        x = &mut b;
    }
    *x = 6;
}
```

### DIFF
--- a/prusti-viper/src/encoder/mir/procedures/encoder.rs
+++ b/prusti-viper/src/encoder/mir/procedures/encoder.rs
@@ -61,6 +61,7 @@ pub(super) fn encode_procedure<'v, 'tcx: 'v>(
     };
     let mir = procedure.get_mir();
     let locals_without_explicit_allocation: BTreeSet<_> = mir.vars_and_temps_iter().collect();
+    let rd_perm = lifetimes.lifetime_count();
     let mut procedure_encoder = ProcedureEncoder {
         encoder,
         def_id,
@@ -70,6 +71,7 @@ pub(super) fn encode_procedure<'v, 'tcx: 'v>(
         check_panics: config::check_panics(),
         locals_without_explicit_allocation,
         fresh_id_generator: 0,
+        rd_perm,
     };
     procedure_encoder.encode()
 }
@@ -86,6 +88,7 @@ struct ProcedureEncoder<'p, 'v: 'p, 'tcx: 'v> {
     /// the entire body of the function.
     locals_without_explicit_allocation: BTreeSet<mir::Local>,
     fresh_id_generator: usize,
+    rd_perm: u32,
 }
 
 impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
@@ -357,11 +360,6 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         Ok(())
     }
 
-    fn read_permission_amount(&mut self) -> u32 {
-        // TODO: count lifetimes, not cfg_edges
-        self.lifetimes.edge_count()
-    }
-
     fn encode_basic_block(
         &mut self,
         procedure_builder: &mut ProcedureBuilder,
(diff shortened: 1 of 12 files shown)


prusti-viper/src/encoder/mir/procedures/encoder.rs
```diff
@@ -104,7 +104,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
             deallocate_returns,
             assert_postconditions,
         );
-        self.encode_body(&mut procedure_builder)?;
+        let rd_perm: u32 = self.lifetimes.lifetime_count();
```
vakaras: Why is this not a field on `ProcedureEncoder`, but an additional argument/parameter that needs to be threaded through?

vir/defs/high/ast/statement.rs
```diff
@@ -188,3 +189,11 @@ pub struct GhostAssignment {
     pub value: Expression,
     pub position: Position,
 }
+
+#[display(fmt = "{} := shorten_lifetime({:?}, {})", target, value, rd_perm)]
+pub struct ShortenLifetime {
+    pub target: VariableDecl,
+    pub value: Expression,
```
vakaras: Is this statement supposed to intersect a series of lifetimes and while doing so modify their `LifetimeTokens`? If yes, then using `Vec<LifetimeConst>` would make it easier to generate `vir_low` than `Expression`.

vakaras (changes requested):
LGTM, except for the two comments I added.

vir/defs/high/operations_internal/identifier/ty.rs
```diff
@@ -128,7 +128,11 @@ impl WithIdentifier for ty::Slice {
 
 impl WithIdentifier for ty::Reference {
     fn get_identifier(&self) -> String {
-        format!("ref${}", self.target_type.get_identifier())
+        format!(
+            "ref$&{}${}",
+            self.lifetime.get_identifier(),
+            self.target_type.get_identifier(),
+        )
```
pascal-huber: @vakaras this change may be causing some problems, looking into it

prusti-viper/src/encoder/mir/procedures/encoder.rs
pascal-huber: yes, done.

vir/defs/high/ast/statement.rs
pascal-huber: I think you're right, this makes no sense. In vir_low I want to create something like this:

```
var old_lft6: Lifetime := lft6
lft6 := lft_tok_sep_take(bw0, bw1, q)
bor_shorten(lft6, old_lft6, q, _4.ref)
```

Having a vector would be better here. And `_4.ref` we will have to look up in vir_low, right?

vakaras: Does `ShortenLifetime` correspond to `lft_tok_sep_take`, to `bor_shorten`, or to both? If `bor_shorten` is needed, then I would say that you have to include `_4` into the struct.

pascal-huber: Originally I was intending to do `lft_tok_sep_take` only - but doing both at the same time seems nicer. I think some work will be required to include `_4`, not sure.

pascal-huber: renamed to `LifetimeTake`.

vakaras (approved):
LGTM

bors r+

PR #970, opened 2022-04-22
Fix a bug in the purification optimization

vakaras:
Fixes a bug in the purification optimization: if a purified type is moved, it is potentially not unfolded:
```
  // [mir] _49 = move _111
  _49 := _111
```
and optimization would incorrectly change it to:
```
  // [mir] _49 = move _111
  _49 := _111.val_int
```
which fails to do permission error because `_111` is not unfolded. This PR fixes the problem by forcing copy assignments for all purifiable types.

### DIFF
--- a/prusti-common/src/vir/optimizations/purification/mod.rs
+++ b/prusti-common/src/vir/optimizations/purification/mod.rs
@@ -17,6 +17,11 @@ pub fn purify_methods(
     methods
 }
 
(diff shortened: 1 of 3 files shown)


vakaras:
bors retry

vakaras:
bors retry

2022-04-22, pn 22:32, bors[bot] ***@***.***> rašė:

Issue #940, opened 2022-04-04
Cache collisions


fpoli:
It's possible to trigger collisions in the caching of verification results.

Step 1: open the Prusti IDE and verify this program
```rust
use prusti_contracts::*;

fn a() {}

#[ensures(false)]
fn test() {}

fn main() {}
```

Step 2: Comment `fn a() {}` and verify the program again.
```rust
use prusti_contracts::*;

//fn a() {}

#[ensures(false)]
fn test() {}

fn main() {}
```

Prusti will fail with an internal error, because the verification request to verify `test` is the same as in step 1, but with different positions. So, the cached verification result will have the wrong position.

JonasAlaif:
Hmm doesn't seem so easy to fix - might need to change how positions are encoded and sent to Viper? Any suggestions @fpoli?

fpoli:
The easiest thing in my opinion is that the hash used for caching should also include the positions, instead of ignoring them.

vakaras:
> The easiest thing in my opinion is that the hash used for caching should also include the positions, instead of ignoring them.

For `vir_low`, you could derive `PositionedHash` to avoid changing the current implementation of the `Hash` trait.

Pointerbender:
I'm not sure how feasible this would be, but it might be preferably to hash a relative position (and translate the relative position back to an absolute position for any warnings/errors emitted), otherwise a change in line number will invalidate the cache for a lot of methods in larger files, which somewhat greatly mitigates the caching benefits. The position could be make relative to the line where the method is first declared, so that moving an identical method around in a file will not invalidate its cached verification result.

fpoli:
The `line` and `column` fields of `Position` do not affect Prusti's error reporting. They are just there to help debugging some internal errors. (I personally never benefit from them and I would be fine with having them removed.) The important field is `id`. Your suggestion still applies to it. Viper programs are equivalent if they differ only for a renaming of the position ids. So, we can normalize the position ids of programs in a pre-processing step in the Prusti server, verify and cache them, then translate the positions used in the verification errors back to the de-normalized version.

[...]

PR #971, opened 2022-04-22
Fix error reporting when caching

JonasAlaif:
Prior to encoding each function, the pos_id will be set a random value seeded from the function name. Thus, a returned cached error positions should match up with the currently encoded one, even if the order in which functions are encoded changes between when the cache was created and now (e.g. by adding a new function). All existing cached errors will be invalidated (not a big deal, due to this issue they were probably invalid anyway) and cache files should be deleted.
Linked to #940, though not a full fix since we don't guarantee a deterministic encoding into VIR nor injectivity of MIR -> VIR (without positions).


fpoli (commented):
* (+) Nice small trick :)
* (-) It doesn't fully solve #940, for the reason that I explained [here](https://github.com/viperproject/prusti-dev/issues/940#issuecomment-1094960799).
* (-) With these changes the caching starts to rely on the internals of how position ids are generated to avoid collisions, and the error manager's implementation mentions "caching" for the first time.

I'm mostly worried about the last point. So far refactoring the error manager has been very easy because there were not many dependencies to/from other parts of the code. I'd still prefer to fix the caching on the server side, making it hash programs without ignoring the position ids.

@vakaras You mentioned that we could automatically derive `PositionedHash`. How hard would that be? Like, 10, 100 or 1000 LOC?

vakaras:
> @vakaras You mentioned that we could you could automatically derive `PositionedHash`. How hard would that be? Like, 10, 100 or 1000 LOC?

The positionless hash is defined [here](https://github.com/viperproject/prusti-dev/blob/3cd79e0fa0fa22c34f59bccbb71e0b9cb5cfa49b/vir-gen/src/deriver/singles.rs#L167-L189)

fpoli:
> > @vakaras You mentioned that we could you could automatically derive `PositionedHash`. How hard would that be? Like, 10, 100 or 1000 LOC?
> 
> The positionless hash is defined [here](https://github.com/viperproject/prusti-dev/blob/3cd79e0fa0fa22c34f59bccbb71e0b9cb5cfa49b/vir-gen/src/deriver/singles.rs#L167-L189)

How can one use it?

JonasAlaif:
@fpoli I think by removing the ignore thingy that I removed in the latest push. Btw do these two changes alleviate your concerns?

fpoli:
> @fpoli I think by removing the ignore thingy that I removed in the latest push. Btw do these two changes alleviate your concerns?

Removing the `ignore` yes. That change alone should be enough to fix https://github.com/viperproject/prusti-dev/issues/940. (I'm curious to see what will break, because the `ignore` was probably there for a reason.)

Removing lines and columns from `Position` it's something that I already wanted to do, nice!

The `reset_pos_id` machinery is still IMO a normalization that should be done on the server side: find the minimum position id and subtract it everywhere in the verification request before computing the hash, then add it to the verification result before sending it back. I'm fine with merging the current solution, but it could be more maintainable in the server.

JonasAlaif:
> The `reset_pos_id` machinery is still IMO a normalization that should be done on the server side: find the minimum position id and subtract it everywhere in the verification request before computing the hash, then add it to the verification result before sending it back. I'm fine with merging the current solution, but it could be more maintainable in the server.

Imo ideally we would create a new `PositionManager` for each function (Viper file) that we verify; why should the `next_pos_id` not start from 1 for each Viper file? The `id` of `Position` doesn't need to be unique between Viper files since we verify and backtranslate each one separately anyway. This would have required more reengineering of `ErrorManager` though, so I've just opted for the simpler choice where I essentially create a new `PositionManager` each time (by deterministically seeding `next_pos_id`), without having to practically do so.

JonasAlaif:
Can't seem to replicate the CI failure locally, just running Prusti on the failing file

fpoli:
> why should the next_pos_id not start from 1 for each Viper file?

In my view, it's a property that the error manager does not need to ensure. The encoder should just know that position ids can be compared for equality, but without knowing that there is a first id nor consecutive or ordered ids. Knowing that the only important thing for ids is to be equal or not, the server can then do some normalization to increase cache hits.

fpoli:
To be a bit more concrete, I was thinking at the following normalization on the server side:
```rust
let pos_id_offset: usize = program.iter_positions().min().unwrap();
program.walk_positions(|p| *p = *p - pos_offset); // Normalize
let result = // Check the cache, verify the program, update the cache
result.walk_positions(|p| *p = *p + pos_offset); // De-normalize
return result
```

Or, alternatively, the following more powerful normalization that remaps positions to consecutive ids from zero:
```rust
let mut original_positions: Vec<usize> = program.iter_positions().collect<FxHashSet<_>>().into_iter().collect();
original_positions.sort();
let normalization: FxHashMap<usize, usize> = original_positions.into_iter().enumerate().map(|(i, p)| (p, i)).collect();
program.walk_positions(|p| *p = normalization[p]); // Normalize
let result = // Check the cache, verify the program, update the cache
result.walk_positions(|p| *p = original_positions[p]); // De-normalize
return result
```

I can give a try at this next week.

JonasAlaif:
I've clearly missed something here, will have to fix it next week

JonasAlaif:
The current `(DefId, Position)`-as-keys approach doesn't work since a failing assertion (e.g. function call) may reference Positions with multiple DefId contexts (e.g. the current fn within which the call happened and the failing precondition of the called fn), but at the error reporting point we only have the DefId of the current fn (and cannot distinguish Positions from this one vs others).

PR #972, opened 2022-04-22
Fix slice with range encoding

### LINKED ISSUE #960: Prusti unexpected internal error
Verifying the following Rust code with Prusti produces an unexpected internal error.

The code is:

```rust
use prusti_contracts::*;

#[requires(array.len() >= 1)]
fn foo(array : &mut [i32]) {
    let mut end = array.len();
    while end > 0 {
        body_invariant!(end >= 1 && end <= array.len());
        let mut start = end - 1;
        while start > 0 {
            body_invariant!(start >= 1);
            start -= 1;
            bar(&mut array[start..end]);
        }
        end -= 1;
    }
}

fn bar(v : &mut [i32]) {}

fn main() {}
```

The error is:
```
- [Prusti internal error] Prusti encountered an unexpected internal error
- This could be caused by too small assertion timeout. Try increasing it by setting the configuration parameter ASSERT_TIMEOUT to a larger value.
- We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
Details: unregistered verification error: [application.precondition:assertion.false; 0] Precondition of function lookup_pure__$TY$__Slice$i32$i32$ might not hold. Assertion j < Slice$len__$TY$__i32$(_LET_1) might not hold. (@0.0)
```

The code should fail the verification anyways, as there is a missing body invariant, and the verification succeeds after adding it.

The correct code - after adding the missing invariant in the inner while loop - is:

```rust
use prusti_contracts::*;

#[requires(array.len() >= 1)]
fn foo(array : &mut [i32]) {
    let mut end = array.len();
    while end > 0 {
        body_invariant!(end >= 1 && end <= array.len());
        let mut start = end - 1;
        while start > 0 {
            body_invariant!(start >= 1 && end <= array.len()); // added body invariant
            start -= 1;
            bar(&mut array[start..end]);
        }
        end -= 1;
    }
}

fn bar(v : &mut [i32]) {}

fn main() {}
```

JonasAlaif:
Fixes #960
Issue is described there. Not sure if this is the correct way to do the error reporting?
It would be really nice to just have a contract for `index` and `index_mut` and report the error as a precondition violation of those. E.g. I really don't want to hardcode the pledge required for `index_mut` by hand.


prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -69,6 +69,9 @@ pub enum ErrorCtxt {
     AssertTerminator(String),
     /// A Viper `assert false` in the context of a bounds check
     BoundsCheckAssert,
+    /// A Viper `assert false` in the context of a hardcoded bounds check (e.g. when we hardcode a `index`)
+    /// TODO: remove this in favor of extern_spec for e.g. the stdlib `fn index(...)`
+    HardcodedBoundsCheckAssert(String),
```
fpoli: What about calling this variant `SliceRangeBoundsCheckAssert`? It took me a minute to understand the meaning of hardcoded in this context.

fpoli (approved):
The error reporting looks good to me. I left a comment.

PR #977, opened 2022-04-26
[Merged by Bors] - Adjust copy check in Environment

vl0w:
`Environment::type_is_copy` currently returns false for e.g. `Option<'&a T>`. This is an attempt to fix this.


vl0w:
The following still causes an ICE (also before this PR):

```rust
#[pure]
#[trusted]
fn foo<'a, T: Copy>(val: &'a T) -> Option<&'a T> {
    unimplemented!()
}
```
Reason (from my very limited understanding): The region `'a` in `Option<&'a T> is escaping and `type_is_copy_modulo_regions` does not like that. I'm not quite sure how to fix this.

Maybe it is out of scope, but if this is an easy fix, any suggestions are appreciated.

vl0w:
Associated types are now supported as well!
```rust
struct Baz<T: Copy>(std::marker::PhantomData<T>);
impl<T: Copy> Foo for Baz<T> {
    type Type = T;

    #[pure]
    #[trusted]
    fn foo(&self) -> Self::Type {
        unimplemented!()
    }
}
```
`Baz::Type` is `Copy` because `T` is `Copy`

vl0w:
> The following still causes an ICE (also before this PR):
> 
> ```rust
> #[pure]
> #[trusted]
> fn foo<'a, T: Copy>(val: &'a T) -> Option<&'a T> {
>     unimplemented!()
> }
> ```
> 
> Reason (from my very limited understanding): The region `'a` in `Option<&'a T> is escaping and `type_is_copy_modulo_regions` does not like that. I'm not quite sure how to fix this.
> 
> Maybe it is out of scope, but if this is an easy fix, any suggestions are appreciated.

This case is now also supported with the latest commit.

prusti-interface/src/environment/mod.rs
```diff
@@ -492,8 +492,8 @@ impl<'tcx> Environment<'tcx> {
             .unwrap_or((called_def_id, call_substs))
     }
 
-    pub fn type_is_allowed_in_pure_functions(&self, ty: ty::Ty<'tcx>, param_env: ty::ParamEnv<'tcx>) -> bool {
-        match ty.kind() {
+    pub fn type_is_allowed_in_pure_functions(&self, ty: ty::Binder<'tcx, ty::Ty<'tcx>>, param_env: ty::ParamEnv<'tcx>) -> bool {
+        match ty.skip_binder().kind() {
             ty::TyKind::Never => {
                 true
             }
```
Aurel300: Let's remove this special case.

prusti-interface/src/environment/mod.rs
```diff
@@ -560,24 +553,23 @@ impl<'tcx> Environment<'tcx> {
...
+    pub fn resolve_assoc_types<T: ty::TypeFoldable<'tcx> + std::fmt::Debug + Copy>(&self, normalizable: T, param_env: ty::ParamEnv<'tcx>) -> T {
+        let norm_res = self.tcx.try_normalize_erasing_regions(
+            param_env,
+            normalizable
+        );
+
+        match norm_res {
+            Ok(normalized) => {
+                debug!("Normalized {:?}: {:?}", normalizable, normalized);
+                normalized},
```
```suggestion
                normalized
            },
```

prusti-interface/src/environment/mod.rs
```diff
@@ -493,14 +493,7 @@ impl<'tcx> Environment<'tcx> {
     }
 
     pub fn type_is_allowed_in_pure_functions(&self, ty: ty::Binder<'tcx, ty::Ty<'tcx>>, param_env: ty::ParamEnv<'tcx>) -> bool {
```
Aurel300: Actually let's just use `type_is_copy` directly then :) (and delete this method)

vl0w: I kind of liked the additional method, but I'm ok with removing it.

PR #980, opened 2022-04-27
WIP: Iterator tracking issues

vl0w:
Some (smaller / bigger) changes needed to enable iterator support
(WIP / tracking PR)

- Moved normalization of constraint param env (for assoc types) for better logging
- Fixed a bug in `SpecGraph`: Preconditions of the base spec were copied to the constrained spec
- Ghost constraint resolving is a bit relaxed (`predicate_may_hold` instead of `predicate_must_hold_considering_regions`) to account for references in associated types. See test `associated_types_6.rs`
- `merge_generics` now also merges lifetimes to accout for lifetimes appearing in `#[refine_trait_spec]`
- Lifetimes in type models are not converted to the anonymous lifetime `'_` anymore
- Do not `#[derive(Copy, Clone]` on type models, because the derive macro "conditionally" implements it when generics are involved. That is, `#[derive(Copy)] struct Foo<T>` adds `impl<T: Copy> Copy for Foo {}`, but we do not want to implement `Copy` only if `T` is `Copy`.
- Quantified variables now support associated types: `forall(|x: Self::SomeAssocType| ...)`


prusti-common/src/config.rs
```diff
@@ -599,4 +600,9 @@ pub fn intern_names() -> bool {
 /// Enables ghost constraint parsing and resolving.
 pub fn enable_ghost_constraints() -> bool {
     read_setting("enable_ghost_constraints")
+}
+
+/// Enables (experimental) iterator support
```
Aurel300: Copy over the (adjusted) explanation from `flags.md`.

docs/dev-guide/src/config/flags.md
```diff
@@ -141,6 +142,12 @@ on a generic type parameter) is satisfied.
 
 **This is an experimental feature**, because it is currently possible to introduce unsound verification behavior.
 
+## `ENABLE_ITERATORS`
+
+Enables support for iterators.
```
Aurel300: Please word this consistently with the other flags, i.e. `When enabled, ...`

Aurel300:
@vakaras Could you have a look at the two failures here? I think they are the last ones before I can merge this and they are happening only in the unsafe core proof version.

PR #982, opened 2022-04-28
Sequences and maps

vakaras:


prusti-contracts/src/lib.rs
```diff
@@ -1,4 +1,7 @@
 #![no_std]
+#![feature(register_tool)]
```
vakaras, line 2: I think these feature annotations are not needed anymore.

prusti-tests/tests/cargo_verify/test_map/.gitignore
```diff
@@ -0,0 +1 @@
+/prusti-*
```
vakaras, line 1: I think it would be better to remove all `cargo_verify` tests and instead have tests in `prusti-tests/tests/verify_overflow/fail/core_proof/mathematical_types/`.

prusti-viper/src/encoder/middle/core_proof/into_low/expression.rs
```diff
@@ -0,0 +1,138 @@
+use crate::encoder::{
```
vakaras, line 1: This file should be deleted.

prusti-viper/src/encoder/middle/core_proof/predicates_owned/interface.rs
```diff
@@ -0,0 +1,338 @@
+use crate::encoder::{
```
vakaras, line 1: This file should be deleted.

PR #986, opened 2022-04-29
Fix #984

### LINKED ISSUE #984: ICE + postcondition might not hold (definition collector issue)
The following Rust program:

```rust
use prusti_contracts::*;

fn main() {}

#[derive(Clone, Copy)]
pub struct A {
    inner: usize,
}

predicate! {
    pub fn a_invariant(a: &A) -> bool {
        a.inner > 10
    }
}

impl A {
    #[pure]
    #[ensures(a_invariant(self) ==> result > 10)]
    pub const fn foo(&self) -> usize {
        self.inner
    }
}

#[derive(Clone, Copy)]
pub struct B {
    a: A,
}

predicate! {
    pub fn b_invariant(b: &B) -> bool {
        a_invariant(&b.a)
    }
}

impl B {
    #[pure]
    #[requires(b_invariant(&self))]
    #[ensures(result > 10)]
    pub const fn inner(&self) -> usize {
        self.a.foo()
    }
    #[pure]
    #[requires(b_invariant(self))]
    pub const fn bar(&self) {
        assert!(self.inner() > 10);
    }
}
```

fails with an internal compiler error:

```
thread 'rustc' panicked at 'assertion failed: env.has_errors() || config::internal_errors_as_warnings()', prusti/src/verifier.rs:73:17

// (...)

note: Prusti or the compiler unexpectedly panicked. This is a bug.

note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new

note: Prusti version: commit 81366ed155 2022-04-26 16:24:52 UTC, built on 2022-04-27 14:59:24 UTC

query stack during panic:
end of query stack
```

The error comes from the Viper program for `B::bar`, at the Viper level the error is:

```
Postcondition of m_foo__$TY$__ might not hold. Assertion !m_a_invariant__$TY$__(_1) || result > 10 might not hold.
```

This is because the definition collector too eagerly turned `m_a_invariant__$TY$__` into an abstract Viper function:

```
function m_a_invariant__$TY$__(_1: Snap$struct$m_A): Bool
  requires true
  requires true
  ensures true
// (no body)
```

but then another pure Viper function for `A::foo` relies on this missing function body:

```
function m_foo__$TY$__(_1: Snap$struct$m_A): Int
  requires true
  requires true
  ensures 0 <= result
  ensures result <= 18446744073709551615
  ensures !m_a_invariant__$TY$__(_1) || result > 10
  ensures [result == mirror_simple$m_foo__$TY$__(_1), true]
{
  Snap$struct$m_A$0$field$f$inner__$TY$__(_1)
}
```

The solution is to update the definition collector to retain the function body of the `a_invariant` Viper predicate in these corner cases. I'll see if I can manage to fix this one and create a PR against this Github issue to fix it.

Aurel300:
Add mirror functions to predicates.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-984.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-984.rs
@@ -1,0 +1,47 @@
+use prusti_contracts::*;
+
+fn main() {}
+
+#[derive(Clone, Copy)]
+pub struct A {
+    inner: usize,
+}
+
+predicate! {
+    pub fn a_invariant(a: &A) -> bool {
+        a.inner > 10
+    }
+}
+
+impl A {
+    #[pure]
+    #[ensures(a_invariant(self) ==> result > 10)]
+    pub const fn foo(&self) -> usize {
+        self.inner
+    }
+}
+
+#[derive(Clone, Copy)]
+pub struct B {
+    a: A,
+}
+
+predicate! {
+    pub fn b_invariant(b: &B) -> bool {
+        a_invariant(&b.a)
+    }
+}
+
+impl B {
+    #[pure]
+    #[requires(b_invariant(&self))]
+    #[ensures(result > 10)]
+    pub const fn inner(&self) -> usize {
+        self.a.foo()
+    }
+    #[pure]
+    #[requires(b_invariant(self))]
+    pub const fn bar(&self) {
+        assert!(self.inner() > 10);
+    }
+}
--- a/prusti-viper/src/encoder/definition_collector.rs
+++ b/prusti-viper/src/encoder/definition_collector.rs
@@ -280,6 +280,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> Collector<'p, 'v, 'tcx> {
                 self.used_mirror_functions
                     .contains(&function.get_identifier().into())
             });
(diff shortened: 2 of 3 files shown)


vakaras (approved):
LGTM
bors r+

Aurel300:
@JonasAlaif This fails on the caching test. From the output it looks like it is the order of functions in the mirror domain that trips it up: is this assumed to be deterministic? Should they be sorted before output?

JonasAlaif:
> @JonasAlaif This fails on the caching test. From the output it looks like it is the order of functions in the mirror domain that trips it up: is this assumed to be deterministic? Should they be sorted before output?

Alternatively we could leave the mirror domain out of the hash, but with the way things are going (i.e. we might start including the positions in the hash) it might be safer the include everything in the hash, but just make it deterministic

fpoli:
> Alternatively we could leave the mirror domain out of the hash, but with the way things are going (i.e. we might start including the positions in the hash) it might be safer the include everything in the hash, but just make it deterministic

I agree. A deterministic encoding should be easier to maintain and debug than special rules in the hash function.

Aurel300:
bors retry

@fpoli There was a timeout (900s) in one of the test crates; maybe we should record also an expected time for crates, make it longer for the large ones?

PR #988, opened 2022-04-30
[Merged by Bors] - Avoid some more warnings about extra braces/parens

dario23:
We insert them in generated code, to be sure not to miss any precedence,
but since it's generated code disable warnings there.


Aurel300:
Thanks!

bors r+

PR #991, opened 2022-05-02
Backlog fixes

viper-admin:
In `test-crates/crates.csv`, replace one `Skip` with `NoCrash` or one `NoCrash` with `NoErrors`. Optionally, also update a crate version. This should be enough to make the `test-crates` test fail because of panics or internal errors. To test a single crate you can use `./target/release/test-crates <FILTER>`, which checks only the crates whose name contains `<FILTER>`.

The purpose of this PR is to fix at least one bug (e.g. convert a panic to an internal error or an internal error to a readable "unsupported" error message).

@Aurel300 could you take care of this?


Aurel300:
Some of the fixes brought `itertools` closer to being verifiable, but some panics still remain (after `could not compile `itertools` due to 402 previous errors; 11 warnings emitted` (: ).

JonasAlaif:
@Aurel300 will you merge this at some point?

PR #998, opened 2022-05-10
[Merged by Bors] - Fix a bug in the purification optimization

vakaras:
Fixes a soundness bug in the purification optimization where the variable of an exhaled predicate was not havocked.


vakaras:
bors retry

Issue #1006, opened 2022-05-17
Consistency error in postcondition when calling pure function that returns a slice


Pointerbender:
The following Rust program:

```rust
use prusti_contracts::*;

#[derive(Copy, Clone)]
pub struct A {
    inner: usize,
}

impl A {
    #[pure]
    pub const fn get(&self) -> usize {
        self.inner
    }
}

pub struct B {
    a: A,
}

impl B {
    #[pure]
    #[trusted]
    pub fn foo(&self, a: usize, b: usize) -> &[u8] {
        unimplemented!()
    }

    #[pure]
    #[ensures(result == self.foo(a.get(), b))]
    pub fn bar(&self, a: A, b: usize) -> &[u8] {
        self.foo(a.get(), b)
    }

    #[pure]
    pub fn baz(&self) -> &[u8] {
        self.bar(self.a, 1)
    }
}
```

Generates the following for error message for the Viper program of `B::baz`:

```
error: [Prusti internal error] Prusti encountered an unexpected internal error
  |
  = note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
  = note: Details: consistency error in prusti_example::B::baz: Consistency error: Local variable _4 not found. (@96.15)
```

This is due to the pure function encoding of `B::bar`:

```
function m_B$$bar__$TY$__(_1: Snap$struct$m_B, _2: Snap$struct$m_A, _3: Int): Snap$Slice$u8
  requires 0 <= _3
  requires _3 <= 18446744073709551615
  requires true
  requires true
  ensures snap$__$TY$__Snap$Slice$u8$(_4.val_ref) == m_B$$foo__$TY$__(_1, m_A$$get__$TY$__(_2), _3)
  ensures [result == mirror_simple$m_B$$bar__$TY$__(_1, _2, _3), true]
{
  m_B$$foo__$TY$__(_1, m_A$$get__$TY$__(_2), _3)
}
```

Here we can see that in the post-condition it mentions `snap$__$TY$__Snap$Slice$u8$(_4.val_ref)`, while it should mention `result` instead. The MIR for `B::bar` is:

![mir3](https://user-images.githubusercontent.com/81013316/168785065-9f9da48c-1b32-4696-a0a1-fe2afa47a79b.svg)

Pointerbender:
It seems that the expression replacement in `fn inline_spec_item` fails here:

[...]

PR #1008, opened 2022-05-18
Fix for #1006 and other minor changes

Pointerbender:
This PR contains a few changes:

1. A fix in `prusti-viper/src/encoder/mir/pure/specifications/encoder_poly.rs` for the bug from #1006 plus a regression test.
2. A new configuration flag `use_new_encoder` which toggles the testing of the new encoder in `prusti-viper/src/encoder/mir/pure/pure_functions/interface.rs`. This flag is enabled by default and is currently selectively disabled for two tests: `issue-1006.rs` and `issue-729-1.rs`.
3. Some additional info logging in `prusti-server/src/process_verification.rs`. This gives some extra feedback during the verification process when the log level is `INFO`. This is handy for large code bases. In my case I have a project with over 300 functions to verify which can take a long time if large parts of the cache get invalidated. This way, it's possible to see the verification results come in in real-time, without having to wait until the entire verification finishes.

### DIFF
--- a/prusti-common/src/config.rs
+++ b/prusti-common/src/config.rs
@@ -105,6 +105,7 @@ lazy_static! {
         settings.set_default("unsafe_core_proof", false).unwrap();
         settings.set_default("only_memory_safety", false).unwrap();
         settings.set_default("enable_type_invariants", false).unwrap();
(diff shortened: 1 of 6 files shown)


vakaras (approved):
LGTM Thanks!

PR #1010, opened 2022-05-18
prusti_assert!

JM4ier:
a macro to allow full prusti syntax in assertions (WIP)

### DIFF
--- a/prusti-interface/src/specs/mod.rs
+++ b/prusti-interface/src/specs/mod.rs
@@ -1,28 +1,32 @@
 use rustc_ast::ast;
-use rustc_hir::{intravisit};
-use rustc_middle::hir::map::Map;
-use rustc_middle::ty::TyCtxt;
-use rustc_span::Span;
 use rustc_errors::MultiSpan;
+use rustc_hir::intravisit;
+use rustc_middle::{hir::map::Map, ty::TyCtxt};
+use rustc_span::Span;
 
-use rustc_hir::def_id::{DefId, LocalDefId};
-use std::collections::HashMap;
-use std::convert::TryInto;
-use std::fmt::Debug;
-use crate::environment::Environment;
-use crate::PrustiError;
-use crate::utils::{has_extern_spec_attr, read_prusti_attr, read_prusti_attrs, has_prusti_attr, has_abstract_predicate_attr};
+use crate::{
+    environment::Environment,
+    utils::{
+        has_abstract_predicate_attr, has_extern_spec_attr, has_prusti_attr, read_prusti_attr,
+        read_prusti_attrs,
+    },
+    PrustiError,
+};
 use log::debug;
+use rustc_hir::def_id::{DefId, LocalDefId};
+use std::{collections::HashMap, convert::TryInto, fmt::Debug};
 
+pub mod checker;
 pub mod external;
 pub mod typed;
-pub mod checker;
 
 use typed::SpecIdRef;
 
-use crate::specs::external::ExternSpecResolver;
+use crate::specs::{
+    external::ExternSpecResolver,
+    typed::{ProcedureSpecification, ProcedureSpecificationKind, SpecGraph, SpecificationItem},
+};
 use prusti_specs::specifications::common::SpecificationId;
-use crate::specs::typed::{ProcedureSpecification, SpecGraph, ProcedureSpecificationKind, SpecificationItem};
 
 #[derive(Debug)]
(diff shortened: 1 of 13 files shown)


prusti-interface/src/specs/mod.rs
```diff
@@ -47,6 +47,7 @@ pub struct SpecCollector<'a, 'tcx: 'a> {
     /// Map from functions/loops and their specifications.
     procedure_specs: HashMap<LocalDefId, ProcedureSpecRefs>,
     loop_specs: Vec<LocalDefId>,
+    prusti_assertion_specs: Vec<LocalDefId>,
```
Aurel300: Why `_specs`?

prusti-interface/src/specs/typed.rs
```diff
@@ -83,6 +88,11 @@ pub struct LoopSpecification {
     pub invariant: LocalDefId,
 }
 
+#[derive(Debug, Clone)]
+pub struct PrustiAssertionSpecification {
```
Aurel300: Here also: why `Specification`?

prusti-tests/tests/verify_overflow/fail/prusti_assert.rs
```diff
@@ -0,0 +1,21 @@
...
+    body_invariant!(false);
+}
+
+fn loop_shouldnt_crash() {
+    loop {
+        prusti_assert!(true);
+        body_invariant!(true);
+    }
+}
+
```
Aurel300, line 51: Add tests with quantifiers and implications.

prusti-viper/src/encoder/mir/specifications/specs.rs
```diff
@@ -74,6 +75,11 @@ impl<'tcx> Specifications<'tcx> {
         self.user_typed_specs.get_loop_spec(def_id)
     }
 
+    pub(super) fn get_assert_spec(&self, def_id: &DefId) -> Option<&PrustiAssertionSpecification> {
+        trace!("Get loop specs of {:?}", def_id);
```
Aurel300: Change comment.

Aurel300:
Looks good so far. For the desugaring to actually do something, you should probably pass the closure created in the rewriter into some special method added to `prusti-contracts` (similar to e.g. `forall`), then intercept the call to it in the procedure encoder and the pure function encoder.

Please also add `prusti_assume!`, which should be basically the same thing.

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -146,6 +146,8 @@ pub enum ErrorCtxt {
     CloseFracRef,
     /// Failed to set an active variant of an union.
     SetEnumVariant,
+    /// Failed Assertion
+    Assertion,
```
Aurel300: Would this not be better as another case in `PanicCause`?

prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
```diff
@@ -1781,13 +1781,46 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+        &mut self,
+        bb: mir::BasicBlock,
+        block: &mir::BasicBlockData<'tcx>,
+        encoded_statements: &mut Vec<vir_high::Statement>,
+    ) -> SpannedEncodingResult<bool> {
+        for stmt in &block.statements {
+            if let mir::StatementKind::Assign(box (
+                _,
+                mir::Rvalue::Aggregate(box mir::AggregateKind::Closure(cl_def_id, cl_substs), _),
+            )) = stmt.kind
```
Aurel300, line 1841: Can you tell these apart from loop invariants? Maybe if the `get_prusti_assertion_specs` call returned `None`, this should return `Ok(false)`?

prusti-tests/tests/verify_overflow/fail/prusti_assert.rs
vakaras: Since lowering of quantifiers is not yet supported, let's land it as it is.

Aurel300:
@vakaras This PR should also handle the old encoding at some point, right?

vakaras:
> @vakaras This PR should also handle the old encoding at some point, right?

No. I do not think Jonas has enough time for that.

vakaras (approved):
LGTM. Thanks!

bors r+

PR #1014, opened 2022-05-21
Bugfix before expiry

JonasAlaif:
See the test case for instance where it the bug would manifest. With current master one would get:
`Details: cannot generate fold-unfold Viper statements. The required permission Pred(old[l2](_0.val_ref).val_ref, read) cannot be obtained.`
This is because it essentially tries to deref twice.


JonasAlaif:
bors retry

Issue #1015, opened 2022-05-21
Implementation of Struct Invariants + Pure is unsound


JonasAlaif:
The invariant is assumed to hold when checking the bodies of pure functions (as a Viper method), but is not enforced when calling these pure functions in other code (as a Viper function). One can therefore prove false:
```rust
use prusti_contracts::*;

#[invariant(false)]
struct S;

#[pure]
#[ensures(false)]
fn foo(s: &S) {}

fn main() {
    let s = S;
    foo(&s);
    assert!(false);
}
```

PR #1021, opened 2022-05-23
Do not encode invariants in contracts of pure functions

Aurel300:
(Mostly temporary for NFM)

Since pure functions are used to define the type invariant, assuming it holds in the precondition of a pure function leads to a circular argument. For now, invariants (or at least parts) have to be repeated explicitly for pure function contracts.


Pointerbender:
Apologies to resurrect an old and closed PR :)

@Aurel300 the PR description says that disabling the type invariants for pure functions is a temporary measure to plug #1015, are you able to share anything about when and in what form might we expect this feature to return for pure functions? :) Thanks!

JonasAlaif:
@Pointerbender I guess @Aurel300 missed this. We haven't discussed this in a while, but I would imagine that we tackle this relatively early when implementing type invariants in the "refreshed" version of Prusti.

PR #1025, opened 2022-05-27
Enable autocompletion of rustc APIs in VS Code.

vakaras:
Thanks to Xavier for pointing out this feature!


JonasAlaif:
Nice, I've had to change this setting manually till now

PR #1028, opened 2022-05-30
Fix for another consistency error (#1006) and tests

Pointerbender:
I hit another consistency error with the same root cause as #1006. In a previous PR #1008 this fixed the corner case for the return arguments, this PR also fixes it for all the arguments ([similar fix](https://github.com/viperproject/prusti-dev/issues/1006#issuecomment-1128839770) as suggested by @Aurel300 but then applied to `fn inline_spec_item` instead of `fn inline_closure`).

Then one more fix in `prusti-viper/src/encoder/mir/pure/pure_functions/interpreter.rs` for a problem I hit with the test case. I'm not entirely sure what I'm doing there is correct, but I copied this approach from the procedure encoder here, which does the same thing:

https://github.com/viperproject/prusti-dev/blob/11c74fcefb30b3bed3570bfdbe8ec4e753a4f8f9/prusti-viper/src/encoder/procedure_encoder.rs#L1293-L1299

The PR also includes another test case that cover both these corner cases.


Aurel300:
Makes sense, thanks.

bors r+

PR #1034, opened 2022-06-02
Add devcontainer settings.

vakaras:
Trying to add support for GitHub's codespaces. To use: `https://github.com/viperproject/prusti-dev` → “Code” → “Codespaces” → “Create codespace on master”.


fpoli:
Cool! The Prusti Assistant extension works fine in the codespace.

I was trying something similar in https://github.com/fpoli/viper-ide-codespace, but for demos, not Prusti development.

PR #1036, opened 2022-06-03
Add an error message if Java is not found

zgrannan:
This is a relatively small thing but if Java is not found the x.py script will fail with an unhelpful error message. This makes the error message more helpful.

### DIFF
--- a/x.py
+++ b/x.py
@@ -104,6 +104,10 @@ def get_linux_env():
         ('JAVA_HOME', java_home),
         ('RUST_TEST_THREADS', '1'),
     ]
+    if java_home is None:
+        error("Could not detect a Java installation.\n" +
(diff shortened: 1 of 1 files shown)


x.py
```diff
@@ -104,6 +104,10 @@ def get_linux_env():
         ('JAVA_HOME', java_home),
         ('RUST_TEST_THREADS', '1'),
     ]
+    if java_home is None:
```
vakaras: I think we should check also for other operating systems.

zgrannan: Done, I haven't been able to test on them but I think it should now raise an error for Windows and Mac as well

vakaras (commented):
Thanks!

bors r+

PR #1038, opened 2022-06-03
Add hint of diff required

JonasAlaif:
bors r+

### DIFF
--- a/prusti-interface/src/specs/external.rs
+++ b/prusti-interface/src/specs/external.rs
@@ -30,7 +30,7 @@ pub enum ExternSpecResolverError {
     InvalidExternSpecForTraitImpl(DefId, Span),
 
     /// Occurs when the extern spec is invalid due to mismatched type params.
-    InvalidGenerics(DefId, Span),
+    InvalidGenerics(isize, DefId, Span),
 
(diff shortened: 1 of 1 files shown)


prusti-interface/src/specs/external.rs
```diff
@@ -214,9 +216,9 @@ impl<'v, 'tcx: 'v> ExternSpecResolver<'v, 'tcx> {
                     ).add_note(err_note, None)
                         .emit(env);
                 }
-                ExternSpecResolverError::InvalidGenerics(def_id, span) => {
+                ExternSpecResolverError::InvalidGenerics(diff, def_id, span) => {
                     let function_name = env.get_item_name(*def_id);
-                    let err_note = format!("Invalid type parameters for method '{}'. The number of type parameters must match the target method.", function_name);
+                    let err_note = format!("Invalid type parameters for method '{}'. The number of type parameters must match the target method; change by {} to get rid of this error.", function_name, diff);
```
Aurel300, line 221: The phrasing "change by X" might be a bit confusing. "Add X additional type parameters" or "Remove X type parameters" would read better. We also need more error reporting here because the type parameter bounds should also match for the specs to make sense…

Issue #1033, opened 2022-06-02
Equality checks fail with generics


Pointerbender:
The following Rust program:

```rust
use prusti_contracts::*;

pub struct Foo<T>(T);

impl<T: Copy + PartialEq + Eq> Foo<T> {
    #[pure]
    #[ensures(result == (left == right))]
    fn eq(left: T, right: T) -> bool {
        left == right
    }
}
```

fails verification in Prusti with the error:

```
error: [Prusti: verification error] postcondition might not hold.
  --> src/lib.rs:29:15
   |
29 |     #[ensures(result == (left == right))]
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the error originates here
  --> src/lib.rs:30:5
   |
30 | /     fn eq(left: T, right: T) -> bool {
31 | |         left == right
32 | |     }
   | |_____^
```

In the corresponding [Viper program](https://github.com/viperproject/prusti-dev/files/8822976/eq.vpr.txt), it looks like the result variable `_0` is first havoc'd, but then never initialized, causing the assertion to fail:

```
  label l1
  // [mir] _0 = <T as std::cmp::PartialEq>::eq(move _3, move _4) -> [return: bb1, unwind: bb2]
  label l2
  assert true
  exhale acc(_3.val_ref, write) && acc(_4.val_ref, write)
  _0 := builtin$havoc_ref()
  inhale acc(bool(_0), write)

  // --- SNIP ---

  // Assert functional specification of postcondition
  assert (unfolding acc(bool(_0), write) in _0.val_bool == ( // <-- This is the failing assertion
    old[pre](snap$__$TY$__Snap$__TYPARAM__$_T$0$__$(_1))
      == old[pre](snap$__$TY$__Snap$__TYPARAM__$_T$0$__$(_2))
  ))
```

Is this a bug or a current limitation of generics in Prusti? :-)

Thanks!

Pointerbender:
p.s. the problem seems to be rooted in the handling of the assignment terminator, here is the corresponding MIR:

![eq](https://user-images.githubusercontent.com/81013316/171613926-bc4d11a9-dd03-43b2-856b-09b86e17ac36.svg)

Pointerbender:
Some extra information about the call path through the Prusti code base:

When it reaches the `TerminatorKind::Call` of block `bb0` it reaches this statement:

[...]

PR #1041, opened 2022-06-07
fix for #1033 with test cases

Pointerbender:
A small fix for #1033 with additional test cases. (See #1033 for a detailed description of the investigation.)


Aurel300:
Looks good for now (i.e. for as long as `Copy` is the same thing as "type usable in pure functions"), thanks!

bors r+

PR #1045, opened 2022-06-09
Implement normalization of position ids in the server

fpoli:
@JonasAlaif This PR implements the second normalization described in https://github.com/viperproject/prusti-dev/pull/971#issuecomment-1146060967. A lot of the new code is boilerplate to visit all positions of a program.

Unfortunately, the `Program` that is seen by the server is an enum of `legacy::Program` and `low::Program`, which makes the code a bit messy. The normalization is implemented only for the `legacy::Program` case, but the same logic can also be implemented for `low::Program`.

The naming of the various mut/fallible walker/visitor/folder passes is not always consistent. I think we should refactor them in the future to only have `visitor`, `visitor_mut`, `fallible_mut_visitor`, removing the "walker".


vir/src/legacy/ast/expr_transformers.rs
```diff
@@ -63,6 +64,34 @@ impl Expr {
...
+            visitor: G,
+        }
+        impl<G: FnMut(&mut Position)> ExprFolder for PositionsMutVisitor<G> {
+            fn fold_position(&mut self, mut pos: Position) -> Position {
+                (self.visitor)(&mut pos);
+                pos
+            }
+        }
+        let expr = mem::replace(self, true.into());
+        *self = PositionsMutVisitor { visitor }.fold(expr);
```
fpoli, line 93: This would be more efficient if we had a `ExprMutVisitor` trait that visits all the subexpressions via mutable references.

vir/src/legacy/cfg/visitor.rs
```diff
@@ -113,6 +127,26 @@ impl CfgMethod {
...
+    /// Mutably visit each expression used in a statement or successor.
+    /// Note: sub-expressions of expressions will not be visited.
+    pub fn walk_expressions_mut<F: FnMut(&mut Expr)>(&mut self, mut walker: F) {
+        self.walk_statements_mut(|s| {
+            // Replace *s with a dummy statement
+            let stmt = std::mem::replace(s, Stmt::Inhale(true.into()));
+            *s = stmt.map_expr(|mut e| {
+                walker(&mut e);
+                e
+            });
```
fpoli, line 139: This would be more efficient if we had a `visit_expressions_mut` method that executes a `FnMut(&mut Expr)` closure on each expression in a statement.

fpoli:
@JonasAlaif feel free to do whatever you want with this PR :)

JonasAlaif:
I'm going to say that these inefficiencies are all pretty small and also in `legacy` so it's ok

PR #1046, opened 2022-06-10
Sanitize generated labels for `impl` arguments

### LINKED ISSUE #1037: Consistency error when using `impl` typed argument
Example:

```rust
fn foo(args: impl IntoIterator) { drop(args) }
fn main() {}
```

Yields:
```
error: [Prusti internal error] Prusti encountered an unexpected internal error
  |
  = note: We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
  = note: Details: consistency error in bug::foo: Consistency error: __TYPARAM__$_impl IntoIterator$0$__ is not a valid identifier. (<no position>)
```

I think this should be fixed by marking `impl` parameters unsupported?

zgrannan:
The type variable names for `impl` function parameters potentially contain spaces, braces, equality signs. This PR sanitizes them so that the resulting viper identifiers are valid.

I'm not quite sure if the resulting encoding is usable for verification; on the other hand making this change prevents Prusti from crashing.

Fixes #1037

### DIFF
--- a/prusti-tests/tests/verify/pass/issues/issue-1037-1.rs
+++ b/prusti-tests/tests/verify/pass/issues/issue-1037-1.rs
@@ -1,0 +1,2 @@
+fn foo(args: impl IntoIterator) { drop(args) }
+fn main() {}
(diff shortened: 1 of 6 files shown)


Aurel300:
Looks good, thanks. Could you also add some tests that actually use an `impl`-generic method, i.e. there is a call to it? If these don't pass, we can `ignore-test` them for now.

Aurel300:
Thanks!

bors r+

zgrannan:
bors try

vakaras (approved):
bors r+

PR #1051, opened 2022-06-13
(WIP) Prusti PCS

markusdemedeiros:
An analysis for a new PCS formulation. 

Recommended that this be put onto a seperate ``prusti-pcs`` branch, since it's experimental.


Cargo.lock
```diff
@@ -1,3209 +0,0 @@
-# This file is automatically @generated by Cargo.
```
Aurel300: I don't think you should delete this file. It version locks our dependencies. If you update any dependencies or (I assume) change the dependency graph in the workspace, there will be changes to this file which you should simply commit.

markusdemedeiros: Done, thanks!

PR #1054, opened 2022-06-15
Lifetime Encoding for Basic Structs with References

pascal-huber:
Add lifetime parameters to `OwnedNonAliased` predicate for structs with reference fields. Also the method bodies of `into_memory_block`, `assign` and `write_place` for structs with references are removed as a temporary solution.


vakaras (approved):
LGTM

bors r+

PR #1056, opened 2022-06-15
Support Generator in `expand_struct_places`.

zgrannan:
Support for Generator type in `expand_struct_places` functions.

Although generators aren't supported yet, this prevents a panic from occurring. Without this PR, the included test case would panic instead of just raising the "unsupported type" error.

I'm not quite sure how to test this functionality currently, so I am uncertain if the modification to `expand_struct_places` will actually do the correct thing (it's based on the handling of closures).


Aurel300:
Thanks for reducing the duplication, I wasn't happy about the method existing in two places. The implementation is correct as far as I can see but you're right that it's probably not something we can properly test yet. Eventually, we should also encode the generator type fully, i.e. similarly to closures for which we encode the field `closure_0..n`. Still, avoiding the panic is a good start.

Is there anything else to be done for this PR, besides merging?

zgrannan:
I don't think there is anything else to do; it can be merged

PR #1057, opened 2022-06-15
Increase assert_timeout for core proof test

zgrannan:
I notice that CI seems to intermittently fail on this test, it seems like increasing the `assert_timeout` fixes it (locally for me, at least)


zgrannan:
Hmm, after trying further it seems like there must be some other underlying reason that this test is failing intermittently...

PR #1060, opened 2022-06-17
Add a --check-timeout flag

zgrannan:
Add the `--check-timeout` flag that is passed to Silicon. 

I think that a too-low timeout is what is causing intermittent verification errors in [arrays.rs](https://github.com/viperproject/prusti-dev/blob/dc6b69046541afc191ef311660e3ad0e9bc9a1e3/prusti-tests/tests/verify_overflow/fail/core_proof/arrays.rs).

Increasing the value seems to make it work: running silicon directly produces different results on the generated viper file when different arguments for `--check-timeout` are used (low values cause it to fail).


vakaras (approved):
LGTM Thanks!

bors r+

PR #1063, opened 2022-06-21
Use field type from projection in encoding

zgrannan:
In `mir_encoder/mod.rs` the type from a field of a closure was taken from `upvar_tys()`. However, it appears that the field does not necessarily correspond to an upvar. In particular the included test raised an internal error because obtaining the type from `upvar_tys()` returns the type of argument to `go` (i.e. `String`), rather than the type of a value returned by the closure (a variant of `ChainRequest`.

It looks like the MIR already includes the field type in the `Field` projection, so this PR uses that instead.

### DIFF
--- a/prusti-tests/tests/verify/fail/unsupported/closure_ref_in_return.rs
+++ b/prusti-tests/tests/verify/fail/unsupported/closure_ref_in_return.rs
@@ -1,0 +1,27 @@
+pub enum TrackingId {
+    Static(&'static str),
+}
+
+pub struct TrackedMsgs {
+    pub msgs: u32,
+    pub tracking_id: TrackingId
+}
+
+pub enum ChainRequest {
+    SendMessagesAndWaitCheckTx {
+        tracked_msgs: TrackedMsgs
+    },
+
+    AddKey {
+        key_name: String
+    },
+}
+
+
+fn go(key_name: String) -> () {
+    || ChainRequest::AddKey { //~ ERROR access to reference-typed fields is not supported
+        key_name
+    };
+}
+
+fn main() {}
--- a/prusti-viper/src/encoder/mir_encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir_encoder/mod.rs
@@ -144,7 +144,7 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
                             encoded_base
                         };
                         let field = &variant_def.fields[field.index()];
-                        let field_ty = field.ty(tcx, subst);
+                        let field_ty = *proj_field_ty;
                         if utils::is_reference(field_ty) {
                             return Err(EncodingError::unsupported(
                                 "access to reference-typed fields is not supported",
--- a/prusti-viper/src/encoder/mir_encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir_encoder/mod.rs
@@ -160,42 +160,19 @@ pub trait PlaceEncoder<'v, 'tcx: 'v> {
                         (encoded_projection, field_ty, None)
                     }
 
-                    ty::TyKind::Closure(def_id, ref closure_subst) => {
+                    ty::TyKind::Closure(def_id, closure_subst) => {
+                        let closure_subst = closure_subst.as_closure();
                         debug!("def_id={:?} closure_subst {:?}", def_id, closure_subst);
 
-                        let closure_subst = closure_subst.as_closure();
-                        debug!("Closure subst: {:?}", closure_subst);
-
-                        // let tcx = self.encoder().env().tcx();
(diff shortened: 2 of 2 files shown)


Aurel300:
I am not sure about this one. Without the change in this PR:

- What `field.index()` do we see in `PlaceEncoder`?
- What does our predicate generated for the closure look like?
- What does Rust print for the representation of the closure type?

What is strange to me is that in the test you added, `key_name` is indeed an upvar (a variable captured by the closure). I see no reason that a projection should be used to get the returned value of the closure (which, when encoding the closure implementation method, is simply `_0`). Is the test not failing for a good reason? The closure is creating an instance of `ChainRequest`, but we cannot encode all of its variants.

The fact that the second argument of `ProjectionElem::Field` is not documented is not great, but I guess it's not a stretch that it is indeed the type we want. Can we use the same `proj_field_ty` in the ADT case as well?

zgrannan:
> - What `field.index()` do we see in `PlaceEncoder`?

The index of the return value of the closure is `0`. Although it looks like the problem comes due to encoding the projection of the `returned_refs` (see below)

> - What does our predicate generated for the closure look like?

I'm not quite sure how to find the predicate, but I think this has the relevant info:
```
ProcedureContractGeneric { def_id: DefId(0:16 ~ closure_ref_in_return[b72e]::go::{closure#0}), args: [Local(1)], returned_refs: [(NormalPlace((*(((_1.0: TrackedMsgs).1: TrackingId).0: &'static str))), Not)], returned_value: Local(0), borrow_infos: [], specification: ProcedureSpecification { span: None, kind: Inherent(Impure), pres: Empty, posts: Empty, pledges: Empty, trusted: Inherent(false) } }
```

> - What does Rust print for the representation of the closure type?

```
def_id=DefId(0:16 ~ closure_ref_in_return[b72e]::go::{closure#0}) closure_subst [i32, extern "rust-call" fn(()) -> ChainRequest, (std::string::String,)]
```

---

> Is the test not failing for a good reason? 

The test should fail, but currently it crashes.

I think the underlying problem comes from encoding the projection for `returned_refs`; in  `mir_encoder/mod.rs`:

```
2022-06-22T14:15:09Z DEBUG prusti_viper::encoder::procedure_encoder] procedure_contract: ProcedureContractGeneric { def_id: DefId(0:16 ~ closure_ref_in_return[b72e]::go::{closure#0}), args: [Local(1)], returned_refs: [(NormalPlace((*(((_1.0: TrackedMsgs).1: TrackingId).0: &'static str))), Not)], returned_value: Local(0), borrow_infos: [], specification: ProcedureSpecification { span: None, kind: Inherent(Impure), pres: Empty, posts: Empty, pledges: Empty, trusted: Inherent(false) } }
[2022-06-22T14:15:09Z DEBUG prusti_viper::encoder::procedure_encoder] Put permission NormalPlace((*(((_1.0: TrackedMsgs).1: TrackingId).0: &'static str))) (Not) in postcondition
```

The problem is that projection `_1.0`, (of type `TrackedMsgs`) is field 0 on the closure; which according to the current implementation is the first upvar, would be the `String` argument. So the problem then occurs when getting `_1.0.1` (i.e. the 2nd field of `TrackedMsgs`); because the `String` type only has one field. Hence an out-of-bounds error occurs:

```
[2022-06-22T14:15:09Z DEBUG prusti_viper::encoder::mir_encoder] encoded_projection: FieldAccess { base: Expr(Local(Local { variable: _1: Ref(closure$m_go$$$opencur$closure$sharp$0$closecur$), position: Position { line: 0, column: 0, id: 0 } })), field: closure_0: Ref(struct$m_std$$string$$String) }
[2022-06-22T14:15:09Z DEBUG prusti_viper::encoder::mir_encoder] base_ty: std::string::String
[2022-06-22T14:15:09Z DEBUG prusti_viper::encoder::mir_encoder] subst []
thread 'rustc' panicked at 'index out of bounds: the len is 1 but the index is 1', prusti-viper/src/encoder/mir_encoder/mod.rs:146:38
```

> The fact that the second argument of ProjectionElem::Field is not documented is not great, but I guess it's not a stretch that it is indeed the type we want.

Agreed, I'm not certain but the implementation of `Debug` in `rustc` seems to indicate that it ought to be the type...

https://github.com/rust-lang/rust/blob/10f4ce324baf7cfb7ce2b2096662b82b79204944/compiler/rustc_middle/src/mir/mod.rs#L2243

> Can we use the same proj_field_ty in the ADT case as well?

I will give this a try

zgrannan:
> Can we use the same `proj_field_ty` in the ADT case as well?

It appears so, the tests passed on my machine.

zgrannan:
@Aurel300 do you think this is OK to merge?

PR #1064, opened 2022-06-23
Fix Lifetime Encoding for Arrays

pascal-huber:
Fixes the lifetime encoding for cases where the lifetimes change in the first statement of a basic block. This is the case when creating a reference to an array item.


vakaras (approved):
LGTM. Thanks!

bors r+

PR #1066, opened 2022-06-25
Start migration to rustc_smir

vakaras:
Project stable MIR is defining a stable version of MIR. As a first step, all compiler consumers are required to collect their APIs into [a single place](https://github.com/rust-lang/project-stable-mir/tree/smir). This PR starts doing that.


Aurel300:
Most of this PR seems related to SMT interaction. Why, are we interacting with Z3 directly for something?

Re: SMIR, we should not switch to this if SMIR ever drifts away from the actual MIR used in the compiler.

vakaras:
> Most of this PR seems related to SMT interaction. Why, are we interacting with Z3 directly for something?

This PR is on top of https://github.com/viperproject/prusti-dev/pull/1062, which adds quantifier instantiation bounds checking for refactored code.

EDIT: Actually, I just realized that I forgot to push…

> Re: SMIR, we should not switch to this if SMIR ever drifts away from the actual MIR used in the compiler.

SMIR is going to be semantically equivalent, so this will not be a problem.

PR #1067, opened 2022-06-26
Lifetime Encoding for Function Call with Returning References

pascal-huber:
Fix lifetime encoding for function calls with a reference as return value.


vakaras (approved):
LGTM

bors r+

PR #1078, opened 2022-07-04
Lifetime encoding Fixes and Extensions

pascal-huber:
- Add `lifetimes` vector to type variants (from `mir`)
 - Clean and fix `erase_lifetimes`
 - Add support for `Trusted`
 - Fix/Enhance lifetime encoding for various predicates/methods/functions

### DIFF
--- a/vir/defs/high/ast/type_decl.rs
+++ b/vir/defs/high/ast/type_decl.rs
@@ -99,6 +99,7 @@ pub struct Enum {
     pub discriminant_bounds: Vec<DiscriminantRange>,
     pub discriminant_values: Vec<DiscriminantValue>,
     pub variants: Vec<Struct>,
+    pub lifetimes: Vec<LifetimeConst>,
 }
 
 #[display(fmt = "{}", name)]
@@ -108,6 +109,7 @@ pub struct Union {
     pub discriminant_bounds: Vec<DiscriminantRange>,
     pub discriminant_values: Vec<DiscriminantValue>,
     pub variants: Vec<Struct>,
+    pub lifetimes: Vec<LifetimeConst>,
 }
 
 #[display(fmt = "Array({}, {})", length, element_type)]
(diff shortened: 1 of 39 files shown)


vir/defs/high/operations_internal/rvalue.rs
```diff
@@ -1,9 +1,94 @@
...
+        position::Position,
+        rvalue::{visitors::RvalueWalker, *},
+        ty::*,
+        variable::*,
+    },
+    ty::Typed,
 };
 
+impl Repeat {
+    pub fn get_lifetimes(&self) -> Vec<LifetimeConst> {
```
vakaras: Would be better to have a trait `WithLifetimes` here that is implemented for all these types.

vir/defs/high/ast/type_decl.rs
```diff
@@ -153,4 +155,5 @@ pub struct Unsupported {
 #[display(fmt = "{}", name)]
 pub struct Trusted {
     pub name: String,
+    pub fields: Vec<FieldDecl>,
```
vakaras: The trusted types should be opaque and encoded as bodyless predicates.
```suggestion
```

vakaras (changes requested):
Modulo the two comments, looks good to me.

vir/defs/high/ast/type_decl.rs
pascal-huber: I see - didn't realize I can completely remove them. The only downside of this is that I can't use `to_procedure_snapshot` on the lifetimes of trusted types - not quite sure why.

vakaras: What error do you get?

pascal-huber: `fn current_snapshot_variable_version` fails to unwrap the snapshot state of the current variables  - seems as if its not yet initialized. But I just realized I shouldn't use `lifetime_const_into_variable` in `owned/encoder.rs` to be more consistent.

vir/defs/high/operations_internal/rvalue.rs
pascal-huber: done

vakaras (approved):
bors r+

PR #1086, opened 2022-07-13
Support for library contracts

### LINKED ISSUE #2: Prusti attributes do not work across crates
**Describe the error**
Attributes such as `#[pure]` or `#[requires]` are not taken into consideration across multiple crates. 

**To Reproduce**
Unpack the attached `.zip` file and run `cargo-prusti` and observe that the `#[pure]` annotation is assumed to be missing even when it's there.

Alternatively, this is pretty easy to set up manually.

1. Create a cargo workspace or have two (or more) empty crates available. (at least one of them as a lib-crate) 
2. add the necessary declarations (`extern crate prusti_contracts`) and have one crate depend on a lib-crate by modifying the `Cargo.toml`.
3. Add a function to the lib crate with any prusti-attributes (`#[pure]`, `#[requires]` and `#[ensures]` have been tested)
4. Use the function from the lib crate inside the "consumer" crate in a way that demonstrates if the attributes are respected (using `#[pure]` function inside of a pre/post-condition, violating the `#[requires]`-clause or `assert!`-ing on `#[ensures]` properties)
5. Run `cargo-prusti` and observe that those attributes do not have the desired effect (not marked as `#[pure]`, no `#[requires]`-violation found or `assert!` fails)

**Expected behavior**
Attributes should work across crate boundaries just as they work within a single crate.

**Reproduction project or code sample**
[pure-across-crates.zip](https://github.com/viperproject/prusti-dev/files/3381838/pure-across-crates.zip)

```rust
// crate producer

#[pure]
pub fn five() -> usize {
    5
}
```

```rust
// crate consumer
extern crate producer;

use producer::five;

#[requires="five() == 5"]
pub fn _test() {}
```

**Software configuration:**
Using `fpoli/prusti-artefact` docker image from `09 July 2019`

jjurm:
Summary: Automatically import specifications from dependency crates and inline them when respective functions or types are referenced in the current crate.

This eliminates the need for `#[extern_spec]` blocks for declarations in other crates as long as the crates are compiled in the same cargo session.

With the proposed changes, the following verifies:
`prusti-tests/tests/cargo_verify/library_contracts_test/src/main.rs`:

```rust
extern crate library_contracts_lib;

use library_contracts_lib::Opt;

// Demonstrating library contracts: the following specification is imported
// and thus the #[extern_spec] block is not needed

/*
use prusti_contracts::*;
#[extern_spec]
impl<T> Opt<T> {
    #[ensures(matches!(*self, Opt::OSome(_)) == result)]
    fn is_some(&self) -> bool;
}
*/

fn main() {
    let a = Opt::OSome(42);
    let b = Opt::ONone::<i32>;

    assert!(a.is_some() == true);
    assert!(b.is_some() == false);
}
```

Closes #2


jjurm:
TO-DO:
- [x] Load specs only from dependency crates instead of loading all compiled crates
- [x] Ensure that serialized specs are usable across subsequent builds

prusti-interface/src/environment/mod.rs
```diff
@@ -277,13 +278,10 @@ impl<'tcx> Environment<'tcx> {
         Procedure::new(self, proc_def_id)
     }
 
-    /// Get the MIR body of a local procedure, monomorphised with the given
-    /// type substitutions.
-    pub fn local_mir(
+   fn local_mir_raw(
```
Aurel300: Indentation: please run `./x.py fmt-all`.

prusti-tests/tests/cargo_verify/library_contracts_lib/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+extern crate prusti_contracts;
+use prusti_contracts::*;
+
+pub enum Opt<T> {
```
Aurel300, line 3: Does this same test work with an `#[extern_spec]` for the real stdlib `Option` in `lib.rs`, used as a specification in the client crate? This would be the primary use of such a crate of specs.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -787,4 +786,13 @@ impl<'v, 'tcx> Encoder<'v, 'tcx> {
...
     }
+
+    pub fn get_mir(&self, def_id: DefId, substs: SubstsRef<'tcx>) -> Rc<mir::Body<'tcx>> {
+        return if let Some(def_id) = def_id.as_local() {
+            self.env().local_mir(def_id, substs)
+        } else {
+            use ty::subst::Subst;
+            ty::EarlyBinder(self.get_local_mir(def_id)).subst(self.env().tcx(), substs)
+        };
+    }
```
Aurel300: Let's not add methods to `Encoder` directly. This looks like something a method that already exists in `Environment` anyway: how does it differ from the type substitution performed there? Is the point of this method to simply to save space and not use the `.env()` on `Encoder`?

prusti-viper/src/encoder/mir/specifications/constraints.rs
```diff
@@ -220,20 +220,22 @@ pub mod trait_bounds {
...
             .unwrap_or_default();
-        let posts: Vec<LocalDefId> = spec
+        let posts: Vec<DefId> = spec
             .posts
             .expect_empty_or_inherent()
             .cloned()
             .unwrap_or_default();
-        for spec_id in pres.iter().chain(posts.iter()) {
+        for spec_id in pres.iter().chain(posts.iter())
+            // Parameter substitution is not yet implemented for external specification
```
Aurel300: Mark this comment with `TODO:` so it is searchable. What actually prevents this from working? Is it just the `get_local_attributes` call that fails? Or can we also not query `param_env` for non-local `DefId`s? (The span is not so important.)

prusti-interface/src/specs/typed.rs
```diff
@@ -5,24 +5,29 @@ use prusti_specs::specifications::common;
...
+pub struct DefSpecificationMap<'tcx> {
     pub proc_specs: HashMap<DefId, SpecGraph<ProcedureSpecification>>,
     pub loop_specs: HashMap<DefId, LoopSpecification>,
     pub type_specs: HashMap<DefId, TypeSpecification>,
     pub prusti_assertions: HashMap<DefId, PrustiAssertion>,
     pub prusti_assumptions: HashMap<DefId, PrustiAssumption>,
     pub ghost_begin: HashMap<DefId, GhostBegin>,
     pub ghost_end: HashMap<DefId, GhostEnd>,
+
+    pub local_mirs: HashMap<DefId, Rc<mir::Body<'tcx>>>,
```
Aurel300: A bit strange that `local_mirs` exists here in `DefSpecificationMap` (causing it to get the `<'tcx>` lifetime throughout the codebase). Can you explain why it is here and not in a separate structure/argument used only when de/serialising?

prusti-interface/src/specs/typed.rs
```diff
@@ -5,24 +5,29 @@ use prusti_specs::specifications::common;
...
 #[derive(Default, Debug, Clone)]
-pub struct DefSpecificationMap {
+pub struct DefSpecificationMap<'tcx> {
     pub proc_specs: HashMap<DefId, SpecGraph<ProcedureSpecification>>,
     pub loop_specs: HashMap<DefId, LoopSpecification>,
     pub type_specs: HashMap<DefId, TypeSpecification>,
     pub prusti_assertions: HashMap<DefId, PrustiAssertion>,
     pub prusti_assumptions: HashMap<DefId, PrustiAssumption>,
     pub ghost_begin: HashMap<DefId, GhostBegin>,
     pub ghost_end: HashMap<DefId, GhostEnd>,
```
Aurel300: These four fields (and `loop_specs`) should not need to be exported, so I was surprised to see that `PrustiAssertion` etc all got the `TyEncodable` derives too. (I see that such a subset is what you called `Lite`.)

prusti-interface/src/environment/mod.rs
```diff
@@ -305,6 +303,23 @@ impl<'tcx> Environment<'tcx> {
                     borrowck_facts: Rc::new(facts),
                 }
             });
+        return body.clone()
+    }
+
+    /// Get the MIR body of a local procedure, without performing any type substitution
```
Aurel300: You can add to the comment that this should be the same as performing the substitution with the `identity_substs` for that `DefId`.

prusti-interface/src/specs/decoder.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    // From rustc
+    fn def_path_hash_to_def_id(&self, hash: DefPathHash) -> DefId {
+        self.tcx.def_path_hash_to_def_id(hash, &mut || panic!("DefPathHash not found in the local crate"))
+    }
+}
+
+// the following two instances are from rustc
```
Aurel300: Add a permalink to the relevant sources of `rustc`?

prusti-interface/src/specs/encoder.rs
```diff
@@ -0,0 +1,108 @@
...
+            interpret_allocs: Default::default(),
+        }
+    }
+
+    pub fn into_inner(self) -> Vec<u8> {
+        self.opaque.finish()
+    }
+}
+
+macro_rules! encoder_methods {
```
Aurel300, line 38: I assume some of this is also from `rustc`? If so, add a comment similar to the decoder one here, also with a permalink.

prusti-interface/src/specs/lite.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// A subset of [DefSpecificationMap]. Used for serialization of
+/// specification data (of the currently compiled crate), loaded by dependent crates
+/// that import external specification (from the current crate).
+#[derive(TyEncodable)]
+pub struct DefSpecificationMapLite<'tcx, 'a> {
+    proc_specs: &'a HashMap<DefId, SpecGraph<ProcedureSpecification>>,
+    type_specs: &'a HashMap<DefId, TypeSpecification>,
+    prusti_assertions: &'a HashMap<DefId, PrustiAssertion>,
+    prusti_assumptions: &'a HashMap<DefId, PrustiAssumption>,
```
Aurel300: These should not need to be exported.

prusti-interface/src/specs/lite.rs
```diff
@@ -0,0 +1,90 @@
...
+use super::{
+    decoder::{DefSpecsBlob, DefSpecsDecoder},
+    encoder::DefSpecsEncoder,
+};
+
+/// A subset of [DefSpecificationMap]. Used for serialization of
+/// specification data (of the currently compiled crate), loaded by dependent crates
+/// that import external specification (from the current crate).
+#[derive(TyEncodable)]
+pub struct DefSpecificationMapLite<'tcx, 'a> {
```
Aurel300: This is not a great name, as "lite" tells me it is a subset maybe, but not its purpose (to be exported across crate boundaries).

prusti-interface/src/specs/lite.rs
```diff
@@ -0,0 +1,90 @@
...
+        File::create(path).and_then(|mut file| file.write(&encoder.into_inner())).map_err(|err| {
+            warn!("could not encode metadata for crate `{:?}`, error: {:?}", "LOCAL_CRATE", err);
+            err
+        })?;
+        Ok(())
+    }
+}
+
+#[derive(TyDecodable)]
+pub struct DefSpecificationMapLiteOwned<'tcx> {
```
Aurel300: I like this even less: that there are two versions of basically the same `struct` which differ only in their derives and thier ownership. Can you push this into a generic parameter somehow and make the two structs into one?

prusti-interface/src/specs/mod.rs
```diff
@@ -93,9 +101,66 @@ impl<'a, 'tcx> SpecCollector<'a, 'tcx> {
...
+    fn get_local_crate_specs_filename(&self) -> String {
+        let name = (&self.tcx.sess().opts.crate_name).as_ref().unwrap();
+        let stable_id = &self.tcx.sess().local_stable_crate_id().to_u64();
+
+        return format!("{}-{:x}.bin", name, stable_id).to_owned();
+    }
+
+    fn get_serialized_mirs_dir(build_output_dir: &PathBuf) -> Box<PathBuf> {
+        let mut output_dir = Box::new(build_output_dir.clone());
+        output_dir.push("serialized_mirs");
```
Aurel300: More descriptive name: `serialized_specs`? Or `serialized_prusti`?

Also I guess what is happening is we dump the serialised def spec map for each crate we compile into a `.bin` file in this directory. Is the TODO you mention [here](https://github.com/viperproject/prusti-dev/pull/1086#issuecomment-1182780570) the fact that we load all `.bin` files from there? If yes, then we should indeed fix this. For one, we get a directory that is never clean up from compiled crates that are no longer used, for two, we might keep loading broken/invalid specifications (though hopefully they would not be used…).

Aurel300 (commented):
Thanks for the contribution! I left some comments that are mostly minor. I agree with your TODOs as being the most urgent things to address.

prusti-interface/src/environment/mod.rs
jjurm: Thanks! I ran `fmt-all` and `clippy`.

prusti-tests/tests/cargo_verify/library_contracts_lib/src/lib.rs
jjurm: Yes, it does! I extended the test crate to demonstrate importing `#[extern_specs]` from another crate.

prusti-viper/src/encoder/mir/specifications/constraints.rs
jjurm: It seems that implementing this was straightforward; no reason to not have that functionality.

prusti-interface/src/specs/typed.rs
jjurm: I renamed the hashmap to `mirs_of_specs` to make it more self-explanatory. The map is first populated with local MIRs only, but when `proc_specs` and `type_specs` are extended with external specs, this hashmap then contains MIRs of those specs, too.

Perhaps the only reason for storing the hashmap within `DefSpecificationMap` is that it is needed approximately wherever non-local specs are needed.

prusti-viper/src/encoder/encoder.rs
jjurm: The difference is, `get_mir` can also return MIRs of non-local `DefId`s. Those are currently stored in `DefSpecificationMap` (see my reasoning behind this in another comment) and so only accessible within `Encoder` and not `Environment`.

I moved the `get_mir` function to `SpecificationsInterface`, which is still implemented for `Encoder` but hopefully makes more sense there.

Please let me know if you think it makes sense to move the serialized-then-deserialized MIRs (imported from other crates) to `Environment` or if you see a better way to implement this.

prusti-interface/src/specs/typed.rs
jjurm: Right, thanks for this, fixed!

prusti-interface/src/specs/lite.rs
jjurm: Renamed to `DefSpecificationMapForExport`, is this better?

prusti-interface/src/specs/lite.rs
jjurm: - The struct with references is needed for serialization since we are only borrowing the specs, but that is enough to serialize them. We could clone the specs to turn them into owned values, which I wanted to avoid.
- On the other hand, deserialization can not only produce owned values (since they are read from a file), but it also needs owned values since the ownership needs to be transferred when extending `DefSpecificationMap`.

I attempted to abstract away the reference/ownership – my idea was to turn `Box`/`Ref` into a generic parameter and implement each function only for struct with the correct wrapper, but I couldn't make it work so far.
Do you think that this is the right direction? And can this potentially work with `TyEncodable` and `TyDecodable`?

prusti-interface/src/specs/mod.rs
jjurm: Fixed, thanks! And yes, your understanding is correct.

I now pushed code that loads specifications only from crates referenced by `tcx.crates(())` (except the local crate). However, I am not sure if this is exactly the set of crates we want – do you know if there is a better way to list all dependency crates?

prusti-interface/src/specs/lite.rs
jjurm: Fixed, thanks!

prusti-interface/src/specs/mod.rs
Aurel300: That sounds like the correct query. I think "dependency crates" is a concept that exists in `cargo`, and not really in `rustc`. This should always include the set of crates we need, but might be larger, e.g. if `Cargo.toml` declares more dependencies, even if they are unused.

A more fine-grained spec import feature would be better still: we would have to locate the imported specs within `tcx.crates(())`, but at least we would only make use of the ones the user specifically requested.

prusti-interface/src/specs/lite.rs
Aurel300: Sure, at least it's more descriptive :)

JonasAlaif:
@jjurm also try changing the `CLEAR_CROSS_CRATE` flag to see what it does, I have no idea but maybe it'll help with the recompilation issue?

JonasAlaif:
If you have time, could you also try to fix the crash I've introduced now - I didn't have time now before going on holiday

prusti-contracts-common/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
Aurel300: Is this the start of a stdlib crate? Is it automatically used by Prusti somehow?

jjurm: Yes, a stdlib crate. Not used by Prusti itself, only in tests. The purse is so that a prusti user can import contracts for some common stdlib functions.

Please let me know if there is a better place for these contracts. Ideally, they would be in `prusti-contracts`, however, `prusti-contracts` does not use `stdlib` (and I guess there is a reason for it), and I somehow couldn't make prusti syntax work right in the `prusti-contracts` crate.

Pointerbender: I worked a bit on the `prusti-contracts` nostd compatibility and can tell a little more about it. `prusti-contracts` and `prusti-conracts-impl` are `#![nostd]`, but `prusti-contracts-internal` is not. You can put any code that relies on the stdlib in `prusti-contracts-internal`, which is only included when the `prusti` feature flag is enabled. This way, libraries and binaries can still be compiled for nostd targets when not built with Prusti.

jjurm: Thank you for the suggestion; putting the contracts to `prusti-contracts-internal` would be great. However, when trying to `pub use extern_specs;` from the crate, I get the following error:
```
`proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
```
is there something we can do to overcome this limitation, or does this mean we still need a new crate for the extern specs?

prusti-viper/src/encoder/encoder.rs
jjurm: Resolved, functions are now in `Environment`.

prusti-contracts-common/Cargo.toml
Pointerbender: Ah right, I forgot about that limitation for proc-macro crates :clown_face: Then that leaves two options I think:

1. A new crate for extern specs like you mentioned. This could then become a dependency of `prusti-contracts`, but needs some special treatment to avoid accidentally linking the stdlib when prusti is not enabled. I think something like this in `Cargo.toml` of `prusti-contracts` should work:

```toml
[dependencies]
prusti-contracts-extern-spec = { path = "../prusti-contracts-extern-spec", optional = true }

[features]
prusti = ["prusti-contracts-internal", "prusti-contracts-extern-spec"]
```

2. Alternatively, since `prusti-contracts` itself is not a proc-macro crate, you could create a Rust module `extern_specs` in `prusti-contracts` and then conditionally include it from `src/lib.rs` when the `prusti` feature is enabled:

```rust
// When the prusti feature is enabled, we need access to the
// stdlib in order to be able to write `#[extern_spec]` for it.
#![cfg_attr(not(feature = "prusti"), no_std)] // instead of the current `#![no_std]` at the top

// When prusti is active, include the module at `src/extern_spec.rs`
// or at `src/extern_spec/mod.rs`, which can in turn include other
// modules if you want to organize the extern specs into modules.
#[cfg(feature = "prusti")]
mod extern_specs; // `pub` keyword is not needed here, I think?
```

Pointerbender:
By the way, will this functionality also work for items that are both in `core` and `std`? For example, the method `core::mem::swap` is also re-exported as `std::mem::swap` by the standard library. If an `#[extern_spec]` is defined for one of these paths, will it also automatically work for the other? This is handy for `#![no_std]` crates that are verified with Prusti, which typically import everything from the `core::*` namespace instead of from the `std::*` namespace.

Aurel300:
@Pointerbender This discussion is starting to overlap quite a bit with @juliand665's thesis. (Also for that reason @jjurm the solution in this PR does not yet have to be perfect.) Our plan is indeed to support both `core` and `std`, maybe in addition to some third-party libraries. For more info, see the [project description](https://ethz.ch/content/dam/ethz/special-interest/infk/chair-program-method/pm/documents/Education/Theses/Julian_Dunskus_MS_Description.pdf).

As you say, `std` re-rexports many modules from `core`, e.g. [here](https://github.com/rust-lang/rust/blob/e2b52ff73edc8b0b7c74bc28760d618187731fe8/library/std/src/lib.rs#L464-L465) for `Option`. Because extern specs bind to a method by its `DefId`, then the methods of option are the one and the same, no matter what path you used to access them.

prusti-interface/src/environment/mod.rs
```diff
@@ -277,36 +287,96 @@ impl<'tcx> Environment<'tcx> {
...
+                .or_insert_with(|| {
+                    self.raw_load_local_mir(def_id)
+                });
+            body
+        };
+
+        self.subst_into_body(&mut body.body, substs)
+    }
+
+    // used only internally
```
Aurel300: This is already implied by the visibility, but what does the function actually do?

prusti-interface/src/environment/mod.rs
```diff
@@ -277,36 +287,96 @@ impl<'tcx> Environment<'tcx> {
...
+            } else {
+                predicate_bodies.get_mut(&def_id)
+            }.unwrap_or_else(|| panic!("Local body of spec {:?} was not loaded with `load_local_spec_mir`!", def_id));
+            self.subst_into_body(&mut body.body, substs)
+        } else {
+            self.external_spec_mir(def_id, substs)
+        }
+    }
+
+    /// Similar to `spec_mir` but also allows lookup in local bodies.
```
Aurel300: I'm not sure I understand this comment. Doesn't this method perform fewer lookups, not more?

prusti-interface/src/environment/mod.rs
```diff
@@ -319,36 +389,60 @@ impl<'tcx> Environment<'tcx> {
...
-        self.bodies.borrow()
+        let res = self.bodies.borrow()
             .get(&def_id)
-            .map(|body| body.borrowck_facts.clone())
+            .map(|body| body.borrowck_facts.clone());
+        if res.is_none() {
+            self.pure_function_bodies.borrow()
+                .get(&def_id)
+                .map(|body| body.borrowck_facts.clone())
+        } else { res }
```
Aurel300: Would `unwrap_or_else` work here?

prusti-contracts-common/Cargo.toml
JonasAlaif: Thanks, that's super useful @Pointerbender! I'm not sure option 1. would work, since the `prusti-contracts-extern-spec` would also need to have `prusti-contracts` as a dependency, but option 2. should work great - I'll try it out now

prusti-launch/src/bin/prusti-rustc.rs
```diff
@@ -42,33 +42,36 @@ fn process(mut args: Vec<String>) -> Result<(), i32> {
...
 
     // Setting RUSTC_WRAPPER causes Cargo to pass 'rustc' as the first argument.
     // We're invoking the compiler programmatically, so we ignore this
-    if !args.is_empty() && Path::new(&args[0]).file_stem() == Some("rustc".as_ref()) {
+    let cargo_invoked = !args.is_empty() && Path::new(&args[0]).file_stem() == Some("rustc".as_ref());
+    if cargo_invoked {
         args.remove(0);
     }
 
-    // filter out the `prusti-contracts` dependency (if present), because we will
```
Pointerbender, line 61: @JonasAlaif In case you get strange compile errors later on, the check for removing a potential prior existing `--extern prusti_contracts={target_dir}/libprusti_contracts.rlib` flag may need to be left intact here. Rustc is a little weird, that when you specify different versions of the same external crate more than once explicitly, then it sometimes selects the wrong version, leading to a compilation error. This happens, because Cargo also adds such a flag when the `prusti-contracts` crate is specified as a dependency already in `Cargo.toml` - and then `prusti-rustc` adds it again. `prusti-tests/tests/cargo_verify/test_no_std/README.md` is also handy for troubleshooting compilation errors, should you encounter any :-)

prusti-launch/src/bin/prusti-rustc.rs
```diff
@@ -52,10 +52,12 @@ fn process(mut args: Vec<String>) -> Result<(), i32> {
...
-    let cargo_invoked =
-        !args.is_empty() && Path::new(&args[0]).file_stem() == Some("rustc".as_ref());
-    if cargo_invoked {
-        args.remove(0);
+    let rustc_pos = args.iter().position(|arg|
+        Path::new(arg).file_stem() == Some("rustc".as_ref())
+    );
+    let cargo_invoked = rustc_pos.is_some();
+    if let Some(rustc_pos) = rustc_pos {
+        args.drain(0..=rustc_pos);
```
Aurel300, line 61: What do the dropped arguments look like?

JonasAlaif: You can see in the failing run (https://github.com/viperproject/prusti-dev/runs/7939531751?check_suite_focus=true): `/home/runner/work/prusti-dev/prusti-dev/target/debug/prusti-rustc /home/runner/.rustup/toolchains/nightly-2022-08-01-x86_64-unknown-linux-gnu/bin/clippy-driver rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg`

JonasAlaif: So `/home/runner/.rustup/toolchains/nightly-2022-08-01-x86_64-unknown-linux-gnu/bin/clippy-driver rustc` gets dropped

JonasAlaif:
Ooof... this should be ready to merge now if CI checks pass. If you want, we can discuss a bit before then @Aurel300?

docs/dev-guide/src/config/flags.md
```diff
@@ -267,6 +268,10 @@ When enabled, logs all SMT wrapper interaction to a file.
 
 Maximum allowed length of a log file name. If this is exceeded, the file name is truncated.
 
+## `MIN_PRUSTI_VERSION`
+
+Minimum required version of Prusti that is allowed to run. If Prusti detects that it's own version is lower than this, it will throw an error and refuse to verify files. Generally [set in a `Prusti.toml` file](providing.md#flags-2) of a crate to enforce a minimum Prusti version.
```
Aurel300:
```suggestion
Minimum required version of Prusti that is allowed to run. If Prusti detects that its own version is lower than this, it will throw an error and refuse to verify files. Generally [set in a `Prusti.toml` file](providing.md#flags-2) of a crate to enforce a minimum Prusti version.
```

prusti-contracts-build/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
Aurel300: Add a comment to explain this file?

prusti-contracts/prusti-contracts-std/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "prusti-contracts-std"
+version = "0.1.0"
+authors = ["Vytautas Astrauskas <vastrauskas@gmail.com>"]
```
Aurel300, line 4: Is everything Prusti-related authored by Vytautas? :) We might want to put something standard in all our crates, e.g. "Prusti contributors" + a mailing list.

prusti-interface/src/environment/body.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    /// Get the MIR body of a local impure function, without any substitutions.
+    pub fn get_impure_fn_body(&self, def_id: LocalDefId) -> MirBody<'tcx> {
+        let mut impure = self.local_impure_fns.borrow_mut();
+        impure.entry(def_id).or_insert_with(|| Self::raw_load_local_mir(self.tcx, def_id)).body.clone()
+    }
+
+    /// Get the MIR body of a local impure function, monomorphised
+    /// with the given type substitutions.
+    pub fn get_impure_fn_body_subs(
```
Aurel300: `subs` here (and all others) stands for "substitutions"? `subst` might be more consistent. However, I would also expect most of the calls to use the substituting variant, in which case the other ones can be called e.g. `get_*_ident` or `_identity`.

prusti-tests/tests/cargo_verify/test_no_std/Prusti.toml
```diff
@@ -0,0 +1 @@
+be_rustc = true
```
Aurel300, line 1: Can you explain this? Does this not just skip Prusti altogether?

prusti-launch/src/lib.rs
```diff
@@ -26,6 +34,12 @@ pub fn get_current_executable_dir() -> PathBuf {
         .to_path_buf()
 }
 
+pub fn get_prusti_contracts_dir(exe_dir: PathBuf) -> PathBuf {
+    let mut prusti_dir = exe_dir.parent().unwrap().parent().unwrap().to_path_buf();
+    prusti_dir.extend(["prusti-contracts", "target", "verify", "release"]);
```
Aurel300, line 39: Can the `release` build be used as a dependency for `debug` builds?

prusti-interface/src/specs/checker/version_checks.rs
```diff
@@ -0,0 +1,73 @@
...
+                    MultiSpan::new(),
+                ));
+            }
+        }
+        check_version.errors
+    }
+}
+
+/// First predicate checks visitor: collect all function items that originate
+/// from predicates
```
Aurel300: `from predicates` -> `from specifications`?

prusti-interface/src/specs/checker/version_checks.rs
```diff
@@ -0,0 +1,73 @@
...
+    PrustiError, utils,
+};
+use prusti_common::config;
+use prusti_rustc_interface::errors::MultiSpan;
+use prusti_rustc_interface::hir;
+use prusti_rustc_interface::middle::hir::map::Map;
+use prusti_rustc_interface::ast::ast::Attribute;
+
+/// Checks for mismatched version issues between `prusti` and `prusti-contracts`/`prusti-specs`
+#[derive(Default)]
```
Aurel300: Is the `Default` needed?

Aurel300:
Generally looks good! I wish it wasn't a PR that combined so many changes:

- support cross-crate specs;
- refactor build process for loading `prusti-contracts` and core/std specs;
- refactor `Environment` into separate parts (+ adapt rest of codebase).

Although you documented `MIN_PRUSTI_VERSION`, I think we should document the contracts feature/crate more thoroughly in the user guide, as there are probably some pitfalls. We can push that onto another PR though, since this one is large and will easily get into conflicts.

prusti-contracts/prusti-contracts-std/Cargo.toml
JonasAlaif: Haha I was wondering if someone would notice this. I agree, but we don't have such a mailing list yet right?

prusti-launch/src/lib.rs
JonasAlaif: From seeing that the tests pass I guess so. The thing is that this will never be an issue with cargo builds (c.f. `get_prusti_contracts_dir` is only called in `if !cargo_invoked { ... }`) and is only relevant for raw `prusti-rustc`, where you can toggle optimisations with the `-O` flag. I doubt this would cause issues there though, since optimisations are only relevant after Prusti/early-MIR.

prusti-tests/tests/cargo_verify/test_no_std/Prusti.toml
JonasAlaif: Yeah, but I don't think this test requires us to run Prusti on it at all, right @Pointerbender? I tried it out with removing the `#![no_std]` from `prusti-contracts` and it failed as expected, so assumed that it's all good.

JonasAlaif: The reason I had to add this is that with the new changes (compiling+verifying `prusti-contracts` with `cargo-prusti`) one can no longer run Prusti on a crate with the dependency `prusti-contracts = { path = "path/to/prusti-contracts/prusti-contracts" }` since you'll get the following error:
```
error: [Prusti: verification error] statement might panic
   --> .../prusti-dev/prusti-contracts/prusti-contracts/src/lib.rs:395:5
    |
395 |     unimplemented!()
    |     ^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the macro `unimplemented` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `prusti-contracts` due to previous error
```
Which is complaining about this fn:
```
#[trusted]
pub fn snap<T>(_x: &T) -> T {
    unimplemented!()
}
```
Instead Prusti will only work on crates with `prusti-contracts = { path = "path/to/prusti-contracts/prusti-contracts", features = ["prusti"] }`. Without the `prusti` feature the `#[trusted]` becomes transparent and Prusti complains about the `unimplemented!()` (we'll likely want to change this behaviour)

Pointerbender: > Yeah, but I don't think this test requires us to run Prusti on it at all, right @Pointerbender? I tried it out with removing the `#![no_std]` from `prusti-contracts` and it failed as expected, so assumed that it's all good.

@JonasAlaif Yep, that's right! (Apologies for the late answer, I just got back from holiday :D )

PR #1091, opened 2022-07-15
Try disabling more complete exhale and see what happens.

vakaras:
At least on some examples, it seems that the more complete exhale does not have anymore its great effect on performance. Since it has incompletenesses, try disabling it.


Aurel300:
@vakaras Since you are merging this, are you 100% sure the performance is overall better? Did @zgrannan get to test this?

vakaras:
> @vakaras Since you are merging this, are you 100% sure the performance is overall better? Did @zgrannan get to test this?

Yes, he did.

vakaras:
bors cancel

PR #1096, opened 2022-07-20
Improve extern spec lifetime handling

zgrannan:
Due to lifetime elision rules, lifetimes that can be elided in a method need to be explicitly annotated in an extern spec. This PR weakens the validity check for extern specs to ignore lifetimes (since they wont match).

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/extern-spec/map.rs
+++ b/prusti-tests/tests/verify_overflow/pass/extern-spec/map.rs
@@ -1,0 +1,32 @@
+use prusti_contracts::*;
+
+#[extern_spec]
+impl<T> std::option::Option<T> {
+    #[pure]
+    pub fn is_some(&self) -> bool;
+}
+
+#[extern_spec]
+impl<K, V, S: std::hash::BuildHasher> std::collections::HashMap<K, V, S> {
+    #[pure]
+    #[ensures(result.is_some() == self.contains_key(k))]
+    pub fn get<'a, Q: ?Sized>(&'a self, k: &Q) -> Option<&'a V>
+    where
+        K: core::borrow::Borrow<Q> + std::cmp::Eq + std::hash::Hash,
+        Q: core::hash::Hash + Eq;
+
+    #[pure]
+    fn contains_key<Q: ?Sized>(&self, k: &Q) -> bool
+    where
+        K: core::borrow::Borrow<Q> + std::cmp::Eq + std::hash::Hash,
+        Q: core::hash::Hash + Eq;
+}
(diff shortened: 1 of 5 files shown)


prusti-tests/tests/verify_overflow/pass/extern-spec/map.rs
```diff
@@ -0,0 +1,32 @@
...
+impl<T> std::option::Option<T> {
+    #[pure]
+    pub fn is_some(&self) -> bool;
+}
+
+#[extern_spec]
+impl<K, V, S: std::hash::BuildHasher> std::collections::HashMap<K, V, S> {
+    #[pure]
+    #[ensures(result.is_some() == self.contains_key(k))]
+    pub fn get<'a, Q: ?Sized>(&'a self, k: &Q) -> Option<&'a V>
```
Aurel300, line 13: So the point is that the `'a` lifetime needs to be declared here, but that does not match the lifetime signature of the real `get` method because it uses a `&self` with an elided lifetime? Is this something we could address specifically for `&self` and `&mut self`? (We translate `self` specially to some `_self` variable.) Or is there an example which benefits from the change in this PR that is not related to `self`?

zgrannan: I'll look into special casing self, as this is specific to the lifetime elision rule for `self`, namely the rule:

> If there are multiple input lifetime positions, but one of them is `&self` or `&mut self`, the lifetime of `self` is assigned to *all* elided output lifetimes.

From https://doc.rust-lang.org/nomicon/lifetime-elision.html

zgrannan: Ah actually it looks like the problem was fixed by proper handling of lifetimes (see [prusti-specs/src/common.rs](https://github.com/viperproject/prusti-dev/pull/1096/files#diff-3e6abe6e1126c31f7da6547a74c472e9a2d52a76045de256fdbde979e418965a))

But I've also made a change to the rewriter to perform the "elision rule 3" conversion explicitly; so users of prusti shouldn't have to add these lifetime annotations (since they'd probably need to know the implementation details of Prusti to know to do that)

Aurel300: Nice, I'm glad to see that old TODO gone!

vakaras:
> Timed out.

For some reason, if the PR is not modified, `bors retry` does not trigger the action resulting in a timeout. The easiest fix is to make a small modification in the PR, for example, rebase on master.

PR #1102, opened 2022-07-27
Retry "fetch" if it failed in test-crates

zgrannan:
Sometimes the `test-crates` task fails due to an intermittent network failure. This PR ensures that it will retry if a failure occurs.


vakaras (commented):
LGTM

PR #1105, opened 2022-07-27
Fix trait selection bug

zgrannan:
For some reason the existing code for trait selection did not find the associated trait for an extern spec in one example. This PR re-uses Rust's existing code for trait selection, which fixes the problem.

### DIFF
--- a/prusti-interface/src/environment/mod.rs
+++ b/prusti-interface/src/environment/mod.rs
@@ -489,7 +489,16 @@ impl<'tcx> Environment<'tcx> {
                 )
             });
             match result {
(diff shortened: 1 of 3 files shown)


Aurel300:
Does this fix https://github.com/viperproject/prusti-dev/blob/master/prusti-tests/tests/verify_overflow/pass/extern-spec/trait-impl/traits-4.rs by any chance?

zgrannan:
> Does this fix https://github.com/viperproject/prusti-dev/blob/master/prusti-tests/tests/verify_overflow/pass/extern-spec/trait-impl/traits-4.rs by any chance?

Yes!

PR #1115, opened 2022-08-02
Termination Measures for Loops and Functions

JM4ier:


prusti-viper/src/encoder/mir/procedures/encoder/loops.rs
```diff
@@ -100,4 +100,81 @@ impl<'p, 'v: 'p, 'tcx: 'v> super::ProcedureEncoder<'p, 'v, 'tcx> {
         )?;
         Ok(statement)
     }
+    pub(super) fn encode_loop_variant(
```
vakaras: As discussed this probably should be part of the `LoopInvariant` desugaring.

prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
```diff
@@ -1801,6 +1925,21 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 .insert(invariant_location, statement);
         }
 
+        // Encode loop variants.
+        let mut loop_variant_blocks = self.specification_blocks.loop_variant_blocks().clone();
```
vakaras: This should be done together with the loop invariant.

prusti-viper/src/encoder/mir/procedures/encoder/specification_blocks.rs
```diff
@@ -20,6 +20,7 @@ pub(super) struct SpecificationBlocks {
     ///
     /// FIXME: Add a check that ensure that the blocks are one after another.
     loop_invariant_blocks: BTreeMap<mir::BasicBlock, LoopInvariantBlocks>,
+    loop_variant_blocks: BTreeMap<mir::BasicBlock, LoopInvariantBlocks>,
```
vakaras: Couldn't these be merged?

vakaras (changes requested):
Overall looks fine to me. Major things to fix:

* [ ] Rebase/merge master.
* [ ] Merge `LoopVariant` into `LoopInvariant`.

vakaras:
@Aurel300 Are there still unmerged PRs that would be invalidated if we rustfmt `prusti-viper` and `prusti-interface`?

prusti-tests/tests/verify/pass/terminating_fib.rs
```diff
@@ -0,0 +1,16 @@
+// compile-flags: -Punsafe_core_proof=true
+
+#![allow(unused)]
+
+use prusti_contracts::*;
+
+fn main() {}
+
+#[pure(Int::new(x))]
```
Aurel300: What is this? Is this the termination measure? Why is it attached to the `pure` attribute?

JM4ier: yes, it is a termination measure.
The PR makes it a requirement for recursive pure functions, or recursive functions with the `#[terminates]` attribute.
I also specifically added it to the `#[pure]` attribute because nonterminating pure expressions in pre/post conditions can cause unsoundness

Aurel300: Still I don't think the UX is good. It should be clear when reading the contract of a method at least roughly what it means. So, if a termination measure is required for a pure function and there is none provided, we should tell the user to write a `#[terminates(...)]` attribute.

prusti-contracts/prusti-contracts-impl/src/lib.rs
```diff
@@ -0,0 +1,92 @@
+#![no_std]
```
vakaras: This file was deleted on `master`.

prusti-contracts/prusti-contracts-internal/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+use proc_macro::TokenStream;
+use prusti_specs::{rewrite_prusti_attributes, SpecAttributeKind};
```
vakaras: This file also does not exist on `master`.

prusti-interface/src/environment/mod.rs
```diff
@@ -134,6 +135,60 @@ impl<'tcx> Environment<'tcx> {
             .unwrap()
     }
 
+    pub fn is_recursive(&self, proc_def_id: ProcedureDefId) -> bool {
```
vakaras: Isn't this function dead code? If yes, please delete it.

prusti-viper/src/encoder/mir/type_layouts/interface.rs
```diff
@@ -11,14 +11,23 @@ pub(crate) trait MirTypeLayoutsEncoderInterface<'tcx> {
         &self,
         ty: ty::Ty<'tcx>,
     ) -> SpannedEncodingResult<vir_high::Expression>;
+    fn encode_type_size_expression_with_reps_vir(
```
vakaras: Dead code?

vakaras (changes requested):
Looks good module the comments. Thanks!

prusti-contracts/prusti-specs/src/rewriter.rs
```diff
@@ -110,12 +112,17 @@ impl AstRewriter {
         //   of a single identifier; without the double negation, the `Return`
         //   terminator in MIR has a span set to the one character just after
         //   the identifier
+        let (return_type, return_modifier) = if spec_type == SpecItemType::Termination {
+            (quote! {Int}, quote! {Int::new(0) + })
```
vakaras: Could you try using `quote_spanned!` here to see whether that fixes the `test_failing_crate` test?

JonasAlaif:
@vakaras I think you still need the enum, what if we decide not to use a `HashMap<u64, VerificationResult>` for the `data`

vakaras:
> @vakaras I think you still need the enum, what if we decide not to use a `HashMap<u64, VerificationResult>` for the `data`

Do we want to support two versions of the cache at the same time? If not, I do not see why we need that.

JonasAlaif:
I mean you need to at least change this line to instead give an empty cache:
```
error!("Failed to read cache from \"{}\": {e}", cache_loc.display()),
```

vakaras:
> I mean you need to at least change this line to instead give an empty cache:
> 
> ```
> error!("Failed to read cache from \"{}\": {e}", cache_loc.display()),
> ```

Why the already existing code does not do that?

fpoli:
> I mean you need to at least change this line to instead give an empty cache:
> 
> ```
> error!("Failed to read cache from \"{}\": {e}", cache_loc.display()),
> ```

That already happens, right? In case of error the `data_res` remains `None`, so later the `unwrap_or_else` creates a fresh cache.

---

The only reason for the enum was to ensure that reading an old version of the cache doesn't succeed by mistake. However, I think that serde doesn't guarantees an error when deserializing with a different type, not even if the new type is in a new variant of the enumeration that we had. It can be, for example, that the serialization just calls `0` the *last* variant of an enumeration, and if so deserializing with the additional variant whould happily succeed. The order of the variants might not even be guaranteed across compiler versions, and when an enum has only one variant the discriminant might not even be serialized.

Adding a version field makes the check more explicit; we no longer rely on the serialization of the variants and I think it's better than before. The only failure that I think is still possible is that
1. We change the `HashMap` to a `Vec` and bump the version of the cache.
2. At the same time, a change in serde or the compiler swaps the order in which fields are deserialized.
3. By chance, decoding the old version of the cache happens to produce a version that matches the bumped version that we expect.

Not knowing exactly what are the safety guarantees of serde, I assume that (3) is very unlikely but still possible. If there was a way to force the serialization to *always* serialize the version field first then I think we would be fully covered.

vakaras (approved):

PR #1116, opened 2022-08-02
Fix Issue 1022

### LINKED ISSUE #1022: Unsound refinement checks
Prusti checks whether trait refinements are valid, i.e. when having code like
```rust
trait SomeTrait {
    // precondition P_T
    // postcondition Q_T
    fn foo(x: i32) -> i32;
}
struct SomeStruct;
#[refine_trait_spec]
    // precondition P_S
    // postcondition Q_S
    fn foo(x: i32) -> i32 { ... }
}
```
Prusti checks whether $P_T \implies P_S$ and $Q_S \implies Q_T$.
By design, Prusti assumes a `#[requires(true)]`(`#[ensures(true)]`) when the precondition (postcondition) is not available.

Having no postconditions on the trait, an implementation of a trait with a postcondition is always a valid refinement. This is, however, not the case for preconditions, as $true \implies P_S$ in most cases never hold for a non-trivial $P_S$.

Prusti however does not conduct the refinement check when there is no precondition on the trait. The following verifies, even though it shouldn't:
```rust
trait SomeTrait {
    fn foo(&self, x: i32) -> i32;
}
struct SomeStruct;
#[refine_trait_spec]
    #[requires(x > 0)]
    fn foo(&self, x: i32) -> i32 {
        if x <= 0 { panic!(); }
        42
    }
}

fn fooer<T: SomeTrait>(fooable: T, x: i32) -> i32 {
    fooable.foo(x)
}

fn main() {
    let s = SomeStruct;
    fooer(s, -1); // panics, but Prusti doesn't catch it
}
```

## Technical problem description
Refinement info is generally tracked via `SpecificationItem`s and computed in `prusti-viper::encoder::mir::specifications::specs` in the method `perform_proc_spec_refinement`.

In case we have no user-provided specs for a trait method, `perform_proc_spec_refinement` will default the trait method spec to an empty `ProcedureSpecification` which results in a `SpecificationItem::Inherent` for the implementation of the trait method -> Prusti looses context that the impl method spec is a refinement of a trait method spec.

The refinement check itself is conducted in `ProcedureEncoder::encode_spec_refinement`, and encodes a the refinement check if it observes a `SpecificationItem::Refined`.

## Proposed solutions
 - (a) (cleaner) The impl spec should be a `SpecificationItem::Refined(trait, impl)`, even though there are no trait specs, where `trait` are the default specs for non-annotated trait methods. This would automatically trigger the encoding in `ProcedureEncoder::encode_spec_refinement`.
 - (b) (maybe easier, a bit uglier) Handle non-present trait method specs somewhere before/in `ProcedureEncoder::encode_spec_refinement`. The reason I consider this *uglier* is because `SpecificationItem` should imo carry the information that we refine an implicit default trait spec.

dewert99:
Fixes #1022.

Adding a precondition (other than `true`) to a trait method with preconditions and implementing an unspecified trait method with a predicate now cause errors.  Additionally adding an `assert_on_expiry` pledge to the implementation of a trait method without pledges now causes a panic with the message "Refining specifications with pledges is not supported" (This already happens with refining a pledge) since it is unsound in general (This should be improved when pledge refining is implemented).


prusti-interface/src/specs/typed.rs
```diff
@@ -669,7 +662,6 @@ mod tests {
         );
 
         refine_success_tests!(
-            refine_from_refined_with_empty: (Refined(1, 2), Empty, Refined(1, 2)),
             // Generally, we can not refine a refined item.
             // In this case it is possible, because the refined-from part does not change.
             refine_from_refined_with_inherited_refinable: (Refined(1, 2), Inherited(1), Refined(1, 2)),
```
dewert99, line 667: In what case do we refine an already refined item

prusti-viper/src/encoder/mir/specifications/specs.rs
```diff
@@ -145,13 +145,9 @@ impl<'tcx> Specifications<'tcx> {
...
-        let trait_spec = self.get_proc_spec(env, trait_query);
-
-        let refined = if let Some(trait_spec_set) = trait_spec {
-            impl_spec.refine(trait_spec_set)
-        } else {
-            impl_spec
-        };
+        let empty = ProcedureSpecification::empty();
+        let trait_spec = self.get_proc_spec(env, trait_query).unwrap_or(&empty);
+        let refined = impl_spec.refine(trait_spec);
```
Aurel300, line 150:
```suggestion
        let trait_spec = self.get_proc_spec(env, trait_query)
            .unwrap_or_else(ProcedureSpecification::empty);
        let refined = impl_spec.refine(trait_spec);
```

(Maybe even `unwrap_or_default` would work here?)

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3888,8 +3888,12 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
         }
 
-        if let SpecificationItem::Refined(_, _) = &procedure_spec.pledges {
-            unimplemented!("Refining specifications with pledges is not supported");
+        if let SpecificationItem::Refined(from, to) = &procedure_spec.pledges {
+            if !from.is_empty() || to.iter().any(|p| p.lhs.is_some()) {
+                unimplemented!("Refining specifications with pledges is not supported");
+            }
+            // If the trait has no pledges and the implementer only makes additional guarantees
+            // then the refinement is safe
```
Aurel300, line 3896: Sounds good, please add a test case for this.

prusti-viper/src/encoder/mir/specifications/specs.rs
dewert99: Currently, `trait_spec` is required to be a reference due to the signature of `Refine::refine`, I could change the signature to take ownership instead which would allow me to simplify things.  I don't think this would have much impact on performance since in almost all cases the individual parts a cloned anyways (or it panics)

Aurel300: Ah, never mind then, I missed the `&`.

PR #1118, opened 2022-08-03
Erase regions for type invariants

zgrannan:
Currently the type invariant encoding for Prusti will panic if it encodes two datatypes that differ only in a lifetime parameters. This is because Prusti encodes them distinct invariants for the two types, but generates the same Viper function name for both.

This implementation resolves this by erasing the memory regions from the types before encoding the invariant. This solves the panic; however, prevents the possibility of having distinct invariants for types only differing by an instantiation of the lifetime parameter. If there is such a use case this design could be re-considered.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/invariants/map.rs
+++ b/prusti-tests/tests/verify_overflow/pass/invariants/map.rs
@@ -1,0 +1,12 @@
+// compile-flags: -Penable_type_invariants=true
+use std::collections::BTreeSet;
+
+// No invariants are used here, but the encoding of the `union` operation
+// introduces datatypes that only vary in a generic lifetime parameter. For the
+// purpose of type invariants, these datatypes should be treated as the same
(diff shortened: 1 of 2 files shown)


Aurel300:
Looks good, thanks.

> [...] the possibility of having distinct invariants for types only differing by an instantiation of the lifetime parameter.

I don't think such a use case exists. I guess this would be something like a type with e.g. `'a, 'b` lifetime parameters and a type invariant that somehow depends on the concrete relationship between these two parameters? This would violate parametricity, at least.

@vakaras might not like yet more regions being erased, but I don't know if this applies to type invariants as well.

bors r+

vakaras:
> @vakaras might not like yet more regions being erased, but I don't know if this applies to type invariants as well.

In polymorphic, this is fine. `vir_high` has a special method `erase_lifetimes` that is used before checking whether the type is encoded or not.

PR #1119, opened 2022-08-04
Fix for quantified reference

zgrannan:
This fixes a fold/unfold error when handling a quantified reference (see the test case).

It appears the problem occurs in the inlining of the closure generated for the spec: the original implementation did not substitute the argument. This appears to be because the substitution map used the value expr of the argument rather than the argument itself; and the value expr of the argument is not used in the spec. Therefore, the substitution is never applied, and the unsubstituted local cannot be accessed by the fold/unfold algorithm. 

This PR changes the map to instead use the argument itself, rather than its value expr.


Aurel300:
Thanks!

bors r+

Aurel300:
bors retry

Issue #178, opened 2020-09-22
Limitations of Prusti attributes recognition


fpoli:
The way `prusti-spec` recognizes Prusti attributes is not perfect. To solve issue #167 we implemented in #177 a technique that works very well in practice, but has some limitations regarding e.g. renaming of Prusti attributes. See the two examples below.

Ideally, we should support such cases or at least report a meaningful error, but both seem hard to implement.

```rust
use prusti_contracts::requires as pre;

// Prusti doesn't recognized the second attribute, because it doesn't know
// that `requires` has been renamed to `pre`, so the compiler runs it as a
// procedural macro and thus loses information on spans.
#[pre(true)]
#[pre(true)] // <-- The error message is incorrectly reported here.
fn main() {
    123 = u32; // <-- The error is here.
}
```

```rust
use prusti_contracts::requires as pre;
use prusti_contracts::ensures as requires;

// Weird edge case: the second attribute is interpreted by Prusti as a
// `requires` even if it should be an `ensures` due to the renaming defined
// at `use prusti_contracts::ensures as requires`.
#[pre(false)]
#[requires(false)] // <-- No verification error is reported at this postcondition.
fn main() {}
```

PR #1128, opened 2022-08-11
Remove `extern crate prusti_contracts` from Prusti

JonasAlaif:


docs/user-guide/src/tour/tour-src/05-chapter-2-3.rs
```diff
@@ -5,9 +5,8 @@
     To this end, we also need to implement a length function for links
 */
 
-// (1) required to enable Prusti's annotations (implemented as Rust macros/attributes)
-extern crate prusti_contracts;
-use prusti_contracts::*; 
+// (1) simplifies writing Prusti's annotations (implemented as Rust macros/attributes)
```
Aurel300, line 8: Are you sure about "simplifies"? When we process nested attributes, do we also allow `prusti_contracts::_`? E.g.

```
#[prusti_contracts::requires(...)]
#[prusti_contracts::ensures(...)]
fn foo() {}
```

Would the second attribute parse successfully? (Related #178)

Aurel300:
Thanks, this was also bugging me.

docs/user-guide/src/tour/tour-src/05-chapter-2-3.rs
JonasAlaif: From what I remember I think things should work, maybe with minor span glitches - it'll just end up running the proc macro twice. Is there actually any reason to handle all of the pres/posts in one go vs. letting each one fire individually? It seems that the latter would be more resilient to this kind of stuff

JonasAlaif: A simple fix would be to handle both cases of `#[requires(...)]` and `#[prusti_contracts::requires(...)]`, see latest push (though `"generic"` and `"concrete"` annotations would still throw an error I guess)

zgrannan:
Finished benchmarking commit (c449775827cf7227253197dd9c2563e04953a62e): [comparison url](http://3.94.193.1:2346/compare.html?start=81c84a036c9ae233806706da639be45165593539&end=c449775827cf7227253197dd9c2563e04953a62e&stat=wall-time).

Benchmark returns the average of 3 consecutive runs of the benchmark suite, performed after an initial 'warm-up' run. An external Prusti server is maintained over the 4 runs.
         
### [Wall time](http://3.94.193.1:2346/compare.html?start=81c84a036c9ae233806706da639be45165593539&end=c449775827cf7227253197dd9c2563e04953a62e&stat=wall-time)
This benchmark run did not return any relevant results for this metric.

### [Instruction count (excludes Viper)](http://3.94.193.1:2346/compare.html?start=81c84a036c9ae233806706da639be45165593539&end=c449775827cf7227253197dd9c2563e04953a62e&stat=instructions:u)
This benchmark run did not return any relevant results for this metric.

### [Cycles (excludes Viper)](http://3.94.193.1:2346/compare.html?start=81c84a036c9ae233806706da639be45165593539&end=c449775827cf7227253197dd9c2563e04953a62e&stat=cycles:u)
<details>
<summary>Results</summary>


- Primary benchmarks: 🎉 relevant improvement found
- Secondary benchmarks: no relevant changes found

|            | mean[^1] | max | count[^2] |
|:----------:|:--------:|:---:|:---------:|
| Regressions 😿 <br /> (primary) | N/A      | N/A | 0         |
| Regressions 😿 <br /> (secondary) | N/A      | N/A | 0         |
| Improvements 🎉 <br /> (primary) | -2.0%    | -2.0% | 1         |
| Improvements 🎉 <br /> (secondary) | N/A      | N/A | 0         |
| All 😿🎉 (primary) | -2.0%    | -2.0% | 1         |
</details>

[^1]: *the arithmetic mean of the percent change*
[^2]: *number of relevant changes*

If you disagree with this performance assessment, please file an issue in [prusti-dev/prusti-perf](https://github.com/prusti-dev/prusti-perf/issues/new).

@rustbot label: -perf-regression

PR #1136, opened 2022-08-20
Document prusti_assert and prusti_assume

zgrannan:
I figured these two macros should be documented, since I didn't realize they existed.

Also I am not sure about the relationship between these and the unsafe core proof; so perhaps that may need some clarification before merging?

### DIFF
--- a/docs/user-guide/src/verify/assert_assume.md
+++ b/docs/user-guide/src/verify/assert_assume.md
@@ -7,12 +7,12 @@ function (via an assumption).
 
 ## Assertions
 
-The `prusti_assert!()` macro instructs Prusti to verify that a certain property
+The `prusti_assert!` macro instructs Prusti to verify that a certain property
 holds at a specific point within the body of a function. In contrast to the
-`assert!()` macro, which only accepts Rust expressions, `prusti_assert!()`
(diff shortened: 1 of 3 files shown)


docs/user-guide/src/verify/assert_assume.md
```diff
@@ -0,0 +1,43 @@
...
+within the body of a function (via an assertion). It is also possible to
+instruct Prusti to assume that a property holds at a certain point within a
+function (via an assumption).
+
+## Assertions
+
+The `prusti_assert!()` macro instructs Prusti to verify that a certain property
+holds at a specific point within the body of a function. In contrast to the
+`assert!()` macro, which only accepts Rust expressions, `prusti_assert!()`
+accepts [specification](../syntax.md) expressions as arguments. Therefore,
```
Aurel300: Although we state in the syntax page that Prusti expressions are side-effect free, we should also say it here. For example `assert!(map.insert(5));` would not work with `prusti_assert!`

Aurel300:
Thanks. As I said on Zulip, these features can be ported back to the current encoding relatively easily. I would prefer to do that first and then merge this documentation.

vakaras:
> I would prefer to do that first and then merge this documentation.

I agree. I think the documentation should reflect the stable features.

PR #1137, opened 2022-08-20
Document snapshot equality

zgrannan:
Please feel free to suggest changes.  Ideally, I'd like to explain the feature without requiring the audience to necessarily understand the snapshot encoding.


Aurel300:
Looks good. Although we indeed might not want to explain snapshots fully (we've had this discussion a couple of times already!), we could at least note that snapshots follow nested fields and references, but not unsafe pointers or `UnsafeCell`s.

Aurel300:
Thanks! I'll merge this now since the documentation is helpful, we can adjust the wording later if anybody finds it confusing.

PR #1138, opened 2022-08-22
Ghost constraint improvements

juliand665:
I reworked ghost constraint parsing to directly use `syn` for the trait bounds and comma, only then handing the rest over to Prusti's parser. Additionally, I extended them to handle `pure` flags (in addition to the status quo of `requires` and `ensures`). The effects are as follows:
- Any trait bounds that are valid rust are now inherently valid in ghost constraints.
- The separating comma between bounds & specs no longer requires a space if preceded by a `>`.
- Ghost constraints can now specify `pure`ness.
- The `ghost_constraint` proc macro is now correctly a no-op when compiling without prusti, rather than removing the function entirely.

Some considerations for code review:
- I just kinda threw in the logic with the rest of it in preparser.rs; perhaps splitting it (or other parts) off into its own file would make sense? It's a pretty long file at this point.
- I wasn't sure how best to translate the existing unit tests now that we no longer represent the traits as an unparsed `TokenStream`. I ended up providing the strings as `quote!` streams instead, but it feels like that's not really testing anything of value anymore (hence also why I dropped it in the new `tuple_generics` test).
- The errors reported for invalid syntax are better in some ways and worse in other ways compared to the status quo. What I most miss is that the old errors often provided an example of how to structure your ghost constraint—perhaps it would make sense to somehow attach this to whatever error syn outputs?

### DIFF
--- a/prusti-specs/src/lib.rs
+++ b/prusti-specs/src/lib.rs
@@ -242,6 +242,10 @@ fn generate_for_pure(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedR
     do_generate_for_pure(item)
 }
 
+/// Generate spec items and attributes to typecheck and later retrieve "pure" annotations.
+/// 
+/// This the actual generating logic called by `generate_for_pure` after checking that the body is empty.
+/// It's exposed separately for use in ghost constraints.
 fn do_generate_for_pure(item: &untyped::AnyFnItem) -> GeneratedResult {
     Ok((
         vec![],
--- a/prusti-specs/src/lib.rs
+++ b/prusti-specs/src/lib.rs
@@ -4,6 +4,7 @@
 #![feature(box_syntax)]
 #![feature(proc_macro_span)]
 #![feature(if_let_guard)]
+#![feature(assert_matches)]
 // This Clippy chcek seems to be always wrong.
 #![allow(clippy::iter_with_drain)]
 
--- a/prusti-specs/src/lib.rs
+++ b/prusti-specs/src/lib.rs
@@ -238,6 +238,10 @@ fn generate_for_pure(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedR
         ));
     }
 
+    do_generate_for_pure(item)
+}
+
+fn do_generate_for_pure(item: &untyped::AnyFnItem) -> GeneratedResult {
     Ok((
         vec![],
         vec![parse_quote_spanned! {item.span()=>
--- a/prusti-specs/src/specifications/preparser.rs
+++ b/prusti-specs/src/specifications/preparser.rs
@@ -51,11 +51,7 @@ pub fn parse_prusti_assert_pledge(tokens: TokenStream) -> syn::Result<(TokenStre
 }
 
 pub fn parse_ghost_constraint(tokens: TokenStream) -> syn::Result<GhostConstraint> {
-    syn::parse2(tokens).map_err(|mut err| {
-        err.combine(error(err.span(),
-            "expected a trait bound and specifications in brackets, e.g.: `ghost_constraint(T: A + B + ..., [requires(...), ...])`"));
-        err
-    })
+    syn::parse2(tokens)
 }
 
 /*
@@ -510,7 +506,7 @@ impl Parse for GhostConstraint {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         Ok(GhostConstraint {
             trait_bounds: parse_trait_bounds(input)?,
-            comma: input.parse()?,
+            comma: input.parse().map_err(with_ghost_constraint_example)?,
             specs: PrustiTokenStream::new(input.parse().unwrap())
                 .parse_rest(|pts| pts.pop_group_of_nested_specs(input.span()))?,
         })
@@ -519,7 +515,7 @@ impl Parse for GhostConstraint {
 
 fn parse_trait_bounds(input: ParseStream) -> syn::Result<syn::PredicateType> {
     use syn::WherePredicate::*;
-    let where_predicate: syn::WherePredicate = input.parse()?;
+    let where_predicate: syn::WherePredicate = input.parse().map_err(with_ghost_constraint_example)?;
     match where_predicate {
         Type(type_bound) => {
             validate_trait_bounds(&type_bound)?;
@@ -552,6 +548,11 @@ fn validate_trait_bounds(trait_bounds: &syn::PredicateType) -> syn::Result<()> {
     }
 
     Ok(())
+}
+
+fn with_ghost_constraint_example(mut err: syn::Error) -> syn::Error {
+    err.combine(error(err.span(), "expected a trait bound and specifications in brackets, e.g.: `ghost_constraint(T: A + B + ..., [requires(...), ...])`"));
+    err
 }
 
 /// A specification enclosed in another specification (e.g. in spec entailments or ghost constraints)
--- a/prusti-specs/src/specifications/preparser.rs
+++ b/prusti-specs/src/specifications/preparser.rs
@@ -413,7 +413,7 @@ impl PrustiTokenStream {
                 other => error(self.source_span, format!("unexpected nested spec type: {}", other).as_ref()),
             }
         } else {
-            error(self.source_span, "expected identifier").into()
+            error(self.source_span, "expected identifier")
         }
     }
 
--- a/prusti-specs/src/specifications/preparser.rs
+++ b/prusti-specs/src/specifications/preparser.rs
@@ -171,7 +171,11 @@ impl PrustiTokenStream {
     {
         let result = f(&mut self)?;
         if !self.is_empty() {
-            let error_span = self.tokens.front().unwrap().span().join(self.tokens.back().unwrap().span()).unwrap();
+            let start = self.tokens.front().unwrap().span();
+            let end = self.tokens.back().unwrap().span();
+            let span = start.join(end);
+            // this is None if the spans are not combinable, which seems to happen when running the cfg(tests) via cargo
+            let error_span = span.unwrap_or(start);
             return error(error_span, "unexpected extra tokens");
         }
         Ok(result)
--- a/prusti-specs/src/specifications/preparser.rs
+++ b/prusti-specs/src/specifications/preparser.rs
@@ -486,17 +486,6 @@ impl PrustiTokenStream {
     }
 }
 
-trait AsPrustiTokenStream {
-    fn as_pts(&self) -> PrustiTokenStream;
-}
-
-// bridge between syn parse streams and prusti token streams
-impl AsPrustiTokenStream for ParseStream<'_> {
-    fn as_pts(&self) -> PrustiTokenStream {
-        PrustiTokenStream::new(self.parse().unwrap())
-    }
-}
-
 // TODO: is there a better place for this type and its logic?
 
 #[derive(Debug)]
@@ -511,7 +500,8 @@ impl Parse for GhostConstraint {
         Ok(GhostConstraint {
             trait_bounds: parse_trait_bounds(input)?,
             comma: input.parse()?,
-            specs: input.as_pts().parse_rest(|pts| pts.pop_group_of_nested_specs(input.span()))?,
+            specs: PrustiTokenStream::new(input.parse().unwrap())
+                .parse_rest(|pts| pts.pop_group_of_nested_specs(input.span()))?,
         })
     }
 }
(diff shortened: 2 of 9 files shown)


juliand665:
@Aurel300 for your consideration :)

prusti-specs/src/lib.rs
```diff
@@ -238,6 +239,10 @@ fn generate_for_pure(attr: TokenStream, item: &untyped::AnyFnItem) -> GeneratedR
         ));
     }
 
+    do_generate_for_pure(item)
+}
+
+fn do_generate_for_pure(item: &untyped::AnyFnItem) -> GeneratedResult {
```
Aurel300, line 249: This is not an amazing name, perhaps add a one-line comment explaining that this one is usable from other parts of `prusti-specs` and that it doesn't receive or check the attribute tokens?

prusti-specs/src/specifications/preparser.rs
```diff
@@ -81,6 +80,7 @@ fn error<T>(span: Span, msg: &str) -> syn::Result<T> {
 struct PrustiTokenStream {
     tokens: VecDeque<PrustiToken>,
     source_span: Span,
+    // TODO: can we somehow update the span after popping stuff?
```
Aurel300, line 88: Would this get us more precise error reporting? I think the span here is for cases where there are no tokens but we still want a span (of the enclosing group, for example).

prusti-specs/src/specifications/preparser.rs
```diff
@@ -161,6 +161,17 @@ impl PrustiTokenStream {
         self.tokens.is_empty()
     }
 
+    fn parse_rest<T, F>(mut self, f: F) -> syn::Result<T>
+    where
+        F: FnOnce(&mut Self) -> syn::Result<T>,
+    {
+        let result = f(&mut self)?;
+        if !self.is_empty() {
+            return error(self.source_span, "unexpected extra tokens");
```
Aurel300: You can use the span of the first token, since you know there is at least one left.

prusti-specs/src/lib.rs
juliand665: Good catch! In any other language, I would have just overloaded it, but coming up with a meaningful name is hard. Maybe we could incorporate the word `unchecked` somehow?

prusti-specs/src/specifications/preparser.rs
juliand665: For example when you use the new `parse_rest` function and it encounters additional tokens past what it needed to parse, I was expecting the span to just cover those unconsumed tokens at that point, and I feel like that would be generally useful. For cases where there are no tokens, I'm not sure the span of "everything parsed as prusti tokens" is more useful than just `Span::call_site()` and such.

prusti-specs/src/specifications/preparser.rs
```diff
@@ -410,32 +391,29 @@ impl PrustiTokenStream {
...
-                    Some(NestedSpec::Ensures(group.clone()))
-                } else {
-                    None
-                }
+    }
+    
+    fn pop_single_nested_spec(&mut self) -> syn::Result<NestedSpec<Self>> {
+        let first = self.tokens.pop_front().map_or_else(|| {
+            error(self.source_span, "expected nested spec")
+        }, Ok)?;
```
Aurel300: A version of the `error` helper method which only returns the `syn::parse::Error` might be nice. Then this can be an `ok_or_else`, and the errors in `expr_bp` would be a bit nicer.

prusti-specs/src/specifications/preparser.rs
```diff
@@ -508,18 +485,87 @@ impl PrustiTokenStream {
...
+trait AsPrustiTokenStream {
+    fn as_pts(&self) -> PrustiTokenStream;
+}
+
+// bridge between syn parse streams and prusti token streams
+impl AsPrustiTokenStream for ParseStream<'_> {
+    fn as_pts(&self) -> PrustiTokenStream {
+        PrustiTokenStream::new(self.parse().unwrap())
+    }
+}
```
Aurel300: This is a trait just to be able to call it as `.as_pts()`?

prusti-specs/src/specifications/preparser.rs
```diff
@@ -508,18 +485,87 @@ impl PrustiTokenStream {
...
+
+// TODO: is there a better place for this type and its logic?
```
Aurel300: Let's keep it here for now. We don't have so many spec syntax components yet.

prusti-specs/src/specifications/preparser.rs
```diff
@@ -907,46 +953,64 @@ mod tests {
...
+            assert_matches!(&constraint.specs[2], NestedSpec::Pure);
+            assert_eq!(constraint.specs.len(), 3);
         }
 
         #[test]
         fn no_specs() {
-            let (bounds, nested_specs) = parse_ghost_constraint(quote!{ T: A, []}).unwrap();
-            assert_eq!(bounds.to_string(), "T : A");
-            assert!(nested_specs.is_empty());
+            // TODO: does it even make sense to compare parsed quotes like this? feels a bit like saying 2 = 2
```
Aurel300: This test is very simple, so the comparison of the trait bound isn't terribly interesting. For the others it makes more sense, and this way it is consistent.

prusti-specs/src/specifications/preparser.rs
```diff
@@ -907,46 +953,64 @@ mod tests {
...
         fn fully_qualified_trait_path() {
-            let (bounds, _) = parse_ghost_constraint(quote!{ T: path::to::A, [requires(true)]}).unwrap();
-            assert_eq!(bounds.to_string(), "T : path :: to :: A");
+            let constraint = parse_ghost_constraint(quote!{ T: path::to::A, [requires(true)]}).unwrap();
+            assert_bounds_eq(constraint.trait_bounds, quote!{ T : path :: to :: A });
+        }
+        
+        #[test]
+        fn tuple_generics() {
+            // just check that parsing succeeds
```
Aurel300, line 1007: Why not `assert!(... .is_ok());`?

prusti-tests/tests/parse/fail/ghost-constraints/trait-bounds-invalid-macro-arguments-1.rs
```diff
@@ -1,6 +1,6 @@
 use prusti_contracts::*;
 
-#[ghost_constraint([ //~ ERROR: Invalid use of macro. Two arguments expected (a trait bound `T: A + B` and multiple specifications `[requires(...), ensures(...), ...]`)
+#[ghost_constraint([ //~ ERROR: expected `,`
```
Aurel300: Yeah this error is not great. Ideally we would show an expected form for the attribute with any error emitted, but I don't see a way to attach "notes" to `syn` errors.

prusti-specs/src/specifications/preparser.rs
juliand665: Good idea! Not sure why I didn't think of that.

prusti-specs/src/lib.rs
juliand665: Actually, what if we get rid of this and instead call `generate_spec_and_assertions` directly from `ghost_constraints::generate`? We'd could either remap `NestedSpec` to `SpecAttributeKind` or directly parse as the latter to get rid of the current (minor) duplicate parsing logic, though this would affect `|=` as well…

prusti-tests/tests/parse/fail/ghost-constraints/trait-bounds-illegal-2.rs
```diff
@@ -2,9 +2,10 @@ use prusti_contracts::*;
 
 trait A<X> { }
 
-#[ghost_constraint(T: for<'a> A<&'a i32> , [ //~ ERROR: lifetimes are not allowed in ghost constraint trait bounds
+#[ghost_constraint(T: for<'a> A<&'a i32>, [ //~ ERROR: lifetimes are not allowed in ghost constraint trait bounds
     ensures(result > 0)
 ])]
+//~| ERROR: expected a trait bound `T: A + B` and specifications in brackets `[requires(...), ensures(...), pure, ...]`
```
Aurel300: Here the extra error is maybe unnecessary, but we would have to check (by string comparison?) that the emitted error is not just a generic `syn` parse error, right?

prusti-specs/src/specifications/preparser.rs
```diff
@@ -51,7 +51,11 @@ pub fn parse_prusti_assert_pledge(tokens: TokenStream) -> syn::Result<(TokenStre
 }
 
 pub fn parse_ghost_constraint(tokens: TokenStream) -> syn::Result<GhostConstraint> {
-    syn::parse2(tokens)
+    syn::parse2(tokens).map_err(|mut err| {
+        err.combine(syn::Error::new(err.span(),
+            "expected a trait bound `T: A + B` and specifications in brackets `[requires(...), ensures(...), pure, ...]`"));
```
Aurel300: This is better, but still doesn't give an actual example of the syntax. I think the error message should contain the expected format `ghost_constraint(T: A + B + ..., [requires(...), ...])`

prusti-tests/tests/parse/fail/ghost-constraints/trait-bounds-illegal-2.rs
juliand665: I have to specify all errors that the test expects to generate; any additional or missing ones will cause it to fail. The way syn outputs these combined errors is as two separate errors with the same span (closest we can get to a note).

Aurel300: I understand that but here our error is specifically that we disallow lifetimes, which is an error we emit ourselves, as opposed to a generic `syn` parsing error.

juliand665: Ah, my bad, I see what you mean now. Yeah, it's kind of silly to attach an error like this in that case. It doesn't do much harm, so I wonder if it justifies the additional complexity of moving down the error to more specific parsing logic—I'll push a commit that does that so we can compare.

prusti-specs/src/specifications/preparser.rs
```diff
@@ -410,44 +401,40 @@ impl PrustiTokenStream {
...
-        if self.tokens.is_empty() {
-            return None;
-        }
-        if self.tokens.len() != 2 {
-            panic!();
+    
+    fn pop_parenthesized_group(&mut self) -> syn::Result<Self> {
+        match self.tokens.pop_front() {
+            Some(PrustiToken::Group(_span, Delimiter::Parenthesis, box group)) => {
+                Ok(group) // TODO: need to clone()?
```
Aurel300, line 408: What is this TODO about?

juliand665: Not sure; that one was there before I touched the code! I would also expect it to be clear whether or not we need to clone from a type safety standpoint, but I don't know enough about syn to be sure.

zgrannan:
Finished benchmarking commit (05b74c27cd49dfb308c33dc9cf363f16dbb4de5b): [comparison url](http://3.94.193.1:2346/compare.html?start=9054e05d9fc5bcfbf0576c40c12dfb31a8b3703c&end=05b74c27cd49dfb308c33dc9cf363f16dbb4de5b&stat=wall-time).

Benchmark returns the average of 3 consecutive runs of the benchmark suite, performed after an initial 'warm-up' run. An external Prusti server is maintained over the 4 runs.
         
### [Wall time](http://3.94.193.1:2346/compare.html?start=9054e05d9fc5bcfbf0576c40c12dfb31a8b3703c&end=05b74c27cd49dfb308c33dc9cf363f16dbb4de5b&stat=wall-time)
This benchmark run did not return any relevant results for this metric.

### [Instruction count (excludes Viper)](http://3.94.193.1:2346/compare.html?start=9054e05d9fc5bcfbf0576c40c12dfb31a8b3703c&end=05b74c27cd49dfb308c33dc9cf363f16dbb4de5b&stat=instructions:u)
This benchmark run did not return any relevant results for this metric.

### [Cycles (excludes Viper)](http://3.94.193.1:2346/compare.html?start=9054e05d9fc5bcfbf0576c40c12dfb31a8b3703c&end=05b74c27cd49dfb308c33dc9cf363f16dbb4de5b&stat=cycles:u)
This benchmark run did not return any relevant results for this metric.

If you disagree with this performance assessment, please file an issue in [prusti-dev/prusti-perf](https://github.com/prusti-dev/prusti-perf/issues/new).

@rustbot label: -perf-regression

PR #1141, opened 2022-08-26
Make rust-analyzer use x.py

juliand665:
[rust-analyzer](https://rust-analyzer.github.io) is a very popular extension for VS Code providing support for Rust. In my setup, however, it failed on the Prusti codebase because it uses standard `cargo` rather than `x.py`. This PR amends the workspace settings file that's already been committed to the repository to configure rust-analyzer appropriately.

The configured values may look non-trivial, but they are in fact exact copies of the default command as specified in their respective setting's comment, e.g. "By default, a cargo invocation will be constructed for the configured targets and features, with the following base command line: `cargo check --quiet --workspace --message-format=json --all-targets`", except of course with `cargo` replaced by `./x.py`.

If any of the authors currently use VS Code, I'd love to get your feedback on this!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,8 @@
 {
     "rust-analyzer.rustc.source": "discover",
     "rust-analyzer.cargo.buildScripts.overrideCommand": [
-        "./x.py",
+        "python3",
+        "x.py",
         "check",
         "--quiet",
         "--workspace",
@@ -9,7 +10,8 @@
         "--all-targets",
     ],
     "rust-analyzer.checkOnSave.overrideCommand": [
-        "./x.py",
+        "python3",
+        "x.py",
         "check",
         "--workspace",
         "--message-format=json",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,18 @@
 {
-    "rust-analyzer.rustc.source": "discover"
+    "rust-analyzer.rustc.source": "discover",
+    "rust-analyzer.cargo.buildScripts.overrideCommand": [
+        "./x.py",
+        "check",
+        "--quiet",
+        "--workspace",
+        "--message-format=json",
+        "--all-targets",
+    ],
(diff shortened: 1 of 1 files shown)


vakaras:
What are the benefits/differences from using `./x.py ide .` to start VS Code?

juliand665:
> What are the benefits/differences from using `./x.py ide .` to start VS Code?

Oh, I had no idea that was an option! This setting shouldn't break it though, and would let you open the workspace however you see fit, so I believe it's a strict improvement.

I usually open the workspace either through macOS restoring the window (after having quit VS Code) or by picking it out of my recents, both of which don't restore the environment `x.py` sets when it launches Code. Always heading into a terminal first and launching Code from there isn't part of my workflow and I'd like to avoid it (even if it's only a minor inconvenience). The only thing the command improves for me is that the environment is set appropriately when using the built-in terminal in VS Code, so in there I can just write `cargo build` rather than `./x.py build`.

With this change, I believe there should be "no wrong way" to open our repo in VS Code. Whichever option you use, including `x.py ide`, rust-analyzer will work.

.vscode/settings.json
```diff
@@ -1,3 +1,18 @@
 {
-    "rust-analyzer.rustc.source": "discover"
+    "rust-analyzer.rustc.source": "discover",
+    "rust-analyzer.cargo.buildScripts.overrideCommand": [
+        "./x.py",
```
vakaras: Will this format work on Windows? Wouldn't it be better to use `python3 x.py`?

juliand665: Great point! I should be more mindful of these things.

vakaras (approved):
LGTM Thanks!

Aurel300:
bors retry

juliand665:
Funky, did CI break?

Aurel300:
Yeah, and in two different ways. This shouldn't affect builds anyway.

PR #1144, opened 2022-09-01
Action to publish `prusti-contracts` to crates.io

### LINKED ISSUE #529: Publish prusti-contracts on crates.io

JonasAlaif:
Fixes #529
- [x] Reorganize `prusti-contracts` workspace


.github/workflows/crates-io.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
```
Aurel300: Why? Python 3 [is pre-installed](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#language-and-runtime) on `ubuntu-*` images.

.github/workflows/crates-io.yml
```diff
@@ -0,0 +1,29 @@
+name: Publish prusti-contracts to Crates.io
```
fpoli:
```suggestion
name: Publish on Crates.io
```

Long workflow names are not fully visible in https://github.com/viperproject/prusti-dev/actions.

.github/workflows/docs-test.yml
```diff
@@ -0,0 +1,27 @@
+name: Check GitHub pages
```
fpoli: What about merging this with `docs.yml`? We can skip the deploy job in PRs with `if: github.event_name == 'push'`.

PR #1146, opened 2022-09-06
Lift requirement of having the `prusti` feature enabled

JonasAlaif:
Having `prusti-contracts = { path = "path/to/prusti-contracts", features = ["prusti"] }` as a dependency breaks regular `cargo build` since the Prusti procedural macros generate `#[prusti::attr = "..."]` attributes and these require `register_tool(prusti)`. One solution would be to require users to have this in their Rust source:
```rust
#[feature(register_tool)]
#[register_tool(prusti)]
```
but it still remains a bit impractical to keep switching the `prusti` feature on and off between prusti verification and regular release builds.

Instead I get Prusti to enable the feature when it runs cargo. This however comes with a caveat that cargo will report:
```
error: none of the selected packages contains these features: prusti-contracts/prusti
```
in the case that `prusti-contracts` isn't a dependency of the crate currently being verified. ~~The only way to catch specifically this error and give the user a hint on how to fix this is to capture the stderr of the running cargo, but we still want all the output cargo spits out printed to the terminal, so we `tee` the stdout and stderr.~~

In the future we can also probably use the cargo config `patch.crates-io` for `prusti-contracts` to ensure that we get the version Prusti was compiled with, e.g. using the commit hash


Aurel300:
Thanks! ~~The stderr tee looks a bit hacky, I think it would have been fine to keep the error message from `cargo` and explain what it means in one of our guides. We'll see over time how stable this is.~~ (tee hack removed)

docs/dev-guide/src/config/providing.md
```diff
@@ -10,3 +10,31 @@ Flags can be set in one of four ways, in increasing order of priority:
...
 
 4. Provided individually as command-line arguments to Prusti with the prefix `-P` (for example, `-Pprint_desugared_specs` for the [`PRINT_DESUGARED_SPECS`](flags.md#print_desugared_specs) flag).
+
+## Multi-crate Cargo Prusti Projects
+
+Setting flags becomes slightly more complicated when Prusti is run on multiple crates as `cargo prusti` (e.g. which `Prusti.toml` file will be used). The three possible approaches to providing flags all behave differently.
+
+### Environment Variables
+
+The environment persists throughout compilation, therefore any flags set through the `DEFAULT_PRUSTI_` or `PRUSTI_` will apply to both the root crate and all dependencies. For example setting `CHECK_OVERFLOWS=false` will disable overflow checks in all dependencies, even if they try to override this with a `Prusti.toml` file.
```
fpoli: > `CHECK_OVERFLOWS=false`

Should be `PRUSTI_CHECK_OVERFLOWS=false`?

> setting `CHECK_OVERFLOWS=false` will disable overflow checks in all dependencies, even if they try to override this with a `Prusti.toml` file.

In a single crate `Prusti.toml` doesn't override `PRUSTI_*` environment variables. Is the quoted sentence correct?

docs/dev-guide/src/config/providing.md
```diff
@@ -10,3 +10,31 @@ Flags can be set in one of four ways, in increasing order of priority:
...
+
+### Prusti Config File
+
+A Prusti config file is loaded prior to running cargo from `PRUSTI_CONFIG` (see default above). This file is used to configure flags in Categories B and C only and these flags apply to the entire compilation.
```
fpoli: > A Prusti config file is loaded prior to running cargo from `PRUSTI_CONFIG` (see default above).

The use case is now that each crate can specify its `Prusti.toml`, right? I'd be happy to remove the `PRUSTI_CONFIG` and hardcode `"Prusti.toml"`. That should simplify the explanation. I think we originally considered having `Prusti.toml` files stored in e.g. the home directory, so configuring the path made sense in that use case, but we never used it.

fpoli: > This file is used to configure flags in Categories B and C only

What are these categories?

docs/dev-guide/src/config/providing.md
```diff
@@ -10,3 +10,31 @@ Flags can be set in one of four ways, in increasing order of priority:
...
+
+Then, as each individual crate is checked (from the roots of the dependency tree) a Prusti config file is loaded from the current working directory (i.e. `./Prusti.toml`). This file is used to configure flags in Category A only — it would not make sense to set e.g. [`NO_VERIFY_DEPS`](flags.md#no_verify_deps) in the `./Prusti.toml` of a dependency. Cargo sets the working directory for each crate as follows:
```
fpoli, line 26: > from the current working directory

Do you mean the directory of the crate of the outermost `cargo prusti` invocation or the directories of each of the dependencies?

docs/dev-guide/src/config/providing.md
```diff
@@ -10,3 +10,31 @@ Flags can be set in one of four ways, in increasing order of priority:
...
+
+1. If the current crate is the (final) root crate, then use the directory from which `cargo prusti` was launched. Note: this case will never apply for workspace projects.
+
+2. If the current crate is in a subdirectory of the root crate/workspace, then use the directory from which `cargo prusti` was launched. Note: this means that any `Prusti.toml` in this subdirectory will be ignored!
```
fpoli: This almost looks like a bug of Prusti. I wonder if we can move the settings to a `Cargo.toml` and then reuse the toml-discovery of cargo instead of implementing it on our own. Vytautas mentioned that rust-analyzer already does that. We could open an issue to implement it in another PR.

docs/dev-guide/src/config/providing.md
```diff
@@ -10,3 +10,31 @@ Flags can be set in one of four ways, in increasing order of priority:
...
+
+3. Else (the crate lives in an unrelated directory or was downloaded from git/crates.io to such a directory) use the directory of this dependency crate. Therefore a `Prusti.toml` placed next to the `Cargo.toml` of such a crate will load flags as expected.
+
+When publishing a crate to git/crates.io one can configure Category A flags by including a `Prusti.toml` file in the root.
+
+### Commandline Arguments
```
fpoli, line 37: What is the override order between this and the two sections above?

prusti-contracts-build/build.rs
```diff
@@ -18,88 +16,57 @@ fn main() {
...
-    // Build `prusti-contracts`
+    // In theory we should build to here (i.e. set `CARGO_TARGET_DIR` to this),
+    // but this is hard to find for linking. So instead build to the `prusti-contracts` dir.
+    let _out_dir = std::env::var("OUT_DIR").unwrap();
+    // println!("cargo:warning=out_dir: {}", _out_dir);
+
+    let target = prusti_contracts.join("target").join("verify");
+    force_reexport_specs(target.as_path());
+
+    // Copy just-built binaries to `target/dir` dir
```
fpoli, line 22: Why is it necessary to copy these files?

prusti-contracts-build/build.rs
```diff
@@ -18,88 +16,57 @@ fn main() {
         }
     }
 
-    // Build `prusti-contracts`
+    // In theory we should build to here (i.e. set `CARGO_TARGET_DIR` to this),
+    // but this is hard to find for linking. So instead build to the `prusti-contracts` dir.
+    let _out_dir = std::env::var("OUT_DIR").unwrap();
+    // println!("cargo:warning=out_dir: {}", _out_dir);
```
fpoli: Do you know if this is still necessary, since the print is commented?

prusti/src/callbacks.rs
```diff
@@ -21,14 +21,19 @@ pub struct PrustiCompilerCalls;
...
-        tcx,
-        ty::WithOptConstParam::unknown(def_id),
-    );
-    // SAFETY: This is safe because we are feeding in the same `tcx` that is
-    // going to be used as a witness when pulling out the data.
-    unsafe {
-        mir_storage::store_mir_body(tcx, def_id, body_with_facts);
+    // Running `get_body_with_borrowck_facts` can be very slow, therefore we avoid it when not necessary:
+    // for crates which won't be verified or spec_fns it suffices to load just the fn body
+    if !config::no_verify() && !is_spec_fn(tcx, def_id.to_def_id()) {
```
fpoli: Have you considered moving the `if !config::no_verify()` condition around the `config.override_queries = Some(override_queries);` statement?

prusti/src/driver.rs
```diff
@@ -175,7 +192,12 @@ fn main() {
         user::message(format!("Prusti version: {}", get_prusti_version_info()));
         info!("Prusti version: {}", get_prusti_version_info());
 
-        rustc_args.push("-Zpolonius".to_owned());
+        // TODO: decide if we want to keep this enabled; it means that we accept rust programs that
+        // rustc wouldn't. For example: prusti-tests/tests/verify/pass/nll-rfc/borrow_first.rs
+        if !config::no_verify() {
+            // Should no longer be needed since `get_body_with_borrowck_facts` doesn't require this
```
fpoli: When Vytatuas implemented `get_body_with_borrowck_facts` the `-Zpolonius` was needed. We can check if this is no longer true, but I think we still need the flag.

fpoli (approved):
See the comments

docs/dev-guide/src/config/providing.md
fpoli: The rust-analyzer code seems to be [here](https://github.com/rust-lang/rust-analyzer/blob/master/crates/project-model/src/lib.rs). I'm not sure how easy is to depend on it.

prusti-contracts-build/build.rs
JonasAlaif: Otherwise they don't live in the same folder and can't find each other

JonasAlaif:
bors retry

PR #1153, opened 2022-09-16
Quantifier optimization fix for conditionals

zgrannan:
The quantifier optimization that moves expressions outside quantifiers by assigning them to let expressions can cause verification failures.

For example, the expression:

`forall y : Int :: (x >= 0 ? sqrt(x) + 1 : y) > 0`

would be optimized to:

`let temp == (sqrt(x) + 1) in forall y : Int :: (x >= 0 ? temp : y) > 0`

Viper will indicate that `sqrt(x)` is not well-formed in the optimized version, because it isn't known that `x >= 0`. 

This PR fixes the above issue by preventing extraction of expressions inside conditional branches. Also, this PR modifiers the behavior of the optimization, such that it will extract the entire conditional expression into a let, provided that the expression does not reference any quantified variables.


fpoli:
(Feel free to merge yourself or to choose a reviewer.)

PR #1155, opened 2022-09-19
Delete backlog.yml

fpoli:
The experiment failed; there were more open than closed PRs.


JonasAlaif:
The main issue for me was that I just couldn't run test crates locally since it only works on Linux

PR #1159, opened 2022-09-20
Remove the verbose flag from cargo builds

fpoli:
The `--verbose` flag of `cargo build` makes the CI logs very long. I think we never actually needed that flag.


fpoli:
bors cancel

PR #1160, opened 2022-09-20
Print Prusti version, clarify error message

fpoli:
I think it was not clear from where the minimum version requirement of prusti comes from. (Let me know if that's wrong.)

### DIFF
--- a/prusti-interface/src/specs/checker/version_checks.rs
+++ b/prusti-interface/src/specs/checker/version_checks.rs
@@ -21,7 +21,8 @@ impl<'tcx> SpecCheckerStrategy<'tcx> for MismatchedVersionsChecker {
             if env.compare_prusti_version(&min_prusti_version).is_lt() {
                 check_version.errors.push(PrustiError::incorrect(
                     format!(
-                        "Running Prusti version '{}' when the min required version is '{min_prusti_version}'! \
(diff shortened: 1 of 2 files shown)


prusti-interface/src/specs/checker/version_checks.rs
```diff
@@ -21,7 +21,8 @@ impl<'tcx> SpecCheckerStrategy<'tcx> for MismatchedVersionsChecker {
             if env.compare_prusti_version(&min_prusti_version).is_lt() {
                 check_version.errors.push(PrustiError::incorrect(
                     format!(
-                        "Running Prusti version '{}' when the min required version is '{min_prusti_version}'! \
+                        "Running Prusti version '{}' when the minimum version required by \
+                        `prusti-contracts`/`prusti-specs` is '{min_prusti_version}'! \
```
fpoli: @JonasAlaif if this change is consistent with the use cases that you looked at, could you `bors r+` the PR?

JonasAlaif: So the thing is that `min_prusti_version` is a config flag - it happens to be set by `prusti-contracts` in a `Prusti.toml` but could be set by any other crate. This error message is too specific

JonasAlaif (dismissed):

PR #1175, opened 2022-09-29
Use debug=1 instead of debug=true

fpoli:
This way the debug binaries are much smaller while still having stacktraces.

For example, on Ubuntu the `target/debug/prusti-driver` binary goes from 1.4 GB to "just" 300 MB. Most of the size was due to a 700 MB `.debug_pubtypes` section. There is a related rustc issue: https://github.com/rust-lang/rust/issues/48762.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,9 @@ members = [
     "jni-gen/systest",
 ]
 
+[profile.dev]
+debug = 1
+
 [profile.release]
 # Enable after fixing https://github.com/viperproject/prusti-dev/issues/383
 # lto = true


fpoli:
On my Ubuntu 22.04 with 16 GB of RAM and a SSD:
* Running `./x.py build` with `debug=1` takes **5m30s** and produces a 3.9 GB `target` folder.
* Running `./x.py build` with `debug=true` takes **23m** and produces a 13 GB `target` folder. In the meantime the computer is laggy, many CPU cores are blocked on IO, RAM usage is high and the OS is swapping.

PR #1181, opened 2022-10-03
Force nightly Rust toolchain in cargo-prusti

fpoli:
Procedural macros behave differently when compiled on stable compared to nightly ([documentation](https://docs.rs/proc-macro2/latest/proc_macro2/struct.Span.html#method.join)). This PR forces `cargo-prusti` to use the same nightly version that we use to compile Prusti.


JonasAlaif:
Thanks, that’s perfect!

PR #1188, opened 2022-10-08
Extern Spec Improvements

juliand665:
As part of [my thesis](https://ethz.ch/content/dam/ethz/special-interest/infk/chair-program-method/pm/documents/Education/Theses/Julian_Dunskus_MS_Description.pdf) under @Aurel300, I have been working with `#[extern_spec]` a lot and had to make some improvements, mostly to that but also to other things:

- `predicate!` functions in traits were forbidden from having bodies defined right in the trait, providing a default for implementations like other trait members. It seems this restriction was unnecessarily strict, and simply removing the check allows this to work as desired.
- `#[extern_spec]` and `#[refine_trait_spec]` parsing required the involved type to be a simple type path. It now supports any combination of paths, references, slices, etc.—I may have missed some, but they would be very simple to add. The reason we can't just support all types at once is that we need to extract the "main" type path for rewriting.
- The dummy bodies generated for `#[extern_spec]` methods forwarded to the original function and then returned an `unimplemented!()`, presumably to make clear that these bodies are dummies. Returning the forwarded result, however, enables type checking against the original method, detecting e.g. differing `where` constraints.
- Guards in `match` patterns/branches caused Prusti to crash due to an assertion about the involved states in their MIR representation, which was just a sanity check. Removing this assertion makes everything work as it should.
- Generic parameters are now forwarded explicitly, rather than relying on argument types to implicitly express them. This broke e.g. `core::mem::size_of::<T>()`

There are currently no regression tests or such for these changes; let me know if I should look add some!


prusti-contracts/prusti-specs/src/common.rs
```diff
@@ -462,6 +462,18 @@ pub(crate) fn add_phantom_data_for_generic_params(item_struct: &mut syn::ItemStr
     }
 }
 
+/// Extracts the main/self type path from a type (e.g. `Foo` from `Foo`, `&Foo`, `[Foo]`, etc.)
```
Aurel300: Hm, unsure I like this. This is a lossy conversion: is it used only to declare trait bounds or can it result in accidentally declaring the wrong extern spec?

prusti-interface/src/specs/checker/predicate_checks.rs
```diff
@@ -28,17 +28,6 @@ impl<'tcx> SpecCheckerStrategy<'tcx> for IllegalPredicateUsagesChecker {
...
-        let body_errors = collected_predicates
-            .abstract_predicate_with_bodies
-            .into_iter()
-            .map(|def_id| {
-                let span = env.query.get_def_span(def_id);
-                PrustiError::incorrect(
-                    "abstract predicates must not have bodies".to_string(),
-                    MultiSpan::from_span(span),
-                )
-            });
```
Aurel300, line 40: Ah, we forbid this originally because behavioural subtyping with predicates is not checked. Maybe we should put a TODO here at least.

prusti-contracts/prusti-specs/src/common.rs
juliand665: I agree that this has a bad smell, but I believe it's just used for rewriting `Self` and associated types to the original type/trait in `#[extern_spec]` and `#[refine_trait_spec]`, where the syntax is traversed until reaching a type path node and only that is replaced. Maybe I should add a note on what it's useful for?

prusti-interface/src/specs/checker/predicate_checks.rs
juliand665: Are you saying abstract predicate implementations should ideally be checked for behavioral subtyping against the default implementation (if present)? I'm not sure how that would work, since the implication would be flipped depending on whether the predicate is used in a pre- or postcondition, for example, right?

prusti-contracts/prusti-specs/src/common.rs
juliand665: Actually, on second look, the suspicion was completely right. I think it would have been an issue and I just didn't think to test a spec on a non-type-path item involving a `Self` type anywhere 🤦
Addressed in the commit I just pushed; `extract_type_path` no longer exists!

PR #1201, opened 2022-10-17
Fix loop encoding when there are no user-defined invariants

### LINKED ISSUE #1200: Internal error: unresolved CFG edges
I encountered an unexpected internal error on the following:

```rust
use std::cmp::min;

/// Find a fft length larger that is at least `minimum_size` in length that is made up of a limited set of factors.
/// The `once_factor` will be used exactly once, the multi factors may be used any number of times to reach the minimum.
fn fast_fft_len(
    minimum_len: usize,
    once_factor: usize,
    multi_factor1: usize,
    multi_factor2: usize,
) -> usize {
    assert!(once_factor >= 1);
    assert!(multi_factor1 > 1);
    assert!(multi_factor2 > 1);

    // Apply once factor
    let mut product = once_factor;

    // apply second factor until at or above minimum
    while product < minimum_len {
        product *= multi_factor2
    }

    // remove second factor one at a time then add enough first factor to reach minimum_size
    // repeat while tracking lowest viable product
    let mut best = product;
    loop {
        match product.cmp(&minimum_len) {
            std::cmp::Ordering::Less => product *= multi_factor1,
            std::cmp::Ordering::Equal => return product,
            std::cmp::Ordering::Greater => {
                best = min(best, product);
                if product % multi_factor2 != 0 {
                    return best;
                }
                product /= multi_factor2;
            }
        }
    }
}
```

fpoli:
Fixes #1200.

When there are no user-defined loop invariants, Prusti was placing a `body_invariant!(true)` after the first exit block (MIR block with an edge that exits the loop). In doing so it was incorrectly assuming that such exit block is the evaluation of a boolean, but #1200 shows that it can actually be the evaluation of a `match`. This PR adds a further condition that checks the number of exit edges.

### DIFF
--- a/prusti-interface/src/environment/loops.rs
+++ b/prusti-interface/src/environment/loops.rs
@@ -114,7 +114,9 @@ impl<'b, 'tcx> Visitor<'tcx> for AccessCollector<'b, 'tcx> {
                 NonMutatingUse(mir::visit::NonMutatingUseContext::Copy) => PlaceAccessKind::Read,
                 NonMutatingUse(mir::visit::NonMutatingUseContext::Move) => PlaceAccessKind::Move,
                 NonMutatingUse(mir::visit::NonMutatingUseContext::Inspect) => PlaceAccessKind::Read,
-                NonMutatingUse(mir::visit::NonMutatingUseContext::AddressOf) => PlaceAccessKind::Read,
+                NonMutatingUse(mir::visit::NonMutatingUseContext::AddressOf) => {
+                    PlaceAccessKind::Read
+                }
                 NonMutatingUse(mir::visit::NonMutatingUseContext::SharedBorrow) => {
                     PlaceAccessKind::SharedBorrow
                 }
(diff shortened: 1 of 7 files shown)


JonasAlaif:
Looks good - so the reason we don't support all [PlaceContext](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/visit/enum.PlaceContext.html) is that we haven't figured out how to encode the permissions for all of them?

JonasAlaif (approved):

fpoli:
> Looks good - so the reason we don't support all [PlaceContext](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/visit/enum.PlaceContext.html) is that we haven't figured out how to encode the permissions for all of them?

I didn't thought too much about it. I suspect almost all remaining `NonMutatingUseContext` can be mapped to a read permission, but I'd need to check what Rust code might produce each `PlaceContext`.

PR #1225, opened 2022-11-01
Disable `rustc_smir` while RA autocomplete is broken for it

JonasAlaif:


vakaras (commented):
LGTM However, I would suggest following up on this issue on Zulip to make sure that it is eventually fixed.

PR #1233, opened 2022-11-08
Add sanity check when decoding extern specs

JonasAlaif:
bors r+

### DIFF
--- a/prusti-interface/src/specs/decoder.rs
+++ b/prusti-interface/src/specs/decoder.rs
@@ -32,7 +32,7 @@ impl<'a, 'tcx> DefSpecsDecoder<'a, 'tcx> {
             cstore.stable_crate_id_to_crate_num(hash.stable_crate_id())
         });
         if result.is_err() {
-            panic!("A compiled dependency is out of sync. Try deleting the target folder, or that of `prusti-contracts`.")
+            panic!("A compiled dependency is out of sync. Try deleting the target folder (`cargo clean`).")
         }
         // Get `DefId`
         self.tcx.def_path_hash_to_def_id(hash, &mut || {
--- a/prusti-interface/src/specs/decoder.rs
+++ b/prusti-interface/src/specs/decoder.rs
@@ -25,8 +25,15 @@ impl<'a, 'tcx> DefSpecsDecoder<'a, 'tcx> {
         }
     }
 
-    // From rustc
     fn def_path_hash_to_def_id(&self, hash: DefPathHash) -> DefId {
+        // Sanity check
+        let cstore = std::panic::AssertUnwindSafe(self.tcx.cstore_untracked());
+        if let Err(_) =
+            std::panic::catch_unwind(|| cstore.stable_crate_id_to_crate_num(hash.stable_crate_id()))
+        {
+            panic!("A compiled dependency is out of sync. Try deleting the target folder, or that of `prusti-contracts`.")
+        }
+        // Get `DefId`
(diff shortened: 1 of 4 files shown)


JonasAlaif:
@fpoli could you check that I updated the required things to move the prebuilt `prusti-contracts/target/verify/...` to `target/verify/...` (esp. for deploying for prusti-assistant)

prusti-interface/src/specs/decoder.rs
```diff
@@ -25,8 +25,16 @@ impl<'a, 'tcx> DefSpecsDecoder<'a, 'tcx> {
...
 
-    // From rustc
     fn def_path_hash_to_def_id(&self, hash: DefPathHash) -> DefId {
+        // Sanity check
+        let cstore = std::panic::AssertUnwindSafe(self.tcx.cstore_untracked());
+        let result = std::panic::catch_unwind(|| {
+            cstore.stable_crate_id_to_crate_num(hash.stable_crate_id())
+        });
+        if result.is_err() {
+            panic!("A compiled dependency is out of sync. Try deleting the target folder, or that of `prusti-contracts`.")
```
fpoli: Regarding "target folder ... of `prusti-contracts`", isn't this PR removing that folder? Or is the message intended for the users that still have `prusti-contracts/target` around from before this PR?

fpoli (approved):

fpoli:
The changes look good to me. Regarding testing with `prusti-assistant`, I find it easier to test the IDE when the nightly Prusti release has been published. I think it is perfectly fine if some nightly releases cannot be used in the IDE while we make changes like in this PR. What is important is that we eventually mark as stable (i.e. non-draft) only the releases that work well in the IDE.

x.py
```diff
@@ -279,17 +279,17 @@ def package(mode: str, package_path: str):
         (f"target/{mode}/prusti-server*", "."),
         (f"target/{mode}/prusti-rustc*", "."),
         (f"target/{mode}/cargo-prusti*", "."),
-        (f"prusti-contracts/target/verify/{mode}/libprusti_contracts.*", "."),
-        (f"prusti-contracts/target/verify/{mode}/deps/libprusti_contracts_proc_macros-*", "deps"),
-        (f"prusti-contracts/target/verify/{mode}/deps/prusti_contracts_proc_macros-*.dll", "deps"),
-        (f"prusti-contracts/target/verify/{mode}/libprusti_std.*", "."),
-        (f"prusti-contracts/target/verify/{mode}/deps/libprusti_contracts-*", "deps"),
-        (f"prusti-contracts/target/verify/{mode}/deps/prusti_contracts-*.dll", "deps"),
+        (f"target/verify/{mode}/libprusti_contracts.*", "."),
```
fpoli, line 282: Uh, actually, why is `verify` in this path?  It might actually be `target/{mode}/...`, here and below. When compiling in debug mode, do the binaries end up in `target/verify/debug/libprusti_contracts.*` or in `target/debug/libprusti_contracts.*`?

fpoli (commented):
See the comment

x.py
JonasAlaif: `cargo-prusti` always adds a `verify` to any cargo target dir you give it, so the `target/verify/...` with prusti is the equivalent of `target/...` for cargo

prusti-interface/src/specs/decoder.rs
JonasAlaif: Ooops yeah I forgot to change the message

Issue #941, opened 2022-04-04
Prusti does not understand the question mark operator


enjhnsn2:
When working with result and option types, it is very convenient to use the `?` operator to propagate errors. Prusti does not crash when you use the `?` operator, but it does not seem to understand that using a value unwrapped via `?` is the same as a value unwrapped via a `match`.
Given a function that returns an option:
```
#[ensures(match result {
    Some(r) => r == 3,
    None => true,  
})]
pub fn three_or_none(x: u32) -> Option<u32> {
    if x == 3 {
        Some(3)
    }
    else {
        None
    }
}
```

This function verifies:
```
// passes
pub fn test_manual_unwrap(x: u32) -> Option<u32> {
    let y = three_or_none(x);
    let y = match y {
        Some(z) => z,
        None => {return None;},
    };
    assert!(y == 3);
    Some(y)
}
```
whereas this function fails:
```
// fails
pub fn test_qm(x: u32) -> Option<u32> {
    let y = three_or_none(x)?;
    assert!(y == 3);
    Some(y)
}
```
while I expect these two assertions to be equisatisfiable.

Its not a huge priority, since it is not too hard to work around, however it would be very convenient if prusti had a better understanding of the `?` operator. 
Thanks for the help.

Aurel300:
Thanks for the suggestion! The reason this doesn't work right now is that the ? operator is desugared by the Rust compiler to a method call and by default that method has no specification and so Prusti has no way to know that internally the method does such a match. This might already be solvable using extern specs (maybe after #882). In any case, in the long term such a specification should of course be a part of the library of standard lib specifications.

enjhnsn2:
Thanks for the fast response!
I would have expected it to have been desugared to a match, but I guess I'll take a look at the underlying MIR to see how its desugared exactly and try to implement an external spec for it. 

I don't need anything more from you, but if you keep this issue open, I'll post an update about whether I got it to work with the external spec if you're interested.

Issue #1221, opened 2022-11-01
Error when using pure functions that return a reference


juliand665:
When you use a pure function that returns a reference to an associated type (like e.g. `Deref::deref` does), you get an error like "There is no procedure contract for loan bw3. This could happen if you are chaining pure functions, which is not fully supported.", even though there doesn't appear to be any chaining going on. I've reduced it to a minimal example with no external types or extern specs:

```rust
trait Foo { // similar to Deref
    type Bar;
    
    // not necessary for error:
    //#[pure]
    fn bar(&self) -> &Self::Bar;
}

impl Foo for i32 {
    type Bar = i32;
    
    #[pure]
    fn bar(&self) -> &Self::Bar {
        &self
    }
}

fn main() {
    let v = 1;
    let _ = v.bar();
}
```

This is the specific error it produces:
<img width="909" alt="image" src="https://user-images.githubusercontent.com/827762/199158968-79af441f-3e2d-491f-b67e-e16e17b39e7b.png">

It doesn't seem to matter whether you use the function as a member or a free function (like `Foo::bar(&v)`). Also—though this might just be a product of how the encoding is performed—the error does not appear at the call site if the value is used later on, e.g. like this:

<img width="635" alt="image" src="https://user-images.githubusercontent.com/827762/199159576-ac14c1b3-6303-45c6-b2e0-0d2646415686.png">

juliand665:
@Aurel300 since you wanted to look into this :)

Patrick-6:
This example causes a very similar error, but it uses a reference to an i64 and not to an associated type:
```
use prusti_contracts::*;

#[pure]
fn fn_a(x: &i64) -> &i64 {
    x
}

#[pure]
fn fn_b(x: &i64, i: u64) {
    if i != 0 {
        fn_b(&fn_a(x), i - 1);
    }
}
```
```
[Prusti internal error] Prusti encountered an unexpected internal error
Details: There is no procedure contract for loan bw6. This could happen if you are chaining pure functions, which is not fully supported.
```

juliand665:
Update—I found an even simpler MWE than my posted one, no traits or associated types needed:

```rust
fn main() {
    let foo = Foo { x: 42 };
    let x_ref = *foo.x_ref();
}

struct Foo {
    x: i32,
}

impl Foo {
    #[pure]
    fn x_ref(&self) -> &i32 {
        &self.x
    }
}
```

So I think @Patrick-6's example is indeed this issue as well.

PR #1249, opened 2022-11-24
Provide Specs for the Standard Library

juliand665:
As part of my thesis under @Aurel300, this pull request adds specifications to the most popular types & functions in the Rust Standard Library. I am using data gathered through [qrates](https://github.com/rust-corpus/qrates) to inform decisions on what specs to prioritize, along with a sense of logical units, e.g. covering all members of the unconstrained `Option<T>` or `Result<T, E>` `impl`s, rather than only specifying the subset of these methods that sees the most usage.

## Outline of Planned & Completed Specifications

These specs are very much a work in progress and far from complete at the moment, though what is there should be sound.

### `core::option::Option`
- [x] main impl

### `core::result::Result`
- [x] main impl

### `core::clone::Clone`
- [x] snapshot equality with opt-out mechanism

### `core::default::Default`
- [x] purity (for usage in specs) with opt-out mechanism

#### specific default values
- [x] `()`, `bool`, `char`, numbers (the types specified [here](https://github.com/rust-lang/rust/blob/872631d0f0fadffe3220ab1bd9c8f1f2342341e2/library/core/src/default.rs#L204-L223))
- [x] other commonly-used types

### `core::ops::Deref`
- [ ] purity (for usage in specs) with opt-out mechanism

This is currently blocked due to #1221.

### `core::ops::Index`/`core::ops::IndexMut`
- [ ] purity (for usage in specs) with opt-out mechanism

This is similarly blocked due to #1221.

### `core::ops::Try`
- [x] impl for `core::option::Option`
- [x] impl for `core::option::Result`

### `core::mem::size_of`
- [x] trait to allow users to specify what this method will return for their type
- [x] via the above, numbers, which are [by far](https://gist.github.com/juliand665/8948233f08f795b7eaade82076ea4fde) the most commonly-passed types

### `core::convert::From`/`Into`
- [x] purity (for usage in specs) with opt-out mechanism
- [x] numeric conversions, which are again [very common](https://gist.github.com/juliand665/486ec64a985de37675dcfd152f2b2bce)

### Slices/Arrays
- [x] indexing with `usize` (built-in)
- [x] length (built-in)
- [ ] "unsizing" arrays to slices

Indexing is currently blocked due to #1221.

### `alloc::vec::Vec`
- [ ] `Deref` as slice (#1221)
- [ ] forwarding of most operations to `as_slice` (#1221)
- [ ] `vec!` macro (requires unsizing from above, but very simple given that!)

### `alloc::string::String`
- [ ] `Deref` as `str` (#1221)
- [ ] forwarding of most operations to `as_str` (#1221)
- [ ] `&str` to `String` conversion using `From`/`Into`
I envision Strings and Vecs to work very similarly, mostly powered by a pure `as_str`/`as_slice` that acts as the source of truth. Unfortunately, that approach (like so much other stuff) is currently blocked, but I think the issue is so fundamental that it's better to just resolve it directly than to find a temporary workaround.

### `core::cmp::PartialEq`/`core::cmp::PartialOrd`
These are currently blocked due to #1311.

### Ranges
- [ ] `contains` on inclusive & half-open ranges, for loops & invariants (#1311)

### `core::ops` Binary Ops for References
- [x] specify how these ops are equivalent to applying to deref'd values
- [ ] the same for comparisons (#1311)

These operations don't have blanket impls for when one or both sides are a reference, but their implementations are [unified using macros](https://github.com/rust-lang/rust/blob/fd815a5091eb4d49cd317f8ad272f17b7a5f550d/library/core/src/ops/arith.rs#L110), which probably makes sense for us to replicate.

### Smart Pointers (e.g. `alloc::rc::Rc`)
This fundamentally relies on `Deref` and is thus also blocked by #1221.
Once that is resolved, specifying their `Deref` implementation as pure and expressing transfer into and out of a box via e.g. `ensures(old(x) === snap(result.deref()))` should go a long way towards specifying this type usefully.
Note that `Box` already has builtin support in Prusti, partly because Rust itself already treats it specially (e.g. that `*box` can move out of the box).


Pointerbender:
> `core::mem::size_of`
>  trait to allow users to specify what this method will return for their type

That's exciting stuff! Would this also be possible for `core::mem::align_of`, by any chance? :)

juliand665:
> Would this also be possible for `core::mem::align_of`, by any chance? :)

Sure! For the values so far, size and alignment seem to be the same, so that should be really simple. Interestingly, it seems that `align_of` [is not commonly called](https://gist.github.com/juliand665/8948233f08f795b7eaade82076ea4fde#file-align_of-txt) with these types. I'll specify it anyway since it's so simple though.

prusti-contracts/prusti-contracts/src/core_spec/default.rs
```diff
@@ -0,0 +1,47 @@
...
+// analogous to https://github.com/rust-lang/rust/blob/872631d0f0fadffe3220ab1bd9c8f1f2342341e2/library/core/src/default.rs#L190-L202
+macro_rules! default_spec {
+    ($t:ty, $v:expr) => {
+        #[extern_spec]
+        impl Default for $t {
+            #[pure]
+            #[ensures(result == $v)]
+            fn default() -> Self {
+                $v
+            }
```
Aurel300: This should not have a body.

prusti-contracts/prusti-contracts/src/core_spec/mem.rs
```diff
@@ -0,0 +1,65 @@
...
+        #[pure]
+        #[ghost_constraint(T: core_spec::mem::KnownSize, [ensures(result == T::size())])]
+        fn size_of<T>() -> usize;
+        
+        #[pure]
+        #[ghost_constraint(T: core_spec::mem::KnownSize, [ensures(result == T::align())])]
+        fn align_of<T>() -> usize;
+		
+		#[ensures(*x === old(snap(y)) && *y === old(snap(x)))]
+		fn swap<T>(x: &mut T, y: &mut T);
```
Aurel300: Use spaces for indentation throughout.

prusti-contracts/prusti-contracts/src/core_spec/mem.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[pure]
+    fn align() -> usize;
+}
+
+macro_rules! known_size_spec {
+    ($t:ty, $size:expr, $align:expr) => {
+        #[refine_trait_spec]
+        impl KnownSize for $t {
+            #[pure]
+            #[ensures(result == $size)]
```
Aurel300: This postcondition should be unnecessary, same for `align` below. (The body of a pure function becomes part of its contract…)

prusti-contracts/prusti-contracts/src/core_spec/default.rs
juliand665: As soon as I switch away from my branch with the change to warn on bodies for such methods lol, good catch.

prusti-contracts/prusti-contracts/src/core_spec/mem.rs
```diff
@@ -0,0 +1,65 @@
...
+known_size_spec!(u8, 1, 1);
+known_size_spec!(u16, 2, 2);
+known_size_spec!(u32, 4, 4);
+known_size_spec!(u64, 8, 8);
+known_size_spec!(u128, 16, 16);
+
+known_size_spec!(f32, 4, 4);
+known_size_spec!(f64, 8, 8);
+
+// usize/isize are not specified exactly because they are platform-dependent and programmers should not depend on their specific values anyway.
```
Aurel300: We could say that the value is one of the three possible for Rust, i.e. [16-, 32- or 64-bit](https://doc.rust-lang.org/src/core/num/mod.rs.html#243-268). However, I think we can do even better: put the `known_size_spec!` under a `cfg`, so when compiling the correct value is chosen.

prusti-contracts/prusti-contracts/src/core_spec/clone.rs
```diff
@@ -0,0 +1,12 @@
+use crate::*;
+
+#[extern_spec]
+trait Clone {
+    #[ghost_constraint(Self: SnapshotEqualClone, [
+		ensures(result === self),
+	])]
```
Aurel300: Use spaces throughout.

prusti-contracts/prusti-contracts/src/core_spec/mem.rs
juliand665: Ah, I was so used to using it in extern specs that that wasn't in my mental model.

prusti-contracts/prusti-contracts/src/core_spec/result.rs
```diff
@@ -0,0 +1,181 @@
...
+#[allow(unused_imports)]
+use core::fmt::Debug;
+#[allow(unused_imports)]
+use core::ops::{Deref, DerefMut};
+#[allow(unused_imports)]
+use core::result::*;
+
+#[extern_spec]
+impl<T, E> Result<T, E> {
+    // TODO: already in prusti-contracts
```
Aurel300: Stale TODO?

prusti-contracts/prusti-contracts/src/core_spec/mem.rs
juliand665: Yeah, I considered something like that. This would allow you to ensure that your code is correct across different word sizes without having to run Prusti for each platform, while specifying the exact size gives information the programmer should maybe not be relying on? But I can absolutely specify it if you think it makes sense.

Aurel300: I think we make the same assumption (that the verification machine = the target machine) when processing MIR [`SizeOf`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.NullOp.html#variant.SizeOf) operations. So for now, I would suggest using the `cfg` approach to be consistent. In the future we might think about "cross-verification".

prusti-contracts/prusti-contracts/src/core_spec/clone.rs
juliand665: Ah right, rustfmt doesn't touch whitespace within attributes because it treats them opaquely. I'll find-replace all tabs with 4 spaces, and try to keep in mind to check occasionally since they seem to like to slip in…

prusti-contracts/prusti-contracts/src/core_spec/default.rs
```diff
@@ -0,0 +1,45 @@
...
+#[extern_spec]
+trait Default {
+    #[refine_spec(where Self: Copy + PureDefault, [pure])]
+    fn default() -> Self;
+}
+
+/// Specifies that `Default::default`, if implemented, is a pure method, allowing its usage in specs.
+///
+/// Does not apply to types that do not implement `Copy`, since pure methods can only involve `Copy` types.
+pub auto trait PureDefault {}
```
Aurel300, line 12: I don't know if I asked before: can `PureDefault` be declared with `Copy` as a super-trait? Or does that not work for auto traits?

prusti-contracts/prusti-contracts/src/core_spec/default.rs
```diff
@@ -0,0 +1,45 @@
...
+        #[extern_spec]
+        impl Default for $t {
+            #[pure]
+            #[ensures(result == $v)]
+            fn default() -> Self;
+        }
+    };
+}
+
+default_spec! { (), () }
```
Aurel300, line 26: Can we also specify the `Default` for non-unit tuples? [It exists for up to 12-item tuples.](https://doc.rust-lang.org/stable/std/primitive.tuple.html#impl-Default-for-(T%2C))

prusti-tests/tests/compiletest.rs
```diff
@@ -198,6 +198,11 @@ fn test_runner(_tests: &[&()]) {
     run_verification_overflow("verify_overflow", &filter);
     save_verification_cache();
 
+    // Test the functionality of built-in specs.
+    println!("[prusti_std]");
```
Aurel300: (As messaged) I would name this directory simply `std`.

prusti-contracts/prusti-contracts/src/core_spec/default.rs
juliand665: They unfortunately cannot have super traits, I just checked again to make sure. The explanation sounds like it's not planned either, rather a principle I'd disagree with.

image

Pointerbender:
Once this PR has laid the ground-work for adding specs to the Rust standard library, how hard/easy would it be for an advanced end-user to contribute standard library specs for iterators to Prusti? :) Would it be as easy as adding a bunch of `#[extern_spec]`s under `prusti-contracts/prusti-std/src/` or would it require some extra development first to be able to `extern_spec` iterator logic? Thanks for building this btw, I'm really looking forward to using this already! :D

prusti-contracts/prusti-std/src/collections.rs
```diff
@@ -0,0 +1,21 @@
+use prusti_contracts::*;
+
+#[extern_spec]
+impl<K, V, S> std::collections::hash_map::HashMap<K, V, S>
+where
+    K: Eq + ::core::hash::Hash,
+    S: ::std::hash::BuildHasher,
+{
+    #[pure]
```
Pointerbender: Is marking this generic method as `#[pure]` also sound in the presence of interior mutability? (e.g. what if a hypothetical `PartialEq::eq` implementation modifies its internals?)

fpoli: That's a good point, I don't think it would be. It's also related to https://github.com/viperproject/prusti-dev/issues/1286#issuecomment-1406247838.

I wonder if with `refine_spec` we can restrict all these specifications to types `K/Q` for which `Eq/Hash/Borrow` is pure.

Pointerbender: Would that also work if `K` is replaced by `&K` which then delegates its impl to a potentially non-structural impl?

fpoli: (It would be nice to add to this PR a test where a key type implements a random `PartialEq` and `Hash`. A client of the `HashMap` should then *not* be able to assert anything about the result of `HashMap::get`.)

fpoli: > Would that also work if `K` is replaced by `&K` which then delegates its impl to a potentially non-structural impl?

I edited my message. The issue is not just interior mutability, but also non-determinism.

Aurel300: > I wonder if with `refine_spec` we can restrict all these specifications to types `K/Q` for which `Eq/Hash/Borrow` is pure.

We could.

I think `get` being deterministic is not controversial for a `HashMap`.

However, I think/hope interior mutability isn't actually such a big problem. In my mental model, an interior-mutable structure can be expected to be snapshot equal. As we encode its snapshot recursively, we will eventually find the `UnsafeCell`, which is where we stop: the cell is a *handle*, which does not tell us anything about the value until we also have e.g. a heap (version) to refer to.

Aurel300:
> Once this PR has laid the ground-work for adding specs to the Rust standard library, how hard/easy would it be for an advanced end-user to contribute standard library specs for iterators to Prusti? :) Would it be as easy as adding a bunch of `#[extern_spec]`s under `prusti-contracts/prusti-std/src/` or would it require some extra development first to be able to `extern_spec` iterator logic? Thanks for building this btw, I'm really looking forward to using this already! :D

Adding specs to `prusti-std` would indeed be the general idea. We do already have an iterator methodology, which I will be submitting to `prusti-std` at some point after this PR. In some cases, we will find issues (some of which Julian has linked in the description of this PR) that would block the specs we want to write. Some of these issues can be quite heavy, for example: supporting references in structs.

juliand665:
Oh yeah, to establish a link: this resolves #941!

prusti-tests/tests/verify_overflow/pass/extern-spec/vec-3.rs
```diff
@@ -1,28 +1,12 @@
...
-    #[requires(self.is_some())]
-    pub fn unwrap(self) -> T;
-}
+use std::{option::Option, vec::Vec};
 
 #[extern_spec]
 impl<T> Vec<T> {
     #[ensures(result.len() == 0)]
-    fn new() -> Vec::<T>;
+    fn new() -> Vec<T>;
```
Aurel300, line 9: `prusti-contracts/prusti-std/src/alloc_spec/vec.rs` also defines a spec for `new`. Why is this not considered a duplicate?

Aurel300: Because in these tests only the `core` specs are used, `std` specs require the `prusti-std` crate as a dependency.

LrWm3:
is this something that could be partially merged?

PR #1259, opened 2022-12-02
Small improvements

Patrick-6:
There were some unnecessary clones in the prusti-common crate that I removed.

In addition, some files in the crate were using the standard library collections HashMap and HashSet. These use the default Hasher that is secure against hashDOS attacks and therefore slower. Since there does not seem to be a need to secure against such attacks, I've changed them to use the FxHashMap and FxHashSet from fxhash, which was already in use in the crate (in program_normalization.rs).
I've also added a Clippy lint to warn about the use of the standard collections HashMap and HashSet inside the crate prusti-common.
Some other crates in the project still use the standard hasher, and some use the FxHash from rustc (e.g. prusti-interface::utils.rs), this may be worth unifying at some point.

I've ran all the tests in the project, and all except for 2 tests pass. In these tests, Viper takes ~6.5 seconds instead of the limit of 5 seconds, but I don't think this is caused by my changes.

The other changes where an additional `not` in x.py, that showed a warning when the package directory was empty, instead of when it contained files.

The last change is to updating the requirements mentioned in the developer guide. I can't test if macOS also needs Java JDK 11 or newer, but I suspect macOS has the same version requirements as Linux.
I also added a comment that `x.py setup` does not install java by itself on Ubuntu.


fpoli:
Thanks! Could you fix the formatting error?

Patrick-6:
It should work now.
I realized that `x.py check-fmt-all` and `x.py fmt-all` did not include `prusti-common`, so I added it now.
The crates `prusti-contracts/prusti-specs`, `prusti-viper` and `test-crates` are still excluded.

fpoli:
Thanks!

PR #1263, opened 2022-12-05
Improve debugging docs

jscissr:
Improve the debugging documentation, by adding non-obvious details and examples. Add a section about running Prusti in a debugger.

Change the verify-test script to add the environment variables to the generated launch.json. It also works without this change if you start VS Code with `./x.py ide .`, but we can avoid this extra manual step with this change.

Also repair the png files that were broken by CRLF -> LF conversion, probably because of a bad `.gitattributes`.


Aurel300:
Thanks!

PR #1265, opened 2022-12-14
Small improvements

fpoli:
This PR contains several small refactorings:
* Add generation of builtin domains in the encoder
* Add convenient methods to VIR poly
    * E.g. `Expr::exhale(..)`
* Add syntax for locals in `vir_expr!`
* Add description of VIR transformations to a README
* Replace `--rustup-only` with `--no-deps` in `x.py`
* Fix encoding of triggers
    * `vir_expr!(forall ... :: { A, B } :: ...)` now translates to `forall ... :: { A, B } ...` in Viper, no longer `forall ... :: { A } { B } ...`.
* Add optional comment field to the domain axioms.
* Add macro to return encoding errors
    * Super useful. `error_internal/unsupported/incorrect!(...)` desugars to `return Err(EncodingError::...(format!(...), ...));`
* Add empty line between the Viper encoding of CFG blocks
* Show stacktrace of encoding errors 
    * Only in debug mode, since it's slow. Very useful for debugging.

### DIFF
--- a/prusti-common/src/vir/low_to_viper/domain.rs
+++ b/prusti-common/src/vir/low_to_viper/domain.rs
@@ -49,6 +49,15 @@ impl<'a, 'v> ToViper<'v, Vec<viper::NamedDomainAxiom<'v>>>
 impl<'a, 'v> ToViper<'v, viper::NamedDomainAxiom<'v>> for (&'a String, &'a DomainAxiomDecl) {
     fn to_viper(&self, context: Context, ast: &AstFactory<'v>) -> viper::NamedDomainAxiom<'v> {
         let (domain_name, axiom) = self;
-        ast.named_domain_axiom(&axiom.name, axiom.body.to_viper(context, ast), domain_name)
+        if let Some(comment) = &axiom.comment {
+            ast.named_domain_axiom_with_comment(
+                &axiom.name,
+                axiom.body.to_viper(context, ast),
+                domain_name,
+                comment,
+            )
+        } else {
+            ast.named_domain_axiom(&axiom.name, axiom.body.to_viper(context, ast), domain_name)
+        }
     }
 }
--- a/prusti-viper/src/encoder/builtin_encoder.rs
+++ b/prusti-viper/src/encoder/builtin_encoder.rs
@@ -304,6 +304,7 @@ impl<'p, 'v: 'p, 'tcx: 'v> BuiltinEncoder<'p, 'v, 'tcx> {
                 vec![vir::Trigger::new(vec![function_app.clone()])],
                 function_app);
             let axiom = vir::DomainAxiom {
+                comment: None,
                 name: format!("{}$axiom", f.get_identifier()),
                 expr: body,
                 domain_name: domain_name.to_string(),
--- a/prusti-viper/src/encoder/builtin_encoder.rs
+++ b/prusti-viper/src/encoder/builtin_encoder.rs
@@ -7,6 +7,7 @@
 use prusti_common::{vir_local, vir_expr};
 use vir_crate::polymorphic::{self as vir};
 use vir_crate::common::identifier::WithIdentifier;
+use super::errors::EncodingResult;
 use super::high::builtin_functions::HighBuiltinFunctionEncoderInterface;
 use super::versioning;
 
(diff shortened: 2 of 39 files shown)


prusti-common/src/vir/low_to_viper/domain.rs
```diff
@@ -35,20 +35,29 @@ impl<'a, 'v> ToViper<'v, viper::DomainFunc<'v>> for (&'a String, &'a DomainFunct
     }
 }
 
-impl<'a, 'v> ToViper<'v, Vec<viper::NamedDomainAxiom<'v>>>
+impl<'a, 'v> ToViper<'v, Vec<viper::AnonymousDomainAxiom<'v>>>
```
vakaras: I see how generating names can be annoying when experimenting, but I find having unique names useful in my encodings, where I have hundreds of axioms. Therefore, I would prefer to revert this at least for `low_to_viper`.

viper/benches/bench_program.rs
```diff
@@ -56,8 +56,7 @@ fn build_program<'a>(ast: &'a AstFactory) -> Program<'a> {
     let wrapper_domain = ast.domain(
         "Wrapper",
         &[wrap_domain_function, unwrap_domain_function],
-        &[ast.named_domain_axiom(
-            "unwrap_wrapper",
+        &[ast.anonymous_domain_axiom_with_comment(
```
vakaras: I think we should test both anonymous and named axioms.

vakaras (dismissed):
> Encode domain axioms as anonymous 

I am not convinced about this one. At least, I want to keep the old behaviour for my encoding.

fpoli:
@vakaras I reverted the anonymous domains and just added an optional comment field.

prusti-viper/src/encoder/builtin_encoder.rs
```diff
@@ -81,7 +80,8 @@ impl<'p, 'v: 'p, 'tcx: 'v> BuiltinEncoder<'p, 'v, 'tcx> {
         }
     }
 
-    pub fn encode_builtin_method_def(&self, method: BuiltinMethodKind) -> vir::BodylessMethod {
+    pub fn encode_builtin_method_def(&self, method: BuiltinMethodKind) -> EncodingResult<vir::BodylessMethod> {
```
Aurel300, line 83: Why was this changed to a `Result` return? There are no error paths. Is this just for consistency?

fpoli: Because I have some `BuiltinMethodKind` that I want to add that need to return a `Result`.

prusti-viper/src/encoder/snapshot/encoder.rs
```diff
@@ -1257,8 +1254,9 @@ impl SnapshotEncoder {
                     });
 
                     vir::DomainAxiom {
+                        comment: None,
                         name: format!("{}$len_of_seq", predicate_type.name()),
-                        expr: vir_expr! { forall data: {seq_type} :: { [len_call], [seq_len] } ([len_call] == [seq_len]) },
+                        expr: vir_expr! { forall data: {seq_type} :: { [len_call] } { [seq_len] } :: ([len_call] == [seq_len]) },
```
Aurel300, line 1259: (I understand this change keeps the current behaviour.) Any chance we actually wanted this to be triggered when *both* expressions are in scope?

Aurel300:
LGTM. I don't find named axioms to be useful. I guess you can only search for them in the generated code @vakaras, but you can do that with comments as well, no?

prusti-viper/src/encoder/snapshot/encoder.rs
fpoli: I tried that, but several `prusti-tests` tests related to arrays and slices were failing.

vakaras:
> LGTM. I don't find named axioms to be useful. I guess you can only search for them in the generated code @vakaras, but you can do that with comments as well, no?

Having axioms without names when there are hundreds of them makes the generated code very hard to understand and navigate. Requiring to write axiom names helps to ensure at least some basic readability of the generated code.

vakaras (approved):
LGTM

PR #1269, opened 2022-12-19
Fix #1202

### LINKED ISSUE #1202: Functions starting or ending with an underscore generate a [non_snake_case] warning
Any function with a pre- or postcondition that either starts or ends with an underscore '_' generates a [non_snake_case] warning: 
<pre>
warning: function `prusti_post_item_test0__04f804db40ba4c8b8a63ba63eaf3d94e` should have a snake case name
--> src/examples/underscore.rs:3:11
  |
3 | #[ensures(true)]
  |           ^^^^ help: convert the identifier to snake case: `prusti_post_item_test0_04f804db40ba4c8b8a63ba63eaf3d94e`    |
 = note: `#[warn(non_snake_case)]` on by default 
</pre>

All of the following functions (except the last 2) generate this warning:

<pre><code>use prusti_contracts::*;

#[ensures(true)]
fn test0_(){}

#[requires(true)]
fn test1_(){}

#[requires(true)]
#[ensures(true)]
fn test2_(){}

#[ensures(true)]
fn _test0(){}

#[requires(true)]
fn _test1(){}

#[requires(true)]
#[ensures(true)]
fn _test2(){}

fn test4_(){} // no warnings
fn _test4(){} // no warnings</code></pre>

Patrick-6:
This is a fix for https://github.com/viperproject/prusti-dev/issues/1202.
The only code change is to add `non_snake_case` to the allow-list for functions created internally by Prusti.

There is one new test for this issue, which passes after the code change, but did not before.
Some existing tests were adjusted to expect the added annotation.


Aurel300:
Thanks!

PR #1271, opened 2023-01-02
Ghost Constraint Stabilization

juliand665:
As part of [my thesis](https://ethz.ch/content/dam/ethz/special-interest/infk/chair-program-method/pm/documents/Education/Theses/Julian_Dunskus_MS_Description.pdf) under @Aurel300, I've been writing specs for the standard library (#1249), which will ship with Prusti for users to benefit from.

## Ghost Constraints → Type-Conditional Spec Refinements

These specs often rely on ghost constraints, e.g. to provide a way to opt out of specs [applied by default](https://github.com/viperproject/prusti-dev/pull/1249/files#diff-b5e647a76ab0c8608a5d7a0ab8ff619d6bbf84217f854e134d5ff60b636c0e0aR5-R7) to certain trait implementations, or to [specify known return values](https://github.com/viperproject/prusti-dev/pull/1249/files#diff-b4d918ba4e39891439964fe0debf2b28f0d9cef6e4417bb0e9321cd34d5b6fb1R9) for e.g. `core::mem::size_of`. Since these specs will be active by default, they would require enabling ghost constraints to use Prusti at all (barring the option of conditionalizing specs on enabled Prusti features somehow). Hence we propose to stabilize ghost constraints, addressing some issues with them to get them to a place where this makes sense.

Ghost constraints are now called type-conditional spec refinements (`TypeCondSpec` for short) and spelled as follows:
```rs
#[refine_spec(where Self: Copy, T: Copy, [
    pure,
    ensures(result === T::default())
])
```
The `where` draws a clear analogy to existing Rust syntax to help users understand this attribute, and it even results in correct syntax highlighting for the constraints. Note also that `pure` is now supported here (correctly—my previous implementation was broken), and that you can apply multiple bounds.

## Extern Spec Improvements

I also polished up & resolved some limitations with extern specs:

- An error is now thrown when extern specs are applied to methods with bodies, rather than stubs.
- They can now be applied to trait impls with generic arguments, which was previously cautiously rejected. There is [a unit test](https://github.com/viperproject/prusti-dev/pull/1271/files#diff-0eda14ab21130110aee12255a5b0425d332f42439cd49d31acd0ab6b4c2408a5) ensuring that this works correctly.
- Extern specs on traits now support bounds in a `where` clause.
- They can now be applied to free functions, taking an argument to specify that function's module path, e.g. `#[extern_spec(core::mem)] fn swap(...)`. This argument also works on traits & modules (`impl`s already allow qualifying involved type paths).

## Miscellaneous

As part of this PR, I enabled auto-formatting for the `prusti-specs` crate, and reformatted it in the process. This crate is one of the last few holdovers that weren't yet auto-formatted, the other big one being `prusti-viper`. It looked like this crate is not currently being majorly worked on by anyone else!

Finally, I fixed a bug where lifetimes would make trait method resolution fail, by instead erasing all regions involved in the resolution process. This made some tests fail before I [reworked](https://github.com/viperproject/prusti-dev/commit/b3be7e81ed2cfb16eaa007f7431b5a65b1b38318) the piece of code that encodes lifetimes for function calls. This bug made it impossible to specify e.g. `impl Add<&i32> for i32`, because Prusti could not resolve the general `Add::add` to this specific implementation (due to the lifetimes necessarily differing between call site & declaration site).


docs/user-guide/src/verify/type_cond_spec.md
```diff
@@ -0,0 +1,37 @@
...
+        // we could write more specs here
+    ])]
+    #[ensures(/* partial equivalence formulas */)]
+    fn eq(&self, other: &Rhs) -> bool;
+}
+```
+
+Thus, any client implementing `Eq` on a custom type can take advantage of the additional semantics of the total equivalence.
+
+> Refinements like these are not scoped. Therefore, considering the previous example, implementing `Eq` on a type implies that the total equivalence contract is always considered on the type, irrespective of whether `Eq` is in scope or not.
```
Aurel300: I find this a bit confusing/redundant: implementation of traits is not scoped in general. At the callsite, the constraint can be resolved either way for sure, because the `impl` exists either on the concrete type or on the trait (+ blanket `impl`s).

juliand665: Yeah, I was a bit torn on this comment, so I just tried to keep things unchanged from the previous state. But you're absolutely right: the compiler knows about the impl with or without an import; that would only be relevant when actually trying to use one of its members. Do you reckon we can just get rid of the quoted part entirely?

prusti-contracts/prusti-specs/src/extern_spec_rewriter/impls.rs
```diff
@@ -340,16 +329,28 @@ mod tests {
...
+            let rewritten = rewrite_extern_spec_internal(&mut inp_impl).unwrap();
+
+            let newtype_ident = &rewritten.generated_struct.ident;
+            let expected_impl: syn::ItemImpl = parse_quote! {
+                impl #newtype_ident <> {
+                    #[prusti::extern_spec = "trait_impl"]
+                    #[allow(unused, dead_code)]
+                    #[prusti::trusted]
+                    fn foo(_self: &mut MyStruct, arg1: I) {
+                        <MyStruct as MyTrait<I>> :: foo :: <>(_self, arg1)
```
Aurel300, line 373: Where is the type parameter `I` declared after rewriting? Is it somehow part of `newtype_ident`?

prusti-contracts/prusti-specs/src/extern_spec_rewriter/traits.rs
```diff
@@ -49,29 +59,42 @@ fn generate_new_struct(item_trait: &syn::ItemTrait) -> syn::Result<GeneratedStru
...
     };
 
     // Add a where clause which restricts this self type parameter to the trait
-    if let Some(where_clause) = &item_trait.generics.where_clause {
-        return Err(syn::Error::new(where_clause.span(), "Where clauses for extern traits specs are not supported"));
-    }
-    let self_where_clause: syn::WhereClause = parse_quote! {
-        where #self_type_ident: #self_type_trait
-    };
+    let self_where_clause: syn::WhereClause =
```
Aurel300: Maybe worth checking: is a `trait Foo where Self: Bar ...` clause handled correctly, i.e. is the `Self` rewritten to `Prusti_T_Self` as well?

prusti-contracts/prusti-specs/src/extern_spec_rewriter/traits.rs
```diff
@@ -203,15 +233,16 @@ impl ProvidedTypeParam {
         match path.segments[0].ident.to_string().as_str() {
             "generic" => Some(ProvidedTypeParam::GenericType(clone_without_attrs())),
```
Aurel300: We should get rid of this mechanism altogether, we only want to allow "generic" generics.

prusti-contracts/prusti-specs/src/rewriter.rs
```diff
@@ -164,6 +170,30 @@ impl AstRewriter {
...
+            &format!("prusti_pure_ghost_item_{}", item.sig().ident),
+            item_span,
+        );
+
+        let spec_id_str = spec_id.to_string();
+        let mut spec_item: syn::ItemFn = parse_quote_spanned! {item_span=>
+            #[allow(unused_must_use, unused_parens, unused_variables, dead_code)]
+            #[prusti::spec_only]
+            #[prusti::spec_id = #spec_id_str]
+            fn #item_name() {} // we only care about this for evaluating
```
Aurel300: "for evaluating" what? The `where` clauses I guess?

prusti-contracts/prusti-specs/src/specifications/mod.rs
```diff
@@ -11,7 +11,7 @@
 ///           ;
 /// requires ::= "requires", "(", prusti_expr, ")" ;
 /// ensures ::= "ensures", "(", prusti_expr, ")" ;
-/// 
+///
 /// This grammar doesn't yet handle the unintuitive precedence difference between `&&` and `||` operators.
```
Aurel300: (This entire comment is outdated now.)

prusti-contracts/prusti-specs/src/specifications/preparser.rs
```diff
@@ -905,14 +946,16 @@ impl RustOp {
 }
 
 fn join_spans(s1: Span, s2: Span) -> Span {
-    // Tests don't run in the proc macro context
-    let is_proc_macro = std::panic::catch_unwind(|| s1.unwrap()).is_ok();
-    if is_proc_macro {
-        // This works even when compiled with stable
-        s1.unwrap().join(s2.unwrap()).expect("Failed to join spans!").into()
-    } else {
+    // Tests don't run in the proc macro context, so this gets a little funky for them
```
Aurel300, line 954: @JonasAlaif I think I recall we talked about this at some point. Does this `cfg` solution look good to you? Did you try this before the `catch_unwind` one?

prusti-interface/src/specs/typed.rs
```diff
@@ -203,6 +203,7 @@ pub struct ProcedureSpecification {
     pub pledges: SpecificationItem<Vec<Pledge>>,
     pub trusted: SpecificationItem<bool>,
     pub terminates: SpecificationItem<Option<LocalDefId>>,
+    pub purity: SpecificationItem<Option<DefId>>, // for type-conditional spec refinements
```
Aurel300, line 206: I found this addition really strange. Why do we need another field just for purity? Isn't the refinability of `kind` already expressed in the `SpecificationItem` type?

prusti-interface/src/specs/typed.rs
```diff
@@ -436,6 +438,23 @@ impl SpecGraph<ProcedureSpecification> {
...
+            None => {
+                unreachable!(
+                    "separate purity defs are only used for type-conditional spec refinements"
+                )
+            }
+            Some(constraint) => {
+                let constrained_spec = self.get_constrained_spec_mut(constraint);
+                constrained_spec.kind =
+                    SpecificationItem::Inherent(ProcedureSpecificationKind::Pure);
+                // need to store this as well, since without pres or posts we couldn't find any def id with the trait bounds
```
Aurel300, line 452: Is this the answer to the comment about the `purity` field? Is it because we still won't generate a spec item *just* for purity?

prusti-tests/tests/parse/ui/and.stdout
```diff
@@ -23,7 +23,7 @@ use prusti_contracts::*;
 #[prusti::spec_only]
 #[prusti::spec_id = "$(NUM_UUID)"]
 fn prusti_pre_item_test1_$(NUM_UUID)() -> bool {
-    !!((true && true): bool)
+    !!(((true) && (true)): bool)
```
Aurel300, line 27: Where did we get the extra parentheses from?

prusti-tests/tests/verify/fail/ghost-constraints/purity.rs
```diff
@@ -0,0 +1,12 @@
+use prusti_contracts::*;
+
+#[refine_spec(where T: Copy, [pure])]
+#[trusted]
+fn test<T>(_t: T) -> bool { true }
+
+#[derive(Clone, PartialEq, Eq)]
```
Aurel300: Unnecessary derive?

prusti-tests/tests/verify/pass/extern-spec/module-arg.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[extern_spec(module::inner)]
+trait Advanced<#[generic] T> where T: Copy {
+    #[pure]
+    fn example() -> T;
+}
+
+#[extern_spec(module)]
+mod inner {
+    use crate::*;
```
Aurel300: I wonder: do we really need to leave these imports in? Do we keep the `mod` structure around when rewriting this kind of extern spec? Or is this just to appease IDEs?

prusti-tests/tests/verify/pass/type-cond-specs/purity.rs
```diff
@@ -0,0 +1,9 @@
+use prusti_contracts::*;
+
+#[refine_spec(where T: Copy, [pure])]
+#[trusted]
+fn test<T>(_t: T) -> bool { true }
+
+fn main() {
+	assert!(test(5) == test(5));
```
Aurel300: Indented with a tab.

prusti-tests/tests/verify/ui/failing-postcondition.stderr
```diff
@@ -1,8 +1,8 @@
 error: [Prusti: verification error] postcondition might not hold.
- --> $DIR/failing-postcondition.rs:8:11
+ --> $DIR/failing-postcondition.rs:8:28
   |
 8 | #[ensures(something_true() && false)]
```
Aurel300: What made the span change here?

prusti-tests/tests/verify_overflow/fail/type-cond-specs/in-non-trusted-function.rs
```diff
@@ -0,0 +1,11 @@
+use prusti_contracts::*;
+
+trait A {}
+
+#[refine_spec(where T: A, [
+	ensures(true)
```
Aurel300: tab

prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
```diff
@@ -1488,11 +1488,9 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
         // The called method might be a trait method.
         // We try to resolve it to the concrete implementation
         // and type substitutions.
+        let query = self.encoder.env().query;
```
Aurel300, line 1491: Why this alias? It doesn't seem to be used later on.

juliand665: Without it, the long line spilled over into a silly-looking chained function call spanning like 5 lines, lol. I considered updating the later usage, but that felt worse. Should I just accept the other formatting rather than polluting this long function's namespace?

prusti-tests/tests/verify_overflow/fail/type-cond-specs/in-non-trusted-function.rs
juliand665: Non-autoformatted files will be my undoing…

prusti-contracts/prusti-specs/src/extern_spec_rewriter/impls.rs
juliand665: I did a double take here too originally—`I` is ([was](https://github.com/juliand665/prusti-dev/commit/76e6f6598f6161d0afa5ce23ce0fa03ebfe90a73)) in fact not a type parameter! Hence why it's not where you'd expect. I think it makes sense to test both a blanket `impl<I>` and a specific `impl` though. I'll update the tests accordingly, and rename `I` in the single case to not break conventions ;)

prusti-contracts/prusti-specs/src/extern_spec_rewriter/traits.rs
juliand665: Ooh yeah, we can entirely get rid of the whole annotation mess thanks to typecondspecs turning stable! Always satisfying to delete code.

prusti-contracts/prusti-specs/src/rewriter.rs
juliand665: Oh yeah, I must have gotten distracted halfway through that comment. I'll replace it with something clearer.

prusti-contracts/prusti-specs/src/specifications/mod.rs
juliand665: Not just because of this PR though, right? Should I just put down a FIXME or something?

prusti-tests/tests/parse/ui/and.stdout
juliand665: They came from [here](https://github.com/viperproject/prusti-dev/pull/1271/commits/d4069dd6e86805754cfa48090f88e08bac2090e8), fixing issues if e.g. `lhs` were `foo || bar`. They are technically avoidable, but I think that would require quite some restructuring to apply initial parentheses to any "untrusted" tokens, then respect precedence levels in the further desugaring. Probably not worth it?

prusti-tests/tests/verify/ui/failing-postcondition.stderr
juliand665: Not sure how that one slipped through! I couldn't find it by just looking so it looks like I'll have to throw in some print statements to track it down.

juliand665: Oh great, looks like [this change](https://github.com/viperproject/prusti-dev/pull/1271/commits/d4069dd6e86805754cfa48090f88e08bac2090e8) messed that up somehow. That doesn't make sense to me because it's already using an explicit span. Any ideas?

docs/user-guide/src/verify/type_cond_spec.md
Aurel300: Yes, I would drop this.

prusti-contracts/prusti-specs/src/specifications/mod.rs
Aurel300: No, I just noticed because of the reformat. But feel free to just remove the comment, it's not valuable if it's not accurate anyway.

prusti-tests/tests/verify/pass/extern-spec/module-arg.rs
juliand665: `#[extern_spec]` [erases](https://github.com/viperproject/prusti-dev/blob/a4ac54bbb747607ca9ab5d24f0013202a2e6e171/prusti-contracts/prusti-contracts-proc-macros/src/lib.rs#L73-L77) whatever it's attached to when not building for prusti, so I don't think this helps IDEs. I believe this is actually necessary because we're currently still preserving the module structure rather than moving everything to the top level. Thus, the `prusti_contracts` attributes are not in scope, and if you do import them, this case still gives an error because it can't resolve `module`, which is also not in scope within `inner`.

prusti-interface/src/specs/typed.rs
juliand665: It is indeed! From what I understand, the [current design](https://github.com/viperproject/prusti-dev/blob/a4ac54bbb747607ca9ab5d24f0013202a2e6e171/prusti-viper/src/encoder/mir/specifications/constraints.rs#L218-L221) (which we deemed out of scope to overhaul) looks at all the spec items, gathers `where` clauses as `ParamEnv`s, checks that there's only one such env (since we currently only support a single typecondspec refinement for any method), and returns it.

The problem is that to do so, it [runs through](https://github.com/viperproject/prusti-dev/blob/a4ac54bbb747607ca9ab5d24f0013202a2e6e171/prusti-viper/src/encoder/mir/specifications/constraints.rs#L234) all the pre- and postconditions listed in the `ProcedureSpecification`. I didn't touch this code initially until I realized it fails for ghost constraints that only establish purity and nothing else (no pres or posts) because the loop has nothing to run through. So [my solution](https://github.com/juliand665/prusti-dev/blob/0684867cc8765c8cc1abee796882393dd5761603/prusti-viper/src/encoder/mir/specifications/constraints.rs#L235) is to store and iterate over the purity spec item as well.

prusti-contracts/prusti-specs/src/extern_spec_rewriter/traits.rs
juliand665: Good catch! It indeed does not. Funnily enough it appears it would still work, if not for the fact that `Self` does not appear to be getting rewritten in pre-/postconditions anyway, so this area might need some more love in general. I'll handle the where clauses though.

juliand665: On second thought, I'm not sure what I needed where clauses in extern specs for anymore? I don't have any code lying around using it, and it doesn't seem useful, since you don't need them to figure out which trait it is, and it seems you already get the type bounds for "free", i.e. you're able to use methods from `Bar` in the specs for `Foo` like this:

```rs
#[extern_spec]
trait Foo {
    #[ensures(result == self.bar())]
    fn foo(&self) -> i32;
}
```

(I initially wrote this as static methods rather than taking `&self`, bringing the `Self`-in-specs issue to my attention.)

Aurel300: Ah right, you're right: the bounds on `Self` should of course be inherited from the actual trait definition. Never mind!

prusti-tests/tests/parse/ui/and.stdout
Aurel300: I see the issue now. Yes, we could somehow detect when the parentheses are not needed, but that's indeed more effort than it's worth.

prusti-contracts/prusti-specs/src/specifications/preparser.rs
JonasAlaif: I guess it's late now anyway, but this solution looks better yeah

PR #1276, opened 2023-01-10
Optimize lookups in code generated by jni-gen

fpoli:
This PR caches the lookup of classes and method ids in the code generated by `jni-gen`, fixing the performance pitfall n.1, "Not caching method IDs, field IDs, and classes", described in IBM's ["Best practices for using the Java Native Interface"](https://developer.ibm.com/articles/j-jni/#notc) article. On large Viper programs (>50000 LOC), this saves tens of seconds in the construction of Java objects. Currently, each constructor/method wrapper uses its own cache. Cached Java classes will prevent the JVM from garbage-collecting the class object, but this should have no observable effect as class definitions are quite small.

One assumption of the generated code is that the Java class resolution doesn't change at runtime (e.g. no JARs are loaded or removed at runtime). This is true for Viper, but various `debug_assert!` will double check it anyway.


jni-gen/README.md
```diff
@@ -8,4 +8,5 @@ The generated code uses the `jni` crate to interact with the JVM.
 Requirements
 ------------
 
-`libjvm.so` needs to be in the `LD_LIBRARY_PATH` environment variable.
+* `libjvm.so` needs to be in the `LD_LIBRARY_PATH` environment variable.
+* The crate that makes use of `jni-gen` should have the `once_cell` dependency.
```
vakaras, line 12: Cannot `once_cell` be a dependency of `jni-gen`?

vakaras (approved):
LGTM

Aurel300:
Looks good, and the performance improvement is promising. I only wondered if the various argument-checking fragments aren't similar enough to refactor into a single function.

Aurel300 (approved):

jni-gen/README.md
fpoli: That would also require to make `viper-sys` depend on `jni-gen`, while at the moment `jni-gen` is only a *build* dependency. I don't particularly like it. `jni-gen` would also need to re-export `once_cell`, but that's fine.

fpoli:
> I only wondered if the various argument-checking fragments aren't similar enough to refactor into a single function.

Probably yes, but I would leave that to the future.

jni-gen/README.md
vakaras: I see.

PR #1280, opened 2023-01-17
Add macros for returning spanned encoding errors

fpoli:
This PR introduces the `span =>` syntax to return *spanned* encoding errors. For example:
```rust
error_unsupported!(span => "unsupported type: {}", array_of_bananas);
```
Its less-readable expansion:
```rust
return Err(SpannedEncodingError::unsupported(format!("unsupported type: {}", array_of_bananas), span)));
```

### DIFF
--- a/prusti-viper/src/encoder/errors/encoding_error.rs
+++ b/prusti-viper/src/encoder/errors/encoding_error.rs
@@ -71,33 +71,3 @@ impl From<SpannedEncodingError> for EncodingError {
         EncodingError::Spanned(other)
     }
 }
-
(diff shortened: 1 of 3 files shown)


Aurel300:
Looks good. I didn't realise before that these macros also included the `return`, but that would probably be the common usage anyway.

fpoli:
> Looks good. I didn't realise before that these macros also included the `return`, but that would probably be the common usage anyway.

Yeah. `throw_unsupported/internal/incorrect!` might be a better name. We can do the renaming in another PR.

PR #1281, opened 2023-01-17
Enable `encode_unsigned_num_constraint` by default

### LINKED ISSUE #1215: Type limits for unsigned integer missing when using `check_overflows = false`
`#[requires(0 <= i)]` generates a warning: `[unused_comparisons] comparison is useless due to type limits.`, but removing it with `check_overflows = false` will cause the verification to fail:
`[Prusti internal error] Prusti encountered an unexpected internal error`
`Details: unregistered verification error: [application.precondition:assertion.false; 0] Precondition of function lookup_pure__$TY$__Slice$i64$i64$Slice$i64$$int$$$int$ might not hold. Assertion 0 <= _3 might not hold. (@0.0)`.

Additionally, the error appears on line 1 column 1 in the lib.rs file, not in the file where the error happened.
<pre><code>use prusti_contracts::*;

// #[requires(0 <= i)]
#[requires(i < s.len())]
fn test(s: &[i64], i: usize) -> i64 {
    s[i]
}
</pre></code>

Patrick-6:
This commit partially fixes https://github.com/viperproject/prusti-dev/issues/1215.
The encoding of unsigned integers now contains the non-negative constraint by default, even if overflow checks are disabled.


prusti-tests/tests/verify/pass/issues/issue-1215.rs
```diff
@@ -0,0 +1,11 @@
+// compile-flags: -Pcheck_overflows=false
```
Aurel300: Files in the `verify` directory already have `check_overflows=false`, set [here](https://github.com/viperproject/prusti-dev/blob/01353f4f7b4b21daafaffa6cc0b8cc779d699dff/prusti-tests/tests/compiletest.rs#L135). Interestingly, our "base" also sets `encode_unsigned_num_constraint` to `true` [here](https://github.com/viperproject/prusti-dev/blob/01353f4f7b4b21daafaffa6cc0b8cc779d699dff/prusti-tests/tests/compiletest.rs#L127). So I think this test would pass even without the change you make in `config`. Please modify `compiletest.rs` such that the unsigned num constraint is not set.

Patrick-6: Resolved in #[50e7c00](https://github.com/viperproject/prusti-dev/pull/1281/commits/50e7c00a049ee53605fe08e64e08f2861cfdaf7f)

prusti-tests/tests/verify/pass/issues/issue-1215-1.rs
```diff
@@ -0,0 +1,11 @@
+// compile-flags: -Pcheck_overflows=false
```
Aurel300: As stated in a previous comment, these `compile-flags` should be removed, since you placed the tests in `verify` and not `verify_overflow`.

Patrick-6: Resolved in #[5c95497](https://github.com/viperproject/prusti-dev/pull/1281/commits/5c954970094fa257cc4ae497580d187895d13337)

PR #1284, opened 2023-01-24
Frontend improvements

### LINKED ISSUE #1283: `prusti_assert_eq` and `prusti_assert_ne`
Prusti currently supports the macros `assert` and `prusti_assert`.
There are also `assert_eq` and `assert_ne`, so support for macros like `prusti_assert_eq` and `prusti_assert_ne` might be useful.
They could check for snapshot equality `===` instead of the standard equality `==`:
```rust
use prusti_contracts::*;

#[requires(a === b)]
fn test_eq<T>(a: T, b: T) {
    prusti_assert_eq!(a, b);
}

#[requires(!(a === b))]
fn test_ne<T>(a: T, b: T) {
    prusti_assert_ne!(a, b);
}
```

### LINKED ISSUE #1297: Predicate function can only take a single expression
The following code fails on the marked line with an `unexpected token` error. The function inside the `predicate!` macro seems to only allow a single expression, but no statements unlike in a normal function:
```rust
use prusti_contracts::*;

fn non_predicate(x: i64) -> bool {
    let y = x + 1;
    y > x
}

predicate! {
    fn working_predicate(x: i64) -> bool {
        {
            let y = x + 1;
            y > x
        }
    }
}

predicate! {
    fn failing_predicate(x: i64) -> bool {
        let y = x + 1; // <-- Error: unexpected token `;`
        y > x
    }
}
```

### LINKED ISSUE #1299: Support for an if-and-only-if operator `<==>`
Support for if-and-only-if operator like `<==>` could make some specifications more readable.
With a lower precedence than other boolean expressions, some specifications could be written with fewer brackets, like in this example:
```rust
use prusti_contracts::*;

#[ensures((x == 5 || x == 7) == (result == 1))]
fn _current(x: u64) -> u64 {
    unimplemented!()
}

#[ensures(x == 5 || x == 7 <==> result == 1)]
fn _suggestion(x: u64) -> u64 {
    unimplemented!()
}
```

Aurel300:
Some changes to specs:

- `prusti_assert_eq!` and `prusti_assert_ne!` added as `macro_rules`, desugaring to a snapshot (in)equality check. Closes #1283. Error messages are a bit noisy because of the `macro_rules` implementation. We probably want macros that affect the backtrace we report in the long term anyway.
- `!==` is snapshot inequality, to avoid having to write `!(a === b)`.
- Fixed spans for implications, now includes the LHS and RHS, not only the `==>` operator.
- Fixed spans for quantifiers, now includes the `forall`/`exists` keyword.
- `<==` and `<==>` are now supported, closes #1299.
- Predicates can have multiple statements like regular functions, closes #1297.
- Predicates can return any `Copy` type, not just Booleans.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2186,7 +2186,7 @@ dependencies = [
 
 [[package]]
 name = "prusti-specs"
-version = "0.1.3"
+version = "0.1.4"
 dependencies = [
  "itertools",
  "proc-macro2",
--- a/docs/user-guide/src/syntax.md
+++ b/docs/user-guide/src/syntax.md
@@ -6,6 +6,8 @@ Prusti specifications are a superset of Rust boolean expressions. They must be d
 | --- | --- |
 | [`old(...)`](#old-expressions) | Value of expression in a previous state |
 | [`... ==> ...`](#implications) | Implication |
+| [`... <== ...`](#implications) | Implication |
+| [`... <==> ...`](#implications) | Biconditional |
 | [`... === ...`](#snapshot-equality) | Snapshot equality |
 | [`... !== ...`](#snapshot-equality) | Snapshot inequality |
 | [`forall(...)`](#quantifiers) | Universal quantifier |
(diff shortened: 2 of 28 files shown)


prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -217,18 +217,22 @@ impl<'tcx> ErrorManager<'tcx> {
...
-                New error context: {:?}",
-                pos.id(),
-                existing_error_ctxt,
-                error_ctxt
-            );
-        }
+        // TODO: disabled in #1284: when a predicate consists only of a call to
+        //   another function, the span of the call (PureFunctionCall) and the
+        //   span of the predicate definition (PureFunctionDefinition)
+        //   are the same
```
fpoli: Even if the span is the same, can't we encode them with different ids (`register_span(..)`)?

Aurel300: Hmm, yes. For some reason when encoding predicates the existing `.pos()` was used instead of registering a new one, which also caused this problem. Now I am creating a new position. I seem to remember there was a reason the position was assigned this way but let's see CI…

PR #1289, opened 2023-01-25
Run test-test crates with ubuntu-22.04

fpoli:
If the failures of `test-crates` are due to the missing libssl1.1 library, this should fix it in the short-term. Let's see.

(Removing the dependency on libssl1.1 seems better for the long-term, I agree.)


fpoli:
Nope

PR #1290, opened 2023-01-25
Add missing check in encoding of pure code

fpoli:


vakaras (approved):
LGTM. Thanks!

Aurel300:
Looks ok for now; maybe add a comment to that test that this should actually be supported in the future?

PR #1303, opened 2023-01-31
implement opt-in verification

### LINKED ISSUE #1187: Attribute and configuration flag to opt-in verification
Currently, Prusti tries to verify all functions that are *not* marked with `#[trusted]`. In large codebases the opposite would make more sense: only verify the functions that *are* marked with `#[verified]`. This way, users can gradually opt-in and verify more and more functions without getting scary "unsupported feature" errors on code that is not meant to be verified.

Pointerbender:
Adds `#[verified]` attribute and `opt_in_verification` configuration option plus tests and documentation, as discussed in #1187.

Closes #1187.

### DIFF
--- a/docs/dev-guide/src/config/flags.md
+++ b/docs/dev-guide/src/config/flags.md
@@ -295,7 +295,7 @@ When enabled, verification is skipped for dependencies. Equivalent to enabling `
 > **Note:** applied to all dependency crates when running with `cargo prusti`.
 
 ## `OPT_IN_VERIFICATION`
-When enabled, verification is only applied to functions that have the `#[verified]` attribute applied and all functions that do not have the `#[verified]` attribute are assumed to be `#[trusted]`, by default.
+When enabled, Prusti will only try to verify the functions annotated with `#[verified]`. All other functions are assumed to be `#[trusted]`, by default. Functions annotated with both `#[trusted]` and `#[verified]` will not be verified.
 
 ## `ONLY_MEMORY_SAFETY`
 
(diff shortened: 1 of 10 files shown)


Aurel300:
LGTM, thanks! CI failure looks spurious.

prusti-interface/src/specs/mod.rs
```diff
@@ -364,7 +365,8 @@ fn get_procedure_spec_ids(def_id: DefId, attrs: &[ast::Attribute]) -> Option<Pro
     );
 
     let pure = has_prusti_attr(attrs, "pure");
-    let trusted = has_prusti_attr(attrs, "trusted");
+    let trusted = has_prusti_attr(attrs, "trusted")
+        || (opt_in_verification() && !has_prusti_attr(attrs, "verified"));
```
fpoli: For consistency; we usually just import `config`.
```suggestion
        || (config::opt_in_verification() && !has_prusti_attr(attrs, "verified"));
```

prusti-interface/src/specs/mod.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
     PrustiError,
 };
 use log::debug;
+use prusti_common::config::opt_in_verification;
```
fpoli: For consistency; we usually just import `config`.
```suggestion
use prusti_common::config;
```

docs/dev-guide/src/config/flags.md
```diff
@@ -293,6 +294,9 @@ When enabled, verification is skipped for dependencies. Equivalent to enabling `
 
 > **Note:** applied to all dependency crates when running with `cargo prusti`.
 
+## `OPT_IN_VERIFICATION`
+When enabled, verification is only applied to functions that have the `#[verified]` attribute applied and all functions that do not have the `#[verified]` attribute are assumed to be `#[trusted]`, by default.
```
fpoli: To specify what happens for `#[trusted, verified]` functions.
```suggestion
When enabled, Prusti will only try to verify the functions annotated with `#[verified]`. All other functions are assumed to be `#[trusted]`, by default. Functions annotated with both `#[trusted]` and `#[verified]` will not be verified.
```

fpoli (approved):
Thanks! I forgot that it was also necessary to define the procedural macros.

PR #1314, opened 2023-02-04
Completion of issue 711 and general code improvements

Patrick-6:
This PR contains general code improvements:
- It should close #https://github.com/viperproject/prusti-dev/issues/711
  - Most uses of `std::collections::HashMap` and `std::collections::HashSet` have been replaced with [`rustc_hash::{FxHashMap, FxHashSet}`](https://docs.rs/rustc-hash/latest/rustc_hash/)
    - The only place that still uses the standard hashmap is [commandline.rs](https://github.com/viperproject/prusti-dev/blob/master/prusti-utils/src/config/commandline.rs#L104), since it is a type requirement for the `config` crates `Source` trait
  - Files using the [fxhash](https://docs.rs/fxhash/latest/fxhash/) crate have also been switched to `rustc_hash`
  - Files using `rustc_data_structures::fx::{FxHashMap, FxHashSet}` have not been changed
  - A Clippy lint has been added to warn if `std::collections::HashMap`, `std::collections::HashSet` or those from `fxhash` are used.
  - Note: There are many uses of `std::collections::{BTreeMap, BTreeSet}`, for some of which a `FxHashMap/Set` should work too
- Many uses of `sort` were replaced with `sort_unstable`, since their input came from a Map/Set so there are no duplicates (stable sort is not required)
- Some vectors now have their capacity pre-allocated to prevent resizing
- There was some code in [to_viper.rs](https://github.com/Patrick-6/prusti-dev/blob/master/prusti-common/src/vir/to_viper.rs#L38) that filled a `Vec`, and then immediately discarded it if `config::verify_only_preamble()` was set
- [error_manager.rs](https://github.com/Patrick-6/prusti-dev/blob/master/prusti-viper/src/encoder/errors/error_manager.rs) had the same long error message copied 4 times, it is now a constant
- The file [trait_register.rs](https://github.com/viperproject/prusti-dev/blob/master/prusti-interface/src/trait_register.rs) was removed, since it is not used anywhere
- Some other minor code improvements
- Some typos were fixed

The tests passed for me locally.

Should the Clippy lint for disallowed types be a warning or an error?


smt-log-analyzer/src/state.rs
```diff
@@ -616,8 +614,7 @@ impl State {
 
         {
             println!(
-                "The largest number of quantifier matches removed in a single \
-                “pop {}” operation: {}",
+                "The largest number of quantifier matches removed in a single pop {}” operation: {}",
```
Aurel300:
```suggestion
                "The largest number of quantifier matches removed in a single “pop {}” operation: {}",
```

Aurel300:
> Should the Clippy lint for disallowed types be a warning or an error?

It can be a warning, we [deny warnings anyway](https://github.com/viperproject/prusti-dev/blob/3dd957cf440edf04504f6b0ca8b57fac64e813a7/.github/workflows/test.yml#L51), so such usage would fail CI.

Issue #1193, opened 2022-10-13
Support for macOS arm64


fcoury:
I get the following error from running Prusti within the VSCode extension:

```
[stderr] dyld[62808]: Library not loaded: '@rpath/libjli.dylib'
  Referenced from: '/Users/fcoury/Library/Application Support/Code/User/globalStorage/viper-admin.prusti-assistant/prustiTools/LatestRelease/prusti/prusti-server-driver'
  Reason: tried: '/opt/homebrew/Cellar/openjdk@17/17.0.4.1_1/libexec/openjdk.jdk/Contents/Home/lib/libjli.dylib' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64)))
```

Seems like it's looking for the `x84_64` version of this library, which I obviously don't have due to my computer's architecture.

```
> $ java -version
openjdk version "17.0.4.1" 2022-08-12
OpenJDK Runtime Environment Homebrew (build 17.0.4.1+1)
OpenJDK 64-Bit Server VM Homebrew (build 17.0.4.1+1, mixed mode, sharing)
```

<img width="698" alt="image" src="https://user-images.githubusercontent.com/1371/195647438-d8bff237-9c04-4f20-97b5-885d454a18c4.png">

Gadiguibou:
I tried installing a x86_64 Java SDK and pointing Prusti Assistant to it using `prusti-assistant.javaHome` and it fixed the original error but I ran into the following error shortly after:

```
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `/Users/gabriel/Library/Application Support/Code/User/globalStorage/viper-admin.prusti-assistant/prustiTools/LatestRelease/prusti/prusti-rustc rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 255)
  --- stderr
  dyld[34221]: Library not loaded: @rpath/librustc_driver-88b66bfaeeb90970.dylib
    Referenced from: <0B8B796C-EA8F-3D3B-AFAB-2988B0AF6C6C> /Users/gabriel/Library/Application Support/Code/User/globalStorage/viper-admin.prusti-assistant/prustiTools/LatestRelease/prusti/prusti-driver
[...]

PR #1317, opened 2023-02-07
Compile for macOS ARM

fpoli:
Now that we no longer depend on `libssl1.1`, compiling for macOS ARM might work.

(Related to https://github.com/viperproject/prusti-dev/issues/1193)


fpoli:
Actually, some parts of Prusti (`prusti-tests` and `test-crates`) still depend on `libssl1.1`. So, compiling for macOS ARM on Github keeps failing: https://github.com/viperproject/prusti-dev/actions/runs/4111929054/jobs/7096233513

PR #1318, opened 2023-02-07
Normalize field types

jscissr:
With associated types, you can have two types that are equal, but you can only see that after you normalize the types. See https://rustc-dev-guide.rust-lang.org/ty.html#comparing-types

I'm not sure if this is the right way or place to do normalization, but at least tests pass.

In the mir encoder, I can't use try_normalize_erasing_regions, because the unsafe core proof needs lifetime information.
Here is an example how the Normalize type op is used in the Rust compiler: https://github.com/rust-lang/rust/blob/19423b59440f464c6cbe6be442d447e37b50fe3c/compiler/rustc_borrowck/src/type_check/free_region_relations.rs#L247

### DIFF
--- a/prusti-tests/tests/verify/pass/generic/associated-types.rs
+++ b/prusti-tests/tests/verify/pass/generic/associated-types.rs
@@ -1,0 +1,34 @@
+use prusti_contracts::*;
+
+trait Model {
+  type Kind;
+}
+
+struct MyModel {}
+
+impl Model for MyModel {
+  type Kind = bool;
+}
+
+struct Guard<M: Model> {
+  kind: M::Kind,
+}
+
+// <MyModel as Model>::Kind is the same type as bool,
+// but it need to be normalized to see that.
+fn test1(gd: Guard<MyModel>) -> bool {
+  gd.kind
+}
+
+// Snapshot types should also be normalized.
+#[pure]
+fn test2(gd: &Guard<MyModel>) -> bool {
+  gd.kind
+}
+
+// In this case, normalization fails, so it should
+// continue with the non-normalized type.
+fn test3<M: Model>(gd: Guard<M>) {}
+
+#[trusted]
+fn main() {}
--- a/prusti-viper/src/encoder/mir/types/encoder.rs
+++ b/prusti-viper/src/encoder/mir/types/encoder.rs
@@ -18,7 +18,10 @@ use prusti_common::config;
 use prusti_rustc_interface::{
     errors::MultiSpan,
     hir::def_id::DefId,
-    middle::{mir, ty},
+    middle::{
+        mir,
+        ty::{self, TypeVisitable},
+    },
 };
 use vir_crate::high::{self as vir, operations::ty::Typed};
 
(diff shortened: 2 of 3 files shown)


prusti-tests/tests/verify/pass/generic/associated-types.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+struct Guard<M: Model> {
+  kind: M::Kind,
+}
+
+// <MyModel as Model>::Kind is the same type as MyKind,
+// but it need to be normalized to see that.
+#[requires(gd.kind === gd.kind)]
+fn test1(gd: Guard<MyModel>){}
```
Aurel300: What does this test? Is it that snapshot equality is not supported for an unresolved associated type, but it is supported for a concrete `struct` type? Could `MyModel` set `type Kind = i32` and then:

```suggestion
#[requires(gd.kind === x)]
fn test1(gd: Guard<MyModel>, x: i32) {}
```

jscissr: I pushed a new version that tests the same thing but makes more sense.

jscissr:
# Summary

When there are associated types, we need to [normalize types](https://rustc-dev-guide.rust-lang.org/ty.html#comparing-types), otherwise we have multiple names for the same type, which causes errors in Prusti.

The standard way to normalize after type checking is `tcx.normalize_erasing_regions(ty)`. As the name says, this does not just normalize, but also removes lifetime (region) information.
If this function is used to normalize, then the test `prusti-tests/tests/verify_overflow/pass/core_proof/types.rs` fails with `Consistency error: Local variable lft_erased not found.`.
This is because the unsafe core proof needs the lifetime information.

There is a way to normalize without erasing types, `type_op::normalize::Normalize`, which is [used in the borrow checker](https://github.com/rust-lang/rust/blob/19423b59440f464c6cbe6be442d447e37b50fe3c/compiler/rustc_borrowck/src/type_check/free_region_relations.rs#L247).
This operation requires an `InferCtxt`. If we just pass an empty `InferCtxt`, then the `core_proof/types.rs` test passes.

However, if there are lifetimes in a type which needs normalization, then we get an `index out of bounds: the len is 0 but the index is 3` panic.
Here is an example where this panic happens:

```rust
trait IntoIterator {
  type IntoIter;
}

struct SomeIntoIter {}

impl IntoIterator for SomeIntoIter {
  type IntoIter = i32;
}

struct FlatMap<'a> {
  inner: (&'a i32, <SomeIntoIter as IntoIterator>::IntoIter),
}

fn test(f: FlatMap) {}

fn main() {}
```

There is a `RegionVid` in a `ty::ReVar` in a `ty::Region` inside the type.
When we call the normalize operation, this `RegionVid` is used as an index into the `UnificationTable` in the `TypeVariableTable` in the `RegionConstraintStorage` in the `InferCtxt`.
And because we passed an empty `InferCtxt`, this table is empty.

Prusti calls [`borrowck::consumers::get_body_with_borrowck_facts`](https://github.com/rust-lang/rust/blob/5157d938c49af1248a21e7ed2fbc5c6f71963276/compiler/rustc_borrowck/src/consumers.rs) to obtain the MIR body. This function creates the `InferCtxt`, performs borrow checking, and then throws away the `InferCtxt`. There is no way to obtain the `InferCtxt` with the public API.

For now, I will therefore use `normalize_erasing_regions`, but only if the unsafe core proof is not enabled, or if the type does not contain lifetimes.

---
Now, all tests on CI pass, except for the test which also fails on master.

PR #1321, opened 2023-02-13
Lithium: a Rust-native back end

JakuJ:
This is a draft PR outlining the work on a new Rust-native back end for Prusti, developed in connection to my Master's thesis. The aim is to, among others, evaluate the performance benefits of going straight from CFG to SMT-LIB and decoupling from Viper.

The project is in its initial stages and so the code is highly volatile. Everything is subject to change. The purpose of this PR is to give Prusti developers an overview of the work being done in order to avoid double work, conflicts etc.


fpoli:
Wow, thanks! I'm not sure why the CI is not running. Even if it's a draft, I was wondering if you considered moving `native_verifier` *outside* `viper`, because the API of a Viper and a native one have various differences (e.g. parsing command line parameters, the type of the `Program`, depending on the JVM...).

JakuJ:
I think the CI doesn't run because of the merge conflicts. I'll move the code to a separate create as soon as I figure out how to avoid circular dependencies.

prusti-server/src/process_verification.rs
```diff
@@ -14,9 +14,88 @@ use prusti_common::{
...
+            return result;
+        }
+    };
+
+    assert!(request.backend_config.backend == VerificationBackend::Lithium);
+
+    // Create a new verifier each time.
+    // A workaround for https://github.com/viperproject/prusti-dev/issues/744
+    // TODO: Is it still needed?
+    let mut stopwatch = Stopwatch::start("prusti-server", "verifier startup");
```
vakaras: `prusti-server` is used for accommodating the JVM startup time and should not be needed with the native backend.

JakuJ: Yes, I intend to purge all my changes from `prusti-server` eventually, but for now I just plugged this in to make the compiler happy.

fpoli: > `prusti-server` is used for accommodating the JVM startup time and should not be needed with the native backend.

Do you mean that the message is misleading? The server does provide caching and a uniform API to verify programs.

vakaras: `prusti-server` is JVM dependent. If we want to use it for caching/network-based verification, I think we should create a server trait that is implemented by Viper and native servers.

fpoli: Good point, the JVM dependency is bad.

prusti-server/src/client.rs
```diff
@@ -30,6 +30,27 @@ impl PrustiClient {
     pub async fn verify(
         &self,
         request: VerificationRequest,
+    ) -> reqwest::Result<VerificationResult> {
```
vakaras: I suggest creating a version of `VerificationResult` that is not Viper dependent and using it here instead.

prusti-server/src/client.rs
```diff
@@ -30,6 +30,27 @@ impl PrustiClient {
...
+        let response = if let VerificationBackend::Lithium = request.backend_config.backend {
+            self.verify_natively(request).await
```
vakaras: Instead of branching on the client side, I would suggest branching on the server side because this allows doing (at least theoretically) the heavy verification task on a different machine. (The user can still disable the Prusti server completely if they want to verify locally and avoid the serialization overhead.)

prusti-server/src/process_verification.rs
```diff
@@ -104,11 +183,21 @@ pub fn process_verification_request<'v, 't: 'v>(
...
+        let backend = request.backend_config.backend;
         let mut verifier =
             new_viper_verifier(program_name, verification_context, request.backend_config);
 
         stopwatch.start_next("verification");
-        let mut result = verifier.verify(viper_program);
+        let mut result = if let VerificationBackend::Lithium = backend {
+            if let prusti_common::vir::program::Program::Low(ref low_program) = request.program {
+                stopwatch.start_next("verification");
+                verifier.verify_vir(low_program)
```
vakaras: I would suggest having a `Verifier` trait that is implemented by `ViperVerifier` and `NativeVerifier`. That trait then would have a `verify` method that accepts a VIR program, which `ViperVerifier` would internally convert to a Viper program.

viper/src/verifier.rs
```diff
@@ -4,345 +4,14 @@
...
-                }
-
-                VerificationResult::Failure(errors)
-            } else {
-                VerificationResult::Success
-            }
-        })
-    }
-}
+    fn verify(&mut self, program: crate::Program) -> VerificationResult;
```
vakaras: I would remove this method because this way, you could have the native backend completely independent of Viper.

vakaras:
The PR looks like a good start. I think that once it is cleaned up, we should merge it in with a stub native verifier so that we do not need to worry about getting large conflicts.

cc @Aurel300 @JonasAlaif @fpoli

Aurel300:
We'll probably soon get some merge conflicts in `prusti-server` since Cedric and Joseph will be finishing the IDE improvements project soon.

prusti-server/src/client.rs
JakuJ: I moved the `VerificationResult` struct to another crate `backend-common`. The `viper` crate and my (future) native verifier crate depend on it.

prusti-server/src/process_verification.rs
JakuJ: All test projects expect a `verify` method that accepts a Viper program. I exposed such method on the `JavaVerifier` so that nothing needs to be renamed in test projects. The method I placed on the trait is named `verify_vir` and accepts a `Program::Legacy` or `Program::Low`, with the `JavaVerifier` transforming it to Viper before passing it to its `verify` method.

vakaras:
@JakuJ In our meeting, we discussed what would be the best way to proceed and agreed that the work should be split into two PRs:
1. A clean-up PR that prepares code to allow easily adding a new backend. This PR should be merged as soon as possible to avoid conflicts.
2. A PR that implements the actual backend – this PR would be merged only if this project worked out.

The clean-up PR should add `VerificationBackend` enum to `prusti-server` that abstracts over Viper and your new backend. We decided that an enum will be a cleaner solution for now than having a `Verifier` trait, which I suggested before.

```rust=
enum VerificationBackend {
    Viper(viper::Viper),
//    Lithium(lithium::Lithium), – only added later once the backend is ready
}

impl VerificationBackend {
    pub fn verify(&self, program: vir::Program) -> VerificationResult {
        match self { /* delegate */ }
    }
}
```

Notice that the `verify` method takes the VIR program (that is either legacy or low) as input. This means that all JAVA stuff should happen inside the `verify` method.

prusti-server/src/process_verification.rs
vakaras: Since the test projects are expected to construct the `JavaVerifier` directly, this sounds good to me.

fpoli:
* Use `git rebase -i master` to remove 4b055b7b4edb49a03141732dc1a64285129b9dcc and 5484ca74c2d742325f4933a61d4c0e20fc5c403c.
* Remove `.DS_Store` (from the history of the PR, not in a new commit) and add it to `.gitignore`.

fpoli:
Is there a clear advantage in splitting `backend-common` out of `viper`? All its current content (java exceptions, verification results, silicon counterexamples) can still be seen as part of Viper, and I'd avoid one more top-level folder if it's not necessary.

JakuJ:
A test or two are still failing, but I'm going to switch from draft mode now. I suppose some quality-of-life improvements are necessary for the general public to find this usable, in particular:

- [ ] Better error handling and error messages
- [ ] A single configuration flag to run the native back end, instead of modifying half the `config.rs` file

I'll be chipping away at these improvements when I have the time, and in the meantime I would appreciate a code review.

fpoli:
I notice that the PR runs `cargo fmt` on `prusti-viper`, but from the diff it's difficult to see if you made more changes in there. Could you try to reformat `prusti-viper` on the `master` branch, opening a single-commit PR that we can immediately merge, and rebasing at the same time this PR on top of that commit?

prusti-server/src/backend.rs
```diff
@@ -36,6 +38,10 @@ impl<'a> Backend<'a> {
                     viper.verify(viper_program)
                 })
             }
+            Backend::Lithium(lithium) => {
+                Stopwatch::start("prusti-server", "vir verification");
```
fpoli: Without this, the stopwatch is immediately dropped and does *not* measure how long the verification takes.
```suggestion
                let _stopwatch = Stopwatch::start("prusti-server", "vir verification");
```

test.sh
fpoli, line 1: Could you move this script into the `scripts` folder?

prusti-utils/src/config.rs
```diff
@@ -190,6 +191,17 @@ lazy_static::lazy_static! {
...
+        if let Ok(true) = env::var("PRUSTI_NATIVE_VERIFIER").map(|s| s.parse::<bool>().unwrap_or(false)) {
+            settings.set_default("viper_backend", "Lithium").unwrap();
+            settings.set_default("verify_specifications_backend", "Lithium").unwrap();
+            settings.set_default("encode_bitvectors", true).unwrap();
+            settings.set_default("enable_purification_optimization", true).unwrap();
+            settings.set_default("unsafe_core_proof", true).unwrap();
+            settings.set_default("verify_core_proof", false).unwrap();
+            settings.set_default("inline_caller_for", true).unwrap();
+        }
+
```
fpoli, line 204: What is the use case of this new `PRUSTI_NATIVE_VERIFIER`?

JakuJ: Grouping all required flags into one, to facilitate switching between the backends. Alternatively I could check if `viper_backend == "Lithium"`.

Why an environment variable? For some reason if I do

```rust
if settings.get_bool("native_verifier").unwrap_or(false) {
   ...
}
```

at the end (after the `-P` arguments) and provide it via `cargo-prusti -- -Pnative_verifier=true` it breaks everything, including the building of `prusti-contracts` 🤨

fpoli: There are various reasons why I'd suggest to undo this change and look for another solution:
* The `native_verifier` and `viper_backend` flags are partially overlapping.
* The code checks the environment variable `PRUSTI_NATIVE_VERIFIER`, but does not work if the same flag is set via `-P`. This goes against the current design of the configuration flags.
* Changing the default of some flags depending on the value of some flags looks like the kind of thing that easily leads to surprises and hard-to-debug issues. We were forced to do it in one case (`DEFAULT_PRUSTI_*`) because of the way cargo works, but in this PR I think that we have more freedom to choose a better solution.

If Lithium *requires* a specific set of flags, what about just implementing runtime checks like we do [here](https://github.com/viperproject/prusti-dev/blob/a5c29c994cee03e1ba02c3bc2c2761803571d3f5/prusti-utils/src/config.rs#L602-L605)? Moreover, if a set of tests require a long combination of flags, what about adding one entry [here](https://github.com/viperproject/prusti-dev/blob/a5c29c994cee03e1ba02c3bc2c2761803571d3f5/prusti-tests/tests/compiletest.rs#L162-L219), with a new test folder such as `prusti-tests/tests/verify_lithium`?

PR #1328, opened 2023-02-18
Enable tracing support

JonasAlaif:
Running with e.g. `PRUSTI_LOG=debug` will now generate a `log/trace.json` file which can be opened in [ui.perfetto.dev](https://ui.perfetto.dev/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,7 @@ dependencies = [
  "serde",
  "serde_json",
  "syn",
+ "tracing 0.1.0",
 ]
 
 [[package]]
@@ -1230,7 +1231,7 @@ dependencies = [
  "slab",
  "tokio",
  "tokio-util",
- "tracing",
+ "tracing 0.1.37",
 ]
 
 [[package]]
@@ -1363,7 +1364,7 @@ dependencies = [
  "socket2",
  "tokio",
  "tower-service",
- "tracing",
+ "tracing 0.1.37",
  "want",
 ]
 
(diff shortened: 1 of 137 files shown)


fpoli:
Does the tracing slow down the encoding, when `PRUSTI_LOG=info`? With the `log` crate that was the case, which is why we had to add this feature to ignore all `debug!` and `trace!` calls when compiling in release mode: https://github.com/viperproject/prusti-dev/blob/8fd6386bbfb3eb36cb0c249049e54ec7e848830f/prusti-viper/Cargo.toml#L14

JonasAlaif:
> Does the tracing slow down the encoding, when `PRUSTI_LOG=info`?

For me running in release on `bst_generics.rs` with `PRUSTI_PRINT_HASH=true` (to stop just before Viper stuff) for two variants: a) `tracing` enabled as normal, b) the `tracing::instrument` macro changed to do nothing (i.e. all tracing removed at compile time) I got:

With `PRUSTI_LOG=trace` (generates 750MB trace file with a):
a) ~3.5s (+-0.5s)
b) ~1.72s (+-0.05s)

With `PRUSTI_LOG=debug` (generates 13.6MB trace file with a):
a) ~1.85s (+-0.05s)
b) ~1.72s (+-0.05s)

With `PRUSTI_LOG=info` (generates 8KB trace file with a):
a) ~1.72s (+-0.05s)
b) ~1.72s (+-0.05s)

Ofc this isn't very rigorous but it seems to me that even though the ability to do tracing being included in to code (and checking if to do it), it doesn't seem to slow Prusti down.

PR #1329, opened 2023-02-19
Refactoring to allow for new verification back ends

JakuJ:
Related to #1321 – this PR aims to abstract away the notion of a verification back end behind an enum, in order to facilitate implementation of new back ends.

Had to make `env` on `viper::Verifier` public to do this. I cannot put the call to `program.to_viper()` in `viper` since that would be a cyclic dependency to `prusti-common`, and returning an `AstFactory` from some method would cause [E0502](https://doc.rust-lang.org/error_codes/E0502.html) on call to `viper.verify()`.

### DIFF
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -6,7 +6,7 @@ pub enum Backend<'a> {
 }
 
 impl<'a> Backend<'a> {
-    pub fn verify(&mut self, program: &prusti_common::vir::Program) -> VerificationResult {
+    pub fn verify(&mut self, program: &prusti_common::vir::program::Program) -> VerificationResult {
         match self {
             Backend::Viper(viper) => {
                 let ast_factory = AstFactory::new(&viper.env);
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -1,0 +1,19 @@
+use prusti_common::vir::ToViper;
+use viper::{AstFactory, VerificationResult};
+
+pub enum Backend<'a> {
+    Viper(viper::Verifier<'a>),
+}
+
+impl<'a> Backend<'a> {
+    pub fn verify(&mut self, program: &prusti_common::vir::Program) -> VerificationResult {
+        match self {
+            Backend::Viper(viper) => {
+                let ast_factory = AstFactory::new(&viper.env);
+                let program =
+                    program.to_viper(prusti_common::vir::LoweringContext::default(), &ast_factory);
+                viper.verify(program)
+            }
+        }
+    }
+}
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -32,6 +32,7 @@ impl<'a> Backend<'a> {
                         );
                     }
 
+                    stopwatch.start_next("viper verification");
                     viper.verify(viper_program)
                 })
             }
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -9,7 +9,7 @@ impl<'a> Backend<'a> {
     pub fn verify(&mut self, program: &prusti_common::vir::program::Program) -> VerificationResult {
         match self {
             Backend::Viper(viper) => {
-                let ast_factory = AstFactory::new(&viper.env);
+                let ast_factory = AstFactory::new(viper.env);
                 let program =
                     program.to_viper(prusti_common::vir::LoweringContext::default(), &ast_factory);
                 viper.verify(program)
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -12,7 +12,7 @@ impl<'a> Backend<'a> {
     pub fn verify(&mut self, program: &prusti_common::vir::program::Program) -> VerificationResult {
         match self {
             Backend::Viper(viper, context) => {
-                let mut stopwatch = Stopwatch::start("prusti-server", "verifier startup");
+                let mut stopwatch = Stopwatch::start("prusti-server backend", "construction of JVM objects");
 
                 let ast_factory = context.new_ast_factory();
                 let viper_program = program.to_viper(LoweringContext::default(), &ast_factory);
--- a/prusti-server/src/backend.rs
+++ b/prusti-server/src/backend.rs
@@ -1,18 +1,26 @@
-use prusti_common::vir::ToViper;
-use viper::{AstFactory, VerificationResult};
+use prusti_common::{
+    vir::{LoweringContext, ToViper},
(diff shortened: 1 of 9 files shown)


vakaras (changes requested):
Could you please move all the Java things into `Backend::Viper`? For example, the [construction of the `Viper` object](https://github.com/viperproject/prusti-dev/blob/8d1e3129a498f7b4c69302481add7edfbae32cba/prusti-server/src/server.rs#L49-L52), [build Viper program closure](https://github.com/viperproject/prusti-dev/blob/8d1e3129a498f7b4c69302481add7edfbae32cba/prusti-server/src/process_verification.rs#L47-L64), etc.

fpoli:
> the [construction of the Viper object](https://github.com/viperproject/prusti-dev/blob/8d1e3129a498f7b4c69302481add7edfbae32cba/prusti-server/src/server.rs#L49-L52)

We just discussed this with Vytautas: it should be enough to store Viper inside an `Arc<Lazy<..>>` instead of inside `Arc<..>` as it is now. There us no need to move the local variable somewhere else. See the Lazy [documentation](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html).

fpoli:
If you remove the ` verification_context: &'v VerificationContext<'t>` argument and make it a field of `
```rust
enum Backend<'v> {
    Viper { verifier: viper::Verifier<'v>, verification_context: Rc<VerificationContext<'v>> }
}
```
then making the `env` field public should no longer be needed. Let us know if this suggestion causes weird lifetime errors.

fpoli:
> etc

The `ast_utils.with_local_frame(16, || { .. })` wrapper can be moved too inside the `Backend::verify` method, so that it's used only for the Viper verifiers.

JakuJ:
I made it so that the `viper` object is now initialized lazily, both on the server as well as in the `prusti-viper` package. Reorganized code around the backend call. Let me know how it looks now.

vakaras (commented):
Looks good to me.

@fpoli Could you please check that `Lazy` is used as you intended?

prusti-server/src/backend.rs
```diff
@@ -0,0 +1,27 @@
...
+impl<'a> Backend<'a> {
+    pub fn verify(&mut self, program: &prusti_common::vir::program::Program) -> VerificationResult {
+        match self {
+            Backend::Viper(viper, context) => {
+                let mut stopwatch = Stopwatch::start("prusti-server", "verifier startup");
+
+                let ast_factory = context.new_ast_factory();
+                let viper_program = program.to_viper(LoweringContext::default(), &ast_factory);
+
+                context.new_ast_utils().with_local_frame(16, || {
```
fpoli: Move `.with_local_frame(16, || {` earlier, so that also `let ast_factory = ` is inside the local frame. We use `with_local_frame` as a safeguard, to be sure that we let the JVM deallocate objects that we constructed while constructing the viper program or while verifying.

prusti-server/src/process_verification.rs
```diff
@@ -97,32 +98,33 @@ pub fn process_verification_request<'v, 't: 'v>(
...
-        let program_name = request.program.get_name();
-
-        // Create a new verifier each time.
-        // Workaround for https://github.com/viperproject/prusti-dev/issues/744
-        let mut stopwatch = Stopwatch::start("prusti-server", "verifier startup");
-        let mut verifier =
-            new_viper_verifier(program_name, verification_context, request.backend_config);
+    // Create a new verifier each time.
+    // Workaround for https://github.com/viperproject/prusti-dev/issues/744
+    let mut backend = match request.backend_config.backend {
```
fpoli, line 105: Add the `Stopwatch::start("prusti-server", "verifier startup");` and `stopwatch.start_next("verification");` that you deleted, so that we still log how much time those operations take.

prusti-server/src/backend.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub enum Backend<'a> {
+    Viper(viper::Verifier<'a>, &'a VerificationContext<'a>),
+}
+
+impl<'a> Backend<'a> {
+    pub fn verify(&mut self, program: &prusti_common::vir::program::Program) -> VerificationResult {
+        match self {
+            Backend::Viper(viper, context) => {
+                let mut stopwatch = Stopwatch::start("prusti-server", "verifier startup");
```
fpoli:
```suggestion
                let mut stopwatch = Stopwatch::start("prusti-server backend", "construction of JVM objects");
```

prusti-server/src/backend.rs
```diff
@@ -0,0 +1,27 @@
...
+
+                let ast_factory = context.new_ast_factory();
+                let viper_program = program.to_viper(LoweringContext::default(), &ast_factory);
+
```
fpoli, line 21: When constructiong a Viper program, we also need to conditionally dump it for debugging. So, add something like
```rust
if config::dump_viper_program() {
    stopwatch.start_next("dumping viper program");
    dump_viper_program(
        &ast_utils,
        viper_program,
        program.get_name_with_check_mode(),
    );
}
```

fpoli (changes requested):
`Lazy` is perfect, thanks! The dump disappeared and the log messages need to be updated, then it's good to be merged.

vakaras (approved):

prusti-server/src/process_verification.rs
```diff
@@ -97,35 +98,43 @@ pub fn process_verification_request<'v, 't: 'v>(
...
+                verification_context,
+                request.backend_config,
+            ),
+            verification_context,
+        ),
+    };
 
-        stopwatch.start_next("verification");
-        let mut result = verifier.verify(viper_program);
+    stopwatch.start_next("verification");
```
fpoli:
```suggestion
    stopwatch.start_next("backend verification");
```

prusti-server/src/backend.rs
```diff
@@ -0,0 +1,40 @@
...
+                    if config::dump_viper_program() {
+                        stopwatch.start_next("dumping viper program");
+                        dump_viper_program(
+                            &ast_utils,
+                            viper_program,
+                            &program.get_name_with_check_mode(),
+                        );
+                    }
+
+                    viper.verify(viper_program)
```
fpoli, line 36:
```suggestion
                    stopwatch.start_next("viper verification");
                    viper.verify(viper_program)
```

fpoli (approved):

fpoli:
I rebased on master.

vakaras:
The CI failure seems to be unrelated to this PR. @fpoli If you agree, please merge.

PR #1331, opened 2023-02-21
Add macro to generate getters/setters for fields via JNI

simon-hrabec:
This PR allows using a new macro `field_getter_setter` in the `viper-sys/build.rs`. For a given java class it will generate a getter/setter (rust code using JNI) for a certain field.

When reviewing this code please note that I have no prior experience with either Rust or Scala - hence there might be beginner mistakes and if you believe I did something because there had to be a specific reason there is probably none.

### DIFF
--- a/jni-gen/src/errors.rs
+++ b/jni-gen/src/errors.rs
@@ -54,5 +54,10 @@ error_chain::error_chain! {
             description("no matching method")
             display("no method '{}' with signature '{}' in class '{}'", method, signature, class)
         }
+
+        NoField(class: String, field: String) {
+            description("no field")
+            display("no field '{}' in class'{}'", field, class)
+        }
     }
 }
--- a/jni-gen/src/errors.rs
+++ b/jni-gen/src/errors.rs
@@ -57,7 +57,7 @@ error_chain::error_chain! {
 
         NoField(class: String, field: String) {
             description("no field")
-            display("no field '{}' in class'{}'", field, class)
+            display("no field '{}' in class '{}'", field, class)
         }
     }
 }
--- a/jni-gen/src/generators/field_getter_setter.rs
+++ b/jni-gen/src/generators/field_getter_setter.rs
@@ -7,46 +7,68 @@
 use crate::{class_name::*, errors::*, utils::*};
 use jni::JNIEnv;
 
-fn generate_type_check(type_signature: &str, type_name: &str, prefix: &str) -> String {
+fn generate_type_check(type_signature: &str, type_name: &str, is_result: bool) -> String {
+    if !type_signature.starts_with('L') {
+        return "".to_string();
+    }
+
     let type_class = &type_signature[1..(type_signature.len() - 1)];
-    vec![
+
+    let type_check = vec![
         "    debug_assert!(".to_string(),
         "        self.env.is_instance_of(".to_string(),
-        format!("            {},", type_name),
-        format!("            self.env.find_class(\"{}\")?,", type_class),
+        format!("            {type_name},"),
+        format!("            self.env.find_class(\"{type_class}\")?,"),
         "        )?".to_string(),
         "    );".to_string(),
-    ].iter().map(|x| format!("{}{}", prefix, x)).collect::<Vec<_>>().join("\n")
+    ];
+
+    if is_result {
+        let indented_type_check = type_check
+            .iter()
+            .map(|x| format!("    {x}"))
+            .collect::<Vec<_>>()
+            .join("\n");
+
+        vec![
+            "    if let Ok(result) = result {".to_string(),
+            indented_type_check,
+            "    }".to_string(),
+        ]
+    } else {
+        type_check
+    }
+    .join("\n")
+        + "\n"
 }
 
 fn generate_result_type_check(return_signature: &str) -> String {
-    vec![
-        "    if let Ok(result) = result {".to_string(),
-        generate_type_check(return_signature, "result", "    "),
-        "    }".to_string(),
-        "".to_string(),
-    ].join("\n")
+    generate_type_check(return_signature, "result", true)
 }
 
 fn generate_parameter_type_check(type_signature: &str, type_name: &str) -> String {
-    generate_type_check(type_signature, type_name, "")+"\n"
+    generate_type_check(type_signature, type_name, false)
 }
 
 fn generate_field_getter(class_name: &ClassName, field_name: &str, type_signature: &str) -> String {
-    let rust_getter_name = format!("call_getter_{}", java_identifier_to_rust(field_name));
+    let rust_getter_name = format!("get_{}", java_identifier_to_rust(field_name));
     let parameter_type = generate_jni_type(type_signature);
 
     vec![
-        format!("/// Returns the field `{}` of the scala class `{}`.", field_name, class_name.full_name()),
+        format!(
+            "/// Returns the field `{}` of the scala class `{}`.",
+            field_name,
+            class_name.full_name()
+        ),
         "///".to_string(),
-        format!("/// Return type and Java signature: `{}` (`{}`)", parameter_type, type_signature),
+        format!("/// Return type and Java signature: `{parameter_type}` (`{type_signature}`)"),
         format!("pub fn {rust_getter_name}("),
         "    &self,".to_string(),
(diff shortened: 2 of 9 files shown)


jni-gen/src/generators/method.rs
```diff
@@ -190,9 +190,12 @@ fn generate(
...
     ));
-    code.push("///".to_string());
-    code.push("/// Type and Java signature of parameters:".to_string());
-    code.push("///".to_string());
+
+    if !parameter_names.is_empty() {
+        code.push("///".to_string());
+        code.push("/// Type and Java signature of parameters:".to_string());
+        code.push("///".to_string());
+    }
```
simon-hrabec, line 198: When a method does not have any parameters it will not print the `/// Type and Java signature of parameters` followed by no parameters.

jni-gen/src/generators/method.rs
```diff
@@ -451,8 +454,3 @@ fn generate_static(
 
     code.join("\n") + "\n"
 }
-
-fn java_method_to_rust(method_name: &str) -> String {
-    method_name.replace('_', "__").replace('$', "_dollar_")
-    // If needed, replace other charachters with "_{something}_"
-}
```
simon-hrabec, line 458: Moved to utils.rs and made it reusable.

jni-gen/src/utils.rs
```diff
@@ -72,11 +72,10 @@ pub fn java_str_to_string(string: &JavaStr) -> Result<String> {
...
 
 pub fn java_str_to_valid_rust_argument_name(string: &JavaStr) -> Result<String> {
-    let mut res = "arg_".to_string();
-    res.push_str(
-        &java_str_to_string(string)?
-            .replace('_', "___")
-            .replace('$', "_d_"),
-    );
-    Ok(res)
+    Ok(format!("arg_{}", java_identifier_to_rust(&java_str_to_string(string)?)))
```
simon-hrabec: Here is a change - I made it all rely on `java_identifier_to_rust`. However, before there was a change that argument name would have 3 underscores instead of 2 and a dollar would be represented as `_d_` instead of `_dollar_`. I unified the format to double underscores `__` for an underscore and `_dollar_` for a dollar sign.

simon-hrabec:
When I added generated the getter/setter for the field `_program` (protected, declared in SilFrontend) for `SiliconFrontend`, it generates the following code:
```rust
/// Returns the field `_program` of the scala class `viper.silicon.SiliconFrontend`.
///
/// Return type and Java signature: `JObject` (`Lscala/Option;`)
pub fn call_getter___program(
    &self,
    receiver: JObject<'a>,
) -> JNIResult<JObject> {
    let class_name = "viper/silicon/SiliconFrontend";
    let field_name = "_program";
    let return_signature = "Lscala/Option;";

    debug_assert!(
        self.env.is_instance_of(
            receiver,
            self.env.find_class(class_name)?,
        )?
    );

    let result = self.env.get_field(
        receiver,
        field_name,
        return_signature,
    ).and_then(|x| x.l());

    if let Ok(result) = result {
        debug_assert!(
            self.env.is_instance_of(
                result,
                self.env.find_class("scala/Option")?,
            )?
        );
    }

    result
}

/// Sets the field `_program` of the scala class `viper.silicon.SiliconFrontend`.
///
/// Type and Java signature of parameters:
///
/// - `new__program`: `JObject` (`Lscala/Option;`)
/// 
/// Return type and Java signature: `()` (`V`)
pub fn call_setter___program(
    &self,
    receiver: JObject<'a>,
    new__program: JObject,
) -> JNIResult<()> {
    let class_name = "viper/silicon/SiliconFrontend";
    let field_name = "_program";
    let return_signature = "Lscala/Option;";

    debug_assert!(
        self.env.is_instance_of(
            receiver,
            self.env.find_class(class_name)?,
        )?
    );

    debug_assert!(
        self.env.is_instance_of(
            new__program,
            self.env.find_class("scala/Option")?,
        )?
    );

    let result = self.env.set_field(
        receiver,
        field_name,
        return_signature,
        JValue::from(new__program)
    );

    result
}
```

simon-hrabec:
Screenshot of the above code
Getter
image

Setter
image

JonasAlaif:
You can locally run `./x.py fmt-all` (will auto-fix) and `./x.py clippy` (won't auto-fix) to resolve the CI issues

JonasAlaif:
Otherwise lgtm quickly scanning over it, I'll take a proper look later today. One change that is definitely required: please add some tests of your macro to `jni-gen/systest/`.

fpoli:
Nice, thanks! What about the following renamings?
- the  `field_getter_setter` macro to just `field`
- the `call_getter___program` getters to e.g. `get___program` ("call" was for methods)
- the `call_setter___program` to e.g. `set___program`

simon-hrabec:
Removed the WIP tag, ready to be reviewed now.

Merged the branch with the update to the macro - no need to specify the signature/type of the field. However, it requires iterating over the class hierarchy as we cannot check private fields that are inherited.

Tried creating a binding for a non-existing field and got the proper error.

Also added some tests to see if I can get a simple (integer) type as well as an object type (in this case it is also inherited, so it tests that functionality).

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -0,0 +1,231 @@
...
+        "        return_signature,".to_string(),
+        format!("        JValue::from({parameter_name})"),
+        "    )".to_string(),
+        "}".to_string(),
+    ]
+    .join("\n")
+        + "\n"
+}
+
+pub fn generate_field_getter_setter(
```
fpoli, line 214: Can you add a doc comment to this `pub fn`, mentioning the hierarchy field lookup?

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        if field_name == lookup_name {
+            return Ok(Some(field));
+        }
+    }
+
+    Ok(None)
+}
+
+fn generate_type_check(type_signature: &str, type_name: &str, is_result: bool) -> String {
```
fpoli: Even if it's private, can you add a doc comment to remember the `!type_signature.starts_with('L')` case and the `is_result` case?

Also, `type_name` looks confusing to me. Could you do this renaming (or abetter one) here and in the two callers?
* `type_signature` -> `expected_variable_signature`
* `type_name` -> `variable_name`

fpoli (commented):
Nice, thanks! Two minor comments.

jni-gen/src/generators/field_getter_setter.rs
simon-hrabec: Added the following comment:
/// Generates Rust code to both retrieve and modify a value of
/// the specified field of a given object of the specified class
///
/// It works also for private fields and for inherited fields
///
/// It determines the type of the field by iterating over the
/// inheritance hierarchy and checking for a field with matching name
///
/// For class type fields it also generates runtime checks verifying
/// that the given object is of the specified class (or its descendant)

jni-gen/src/generators/field_getter_setter.rs
simon-hrabec: Added the following comment and changed the variable names:
```
// Generates a runtine check that the object is of the expected class - applies only to object types (starting with L)
// If this value is also the returned value in the setter, unwrap it first (since it is of type Result<T>)
```

EDIT: I see there is a typo

fpoli:
See the failing fmt-check test

jni-gen/src/errors.rs
```diff
@@ -54,5 +54,10 @@ error_chain::error_chain! {
             description("no matching method")
             display("no method '{}' with signature '{}' in class '{}'", method, signature, class)
         }
+
+        NoField(class: String, field: String) {
+            description("no field")
+            display("no field '{}' in class'{}'", field, class)
```
fpoli:
```suggestion
            display("no field '{}' in class '{}'", field, class)
```

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -0,0 +1,245 @@
+// © 2019, ETH Zurich
```
fpoli:
```suggestion
// © 2023, ETH Zurich
```

PR #1304, opened 2023-02-01
Update dependencies (rustc nightly-2023-02-01, viper v-2023-01-31-0912)

viper-admin:
* [x] Update Viper version to `v-2023-01-31-0912`.
* [x] Update rustc version to `nightly-2023-02-01`.
* [x] Run `cargo audit` and fix the issues.
* [x] Manualy update outdated dependencies (see the list below).
* [x] Manualy run `cargo update`.

<details><summary>List of direct outdated dependencies:</summary>

```
$ mv prusti-contracts/Cargo.toml prusti-contracts/Cargo_disabled.toml
$ cargo outdated --root-deps-only --workspace

info: syncing channel updates for 'nightly-2023-02-01-x86_64-unknown-linux-gnu'
warning: Signature verification failed for 'https://static.rust-lang.org/dist/2023-02-01/channel-rust-nightly.toml'
info: latest update on 2023-02-01, rust version 1.69.0-nightly (dc1d9d50f 2023-01-31)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'llvm-tools'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustc-dev'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'llvm-tools'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustc-dev'
info: installing component 'rustfmt'
    Updating git repository `https://github.com/rust-lang/cargo.git`
warning: skipping duplicate package `cargo-list-test-fixture` found at `/home/runner/.cargo/git/checkouts/cargo-e7ff1db891893a9e/17f8088/tests/testsuite/cargo_add/overwrite_default_features/in`
warning: skipping duplicate package `cargo-list-test-fixture` found at `/home/runner/.cargo/git/checkouts/cargo-e7ff1db891893a9e/17f8088/tests/testsuite/cargo_add/overwrite_default_features_with_no_default_features/in`
warning: skipping duplicate package `cargo-list-test-fixture` found at `/home/runner/.cargo/git/checkouts/cargo-e7ff1db891893a9e/17f8088/tests/testsuite/cargo_add/add-basic.in`
warning: skipping duplicate package `cargo-list-test-fixture` found at `/home/runner/.cargo/git/checkouts/cargo-e7ff1db891893a9e/17f8088/tests/testsuite/cargo_add/require_weak/in`
[...]

PR #1332, opened 2023-02-21
Support unevaluated constants in specs

### LINKED ISSUE #1268: Unsupported constant type `Ref(...)`
Logical Equality currently does not work for enums:
```
use prusti_contracts::*;

pub enum TestEnum {
    Nil,
}

#[ensures(TestEnum::Nil === TestEnum::Nil)]
fn test() {}
```
Error on `=== TestEnum::Nil`:
```
[Prusti: unsupported feature]
unsupported constant type Ref(ReErased, bug::unsupported_enum_equality::TestEnum, Not)
```

Aurel300:
Fix #1268.

Constant encoding is a bit of a mess. There are some functions in `Enconder` to specifically handle scalar constants, there is `encode_snapshot_constant`, and there is the `ConstantsEncoderInterface` (@vakaras) which is a copy of the `Encoder` version but without using the `encode_snapshot_constant`…

Anyway, this PR adds support for unevaluated constants (for example, a reference to a constant) in the pure function encoder. In the MIR these are represented as "promoted" items, i.e. sub-bodies of MIR bodies identified by a pair the owner `DefId` and a [`Promoted`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Promoted.html) index. We can run such bodies through the pure function encoder to get a snapshot encoding of the "execution" of such a constant. Since #1304 this works for enums too, thanks to https://github.com/rust-lang/rust/pull/107267.

### DIFF
--- a/prusti-tests/tests/verify_overflow/pass/issues/issue-1268.rs
+++ b/prusti-tests/tests/verify_overflow/pass/issues/issue-1268.rs
@@ -1,0 +1,13 @@
+use prusti_contracts::*;
+
+enum TestEnum {
+    Nil,
+}
+
+#[ensures(TestEnum::Nil === TestEnum::Nil)]
+fn test() {}
+
+fn main() {
+    let foo = 5;
+    prusti_assert!(foo === 5);
+}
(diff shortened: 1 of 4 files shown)


fpoli:
I rebased. `test-crates` should work now.

Issue #1261, opened 2022-12-04
cargo-prusti with no_verify flag will case subsequent verifications to succeed


cedihegi:
I noticed some unexpected behaviour of cargo-prusti when using the no_verify flag. In particular cargo-prusti will succesfully verify projects that should not verify, if I invoke cargo-prusti with the no_verify flag beforehand. 

In more detail, I ran the following sequence of commands in a rust-project that has verification errors:
(commands executed in the directory of that project)

`./path-prusti/cargo-prusti`
<details>
<summary> Prusti Output (failing Verification) </summary>

```
error: [Prusti: unsupported feature] unsupported constant type Ref('_#17r, [&str; 1], Not)
 --> src/main.rs:5:14
  |
5 |     println!("hello world :)");
  |              ^^^^^^^^^^^^^^^^

error: [Prusti: verification error] postcondition might not hold.
  --> src/main.rs:9:11
   |
9  | #[ensures(result)]
   |           ^^^^^^
   |
note: the error originates here
  --> src/main.rs:10:1
   |
10 | / pub fn add(x: i32, y: i32) -> bool {
11 | |     let a = 3;
12 | |     let _z = x.abs();
13 | |     if x != 42 {
...  |
17 | |     }
18 | | }
   | |_^

error: [Prusti: verification error] postcondition might not hold.
  --> src/main.rs:26:11
   |
26 | #[ensures(result == s.x + s.y)]
   |           ^^^^^^^^^^^^^^^^^^^
   |
note: the error originates here
  --> src/main.rs:27:1
   |
27 | / pub fn add_struct(s: SomeStruct) -> i32 {
28 | |     if s.x != 42 {
29 | |         s.x + s.y
30 | |     } else {
31 | |         s.x
32 | |     }
33 | | }
   | |_^

warning: `testcrate` (bin "testcrate") generated 1 warning
error: could not compile `testcrate` due to 3 previous errors; 1 warning emitted
```

</details>

`> export PRUSTI_NO_VERIFY=true`

`> ./path-to-prustifiles/cargo-prusti`
Here prusti reports no errors, but this is expected

`> export PRUSTI_NO_VERIFY=true`

`> ./path-to-prustifiles/cargo-prusti`
But here is the unexpected part:

**Prusti Output (Verification still skipped and no errors reported)**

```
warning: `testcrate` (bin "testcrate") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

[...]

PR #1334, opened 2023-02-22
[WIP] Various IDE improvements and new features

cedihegi:
Changes to support various new features in prusti-assistant. 

@Aurel300 
@jthomme1


prusti-server/src/process_verification.rs
```diff
@@ -4,24 +4,110 @@
...
+
+pub struct VerificationRequestProcessing {
+    mtx_rx_servermsg: lock::Mutex<mpsc::Receiver<ServerMessage>>,
+    mtx_tx_verreq: sync::Mutex<mpsc::Sender<ServerRequest>>,
+}
+
+// one structure that lives for all the requests and has a single thread working on all the
+// requests sequentially
+// on reception of a verification request, we send it through a channel to the already running
+// thread
```
fpoli: Use `///` for [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html)

viper/src/verification_result.rs
```diff
@@ -6,20 +6,42 @@
...
+            result_type: VerificationResultType::Success,
+            cached: false,
+            time_ms: 0,
+        }
+    }
+}
 /// The result of a verification request on a Viper program.
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
-pub enum VerificationResult {
+pub enum VerificationResultType{
```
fpoli: The `...Kind` suffix is a common choice in Rust to avoid naming `...Type` a type.

prusti-viper/src/encoder/encoder.rs
```diff
@@ -103,6 +104,7 @@ pub struct Encoder<'v, 'tcx: 'v> {
     /// this requires special care when encoding array/slice accesses which may come with
     /// bound checks included in the MIR.
     pub(super) is_encoding_trigger: Cell<bool>,
+    pub spans_of_call_contracts: RefCell<Vec<SpanOfCallContracts>>,
```
fpoli, line 110: Can you add a doc comment?

prusti-interface/src/prusti_error.rs
```diff
@@ -40,6 +40,7 @@ pub enum PrustiErrorKind {
     Warning,
     /// A warning which is only shown if at least one error is emitted.
     WarningOnError,
+    Message,
```
fpoli, line 43: If there are no Prusti functions that expect `PrustiErrorKind` to always be causes by an error, at this point I think we should rename `PrustiErrorKind` to something like `PrustiDiagnosticKind`.

prusti-smt-solver/src/context.rs
```diff
@@ -15,6 +16,7 @@ pub(super) struct Context {
 
 impl Context {
     pub(crate) async fn new() -> Self {
+        trace!("BBBB");
```
fpoli: A leftover to remove, I guess?

prusti-viper/src/ide/encoding_info.rs
```diff
@@ -0,0 +1,46 @@
+use prusti_rustc_interface::span::{source_map::SourceMap, Span};
+use serde::Serialize;
+
+use super::vsc_span::VscSpan;
+/// stores the spans of a calls contracts.
+/// obtained during encoding
+
```
fpoli, line 4: Put an empty line before the doc comment and remove the empty line after.

prusti/src/callbacks.rs
```diff
@@ -127,8 +133,42 @@ impl prusti_rustc_interface::driver::Callbacks for PrustiCompilerCalls {
...
+
+            if !config::no_verify() && !config::skip_verification() {
+                if config::selective_verify().is_none() {
+                    let verification_task = VerificationTask {
+                        procedures: annotated_procedures,
+                        types,
+                    };
+                    verify(env, def_spec, verification_task);
+                } else {
+                    let target_def_path = config::selective_verify().unwrap();
```
fpoli: Consider if it's better to rewrite the branch to `if let Some(..) = config::selective_verify() { .. } else { .. }` to avoid the dependency between `is_none()` and `unwrap()`.

prusti-server/src/server.rs
```diff
@@ -41,61 +41,53 @@ pub fn spawn_server_thread() -> SocketAddr {
     receiver.recv().unwrap()
 }
 
+// initialize on first access
```
fpoli: The comment is implied by the `Lazy` type. Can you instead summarize what this global does?

fpoli (commented):
See the comments.

fpoli:
See also https://cbea.ms/git-commit/

viper/src/verification_result.rs
```diff
@@ -6,9 +6,30 @@
...
+pub struct VerificationResult {
+    pub item_name: String,
+    pub result_type: VerificationResultKind,
+    pub cached: bool,
+    pub time_ms: u128,
+}
+
+impl VerificationResult {
+    pub fn is_success(&self) -> bool {
+        matches!(self.result_type, VerificationResultKind::Success)
```
Aurel300:
```suggestion
        self.result_type.is_success()
```

viper/src/verification_result.rs
```diff
@@ -6,9 +6,30 @@
 
 use crate::{silicon_counterexample::SiliconCounterexample, JavaException};
 
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct VerificationResult {
+    pub item_name: String,
+    pub result_type: VerificationResultKind,
```
Aurel300:
```suggestion
    pub kind: VerificationResultKind,
```

prusti/src/verifier.rs
```diff
@@ -18,11 +22,6 @@ pub fn verify(env: Environment<'_>, def_spec: typed::DefSpecificationMap) {
         debug!("Prepare verification task...");
         // TODO: can we replace `get_annotated_procedures` with information
         // that is already in `def_spec`?
```
Aurel300: Please move the TODO comment to where you now invoke `get_annotated_procedures_and_types`.

prusti-utils/src/config.rs
```diff
@@ -102,6 +102,9 @@ lazy_static::lazy_static! {
         settings.set_default("quiet", false).unwrap();
         settings.set_default("assert_timeout", 10_000).unwrap();
         settings.set_default("smt_qi_eager_threshold", 1000).unwrap();
+        settings.set_default("smt_qi_profile", true).unwrap();
+        settings.set_default("smt_qi_profile_freq", 10000).unwrap();
```
Aurel300: Will these be passed to Z3 when not using the IDE? And in such a case, will the messages not be reported to the user? If so, QI profiling should be disabled by default.

prusti-utils/src/config.rs
```diff
@@ -1017,3 +1043,29 @@ pub fn cargo_command() -> String {
...
+/// Very similar to no_verify but should only be used by prusti-assistant
+/// to avoid interference
+pub fn skip_verification() -> bool {
+    read_setting("skip_verification")
+}
+
+/// Used for selective verification, can be passed a String containing
+/// the DefPath of the method to be verified
+pub fn selective_verify() -> Option<String> {
+    read_setting("selective_verify")
```
Aurel300: Rename to `verify_only_defpath` to be consistent with `verify_only_basic_block_path`.

prusti-utils/src/config.rs
```diff
@@ -102,6 +102,9 @@ lazy_static::lazy_static! {
         settings.set_default("quiet", false).unwrap();
```
Aurel300, line 103: Make sure to update the `flags.md` file in the dev guide (`docs/dev-guide` in the repo) with all the flags you've added.

prusti-utils/src/config.rs
```diff
@@ -494,6 +504,22 @@ pub fn smt_qi_eager_threshold() -> u64 {
     read_setting("smt_qi_eager_threshold")
 }
 
+/// Set `qi.profile` value to the given one.
```
Aurel300: (Also for `smt_qi_profile_freq`:) please actually explain the effect of this setting. Reading the description here a user might not realise there is any connection to Z3, since you don't mention it and they might not be aware that Z3 has such a setting.

prusti-utils/src/config.rs
```diff
@@ -1017,3 +1043,29 @@ pub fn cargo_command() -> String {
...
 }
+
+/// When enabled, prusti should return various Data structures that are
+/// used by prusti-assistant, such as a list of method calls,
+/// a list of all procedures to be verified, etc.
+pub fn show_ide_info() -> bool {
+    read_setting("show_ide_info")
+}
+
+/// Very similar to no_verify but should only be used by prusti-assistant
```
Aurel300: "Very similar"—so is there a difference? What is it? Would it make more sense to make "verification mode" into an enum, which would contain IDE modes as well as the "no verify" mode and maybe other test modes?

prusti-smt-solver/Cargo.toml
```diff
@@ -12,7 +12,8 @@ doctest = false # and no doc tests
 
 [dependencies]
 futures = "0.3"
+log = { version = "0.4", features = ["release_max_level_info"] }
```
Aurel300: I assume this change is not needed: there is no actual change to `prusti-smt-solver` in this PR.

prusti-utils/src/config.rs
Aurel300: (After looking at how they are used in the other file:) please make these be `Option<_>`, such that they are only passed as arguments when a value is set. The IDE can set these if they are needed.

prusti-server/src/verification_request.rs
```diff
@@ -56,9 +56,11 @@ impl ViperBackendConfig {
                     // model.partial changes the default case of functions in counterexamples
                     // to #unspecified
                     format!(
-                        "smt.qi.eager_threshold={} model.partial={}",
+                        "smt.qi.eager_threshold={} model.partial={} smt.qi.profile={} smt.qi.profile_freq={}",
```
Aurel300: (See comment in `config.rs`.)

prusti-server/tests/basic_requests.rs
```diff
@@ -61,10 +60,21 @@ where
         ),
     };
 
+    let get_result_type = |acc: VerificationResultKind, msg: ServerMessage| async move {
```
Aurel300: Why is this a closure, why `async`, why `move`?

prusti-server/tests/basic_requests.rs
```diff
@@ -61,10 +60,21 @@ where
...
         .enable_all()
         .build()
         .expect("failed to construct Tokio runtime")
-        .block_on(client.verify(request))
-        .expect("Verification request failed")
+        .block_on(async {
+            PrustiClient::verify(SERVER_ADDRESS.clone(), request)
+                .await
+                .fold(VerificationResultKind::Failure(vec![]), get_result_type)
+                .await
```
Aurel300: This is not very pretty… Is there an easy way to `await` all the stream items to get an iterator (or a collection), then just use [`Iterator::find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) to find the termination message and [`unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) to get a failure if there was none?

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3387,6 +3388,14 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 substs,
             ).with_span(call_site_span)?
         };
+        // store spans of items of the contracts for prusti-assistant
+        self.store_contract_spans(
```
Aurel300: Can this only be called when some config flag is enabled?

prusti-viper/src/encoder/errors/position_manager.rs
```diff
@@ -39,7 +39,6 @@ impl<'tcx> PositionManager<'tcx>
         let span = span.into();
         let pos_id = self.next_pos_id;
         self.next_pos_id += 1;
-        trace!("Register position id {} for span {:?} in {:?}, ", pos_id, span, def_id);
```
Aurel300: Why delete?

prusti-viper/src/ide/vsc_span.rs
```diff
@@ -0,0 +1,41 @@
...
+                file_name: fname,
+                // the following 3 are not relevant here, we just want to be
+                // able to reuse the existing methods and the parser
+                // for spans in VSCode
+                is_primary: false,
+                label: None,
+                expansion: None,
+            })
+        } else {
+            None
```
Aurel300: How common is this? Do we reasonably expect every span that is passed here to be a valid span? If so, this could maybe be an `Err` rather than `None`, or even a panic? If it is an error (`Result` return type), the method should be named `try_from_span`.

prusti-viper/src/ide/encoding_info.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub contracts_spans: Vec<VscSpan>,
+}
+
+impl SpanOfCallContracts {
+    pub fn new(
+        defpath: String,
+        call_span: Span,
+        contracts_spans: Vec<Span>,
+        source_map: &SourceMap
+    ) -> Option<Self> {
```
Aurel300: A function called `new` should not return an `Option`. Please check the comment in `VscSpan` for whether this even needs to be an `Option`/`Result` at all.

prusti-viper/src/ide/encoding_info.rs
```diff
@@ -0,0 +1,45 @@
+use prusti_rustc_interface::span::{source_map::SourceMap, Span};
+use serde::Serialize;
+use super::vsc_span::VscSpan;
+
+/// stores the spans of a calls contracts.
+/// obtained during encoding
```
Aurel300: Please reword and expand.

prusti-viper/src/ide/ide_verification_result.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+use viper::VerificationResult;
+
+// since the existing verification result
+// is not as trivially passed in json
```
Aurel300: Reword, expand.

prusti-viper/src/ide/ide_verification_result.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Serialize)]
+pub struct IdeVerificationResult {
+    item_name: String,
+    success: bool,
+    time_ms: u128,
+    cached: bool,
+}
+
+impl IdeVerificationResult {
+    pub fn from_res(res: &VerificationResult) -> Self {
```
Aurel300: Should this be a [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) implementation?

prusti-viper/src/ide/vsc_span.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Serialize, Clone)]
+pub struct VscSpan {
+    column_end: usize,
+    column_start: usize,
+    line_end: usize,
+    line_start: usize,
+    file_name: String,
+    is_primary: bool,
+    label: Option<()>,
+    expansion: Option<()>,
```
Aurel300: Can you explain these two fields? Why are they not just `bool`s?

prusti-server/src/server_message.rs
```diff
@@ -0,0 +1,22 @@
+// © 2023, ETH Zurich
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use viper::VerificationResult;
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum ServerMessage {
```
Aurel300, line 12: Can you add a comment explaining what this is, where it comes from? Each variant should also have a brief explanation.

prusti-viper/src/encoder/procedure_encoder.rs
```diff
@@ -3528,6 +3537,73 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
...
+                if let Some(lhs_defid) = pledge.lhs {
+                    let lhs_span = self.encoder.env().query.get_def_span(lhs_defid);
+                    let rhs_span = self.encoder.env().query.get_def_span(pledge.rhs);
+                    lhs_span.to(rhs_span)
+                } else {
+                    self.encoder.env().query.get_def_span(pledge.rhs)
+                }
+            })
+            .collect();
+        let mut result = vec![];
```
Aurel300: It looks to me like you have a bunch of `DefId`s, for each of them you end up calling `get_def_span`. How about: create an iterator chain of all of the `DefId`s, then `map` `get_def_span` over it once, then `collect` into a `Vec` rather than append to it like in the following lines…

prusti-viper/src/verifier.rs
```diff
@@ -74,204 +86,318 @@ impl<'v, 'tcx> Verifier<'v, 'tcx> {
...
 
-        for prusti_error in prusti_errors {
-            debug!("Prusti error: {:?}", prusti_error);
-            if prusti_error.is_disabled() {
-                prusti_error.cancel();
-            } else {
+        // if we are in an ide, we already emit the errors asynchronously, otherwise we wait for
+        // all of them in order to sort them
+        if !config::show_ide_info() {
+            prusti_errors.sort();
```
Aurel300, line 169: The comment about sortedness (just before `let mut prusti_errors`) should be here.

prusti-viper/src/verifier.rs
```diff
@@ -74,204 +86,318 @@ impl<'v, 'tcx> Verifier<'v, 'tcx> {
...
-            config::viper_backend()
-        }.parse().unwrap();
-        let request = VerificationRequest {
-            program,
-            backend_config: ViperBackendConfig::new(backend),
-        };
-        (program_name, request)
-    });
-    if let Some(server_address) = config::server_address() {
+    /// Returns a list of (program_name, verification_requests) tuples.
```
Aurel300: Returns verification requests?

prusti-viper/src/verifier.rs
```diff
@@ -74,204 +86,318 @@ impl<'v, 'tcx> Verifier<'v, 'tcx> {
...
+                                serde_json::to_string(&IdeVerificationResult::from_res(&result))
+                                    .unwrap()
                             ),
-                            None,
-                        );
+                            DUMMY_SP.into(),
+                        )
+                        .emit(&self.env.diagnostic);
+                    }
+                    match result.result_type {
```
Aurel300: The `match` arms here are quite deeply nested. Does it make sense to extract handling for the cases into separate methods? It would also be good to be able to quickly see how much of the scope is actually needed for each (i.e. just the result kind + `prusti_errors`? `self`? `quantifier_instantiations`?).

prusti-server/src/client.rs
```diff
@@ -4,58 +4,76 @@
...
-}
+pub struct PrustiClient;
 
 impl PrustiClient {
-    pub fn new<S: ToString>(server_address: S) -> Result<Self, ParseError> {
+    pub async fn verify<S: ToString>(
+        server_address: S,
+        request: VerificationRequest,
+    ) -> impl Stream<Item = ServerMessage> {
+        // TODO: do proper error handling
```
Aurel300, line 26: How "improper" is it? For example, is it just resilience against the server unexpectedly quitting? If we panic in such a case that might be ok.

prusti-server/src/process_verification.rs
```diff
@@ -120,11 +266,129 @@ pub fn process_verification_request<'v, 't: 'v>(
...
+                    debug!("QuantifierInstantiationsMessage: {} {}", q_name, q_inst);
+                    // also matches the "-aux" and "_precondition" quantifiers generated
+                    // we encoded the position id in the line and column number since this is not used by
+                    // prusti either way
+                    if q_name.starts_with("prog.l") {
+                        let no_pref = q_name.strip_prefix("prog.l").unwrap();
+                        let stripped = no_pref
+                            .strip_suffix("-aux")
+                            .or(no_pref.strip_suffix("_precondition"))
+                            .unwrap_or(no_pref);
```
Aurel300: Where do these alternatives come from?

prusti-server/src/process_verification.rs
```diff
@@ -120,11 +266,129 @@ pub fn process_verification_request<'v, 't: 'v>(
...
+                    let viper_quant = jni.unwrap_result(msg_wrapper.call_quantifier(msg));
+                    let viper_quant_str =
+                        jni.get_string(jni.unwrap_result(obj_wrapper.call_toString(viper_quant)));
+                    // we encoded the position id in the line and column number since this is not used by
+                    // prusti either way
+                    let pos = jni.unwrap_result(positioned_wrapper.call_pos(viper_quant));
+                    let pos_string =
+                        jni.get_string(jni.unwrap_result(obj_wrapper.call_toString(pos)));
+                    let pos_id_index = pos_string.rfind('.').unwrap();
+                    let pos_id = pos_string[pos_id_index + 1..].parse::<u64>().unwrap();
```
Aurel300: Why is the position handling different here and in the quantifier instantiations message? Can they share code?

prusti/src/ide_helper/call_finder.rs
```diff
@@ -0,0 +1,124 @@
...
+                    if method_def_id.as_local().is_none() {
+                        if defpath_unresolved == defpath_resolved {
+                            self.called_functions
+                                .push((defpath_resolved, resolved_def_id, sp));
+                        } else {
+                            // in this case we want both
+                            self.called_functions
+                                .push((defpath_resolved, resolved_def_id, sp));
+                            self.called_functions
+                                .push((defpath_unresolved, method_def_id, sp));
```
Aurel300, line 84: Should the entries in `called_functions` also contain some indication of what kind of method it is? For example, the IDE could display "extern spec for trait method X" vs "extern spec for trait impl method X"?

prusti/src/ide_helper/call_finder.rs
```diff
@@ -0,0 +1,124 @@
...
+                        }
+                    }
+                }
+            }
+            ExprKind::Binary(..) | ExprKind::AssignOp(..) | ExprKind::Unary(..) => {
+                let resolve_res = self.resolve_expression(expr);
+                // this will already fail for standard addition
```
Aurel300: As in, for example, `i32` addition, because it is a built-in operation?

prusti/src/ide_helper/compiler_info.rs
```diff
@@ -0,0 +1,128 @@
...
+/// This struct will be passed to prusti-assistant containing information
+/// about the program that is currently being verified
+#[derive(Serialize)]
+pub struct IdeInfo {
+    procedure_defs: Vec<ProcDef>,
+    function_calls: Vec<ProcDef>,
+    queried_source: Option<String>,
+    // additionally this will contain:
+    // function_calls:
+    // ... we'll see
```
Aurel300: ?

prusti/src/ide_helper/compiler_info.rs
```diff
@@ -0,0 +1,128 @@
...
+                defid,
+                span: VscSpan::from_span(&sp, source_map).unwrap(),
+            })
+            .collect();
+
+        // For declaring external specifications:
+        let queried_source = query_signature::collect_queried_signature(env.tcx(), &fncalls);
+        Self {
+            procedure_defs: procs,
+            function_calls: fncalls,
```
Aurel300: Rename `procs` to `procedure_defs` and `fncalls` to `function_calls`.

prusti/src/ide_helper/compiler_info.rs
```diff
@@ -0,0 +1,128 @@
...
+        let procs = collect_procedures(env, procedures, def_spec);
+        let source_map = env.tcx().sess.source_map();
+        let fncalls: Vec<ProcDef> = collect_fncalls(env)
+            .into_iter()
+            .map(|(name, defid, sp)| ProcDef {
+                name,
+                defid,
+                span: VscSpan::from_span(&sp, source_map).unwrap(),
+            })
+            .collect();
```
Aurel300:
```suggestion
            .collect::<Vec<_>>();
```
(bit nicer than explicitly annotating `fncalls` with a type hint)

prusti/src/ide_helper/compiler_info.rs
```diff
@@ -0,0 +1,128 @@
...
+/// A struct that contains either a reference to a procedure that can be verified
+/// (for selective verification) or a function call (so a user can query
+/// external_spec blocks for it). The name contains the defpath.
+pub struct ProcDef {
+    pub name: String,
+    pub defid: DefId,
+    pub span: VscSpan,
+}
+
+impl Serialize for ProcDef {
```
Aurel300: You can add [`#[serde(skip)]`](https://serde.rs/field-attrs.html#skip) to the `defid` field to skip it, instead of manually implementing the trait.

prusti/src/ide_helper/query_signature.rs
```diff
@@ -0,0 +1,432 @@
...
+use std::{collections::HashMap, fmt};
+
+use super::compiler_info::ProcDef;
+use prusti_rustc_interface::{
+    hir::{def::DefKind, def_id::DefId},
+    middle::ty::{self, Clause, DefIdTree, ImplSubject, PredicateKind, TyCtxt},
+};
+
+/// data structure to represent the code we want to generate
+/// since we didnt manage to this with syn etc..
```
Aurel300: Reword, expand. The reference to `syn` only makes sense to us because we discussed it…

prusti/src/ide_helper/query_signature.rs
```diff
@@ -0,0 +1,432 @@
...
+            DefKind::Fn => {
+                let parent_chain = get_parent_chain(defid, tcx);
+                Some(ExternSpecBlock::StandaloneFn {
+                    parent_chain,
+                    function_sig: signature,
+                })
+            }
+            DefKind::AssocFn => {
+                // this will be None for traits
+                let impl_defid_opt = tcx.impl_of_method(defid);
```
Aurel300: Why the local?

prusti/src/ide_helper/query_signature.rs
```diff
@@ -0,0 +1,432 @@
...
+                        Some(ExternSpecBlock::Impl {
+                            name: self_ty,
+                            trait_name,
+                            generics,
+                            function_sig: signature,
+                        })
+                    }
+                    None => {
+                        // function is a Trait (or something else?)
+                        // are there other cases for AssocFns?
```
Aurel300, line 77: I don't think so.

prusti/src/ide_helper/query_signature.rs
```diff
@@ -0,0 +1,432 @@
...
+                let trait_defpath = tcx.def_path_str(trait_defid);
+
+                let item_name = tcx.item_name(item_id);
+
+                let projection_term = format!("{}={}", item_name, p.term);
+                let genarg = generic_params.get_mut(&self_ty).unwrap();
+                // not sure if this is actually completely correct.
+                // can a single generic type have the same traitbound
+                // more than once with different arguments?
+                // I would imagine yes.. may be a todo
```
Aurel300: Like `T: Foo<i32> + Foo<u32>`? Yes, this is possible.

Aurel300 (changes requested):

prusti-server/src/process_verification.rs
jthomme1: The handling is different due to the different message types. In the QuantifierInstantiationsMessage, we get a Z3 quantifier name (e.g. a string `prog.l234-aux`) while in the QuantifierChosenTriggersMessage, we get an actual `Positioned` jObject, where we extract the position id from the string of the position (because we needed less Java boiler plate code, but maybe it would be better to get it from the actual line member).

prusti-server/src/process_verification.rs
jthomme1: I did not investigate too much, but as far as I remember they are introduced by Z3

prusti-server/src/client.rs
jthomme1: Atm, we panic if the server address is invalid and if the connection is torn down/fails somehow unexpectedly. However, this was basically the same before: just that there this method returned a result which was `.expect`ed.

prusti-utils/src/config.rs
cedihegi: I will add a comment and describe it in the dev-guide, but yes there are some important differences. 
It is needed because of issue #1261. My solution for that problem was to generate a fake error when we skip verification.
The problem with `no_verify` is that it's value is not just what we pass via the command line. If no-verify is set to false by the user, it will still evaluate to true when compiling dependencies, and if I start throwing fake errors at that point, this will not work. 
If issue #1261 is resolved, this flag and the fake errors will no longer be necessary.
Let me know if you still think creating this enum you described would be a good idea, but I don't think that this skip_verification variable should be in there since it's probably temporary.

prusti-viper/src/ide/vsc_span.rs
cedihegi: The idea was to allow easy serialization here, and deserialization in typescript, where spans of this form are already parsed. (this is the form they have when emitting compiler messages with spans).

`label` and `expansion` are both optional fields in the type defined in prusti-assistant, so putting in a boolean would not work. Option fields with value None, when serialized, will result in no field. But since I don't use them at all I can just delete them.

prusti-viper/src/ide/vsc_span.rs
cedihegi: I think panicking here should be reasonable.

prusti/src/ide_helper/call_finder.rs
cedihegi: yes exactly. I will reword it to built-in operations instead of standard addition.

prusti-viper/src/encoder/procedure_encoder.rs
cedihegi: I somewhat adjusted it, but it's still not very pretty.
I tried to generate a chain of iterators, but I did not manage to make this prettier at all. 
For the pledges I still have to merge them after finding the spans, so I can not include them. For the others it's also not quite as simple as it might look since there are 2 vecs and 2 `Option<Option<DefId>>`. I put them all into a vector now before querying the spans. I hope that's alright.

prusti-viper/src/ide/ide_verification_result.rs
cedihegi: yes!

prusti/src/ide_helper/compiler_info.rs
cedihegi: left-over comment

PR #1345, opened 2023-03-01
[WIP] JNI bindings - print actually class name on type missmatch

simon-hrabec:
**This PR is not yet to be reviewed**. I want to discuss where to put the logic.


simon-hrabec:
I tried to call a method on the wrong instance in the tests to see if the `debug_assert!` will provide the correct error message:
image
It showed the correct message (but the 2 formats are different tho).

PR #1347, opened 2023-03-06
JNI bindings - generate all dynamic type check at one place

simon-hrabec:
The JNI bindings have runtime checks to verify that the given object (receiver, parameter, result) is of an expected type. If not, an exception is thrown. The concept is delineated in the [draft PR - [WIP] JNI bindings - print actually class name on type missmatch](https://github.com/viperproject/prusti-dev/pull/1345).

This PR does the necessary prerequisite - making all the dynamic type checks generated by a call to a single function (+ a few fixes to formating/indentation of the generated code).


jni-gen/src/utils.rs
```diff
@@ -82,3 +82,44 @@ pub fn java_identifier_to_rust(name: &str) -> String {
...
+            "    }".to_string(),
+        ]
+    } else {
+        type_check
+    }
+    .join("\n")
+        + "\n"
+}
+
+pub fn generate_result_type_check(variable_type_name: &str) -> String {
```
fpoli: Could you add a doc comment stating that `variable_type_name` should be a Java class type signature, in this  and in `generate_variable_type_check`?

Thinking at the argument name again, would `return_signature` and `variable_signature` be better than `..._type_name`?

jni-gen/src/utils.rs
```diff
@@ -82,3 +82,44 @@ pub fn java_identifier_to_rust(name: &str) -> String {
...
     name.replace('_', "__").replace('$', "_dollar_")
 }
+
+pub fn is_signature_of_class_type(signature: &str) -> bool {
+    signature.starts_with('L')
+}
+
+/// Generates a runtime check that the object is of the expected class - applies only to object types (starting with L)
+/// If this value is also the returned value in the setter, unwrap it first (since it is of type Result<T>)
+fn generate_type_check(variable_name: &str, variable_type_name: &str, is_result: bool) -> String {
```
fpoli: ~~Can you add a `debug_assert!(..)` to check that the signature is of a Java class?~~

In the doc comment, can you update "applies only to object types ..." to clarify that the variable is now *required* to refer to a Java class?

fpoli (approved):
I'm not sure how this is a preparation for https://github.com/viperproject/prusti-dev/pull/1345. Independently from that, this PR is a nice refactoring on its own.

jni-gen/src/utils.rs
simon-hrabec: I think the interface/names/docs changes I made are unclear, sorry for the confusion.

I decided to declare all the class names at the beginning of the functions (e.g. arg_arg0_class_name)
image
or 
image
If the parameter/return type is not of a class type then both generating the variable with a class name for the respective parameter/return types and generating the check are skipped (in the second screenshot one parameter is skipped)

Then in the check, the name of the variable is used, rather than the string literal:
image

 
Before both ways were used (e.g. sometimes class_name, sometimes "java/lang/String") and I was considering hardcoding the string literals with the class names there but decided to rather use variables for clarity (once the class names are put there as literals it takes more time to backtrack how did they get there and if they are there by an accident). I also thought to allow to use both (by passing an enum) but thought it might be overengineering.

This has the problem that at the time of generation the `generate_type_check` function cannot do the check.

BTW there is also a bug (get checks the return type with signature, not the class name).

simon-hrabec:
> I'm not sure how this is a preparation for #1345. Independently from that, this PR is a nice refactoring on its own.

The idea was to generate all the runtime `is_instance_of` runtime checks from one place so I can easily modify it from the current
image
to something that prints also the message with the actual type:
image

jni-gen/src/utils.rs
fpoli: Thanks for the explanation. I didn't notice that the argument of `generate_type_check` is now a `/`-separated class name instead of a signature. Can you write this in the doc comment? I agree that an enum would be an overengineering.

fpoli: > This has the problem that at the time of generation the `generate_type_check` function cannot do the check.

Do you mean the `debug_assert!(..)` that I suggested to check that the signature is of a Java class? Sure, if the argument is a class name then that check does not make sense.

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -62,143 +62,86 @@ fn class_field_lookup<'a>(
...
+    code.push("        receiver,".to_string());
+    code.push("        field_name,".to_string());
+    code.push("        return_signature,".to_string());
+    code.push(format!(
+        "    ).and_then(|x| x.{}());",
+        generate_jni_type_char(type_signature)
+    ));
+    code.push("".to_string());
+    if is_signature_of_class_type(type_signature) {
+        code.push(generate_result_type_check("return_signature"));
```
fpoli: > BTW there is also a bug (get checks the return type with signature, not the class name).

I see, yes.

simon-hrabec: fixed

jni-gen/src/utils.rs
simon-hrabec: Updated the docs

jni-gen/src/utils.rs
simon-hrabec: Changed the parameter names

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -62,143 +62,90 @@ fn class_field_lookup<'a>(
...
+        "/// Return type and Java signature: `{parameter_type}` (`{type_signature}`)"
+    ));
+    code.push(format!("pub fn {rust_getter_name}("));
+    code.push("    &self,".to_string());
+    code.push("    receiver: JObject<'a>,".to_string());
+    code.push(format!(") -> JNIResult<{parameter_type}> {{"));
+    code.push(format!("    let class_name = \"{}\";", class.path()));
+    code.push(format!("    let field_name = \"{field_name}\";"));
+    code.push(format!("    let return_signature = \"{type_signature}\";"));
+    if is_signature_of_class_type(&type_signature) {
```
fpoli:
```suggestion
    if is_signature_of_class_type(type_signature) {
```

PR #1348, opened 2023-03-06
Try fix Windows CI

JonasAlaif:


fpoli:
What about moving the `let _setup = job::setup().unwrap();` into the `prusti-server` binary? As far as I know `compiletest.rs` does not start new processes for the server.

PR #1350, opened 2023-03-06
Make Prusti work with build scripts and AnonConst fns

JonasAlaif:
Prusti would crash when trying to get the MIR body of `AnonConst` functions, or any functions when compiling a build script. This change ignores both, in the future we might want to figure out if we actually want to handle these cases, and, if so, how.

### DIFF
--- a/prusti/src/callbacks.rs
+++ b/prusti/src/callbacks.rs
@@ -6,7 +6,7 @@ use prusti_interface::{
 };
 use prusti_rustc_interface::{
     driver::Compilation,
-    hir::def_id::LocalDefId,
+    hir::{def::DefKind, def_id::LocalDefId},
     interface::{interface::Compiler, Config, Queries},
     middle::ty::{
         self,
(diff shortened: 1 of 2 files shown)


Aurel300:
What are these? Do you have an example of Rust code that generates these?

JonasAlaif:
@Aurel300 these were the two crashes I ran into with Prusti using my [top-crates harness](https://github.com/viperproject/prusti-dev/blob/24aff3598140bd7359bfd8cd1b223c8cad2082cf/micromir/tests/top_crates.rs), you can try taking that file and running it with current prusti and `no_verify=true`. I'll document some examples here before merging as well then.
Edit: I didn't spend the time to find the exact definitions of AnonConst fns, but there are many in the `ppv-lite86` crate (version 0.2.17).

Aurel300:
Ok, now I see what they are. I don't think we'll need borrow check facts for these, we evaluate such constants using the pure function encoder or compiler APIs anyway.

PR #1353, opened 2023-03-07
Try fix Windows CI v2

JonasAlaif:


fpoli:
What about moving the `let _setup = job::setup().unwrap();` into the `prusti-server` binary? As far as I know `compiletest.rs` does not start new processes for the server.

fpoli:
Moving the job to the `prusti-server` binary seems to work. I'll wait again for the CI and then merge https://github.com/viperproject/prusti-dev/pull/1361.

PR #1355, opened 2023-03-08
Rework User-Guide and Documentation

Patrick-6:
As part of my Bachelor's Thesis, I've been reworking the user guide and the guided tour for Prusti.

Some things that were done:

- Reworked the guided tour for Prusti
- Added or expanded documentations for some existing Prusti features
- Added a `Limitations and Capabilities` chapter documenting the current state of features that Prusti supports
- Added instructions for writing future documentations in `/docs/readme.md`
- Made the user-guide mdbook ready for [documentation testing](https://rust-lang.github.io/mdBook/cli/test.html)
- Moved some of the bigger inline code examples in the book to the `prusti-tests/tests/` directory to be automatically tested by Prusti

There are some tasks left:

- [x] Add `mdbook test` to the Prusti CI (for both dev-guide & user-guide)
- [ ] Rework/Remove the `Testing` chapter in the guided tour
- [ ] Check and expand the `Limitations` chapter
- [x] Expand file `verify/print_counterexample.md` to also explain counterexamples in general
- [x] Document the `refine_trait_spec` annotation

Future tasks:
- Write the capabilities chapter
- Move other code blocks in the user guide to the `prusti-tests` directory, so they don't go out of date in future Prusti versions
- Make the `dev-guide` ready for documentation testing (and remove all the `ignore` from the code blocks)
- Maybe find a better way to get `prusti_contracts` dependency when running `mdbook test` locally without using the `docs/dummy` crate (and update explanation in readme.md)


docs/user-guide/src/tour/testing.md
```diff
@@ -1,8 +1,74 @@
...
+// Prusti: verifies
 ```
+<!-- We can also have tests that take arguments and also have pre- and postconditions:
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/pass/user-guide/testing_initial_code.rs:test_2}}
+// Prusti: verifies
+``` -->
+
+Note, the function `test_list` would still be available when the program is compiled normally. There may be a possibility to do have something similar to the `#[cfg(test)]` annotation in Prusti. This would allow you to only have `test_list` during verification, see [this section](../capabilities/limitations.md#conditional-compilation-for-verification-vs-normal-compilation) of the limitations chapter.
```
Aurel300: "to do have something similar"

docs/user-guide/src/tour/testing.md
```diff
@@ -1,8 +1,74 @@
...
+## Differing guarantees of verification and testing
+
+Static verification has stronger guarantees than testing.
+Running tests is only be possible for a small subset of all possible input values.
+Take as an example a function taking a single value of type `u64`. The range of potential inputs is `0` to `2^64 - 1`, or `2^64` total values. Assuming each value takes 1 nano-second to test, it would take approximately `584.5` years to exhaustively test just this single function.
+<!-- (2^64)/1000000000 / 60 / 60 / 24 / 365.25 = 584.54204609062639795168200370117 -->
+
+In contrast, a static verifier like Prusti is able to check the entire input space of a function with the help of the specifications of each function.
+
+When verification succeeds, you are guaranteed to not have a bug like a crash, overflow, or return value not fitting the specification.
```
Aurel300, line 19: This is often summarised as "testing can show the *presence* of bugs, verification can show the *absence* of bugs".

docs/user-guide/src/tour/summary.md
```diff
@@ -38,15 +36,16 @@ devices.
...
-4. [Pop](pop.md): Exercise with similar concepts as for push
-5. [Testing](testing.md): More runtime errors caught by Prusti
-6. [A Bad Stack](bad-stack.md): Wrap-up of the second chapter of
-   [Learn Rust with Entirely Too Many Linked Lists](https://rust-unofficial.github.io/too-many-lists/).
-7. [Options](options.md): Verification with option types
-7. [Generics](generics.md): Prusti and generics
-8. [Peek](peek.md): Exercise
-9. [Final Code](final.md): Final code with solution of exercise
-10. [Pledges](pledges.md): Bonus demonstrating Prusti's pledges
+1. [Setup](setup.md) How to add Prusti to a Rust project
```
Aurel300:
```suggestion
1. [Setup](setup.md): How to add Prusti to a Rust project
```

docs/user-guide/src/tour/summary.md
```diff
@@ -38,15 +36,16 @@ devices.
...
+4. [Push](push.md): Preconditions, external specifications, trusted functions, old expressions, quantifiers, snapshots, structural/snapshot equality
+5. [Pop](pop.md): Similar concepts as in [Push](push.md), predicates
+6. [Testing](testing.md): Showing guarantees of verification vs running tests, and how to test specifications
+7. [Option](option.md): Changing `Link` to use `Option` type
+8. [Generics](generics.md): Prusti and generics
+9.  [Peek](peek.md): Verifying a `peek` function
+10. [Pledges (mutable peek)](pledges.md): Demonstrate Prusti's pledges for functions returning mutable references
+11.  [Final Code](final.md): Final code for the verified linked list
+12.  [Loop Invariants](loop_invariants.md): Verifying code containing loops by writing loop invariants
+13.  [Counterexamples](counterexamples.md): Get a counterexample for a failing assertions
```
Aurel300: This and pledges: `Getting a counterexample…` and `Demonstrating Prusti's pledges…`, respectively, to be consistent with the other items. Setup can also say `Adding Prusti to…`.

docs/dummy/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+use prusti_contracts::*;
+
+// This crate is just used to provide the needed dependencies
+// for the mdbook documentation tests
```
Aurel300, line 4: Put this comment in a `README.md` file in `docs/dummy`.

docs/user-guide/src/capabilities/limitations.md
```diff
@@ -0,0 +1,276 @@
...
+Workaround for the `range` iterator `0..100`:
+```rust,noplaypen,ignore
+let mut i = 0;
+while i < 100 {
+    body_invariant!(true);
+    /* loop body */
+    i += 1;
+}
+```
+
```
Aurel300, line 51: Same comment as for closures: postponed until after the stdlib PR.

docs/user-guide/src/capabilities/limitations.md
```diff
@@ -0,0 +1,276 @@
...
+Using a [`Union`](https://doc.rust-lang.org/reference/items/unions.html) in Rust requires unsafe code anyways, but even declaring them is not yet supported.
+
+See [this issue](https://github.com/viperproject/prusti-dev/issues/1341) for updates.
+
+
+## Quantifiers and triggers
+
+### TODO
+Quantifiers need to be correctly triggered during verification. These triggers are either user-supplied or automatically generated by Prusti. This is generation is not perfect at the moment, and the user is also not informed about which triggers have been selected. One place this can show up is in nested loops, which are currently hard to verify.
```
Aurel300: The automatic trigger choice is done by Silicon, not Prusti.

docs/user-guide/src/capabilities/limitations.md
```diff
@@ -0,0 +1,276 @@
...
+
+Parameters and return types of `#[pure]` functions currently need to implement the [`Copy` trait](https://doc.rust-lang.org/std/marker/trait.Copy.html).
+
+- Allowed types:
+  - `i8`, `i16`, `i32`, `i64`, `i128`
+  - `u8`, `u16`, `u32`, `u64`, `u128`
+  - `bool`
+  - `f32`, `f64` (but floating point is not fully supported yet)
+  - `&T`
+  - `struct`, `enum`, slices and stuples that implement `Copy`
```
Aurel300:
```suggestion
  - `struct`, `enum`, slices and tuples that implement `Copy`
```

docs/user-guide/src/tour/counterexamples.md
```diff
@@ -0,0 +1,36 @@
+# Counterexamples
+
+Let's take the summation function from the last chapter, which adds up all the numbers from 1 to `x`. Let's suppose we forgot to add the non-negative postcondition for `x`:
```
Aurel300:
```suggestion
Let's take the summation function from the last chapter, which adds up all the numbers from 1 to `x`. Let's suppose we forgot to add the non-negativity precondition for `x`:
```

docs/user-guide/src/tour/final.md
```diff
@@ -1,3 +1,10 @@
...
 
-**TODO**
+## Full Code
+
+```rust,noplaypen
+// Expand to see the full code
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/pass/user-guide/peek_mut_pledges.rs:nothing}}
+```
+
+The next 2 chapters will explain loop invariants and counterexamples, but we will not use the linked list from our previous chapters.
```
Aurel300: I would put this sentence at the beginning of the next chapter rather than here. It is easy to adjust the chapters and forget to update this.

docs/dummy/README.md
```diff
@@ -0,0 +1,2 @@
+// This crate is just used to provide the needed dependencies
+// for the mdbook documentation tests
```
Aurel300:
```suggestion
This crate is only used to provide the needed dependencies for the mdbook documentation tests.
```

This is a normal markdown file, the comment syntax looks strange when it is rendered.

docs/user-guide/src/basic.md
```diff
@@ -15,15 +15,21 @@ See the [following chapter](verify/summary.md) for a list of verification featur
 
 To run Prusti on a file using the command-line setup:
 
-```bash
-$ prusti-rustc --edition=2018 path/to/file.rs
+```sh
+prusti-rustc --edition=2018 path/to/file.rs
+```
+
+To run Prusti on a Rust crate:
```
Aurel300: - "crate *or workspace*"
- Add a note that this command should be invoked from within the crate/workspace root.
- If `cargo-prusti` is in `PATH`, then I think you can run `cargo prusti` (i.e. it behaves like `clippy` or any other third-party `cargo` command).

docs/user-guide/src/capabilities.md
```diff
@@ -0,0 +1,6 @@
+# Prusti capabilities summary
+
+This chapter documents capabilities and limitations of Prusti:
+
+- [Limitations of Prusti](capabilities/limitations.md)
+- [Supported Rust and verification features of Prusti](capabilities/supported.md)
```
Aurel300: These are not in the same order as in `SUMMARY`, swap; also make the titles match everywhere.

docs/user-guide/src/syntax.md
```diff
@@ -14,33 +14,78 @@ Prusti specifications are a superset of Rust boolean expressions. They must be d
 | [`exists(...)`](#quantifiers) | Existential quantifier |
 | [<code>... &#x7C;= ...</code>](#specification-entailments) | Specification entailment |
 
-## Old expressions
 
-Old expressions are used to refer to the value that a memory location pointed at by a mutable reference had at the beginning of the function:
+## `result` Variable
+
+When using Prusti, `result` is used to refer to what a function returns.
+`result` can only be used inside a postcondition, meaning that variables called `result` used in a function need to be renamed.
```
Aurel300: Only arguments, local variables don't matter (you can't refer to them from specifications anyway).

docs/user-guide/src/syntax.md
Aurel300, line 1: The table at the top of the file should be extended with the new sections. However, this section is really meant to be about the syntax (and in particular, our extensions over Rust syntax). So "Adding specification in trait `impl` blocks" does not belong here.

docs/user-guide/src/tour/loop_invariants.md
```diff
@@ -0,0 +1,25 @@
...
+
+Lets write a function that takes an integer `x` and sums up all values from 0 to that value in a loop.
+For non-negative inputs, the result will be equal to `x * (x + 1) / 2`:
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/fail/user-guide/loop_invariants.rs:specification}}
+// Prusti: fails
+```
+
+We cannot verified this code yet, because Prusti does not know what the `while` loop does to `sum` and `i`. For that, we need to add a [`body_invariant`](../verify/loop.md). Body invariants are expressions that always hold at the beginning and end of the loop body. In our case, the invariant is that `sum` contains the sum of all values between 1 and `i`. Since `i` starts at 1 and not at 0, we have to slightly adjust the formula by using `i - 1` instead of `i`, so we get: `sum == (i - 1) * i / 2`.
```
Aurel300, line 14: "cannot verify"

Also the part "always hold at the beginning and end of the loop body" is not exactly true. The invariant is assumed or checked in the place it is written in the code (it could be in the middle of a loop). Then:

- When entering the loop on the first iteration, the invariant must hold when reached (which implies the loop condition was true);
- when re-entering the loop on an arbitrary iteration, the invariant is assumed, then, executing the rest of the loop and the next iteration's beginning (until the invariant), the invariant must hold again;
- when exiting the loop, the invariant is assumed.

… which can sound rather confusing without a good diagram and should rather be explained in the body invariant section proper.

docs/user-guide/src/tour/loop_invariants.md
```diff
@@ -0,0 +1,25 @@
...
+After adding the `body_invariant`, we get this code:
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/pass/user-guide/loop_invariants_final.rs:specification}}
+// Prusti: verifies
+```
+
+This body invariant is enough to verify the postcondition. After the loop, `i == x + 1` will hold. Plugging this into our `body_invariant!(sum == (i - 1) * i / 2)`, we get `sum == x * (x + 1) / 2`, which is our postcondition.
+
+Note that we did not have to add `body_invariant!(1 <= i && i <= x)`. Prusti can sometimes figure out the correct range for `i` by itself, as long as at least one other `body_invariant` is present in the loop.
```
Aurel300, line 25: This is related to the previous comment in that the position of the body invariant matters, which is why "as long as at least one other `body_invariant` is present in the loop" isn't a completely arbitrary restriction. You should note that you can write `body_invariant!(true)` as well.

docs/user-guide/src/tour/new.md
```diff
@@ -132,52 +55,18 @@ error: cannot find attribute `ensures` in this scope
 
 Prusti's specifications consist of Rust
 [macros and attributes](https://doc.rust-lang.org/reference/procedural-macros.html)
-that are defined in a separate crate called `prusti_contracts`.
+that are defined in a separate crate called `prusti_contracts`. To see how to add this crate to your project, check out [Setup](setup.md).
```
Aurel300, line 58:
```suggestion
that are defined in a separate crate called `prusti_contracts`. To see how to add this crate to your project, see [Setup](setup.md).
```

docs/user-guide/src/tour/new.md
```diff
@@ -248,66 +102,54 @@ error: [Prusti: invalid specification] use of impure function "Link::len" in pur
...
 
 Prusti now manages to verify that `new()` always returns
 a list for which the method `len()` returns 0. (notice
-this is hardly surprising since `len()` ultimately always returns 0 but we will change
-this soon.)
+this is hardly surprising since `len()` ultimately always returns 0.)
+
+## Proper implementation of `len`
+
+We will now properly implement `len()`, and while we're at it, `is_empty()` for `Link`. Both of them are pure functions, so we will add the `#[pure]` annotation. Both functions can be called without any restrictions, so they have the default postcondition `#[requires(true)]`, which we don't have to add manually. We also don't need to add any additional postconditions, since pure functions will be inlined wherever they are used during verification.
```
Aurel300, line 128: "inlined wherever they are used during verification" is not completely accurate. This might suggest recursion is not possible, for example. You should rather say that the body of the pure function is implicitly part of its specification.

docs/user-guide/src/tour/new.md
```diff
@@ -248,66 +102,54 @@ error: [Prusti: invalid specification] use of impure function "Link::len" in pur
...
+```
+
+Here we use the [`matches` macro](https://doc.rust-lang.org/std/macro.matches.html) in `is_empty`, which is true if and only if the first argument matches the pattern in the second argument.
+
+We can now check if the specification is working, by writing a function that panics if the specification is wrong:
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/pass/user-guide/impl_new_full_code.rs:test_len}}
+```
+
+The last line asserts, that the `is_empty` function only returns `true`, if the `len` function returns `0`.
```
Aurel300, line 141:
```suggestion
The last line asserts that the `is_empty` function only returns `true` if the `len` function returns `0`.
```

docs/user-guide/src/SUMMARY.md
```diff
@@ -4,17 +4,19 @@
...
-  - [Options](tour/options.md)
+  - [Option](tour/option.md)
   - [Generics](tour/generics.md)
   - [Peek](tour/peek.md)
+  - [Pledges (mutable peek)](tour/pledges.md)
   - [Final Code](tour/final.md)
-  - [Pledges](tour/pledges.md)
+  - [Loop Invariants](tour/loop_invariants.md)
+  - [Counterexample](tour/counterexamples.md)
 - [Verification Features](verify/summary.md)
```
Aurel300, line 20: This should also include `prusti-feature.md` (and maybe other changes to the ToC I missed?)

docs/user-guide/src/tour/setup.md
```diff
@@ -0,0 +1,52 @@
...
+```sh
+cargo new prusti_tutorial
+```
+Then you can change the current working directory to the project's directory:
+```sh
+cd ./prusti_tutorial/
+```
+
+
+To use the additional syntax used for verification with Prusti, you need to add the [prusti_contracts](https://docs.rs/prusti-contracts/latest/prusti_contracts/) crate to your project. If you have at least Cargo version 1.62.0, you can use this command to add the dependency:
```
Aurel300, line 14:
```suggestion
To use the additional syntax used for verification with Prusti, you need to add the [`prusti-contracts`](https://docs.rs/prusti-contracts/latest/prusti_contracts/) crate to your project. If you have at least Cargo version 1.62.0, you can use this command to add the dependency:
```

docs/user-guide/src/tour/setup.md
```diff
@@ -0,0 +1,52 @@
...
+```sh
+cargo add prusti-contracts
+```
+For older versions of Rust, you can manually add the dependency in your Cargo.toml file:
```
Aurel300, line 18:
```suggestion
For older versions of Rust, you can manually add the dependency in your `Cargo.toml` file:
```

docs/user-guide/src/tour/setup.md
```diff
@@ -0,0 +1,52 @@
...
+```toml
+[dependencies]
+prusti-contracts = "0.1.4"
+```
+
+To use prusti-contracts in a Rust code file, just add the following line:
```
Aurel300, line 24:
```suggestion
To use `prusti-contracts` in a Rust code file, just add the following line:
```

docs/user-guide/src/tour/setup.md
```diff
@@ -0,0 +1,52 @@
...
+```rust,ignore
+use prusti_contracts::*;
+```
+
+
+To simplify the tutorial, overflow checks by Prusti will be disabled. To do that, create a file called `Prusti.toml` in your projects root directory (where `Cargo.toml` is located).
```
Aurel300, line 30:
```suggestion
To simplify the tutorial, overflow checks by Prusti will be disabled. To do that, create a file called `Prusti.toml` in your project's root directory (where `Cargo.toml` is located).
```

docs/user-guide/src/tour/setup.md
```diff
@@ -0,0 +1,52 @@
...
+```toml
+check_overflows = false
+```
+
+**Note**: Creating a new project will create a `main.rs` file containing a `Hello World` program. Since Prusti does not yet support Strings (see [Prusti Limitations](../capabilities/limitations.md#strings-and-string-slices) chapter), verification will fail on `main.rs`. To still verify the code, remove the line `println!("Hello, world!");`.
+
+
+## Standard library annotations
+
+Annotations for functions and types in the Rust standard library will be available in the [prusti-std crate](https://crates.io/crates/prusti-std) after [this PR](https://github.com/viperproject/prusti-dev/pull/1249) is merged.
```
Aurel300, line 41:
```suggestion
Annotations for functions and types in the Rust standard library will be available in the [`prusti-std` crate](https://crates.io/crates/prusti-std) after [this PR](https://github.com/viperproject/prusti-dev/pull/1249) is merged.
```

docs/user-guide/src/tour/testing.md
```diff
@@ -1,8 +1,76 @@
...
+This assumes that you have manually verified any `#[trusted]` functions and have checked for correct termination of all functions.
+If the verification fails, you may have a bug, or your specifications are not strong enough.
+
+In other words: Testing can show the *presence* of bugs, verification can show the *absence* of bugs.
```
Aurel300, line 23: I think this paragraph should have been added after the "The guarantees of testing…" one.

docs/user-guide/src/verify/assert_assume.md
```diff
@@ -7,33 +7,50 @@ function (via an assumption).
...
 
-The `prusti_assert!` macro instructs Prusti to verify that a certain property
-holds at a specific point within the body of a function. In contrast to the
-`assert!` macro, which only accepts Rust expressions, `prusti_assert!` accepts
-[specification](../syntax.md) expressions as arguments. Therefore, quantifiers
-and `old` expressions are allowed within a call to `prusti_assert!`, as in
-the following example:
-
-```rust,noplaypen
+The macros `prusti_assert!`, `prusti_assert_eq` and `prusti_assert_ne` instruct Prusti to verify that a certain property holds at a specific point within the body of a function. In contrast to the `assert!`, `assert_eq` and `assert_ne` macros, which only accept Rust expressions, the Prusti variants accept [specification](../syntax.md) expressions as arguments. Therefore, [quantifiers](../syntax.md#quantifiers), [`old`](../syntax.md#old-expressions) expressions and other Prusti specification syntax is allowed within a call to `prusti_assert!`, as in the following example:
```
Aurel300, line 10: Be consistent with the exclamation mark: `prusti_assert_eq!`, etc. (Also in the rest of the file.)

docs/user-guide/src/verify/counterexample.md
```diff
@@ -0,0 +1,56 @@
...
+
+Note 1: There are no guarantees on which value gets returned for the counterexample. The result will be an arbitrary value that fails the assertion (in this case any value in the range `i32::MIN..=9`).
+Note 2: Verification will be slower with `counterexamples = true`.
+
+
+# Customizable counterexamples
+
+A counterexample for structs and enums can be formatted by annotating the type with `#[print_counterexample()]`. This is only available if the [`unsafe_core_proof`](https://viperproject.github.io/prusti-dev/dev-guide/config/flags.html#unsafe_core_proof) flag is set to `true`.
+
+## Syntax structs
```
Aurel300, line 28:
```suggestion
## Syntax: structs
```
(same for enums below)

docs/user-guide/src/tour/pledges.md
```diff
@@ -1,3 +1,73 @@
...
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/fail/user-guide/peek_mut_pledges.rs:peek_mut_code}}
+```
+
+Note that `peek_mut` cannot be `#[pure]`, since it returns a mutable reference.
+
+## Writing a test for our specification
+
+Lets write a test to see if our specification works:
```
Aurel300, line 22:
```suggestion
Let's write a test to see if our specification works:
```

(did you mean to keep this part?)

docs/user-guide/src/tour/pledges.md
```diff
@@ -1,3 +1,73 @@
...
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/fail/user-guide/peek_mut_pledges.rs:test_peek_mut}}
+```
+
+But this fails, Prusti cannot verify any of our last three `prusti_assert` statements. This is where `pledges` come in. We have to tell Prusti how the `result` affects the original list. Without this, Prusti assumes that changes to the reference can change every property of the original list, so nothing can be known about it after the reference gets dropped.
+
+## Writing the pledge
+
+The pledge gets written with an annotation like for `ensures` and `requires`, but with the keyword `after_expiry`.
```
Aurel300, line 39:
```suggestion
The pledge is written in an `after_expiry` attribute.
```

docs/user-guide/src/tour/pledges.md
```diff
@@ -1,3 +1,73 @@
...
+```
+
+We have 3 conditions here:
+1. The list will have the same length afterwards
+2. Any element of the list with index `1..list.len()` will not be changed
+3. The element at the head of the list is the value that was assigned to the returned reference. This is denoted with the `before_expiry` function
+
+With these 3 conditions, our test verifies successfully!
+
+## Assert on expiry
```
Aurel300, line 53: Try to emphasise a bit more that the pledge precondition is an obligation that the *caller* needs to uphold.

docs/user-guide/src/tour/pop.md
```diff
@@ -1,13 +1,172 @@
...
+There is one change to the code from the original tutorial:
+We will rename the `pop` function to `try_pop`. The return type is still `Option<i32>` and `try_pop` will return `Some(item)` if the list has elements, and `None` otherwise. We will then add a new `pop` function, which has the return type `i32`, and will panic if it is called with an empty list. However, by using the correct precondition, we can prevent the `pop` function from ever panicking! Here is the code:
+
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/fail/user-guide/pop.rs:initial}}
+```
+
+For the implementation of our `pop` method, we can reuse the implementation of `try_pop`. We call `try_pop` on the list, then call [`unwrap`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap) on the result. `unwrap` will return the inner value of the `Option` if it is `Some`, and will panic otherwise.
+Normally, it is bad form to use `unwrap` in production code, where you should handle potential errors, instead of just panicking.
+But, since we are verifying that there will never be `None` passed to `unwrap`, we should be able to get away with it here.
```
Aurel300, line 17: (I wonder if this point is made anywhere yet:) we are indeed verifying the function and it will not panic *if* the precondition holds. When verifying an application, this is fine. However, if we are verifying library code, then the code using the library must also be verified by Prusti! Otherwise, nothing enforces that the user code is using the library correctly, and it could e.g. contain a call to `pop` over an empty list.

docs/user-guide/src/tour/pop.md
```diff
@@ -1,13 +1,172 @@
...
+
+## Properties of `try_pop`
+
+Lets start by (informally) listing the properties we want our `try_pop` method to have.
```
Aurel300, line 21:
```suggestion
Let's start by (informally) listing the properties we want our `try_pop` method to have.
```

(please also check throughout the guides)

docs/user-guide/src/tour/pop.md
```diff
@@ -1,13 +1,172 @@
...
+```rust,noplaypen
+{{#rustdoc_include ../../../../prusti-tests/tests/verify/fail/user-guide/pop.rs:pop_precondition}}
+```
+
+`try_pop` does not require a precondition, since it will never panic.
+
+### `try_pop` postcondition for empty lists
+
+Now we will implement the two conditions that hold for `try_pop` if you pass an empty list to it.
+To ensures that these are only checked for empty lists, we use the implication operator [`==>`](../syntax.md#implications):
```
Aurel300, line 79:
```suggestion
To ensure that these are only checked for empty lists, we use the implication operator [`==>`](../syntax.md#implications):
```

PR #1360, opened 2023-03-08
Implement export/import of quantifiers

jscissr:
When importing modules which contain specifications, these specs are also imported. The spec functions are serialized and written to a .specs file when the module is verified, and can be loaded from this file where the module is imported.
However, if there are quantifiers in the exported specs, this did not work before. Quantifiers are represented as closures, and are not included the spec functions themselves, they need to be exported separately. This change implements that.

There already exists a closure collecting visitor in prusti-interface/src/environment/collect_closure_defs_visitor.rs, but this is used for a different purpose, so it might not be a good idea to reuse it.

---

However, this changes the format of the .specs file. If you still have old .specs files in your target folder and run the new Prusti, it will panic trying to parse the old .specs file. This can be solved by deleting the target folder, but that is not a nice user experience. We would like everything to be re-verified when Prusti is updated.

Cargo uses various pieces of information to determine whether a recompile is needed, among them the rustc version. Cargo runs `rustc -vV` to obtain that. By replacing rustc with prusti-rustc, and outputting the Prusti version when running `prusti-rustc -vV`, we can achieve our goal. Everything is re-verified when the Prusti version changes.

Prusti previously set RUSTC_WRAPPER instead of RUSTC, but cargo does not use the wrapper when querying the version. I don't know why RUSTC_WRAPPER was used instead of RUSTC. Switching to RUSTC does not seem to cause problems.

### DIFF
--- a/prusti-tests/tests/cargo_verify/library_contracts_test/library_contracts_lib/src/lib.rs
+++ b/prusti-tests/tests/cargo_verify/library_contracts_test/library_contracts_lib/src/lib.rs
@@ -21,3 +21,21 @@ impl<T> Opt<T> {
         }
     }
 }
+
+#[trusted]
+#[pure]
+pub fn demo_fn(_a: u8) -> bool { unimplemented!() }
+
+// Test exporting quantifiers in specs.
+// Also test exporting triggers and nested quantifiers.
+#[trusted]
+#[ensures(forall(|i: u8| i == 0 ==> demo_fn(i)))]
+#[ensures(forall(|i: u8| i == 1 ==> forall(|j: u8| j == 1 ==> i == j), triggers=[(demo_fn(i),)]))]
+pub fn quantifier_test() {}
+
+// Test exporting quantifiers in predicates.
+predicate! {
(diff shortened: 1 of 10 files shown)


prusti-tests/tests/cargo_verify/library_contracts_test/library_contracts_lib/src/lib.rs
```diff
@@ -21,3 +21,21 @@ impl<T> Opt<T> {
...
+#[trusted]
+#[ensures(forall(|i: u8| i == 0 ==> demo_fn(i)))]
+#[ensures(forall(|i: u8| i == 1 ==> forall(|j: u8| j == 1 ==> i == j), triggers=[(demo_fn(i),)]))]
+pub fn quantifier_test() {}
+
+// Test exporting quantifiers in predicates.
+predicate! {
+  pub fn quantifier_predicate(i: u8) -> bool {
+    forall(|j: u8| j < i ==> demo_fn(j))
+  }
```
Aurel300: Nit: indent with 4 spaces.

PR #1364, opened 2023-03-12
JNI bindings - print actual class name on type mismatch

simon-hrabec:
This PR is the follow-up to https://github.com/viperproject/prusti-dev/pull/1345

For all generated dynamic checks/assertions for Java Objects (that they are of expected type) it now prints what the actual type was in case of mismatch (for potential debugging purposes).

**This PR Includes:**

- Code to generate the name of the type given a JObject
- A change to the runtime check generation to include a custom message for the assert including the actual type name
- Tests trying wrong objects to trigger all possible checks (result is not checked as it is not an input of the functions)
- fixed bug - missing check for result in field getter
- Code to copy the util functions file to gen folder

**What needs to be changed, fixed, or verified:**
- Get the relative path to builtins/utils.rs - ✔ workaround (generating the file)
- Avoid the same initialization in every test (DRY but also 2 tests in one file cause a failure - `JNI call failed source: Some(AlreadyCreated)`) - ✔ fixed
- Perhaps copy the whole builtins dir rather than its single file ✔ workaround (generating the file)
- Have shared functions for tests ✔ done
- In the generated files I have `use crate::wrappers::builtins::*;` - is this the correct approach? ✔ changed


jni-gen/builtins/utils.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub fn get_jobject_class_name<'a, O>(env: &'a JNIEnv<'a>, object: O) -> String
+where
+    O: Into<JObject<'a>>
+{
+    match get_jobject_class_name_inner(env, object) {
+        Ok(result) => result,
+        Err(_) => "Cannot be determined".to_string(),
+    }
+}
```
simon-hrabec: I had issues with Error types conversions when not having a dependency on `jni-gen`

jni-gen/src/generators/field_getter_setter.rs
```diff
@@ -134,13 +134,27 @@ fn generate_field_setter(class: &ClassName, field_name: &str, type_signature: &s
...
+        ));
+    }
     code.push("".to_string());
     code.push(generate_variable_type_check("receiver", "class_name"));
+    if is_signature_of_class_type(type_signature) {
+        code.push(generate_variable_type_check(
+            &parameter_name,
+            &format!("{parameter_name}_class_name"),
+        ));
+    }
```
simon-hrabec, line 153: This thing (dynamic check for result/returned value) was missing by accident.

jni-gen/src/wrapper_generator.rs
```diff
@@ -62,6 +66,9 @@ impl WrapperGenerator {
 
         //remove_dir_all(out_dir)?;
         create_dir_all(out_dir)?;
+        create_dir_all(out_dir.join("builtins"))?;
+        let builtins_utils = Path::new("/home/simon/repo/prusti-dev/jni-gen/builtins/utils.rs");
```
simon-hrabec: This will work only on my machine. Is there an elegant way to get a path to the current file? I could come up with some working solution (like taking the path to the current crate, go up until you find the root directory of the repository (e.g. by cargo.toml having workspace) and then adding the relative path from the root of the repository, but I guess that's not a very robust/clean solution.

jni-gen/builtins/utils.rs
fpoli: Could you show one of those errors?

fpoli:
```suggestion
}

```

jni-gen/builtins/utils.rs
```diff
@@ -0,0 +1,39 @@
...
+                 .l()?
+                 .into(),
+        )?,
+    );
+
+    match res {
+        Ok(str) => Ok(str.replace(".", "/")),
+        Err(_) => Ok("Cannot be determined".to_string()),
+    }
+}
```
fpoli: Could you add a check of the type of the `object` parameter, executed only when compiling in debug mode?

jni-gen/src/wrapper_generator.rs
fpoli: See this to obtain the path of the current module: https://doc.rust-lang.org/std/macro.module_path.html. Then, you can join `../builtins`. For maintainability of `jni-gen`, it's better to never make assumptions about where the crate is placed in the repository or how Prusti uses it.

jni-gen/systest/tests/field_setter_check_fails.rs
```diff
@@ -0,0 +1,85 @@
...
+#[test]
+#[should_panic(expected = "Java binding type failure. Expected object of class java/lang/String, but got java/lang/Integer instead")]
+fn field_setter_should_fail_on_wrong_argument() {
+    let jvm_args = InitArgsBuilder::new()
+        .version(JNIVersion::V8)
+        .option("-Xcheck:jni")
+        .option("-Xdebug")
+        .option("-XX:+CheckJNICalls")
+        //.option("-verbose:jni")
+        //.option("-XX:+TraceJNICalls")
```
fpoli: (Here and in the other new tests) Did you find all of some of these two options useful for debugging the tests? If not, could you remove those that you didn't need instead of commenting them?

jni-gen/systest/tests/method_runtime_check_fails.rs
```diff
@@ -0,0 +1,78 @@
...
+        let integer_object = integer_wrapper.new(1337)?;
+        let error_wrapper = java::lang::Error::with(&env);
+        let error_object = error_wrapper.new(string_to_jobject(&env, "error message")?)?;
+        let _result = integer_wrapper.call_compareTo(integer_object, error_object);
+        Ok(JObject::null())
+    }).unwrap_or_else(|e| {
+        print_exception(&env);
+        panic!("{} source: {:?}", e, std::error::Error::source(&e));
+    });
+}
```
fpoli:
```suggestion
}

```

jni-gen/src/wrapper_generator.rs
```diff
@@ -62,6 +66,9 @@ impl WrapperGenerator {
...
+        copy(builtins_utils, out_dir.join("builtins/utils.rs"))?;
```
fpoli: Please use the `walkdir` crate to recursively copy the content of the folder
```rust
for entry in walkdir::WalkDir::new(builtins_dir) { /* do the copy */ }
```

jni-gen/src/generators/module.rs
```diff
@@ -43,6 +43,8 @@ pub fn generate_module(class_names: Vec<&ClassName>) -> String {
     vec![
         "//! Automatically generated code\n".to_string(),
         "#![allow(non_snake_case)]\n".to_string(),
+        "#[path = \"builtins/utils.rs\"]".to_string(),
```
fpoli:
```suggestion
```
If you rename `utils.rs` to `mod.rs`, then you can import the builtins without having to specify a path.

jni-gen/src/generators/class.rs
```diff
@@ -69,6 +69,7 @@ impl<'a> ClassGenerator<'a> {
             "use jni::signature::*;",
             "use once_cell::sync::OnceCell;",
             "use std::str::FromStr;",
+            "use crate::wrappers::builtins::*;",
```
fpoli:
```suggestion
            "use super::builtins::*;",
```
When builting the `modules_tree` in `generate_module`, in every module you can import `builtins` from the parent module, so that it's always available:
```rust
res.push(format!("pub mod {name} {{\n"));
```
:arrow_down: 
```rust
res.push(format!("pub mod {name} {{\n"));
res.push(format!("use super::builtins;\n"));
```

fpoli:
> * Get the relative path to builtins/utils.rs

See the comment.

> * Perhaps copy the whole builtins dir rather than its single file

Yes, I'd not merge without that.

> * In the generated files I have `use crate::wrappers::builtins::*;` - is this the correct approach?

See the comment. `jni-gen` should not assume that the generated code always ends up in `crate::wrappers`.

> * Avoid the same initialization in every test (DRY but also 2 tests in one file cause a failure - `JNI call failed source: Some(AlreadyCreated)`)
> * Have shared functions for tests

Have you tried defining a `jni-gen/tests/lib.rs` with the shared functions, importing them with `use self::lib::*;` in the tests?

jni-gen/builtins/utils.rs
fpoli: Maybe you miss something like a `use jni::errors::{Error, Result as JniResult};`

jni-gen/src/wrapper_generator.rs
simon-hrabec: I checked the `module_path!` macro, but it returns the path in terms of modules - `jni_gen::wrapper_generator`
I also tried the `file!` macro, but it only returns the path to the current file within the crate (`jni-gen/src/wrapper_generator.rs`).
However, I am unable to obtain the absolute path to the `jni_gen` crate.
I can get the absolute path of the crate that is being built by `std::env::current_dir()` (which should be the same as `env::var("CARGO_MANIFEST_DIR")`).
This can be `/home/simon/repo/prusti-dev/viper-sys` or `/home/simon/repo/prusti-dev/jni-gen/systest` (the things before `jni-gen` are just for my PC).

but I don't know how to get the path to the `jni-gen` crate. I (incorrectly, as you stated) assume that the `jni-gen` is in the root of the repository and get the repository root by truncating the `current_dir()` path.

Aurel300: Instead of *copying* the file using `jni-gen`, you can compile the file *into* `jni-gen` with something like [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) and just make it write the file.

fpoli: > I checked the module_path! macro, but it returns the path in terms of modules - jni_gen::wrapper_generator

Oh, my bad.

> I also tried the file! macro, but it only returns the path to the current file within the crate (jni-gen/src/wrapper_generator.rs).

That sounds good, right? Can `file!()` joined with `../../builtins` be used as path of the source folder of the copy?

jni-gen/builtins/utils.rs
simon-hrabec: My problems, in this case, were that some things are reliant on `jni-gen`

```
pub fn java_str_to_string(string: &JavaStr) -> JniResult<String> {
    unsafe { Ok(CStr::from_ptr(string.get_raw()).to_str()?.to_string()) }
}
```
fails with:
```
error[E0277]: `?` couldn't convert the error to `jni::errors::Error`
  --> jni-gen/systest/src/../gen/builtins/utils.rs:14:58
   |
13 | pub fn java_str_to_string(string: &JavaStr) -> JniResult<String> {
   |                                                ----------------- expected `jni::errors::Error` because of this
14 |     unsafe { Ok(CStr::from_ptr(string.get_raw()).to_str()?.to_string()) }
   |                                                          ^ the trait `From<Utf8Error>` is not implemented for `jni::errors::Error`
```
Which can be solved with `use jni_gen::errors::*`, but this requires then to have `jni_gen` as a regular (not just build) dependency for both systest and vyper-sys. Is that ok? The change would look like this (pushed to this branch) https://github.com/viperproject/prusti-dev/pull/1364/commits/781ea03369e27954742a68826525d5d25c7bb9f9

jni-gen/src/wrapper_generator.rs
simon-hrabec: > That sounds good, right? Can `file!()` joined with `../../builtins` be used as path of the source folder of the copy?

That gives a relative path within the `jni-gen`, but when the file is copied it is during the build (running build.rs) of either `systest` or `viper-sys`, but I would then need to combine it with path to `jni-gen`, which I don't have (basically the path that is written in the `cargo.toml` files as a path to `jni-gen`).

I think the @Aurel300's approach might be an easy fix for this.

jni-gen/src/generators/class.rs
simon-hrabec: Thanks, I could not figure this out.

jni-gen/src/wrapper_generator.rs
simon-hrabec: Not applicable anymore as changed the code ([commit](https://github.com/viperproject/prusti-dev/pull/1364/commits/faade4e4854e610090de7fd487a5d15cacab2451)) according to Aurel's [suggestion](https://github.com/viperproject/prusti-dev/pull/1364#discussion_r1135616848).

jni-gen/builtins/utils.rs
fpoli: Adding a dependency on `jni-gen` is not necessary and would break the high-level design of `jni-gen`; I would avoid that. Your issue is that `to_str` returns a `Result<&str, Utf8Error>`, but you declared the return type of `java_str_to_string` to be `Result<String, jni::errors::Error>`. The compiler complains that there is no automatic convertsion from `Utf8Error` to `jni::errors::Error`. Since `utils.rs` is only used in debug assertions I think the easier solution is to replace every `?` of non-jni errors with an `.unwrap_or_else(|err| panic!("Failed to ...: {err:?}"))`.

jni-gen/src/wrapper_generator.rs
fpoli: Oh, I see. Yes, Aurel's approach sounds good. Alternatively, to obtain the path of the crate being compiled you can use `std::env!("CARGO_MANIFEST_DIR")`, which evaluates at *compilation time* the environment variable set by cargo.

jni-gen/systest/tests/field_setter_check_fails.rs
simon-hrabec: I just copied the code from https://github.com/viperproject/prusti-dev/blob/master/jni-gen/systest/tests/jvm_builtin_classes.rs#L26 without any changes. Removed in this version

simon-hrabec:
> > * Avoid the same initialization in every test (DRY but also 2 tests in one file cause a failure - `JNI call failed source: Some(AlreadyCreated)`)
> > * Have shared functions for tests
> 
> Have you tried defining a `jni-gen/tests/lib.rs` with the shared functions, importing them with `use self::lib::*;` in the tests?

I tried to implement that in the last commit - https://github.com/viperproject/prusti-dev/pull/1364/commits/99554e54839e3682a8646db47c05b7cbea0a83f8

It requires some `unsafe`, not sure if it is the right approach or if some `mutex`/`lazy_static` should be used (as shown in [this SO question](https://stackoverflow.com/questions/27791532/how-do-i-create-a-global-mutable-singleton)).

fpoli:
> > > * Avoid the same initialization in every test (DRY but also 2 tests in one file cause a failure - `JNI call failed source: Some(AlreadyCreated)`)
> > > * Have shared functions for tests
> > 
> > 
> > Have you tried defining a `jni-gen/tests/lib.rs` with the shared functions, importing them with `use self::lib::*;` in the tests?
> 
> I tried to implement that in the last commit - [99554e5](https://github.com/viperproject/prusti-dev/commit/99554e54839e3682a8646db47c05b7cbea0a83f8)
> 
> It requires some `unsafe`, not sure if it is the right approach or if some `mutex`/`lazy_static` should be used (as shown in [this SO question](https://stackoverflow.com/questions/27791532/how-do-i-create-a-global-mutable-singleton)).

If you have time, using [`LazyLock`](https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html) seems a better solution because it doesn't need unsafe code. (It requires to add `#![feature(once_cell)]` to every test.) If you don't have time, feel free to just add a `// TODO: Refactor using a LazyLock<JavaVM>`.

As a simpler alternative, you can avoid the complication of the global variable and just define a `pub fn build_jvm() -> Result<JavaVM>` in `lib.rs` that can be used to initialize a local variable.

jni-gen/builtins/utils.rs
simon-hrabec: Removed the dependency

simon-hrabec:
> variable and just define a `pub fn build_jvm() -> Result<JavaVM>` in `lib.rs` that can be used to initialize a local variable

I am quite limited by time, hence I added the comment.

When I just copy-pasted the initiation in the beginning of every test function I had problems when there were more tests in a single file only the first test would pass and the second would fail due to failure to initialize the JVM. I thought having it as a function (just for DRY) would not solve that problem.

jni-gen/builtins/utils.rs
simon-hrabec: I am sorry I am not sure I understand what you mean by this comment.

This function is called (indirectly) to make the argument for the debug_assert! macro (the error message). Do I understand it correctly that it will not be invoked in release mode?

fpoli: I thought that the argument had to be an instance of a `java.lang.Class` and I was asking for an `assert_debug!` subtyping check, but I now see that I was wrong because `get_object_class` takes care of retrieving the class. So, nothing to change here.

jni-gen/src/wrapper_generator.rs
fpoli: The PR uses `include_str!("../builtins/mod.rs");`, so we can close this thread.

jni-gen/builtins/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+                 .into(),
+        )?,
+    )
+}
+
+pub fn get_jobject_class_name<'a, O>(env: &'a JNIEnv<'a>, object: O) -> String
+where
+    O: Into<JObject<'a>>
+{
+    get_jobject_class_name_inner(env, object).unwrap_or("Cannot be determined".to_string()).replace('.', "/")
```
fpoli:
```suggestion
    get_jobject_class_name_inner(env, object).unwrap_or("<unknown>".to_string()).replace('.', "/")
```

fpoli:
@simon-hrabec Is "[WIP]" still in the title because you want to do more changes, or can it be merged?

simon-hrabec:
> @simon-hrabec Is "[WIP]" still in the title because you want to do more changes, or can it be merged?

Sorry, I forgot to change it yesterday (remove the WIP tag). It is mergable with the last change proposal you made (<unknown>). I am not on my PC right now to update it, feel free to do so (including merging it).

PR #1365, opened 2023-03-14
Add prusti_refute! macro

simon-hrabec:
This PR introduces 2 things:

1. It adds prusti_refute! macro
2. It changes how Prusti works with the verifier. Instead of directly having the verifier, it makes a frontend that encapsulates it so we can take advantage of the plugins (which are needed for processing refute statements).

It makes it possible to use the `prusti_refute!` macro in your rust code. There is [documentation](https://viperproject.github.io/prusti-dev/user-guide/verify/assert_assume.html) about `prusti_assert!` and `prusti_assume!` - details on the `prusti_refute!` macro have also been added to [the docs (user guide)](https://github.com/viperproject/prusti-dev/pull/1365/files#diff-e0169b89f7df5485fd3ad11823bbc48ecff66611d7f00fb9886942091d701f83). The macro does not change the function of the program, it is there only for verification purses. It can be used similarly to `prusti_assert!`,


viper/src/verifier.rs
```diff
@@ -80,7 +121,7 @@ impl<'a> Verifier<'a> {
         match backend {
             VerificationBackend::Silicon => silicon::Silicon::with(env).new(reporter, debug_info),
             VerificationBackend::Carbon => {
-                carbon::CarbonVerifier::with(env).new(reporter, debug_info)
+                panic!("Trying to use Carbon backend - now made compatible just with Silicon");
```
fpoli: The subject of "now made compatible just with Silicon" appears to be Carbon, but should it be something like "the viper crate"?

fpoli: It would be useful to add before the panic a comment that explains why Carbon is no longer supported, so that we know what needs to be fixed.

simon-hrabec:
It seems there is an issue that I am not sure how to easily resolve:

Scala has traits that translate to Java interfaces. However, Scala allows adding fields to traits, but in Java, interfaces have no fields. If I understand it correctly, Scala will just put those fields into the classes that extend the trait (implement the interface). This problem manifests when I want to modify a field declared in a trait via a wrapper that represents the trait and holds one of several objects of classes that extend the trait.

More specifically:
The `DefaultFrontend` defines the field `_program` - https://github.com/viperproject/silver/blob/master/src/main/scala/viper/silver/frontend/Frontend.scala#L168

When I want to use the `field!` macro, I can use it only on the concrete classes such as `SiliconFrontend` and `CarbonFrontend`. However, I wanted to have a variable of type `SilFrontend`, but the macro `field!` fails in this case, because it does not work with interfaces but also because interfaces don't have fields (scala only generates some helper methods for each field).

This prevents polymorphic access to the frontend (that's why I put replaced all calls to the frontend_wrapper with a match statement that calls the respective one).

simon-hrabec:
Also getting Prusti using Carbon not working with the following code:
```
fn sign(val: i32) -> i32 {
	prusti_refute!(false);
    if val <= 0 {
        -1
    } else if val >= 0 {
        1
    } else {
        0
    }
}
```
However, I get the same failures when trying to verify the dumped viper program (Silicon passing, Carbon failing).

simon-hrabec:
> It seems there is an issue that I am not sure how to easily resolve:
> 
> Scala has traits that translate to Java interfaces. However, Scala allows adding fields to traits, but in Java, interfaces have no fields. If I understand it correctly, Scala will just put those fields into the classes that extend the trait (implement the interface). This problem manifests when I want to modify a field declared in a trait via a wrapper that represents the trait and holds one of several objects of classes that extend the trait.
> 
> More specifically: The `DefaultFrontend` defines the field `_program` - https://github.com/viperproject/silver/blob/master/src/main/scala/viper/silver/frontend/Frontend.scala#L168
> 
> When I want to use the `field!` macro, I can use it only on the concrete classes such as `SiliconFrontend` and `CarbonFrontend`. However, I wanted to have a variable of type `SilFrontend`, but the macro `field!` fails in this case, because it does not work with interfaces but also because interfaces don't have fields (scala only generates some helper methods for each field).
> 
> This prevents polymorphic access to the frontend (that's why I put replaced all calls to the frontend_wrapper with a match statement that calls the respective one).

I propose either:

- Change the jni-gen field! macro in some way to allow to generate a wrapper for the frontend with the `_program` field (e.g. not do the check when generating the wrapper) and deal with the possible absence of the field at runtime.
- Create some wrapper that would do the dispatching to either carbon frontend or silicon frontend.
- Keep the current code where all calls to the frontend and wrapped in the match statement.

fpoli:
I think it's much simpler than that. I see that the compiled `DefaultFrontend` has an abstract method called `_program` that returns a `scala.Option<viper.silver.ast.Program>` . So, my understanding is that Scala traits are desugared to Java interfaces, and trait fields to abstract methods.

Here is how I learded that. To list all the methods in a compiled Java class or interface the important command to know is `javap`:
```bash
unzip -q viperserver.jar -d viperserver_classes
cd viperserver_classes
javap -s 'viper.silver.frontend.DefaultFrontend'
```
In the output I see:
```
Compiled from "Frontend.scala"
public interface viper.silver.frontend.DefaultFrontend extends viper.silver.frontend.DefaultPhases,viper.silver.frontend.SingleFileFrontend {
  ...
  public abstract scala.Option<viper.silver.ast.Program> _program();
    descriptor: ()Lscala/Option;
  ...
}
```

simon-hrabec:
> I think it's much simpler than that. I see that the compiled `DefaultFrontend` has an abstract method called `_program` that returns a `scala.Option<viper.silver.ast.Program>` . So, my understanding is that Scala traits are desugared to Java interfaces, and trait fields to abstract methods.
> 
> Here is how I learded that. To list all the methods in a compiled Java class or interface the important command to know is `javap`:
> 
> ```shell
> unzip -q viperserver.jar -d viperserver_classes
> cd viperserver_classes
> javap -s 'viper.silver.frontend.DefaultFrontend'
> ```
> 
> In the output I see:
> 
> ```
> Compiled from "Frontend.scala"
> public interface viper.silver.frontend.DefaultFrontend extends viper.silver.frontend.DefaultPhases,viper.silver.frontend.SingleFileFrontend {
>   ...
>   public abstract scala.Option<viper.silver.ast.Program> _program();
>     descriptor: ()Lscala/Option;
>   ...
> }
> ```

(Wiriting froma gym from a phone)
IIRC the issue was in setting some variable/field for which there was no method.

I will check it when I get home.
Btw I was using the JD-GUI to inspect the jar.

fpoli:
> IIRC the issue was in setting some variable/field for which there was no method.

This might be the method to *assign* a `Program` to the field. Just a guess based on the name and signature.
```
  public abstract void _program_$eq(scala.Option<viper.silver.ast.Program>);
    descriptor: (Lscala/Option;)V
```

> Btw I was using the JD-GUI to inspect the jar.

Nice!

simon-hrabec:
> > IIRC the issue was in setting some variable/field for which there was no method.
> 
> This might be the method to _assign_ a `Program` to the field. Just a guess based on the name and signature.
> 
> ```
>   public abstract void _program_$eq(scala.Option<viper.silver.ast.Program>);
>     descriptor: (Lscala/Option;)V
> ```
> 
> > Btw I was using the JD-GUI to inspect the jar.
> 
> Nice!

right right, you are correct. I think I can change it so that if the class is an interface it is gonna look for those 2 methods instead and generate the functions for the wrapper a bit differently.

fpoli:
> I think I can change it so that if the class is an interface it is gonna look for those 2 methods instead and generate the functions for the wrapper a bit differently.

If `_program` is the only trait field that you need to use, then I suggest to just call the `_program()` and `_program_$eq(..)` where needed in the `viper` crate and stop there. If you have time and want to work more on `jni-gen`  (e.g. an additional `trait_field!` macro, or a smarter `field!` macro) then I think we have to first briefly internally sync on the design of `jni-gen`.

The *current* design of `jni-gen` is very simple. It doesn't try to hide from the user that the Scala compiler did some desugaring into Java. It mainly just generate a Rust function for every compiled Java method (and two Rust functions per Java field). This is enough to generate `viper-sys`, which is the *low-level* interface to talk to Viper. The *high-level* interface should go in `viper`. Since this keep-it-simple design worked very well so far I'd not change it unless we have a strong reason.

simon-hrabec:
> > I think I can change it so that if the class is an interface it is gonna look for those 2 methods instead and generate the functions for the wrapper a bit differently.
> 
> If `_program` is the only trait field that you need to use, then I suggest to just call the `_program()` and `_program_$eq(..)` where needed in the `viper` crate and stop there. If you have time and want to work more on `jni-gen` (e.g. an additional `trait_field!` macro, or a smarter `field!` macro) then I think we have to first briefly internally sync on the design of `jni-gen`.
> 
> The _current_ design of `jni-gen` is very simple. It doesn't try to hide from the user that the Scala compiler did some desugaring into Java. It mainly just generate a Rust function for every compiled Java method (and two Rust functions per Java field). This is enough to generate `viper-sys`, which is the _low-level_ interface to talk to Viper. The _high-level_ interface should go in `viper`. Since this worked very well so far I'd not change it unless we have a strong reason.

That will be an easy solution. I will push my current work on the field! into a separate branch so if someone wants to continue they can start there.

simon-hrabec:
> > I think I can change it so that if the class is an interface it is gonna look for those 2 methods instead and generate the functions for the wrapper a bit differently.
> 
> If `_program` is the only trait field that you need to use, then I suggest to just call the `_program()` and `_program_$eq(..)` where needed in the `viper` crate and stop there. If you have time and want to work more on `jni-gen` (e.g. an additional `trait_field!` macro, or a smarter `field!` macro) then I think we have to first briefly internally sync on the design of `jni-gen`.
> 
> The _current_ design of `jni-gen` is very simple. It doesn't try to hide from the user that the Scala compiler did some desugaring into Java. It mainly just generate a Rust function for every compiled Java method (and two Rust functions per Java field). This is enough to generate `viper-sys`, which is the _low-level_ interface to talk to Viper. The _high-level_ interface should go in `viper`. Since this keep-it-simple design worked very well so far I'd not change it unless we have a strong reason.

Done as suggested - works the same way, cleaner.

viper/src/verifier.rs
simon-hrabec: Fixed (reverted)

fpoli:
Thanks! Could you also add to the description of this PR the instructions to make use of the `prusti_refute!` statement? I don't know them, but I'd expect 2/3 steps like "compile the Viper plugin `github.com/viperproject/xyz` using the command `sbt xyz`; copy the jar `xyz/refute_plugin.jar` into `prusti_tools/xyz`; run Prusti with the special flag `xyz=xyz`".

simon-hrabec:
> Thanks! Could you also add to the description of this PR the instructions to make use of the `prusti_refute!` statement? I don't know them, but I'd expect 2/3 steps like "compile the Viper plugin `github.com/viperproject/xyz` using the command `sbt xyz`; copy the jar `xyz/refute_plugin.jar` into `prusti_tools/xyz`; run Prusti with the special flag `xyz=xyz`".

Added some tests with `prusti_refute!` and also updated the documentation as well as added a few sentences to the PR description.

There is no special setup required to use the macro in your code besides the code updates of Prusti in this PR.

simon-hrabec:
> > I think I can change it so that if the class is an interface it is gonna look for those 2 methods instead and generate the functions for the wrapper a bit differently.
> 
> If `_program` is the only trait field that you need to use, then I suggest to just call the `_program()` and `_program_$eq(..)` where needed in the `viper` crate and stop there. If you have time and want to work more on `jni-gen` (e.g. an additional `trait_field!` macro, or a smarter `field!` macro) then I think we have to first briefly internally sync on the design of `jni-gen`.
> 
> The _current_ design of `jni-gen` is very simple. It doesn't try to hide from the user that the Scala compiler did some desugaring into Java. It mainly just generate a Rust function for every compiled Java method (and two Rust functions per Java field). This is enough to generate `viper-sys`, which is the _low-level_ interface to talk to Viper. The _high-level_ interface should go in `viper`. Since this keep-it-simple design worked very well so far I'd not change it unless we have a strong reason.

I tried to modify the `field!` macro as described and made it work. I create a WIP/draft PR that shows can it can be done - https://github.com/viperproject/prusti-dev/pull/1368. There is still a lot of work to be done on it, especially it would be better to extract some of the reflection calls into separate functions. Also, I just reuse the same error as a placeholder.

prusti-viper/src/encoder/errors/error_manager.rs
```diff
@@ -696,6 +698,20 @@ impl<'tcx> ErrorManager<'tcx> {
...
+                    error_span,
+                )
+            }
+
+            ("assert.failed:assertion.false", ErrorCtxt::Panic(PanicCause::Refute)) => {
+                PrustiError::verification(
+                    "the refuted expression holds in all cases or could not be reached",
+                    error_span,
+                )
+            }
```
JonasAlaif: @simon-hrabec are both of these needed? Won't Viper always return either `"assert.failed:assertion.false"` or `"refute.failed:refutation.true"` (I suspect the latter)?

simon-hrabec: I remember in the past I somehow got the `assert.failed:assertion.false` but now cannot reproduce it so either I was wrong or it was triggered by some other issue I fixed in the meantime. Seems only the `refute.failed:refutation.true` is sufficient.

PR #1368, opened 2023-03-21
Make trait_field! macro work for trait fields

simon-hrabec:
In the previous PR https://github.com/viperproject/prusti-dev/pull/1365 in the [comment](https://github.com/viperproject/prusti-dev/pull/1365#issuecomment-1473465741) I described what limitations the `field!` macro has and in this PR I propose a way to make it work even when the field comes from scala trait.

Note - this is an extension of the PR, but I cannot open a PR within those two branches, because I cannot push to this repository. Therefore the changes also contain all the changes in the `prusti_refute!` macro PR.

This is nowhere complete - it is more a mere draft/proposal.


fpoli:
@JonasAlaif my suggestion would be to introduce a new `trait_field!(..)` macro instead of adding more semantics to `field!(..)` with the `if is_interface(env, class)?` in `jni-gen/src/generators/field_getter_setter.rs` (~KISS).

PR #1416, opened 2023-06-23
Simplify creation of pages

fpoli:
* Delete the benchmark workflow, which we never used so far.
    * The old benchmark's data will remain available in the [`old-benchmarks`](https://github.com/viperproject/prusti-dev/tree/old-benchmarks) branch.
* Merge the workflows that generate GitHub pages into one.
* Change the index page to redirect to the repository.


fpoli:
`cargo doc` reports a bunch of useful warnings. We should fix them and change the command to fail on warnings at some point.

PR #1418, opened 2023-06-26
Use position IDs as Viper line numbers

fpoli:
Another attempt at https://github.com/viperproject/prusti-dev/pull/1389.

This PR ensures that Viper positions have unique (line, column) pairs, since Silicon does not use the identifier when deduplicating verification errors from multiple execution paths. Related discussion: https://github.com/viperproject/silicon/issues/732.

The line and column of VIR positions will still refer to the source Rust program, because Prusti uses them to deduplicate counterexamples. Explanation: https://github.com/viperproject/prusti-dev/pull/1389#issuecomment-1511728866. They are then ignored when translating to Viper.

This PR adds a `num_errors_per_function` configuration flag to set `--numberOfErrorsToReport` in Silicon. However, I still can't get more than one error per function. :thinking:


fpoli:
See https://github.com/viperproject/silicon/issues/735 for the explanation of why we don't get multiple verification errors from Silicon.

PR #1430, opened 2023-07-20
Update dependencies (rustc nightly-2023-07-15)

vakaras:


vakaras:
@Aurel300 Your turn.

vakaras:
@Aurel300 What is the status of this?

Aurel300:
@vakaras Can you reproduce the error? Locally, when testing `error_position.rs`, I get the expected error locations, but on the CI they are one line off, it seems?

fpoli:
> @vakaras Can you reproduce the error? Locally, when testing `error_position.rs`, I get the expected error locations, but on the CI they are one line off, it seems?

Did you re-download the `viper_tools`? Maybe it'd due to a change in Viper and you still have the old JARs.

Aurel300:
Ah, I actually did not, my bad!

fpoli:
It would actually be nice to make `x.py` auto-update the `viper_tools` when `viper-toolchain` changes... Adding a metadata file with the tag into all released Viper zip packages would be one way to do that.

vakaras:
@Aurel300 What is the status of this?

Aurel300:
@vakaras I am still waiting for #1434: I see `viper-sys` is failing to build in the PR. Do you want me to take over that PR as well? Otherwise, this is either blocked or we don't upgrade the Viper version yet.

vakaras:
> @vakaras I am still waiting for #1434: I see `viper-sys` is failing to build in the PR. Do you want me to take over that PR as well? Otherwise, this is either blocked or we don't upgrade the Viper version yet.

Marco is still working on it, so I would suggest not updating Viper.

vakaras:
> It would actually be nice to make `x.py` auto-update the `viper_tools` when `viper-toolchain` changes... Adding a metadata file with the tag into all released Viper zip packages would be one way to do that.

You do not need to put the meta-data into the ZIP file. It should be sufficient to just change `x.py setup` to copy `viper_toolchain` to `viper_tools` folder after a successful update.

PR #1435, opened 2023-08-04
Update GitHub actions and fix release builds

vakaras:
Update GitHub actions to fix deprecation warnings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -263,6 +263,40 @@ jobs:
           cd prusti-contracts/prusti-contracts-test/
           cargo +stable build
 
+  release-build:
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04, windows-latest, macos-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -258,6 +263,40 @@ jobs:
           cd prusti-contracts/prusti-contracts-test/
           cargo +stable build
 
+  release-build:
```
fpoli, line 266: Avoid running the job if any of the quick checks fails.
```suggestion
  release-build:
    needs: [fmt-check, clippy-check, check-deps, smir-check, quick-tests]
```

fpoli, line 266: Important: the new `release-build` job has to be added to the `needs` key of the `can-merge` job. Without that, pull requests will be "green" even if the `release-build` fails or is canceled.

fpoli:
Found it! The linker error seems to be due to rust-lang/rust#111888, because setting `codegen-units = 1` fixes it :smile:

.github/workflows/test.yml
```diff
@@ -258,6 +263,34 @@ jobs:
...
+    needs: [fmt-check, clippy-check, check-deps, smir-check, quick-tests]
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04, windows-latest, macos-latest]
```
fpoli:
```suggestion
        os: [ubuntu-latest, windows-latest, macos-latest]
```

fpoli:
I'm not convinced we should keep the `build-release` job, because it raises the number of VMs used in parallel to **11 per CI run**, and GitHub only provides a maximum of 20 parallel VMs across all projects of our organization.

I'll merge for the moment, but we should follow up on this.

Issue #1214, opened 2022-10-25
Internal error from mutually recursive pure functions


Patrick-6:
Mutually recursive pure functions cause an error:
`[Prusti internal error] Prusti encountered an unexpected internal error`
`Details: Rust function m_bug$$mutually_recursive_pure_fn$$f encoded both as a Viper method and function`.
There are no errors if both are not marked with `#[pure]`:
<pre><code>use prusti_contracts::*;

#[pure]
#[ensures(result == 0)]
fn f(n: u64) -> u64 {
    if n == 0 {
        0
    } else {
        g(n - 1)
    }
}

#[pure]
#[ensures(result == 0)]
fn g(n: u64) -> u64 {
    if n == 0 {
        0
    } else {
        f(n - 1)
    }
}
</pre></code>

fpoli:
Thank you for the report!

nokunish:
Hi, 

There might be a similar issue with mutually referential specifications. For instance, if we have 

```
#[pure]
#[trusted]
#[ensures(result == nested_fn(_u))]
pub fn is_true(_u: usize) -> bool {
    return true
}

#[pure]
#[trusted]
#[ensures(result == is_true(u))]
pub fn nested_fn(u: usize) -> bool {
    is_true(u)
}
```
then
1. If neither are trusted, it crashes Prusti
2. If only one of them is trusted, then it produces the error:
    ```Details: Rust function m_test_v1$$external_spec$$bug$$spec_bug$$SomeVec$$$openang$T$closeang$$$len encoded both as a Viper method and function```
4. If both are trusted, it verifies

PR #1440, opened 2023-08-14
Fix encoding of mutually recursive pure functions

### LINKED ISSUE #1267: Internal error from pure function calling itself in its contract
The following code produces an internal error in addition to the invalid specification error. It also marks `fn double(x: i64) -> i64` as the error source and not the `double(x)` call in the postcondition:
```
use prusti_contracts::*;

#[pure]
#[ensures(double(x) % 2 == 0)]
fn double(x: i64) -> i64 {
    2 * x
}
```
Produced error:
```
[Prusti: invalid specification] only trusted functions can call themselves in their contracts
[Prusti internal error] Prusti encountered an unexpected internal error
fn_calling_itself_in_specs.rs(5, 1): We would appreciate a bug report: https://github.com/viperproject/prusti-dev/issues/new
fn_calling_itself_in_specs.rs(5, 1): Details: Rust function m_uncategorized_problems$$fn_calling_itself_in_specs$$double encoded both as a Viper method and function
```

fpoli:
Use the `f_` prefix for encoding pure function names instead of `m_`.

Partially addresses #1214.
Fixes #1267.

### DIFF
--- a/prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
+++ b/prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
@@ -1717,9 +1717,6 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 {
                     // If we are verifying a pure function, we always need
                     // to encode it as a method
-                    //
-                    // FIXME: This is not enough, we also need to handle
-                    // mutual-recursion case.
                     let (function_name, return_type) = self
                         .encoder
                         .encode_pure_function_use_high(called_def_id, self.def_id, call_substs)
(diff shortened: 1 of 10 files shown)


fpoli:
@vakaras The fix made the purification test `fail/core_proof/model2.rs` fail with 3 more errors of the kind "the termination measure of this call is not necessarily lower". I have no clue what caused them, so I [excluded](https://github.com/viperproject/prusti-dev/pull/1440/commits/3bc1d2079d2238131a924daa2bee797ae1319141) your encoder from this PR.

prusti-viper/src/encoder/mir/procedures/encoder/mod.rs
```diff
@@ -1717,9 +1717,6 @@ impl<'p, 'v: 'p, 'tcx: 'v> ProcedureEncoder<'p, 'v, 'tcx> {
                 {
                     // If we are verifying a pure function, we always need
                     // to encode it as a method
-                    //
-                    // FIXME: This is not enough, we also need to handle
-                    // mutual-recursion case.
```
vakaras, line 1722: This should be kept because the soundness check is still missing.

PR #1490, opened 2024-01-17
Disable the "more complete" exhale

fpoli:
Let's see what happens if we disable this; Viper got better in the meantime.


fpoli:
It seems that only counterexample-related UI tests are failing, with "Java exception: java.lang.RuntimeException: Silicon hasn't been started yet - java.lang.RuntimeException: Silicon hasn't been started yet".

Aurel300:
@fpoli What happened here? The runtime error seems unrelated to counterexamples. (Or perhaps related to passing the wrong flags?)

PR #1492, opened 2024-02-08
Fix doc

pixelshot91:


Aurel300:
Thanks! The changes look good, but why the new anchor? I don't see it used anywhere.

pixelshot91:
The anchor is used here: https://viperproject.github.io/prusti-dev/user-guide/tour/pledges.html#writing-a-test-for-our-specification
Right now the snippet only shows this:
```rust
#[cfg(prusti)]
mod prusti_tests {
    use super::*;
```

fpoli:
Can we merge, @Aurel300?

Aurel300:
> The anchor is used here: [...]

Ah, my bad. Thanks!

PR #1511, opened 2024-03-25
Fix issue #1505

### LINKED ISSUE #1505: Wrong encoding of signed divisions
Given `a / b`, for `a < 0` in Rust the result is still rounded toward 0, while in Viper it is rounded away from 0. Prusti currently encodes signed integer divisions using Viper's semantics, which is wrong.

```rust
use prusti_contracts::*;

// To inhibit constant-propagation optimizations:
#[pure]
fn id<T>(x: T) -> T { x}

pub fn main(){
    assert!(id(3i32) / 2 == 1);
    assert!(id(-3i32) / 2 == -1);
    assert!(id(3i32) / -2 == -1);
    assert!(id(-3i32) / -2 == 1);
    prusti_assert!(id(3i32) / 2 == 1);
    prusti_assert!(id(-3i32) / 2 == -1);
    prusti_assert!(id(3i32) / -2 == -1);
    prusti_assert!(id(-3i32) / -2 == 1);
    assert!(false); // Smoke test, the only one that should fail
}
```

If someone wants to work on this, the fix can be done similarly to how `fn rem` is used to encode the correct semantics of signed modulo operations.

fpoli:
Fixes #1505

~~Note that this PR also changes the default of `simplify_encoding` to `false`, because Viper's optimization of signed integer divisions is wrong (https://github.com/viperproject/silver/issues/782).~~

### DIFF
--- a/vir/defs/polymorphic/ast/expr.rs
+++ b/vir/defs/polymorphic/ast/expr.rs
@@ -151,7 +151,7 @@ __binary_op__! {
     add Add,
     sub Sub,
     mul Mul,
-    div Div,
+    viper_div Div,
     modulo Mod,
     and And,
     or Or,
@@ -316,7 +316,19 @@ impl Expr {
     }
 
     #[allow(clippy::should_implement_trait)]
-    /// Encode Rust reminder. This is *not* Viper modulo.
+    /// Encode Rust's division. This is *not* Viper's division.
+    pub fn div(left: Expr, right: Expr) -> Self {
+        Expr::ite(
+            Expr::ge_cmp(left.clone(), 0.into()),
+            // positive value or left % right == 0
+            Expr::viper_div(left.clone(), right.clone()),
+            // negative value
+            Expr::minus(Expr::viper_div(Expr::minus(left), right)),
+        )
+    }
+
+    #[allow(clippy::should_implement_trait)]
+    /// Encode Rust's signed reminder. This is *not* Viper's modulo.
     pub fn rem(left: Expr, right: Expr) -> Self {
         let abs_right = Expr::ite(
             Expr::ge_cmp(right.clone(), 0.into()),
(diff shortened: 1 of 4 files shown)


fpoli:
Disabling `simplify_encoding` requires too many changes. It's better to fix this in Viper and wait for a new nightly release.

vir/defs/polymorphic/ast/expr.rs
```diff
@@ -316,7 +316,22 @@ impl Expr {
...
 
     #[allow(clippy::should_implement_trait)]
-    /// Encode Rust reminder. This is *not* Viper modulo.
+    /// Encode Rust's division. This is *not* Viper's division.
+    pub fn div(left: Expr, right: Expr) -> Self {
+        Expr::ite(
+            Expr::or(
+                Expr::ge_cmp(left.clone(), 0.into()),
+                Expr::eq_cmp(Expr::modulo(left.clone(), right.clone()), 0.into()),
+            ),
```
fpoli:
```suggestion
            Expr::ge_cmp(left.clone(), 0.into()),
```

vir/src/legacy/ast/expr.rs
```diff
@@ -570,6 +571,22 @@ impl Expr {
...
     }
 
+    #[allow(clippy::should_implement_trait)]
+    /// Encode Rust's division. This is *not* Viper's division.
+    pub fn div(left: Expr, right: Expr) -> Self {
+        Expr::ite(
+            Expr::or(
+                Expr::ge_cmp(left.clone(), 0.into()),
+                Expr::eq_cmp(Expr::modulo(left.clone(), right.clone()), 0.into()),
+            ),
```
fpoli:
```suggestion
            Expr::ge_cmp(left.clone(), 0.into()),
```

